PR #1, opened 2021-01-14
added baseline funcs from old repo

woutslakhorst:
stripped all that is no longer needed


crypto/api/v1/api.go
```diff
@@ -0,0 +1,100 @@
...
+// SignJwt handles api calls for signing a Jwt
+func (w *Wrapper) SignJwt(ctx echo.Context) error {
+	var signRequest = &SignJwtRequest{}
+	err := ctx.Bind(signRequest)
+	if err != nil {
+		log.Logger().Error(err.Error())
+		return echo.NewHTTPError(http.StatusBadRequest, err.Error())
+	}
+
+	if len(signRequest.Kid) == 0 {
```
reinkrul, line 49: put this in a validate function on `SignJwtRequest` (also easier testing)?

crypto/api/v1/api.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+
+	if len(signRequest.Claims) == 0 {
+		return echo.NewHTTPError(http.StatusBadRequest, "missing claims")
+	}
+
+	sig, err := w.C.SignJWT(signRequest.Claims, signRequest.Kid)
+
+	if err != nil {
+		log.Logger().Error(err.Error())
```
reinkrul, line 60: maybe we can have middleware log errors?

crypto/api/v1/api.go
```diff
@@ -0,0 +1,100 @@
...
+		return echo.NewHTTPError(http.StatusBadRequest, err.Error())
+	}
+
+	return ctx.String(http.StatusOK, sig)
+}
+
+// PublicKey returns a public key for the given urn. The urn represents a legal entity. The api returns the public key either in PEM or JWK format.
```
reinkrul, line 67: urn -> key ID

crypto/api/v1/client.go
```diff
@@ -0,0 +1,89 @@
...
+		panic(err)
+	}
+	return response
+}
+
+func (hb HTTPClient) client() ClientInterface {
+	return hb.clientWithRequestEditor(nil)
+}
+
+// GetPublicKey returns a PrivateKey from the server given a kid
```
reinkrul, line 60: wiat, what?

crypto/crypto.go
```diff
@@ -0,0 +1,163 @@
...
+		}
+	})
+	return instance
+}
+
+// Configure loads the given configurations in the engine. Any wrong combination will return an error
+func (client *Crypto) Configure() error {
+	var err error
+	client.configOnce.Do(func() {
+		if core.NutsConfig().Mode() != core.ServerEngineMode {
```
reinkrul, line 115: do we still need this?

crypto/crypto.go
```diff
@@ -0,0 +1,163 @@
...
+		return errors.New("only fs backend available for now")
+	}
+	var err error
+	if client.Storage, err = storage.NewFileSystemBackend(client.Config.getFSPath()); err != nil {
+		return err
+	}
+	return nil
+}
+
+// New generates a new key pair. If a key pair with the same identifier already exists, it is overwritten.
```
reinkrul, line 136: "If a key pair with the same identifier already exists, it is overwritten."

Why not error? I think it should never be allowed to happen?

crypto/crypto.go
```diff
@@ -0,0 +1,163 @@
...
+func generateECKeyPair() (*ecdsa.PrivateKey, error) {
+	return ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
+}
+
+// PrivateKeyExists checks storage for an entry for the given legal entity and returns true if it exists
+func (client *Crypto) PrivateKeyExists(kid string) bool {
+	return client.Storage.PrivateKeyExists(kid)
+}
+
+// GetPublicKey loads the key from storage and returns it as PEM encoded. Only supports RSA style keys
```
reinkrul, line 164: think thiscomment is incorrect

crypto/engine/engine.go
```diff
@@ -0,0 +1,145 @@
...
+
+import (
+	"crypto"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"time"
+
+	"github.com/nuts-foundation/nuts-node/core"
+	crypto2 "github.com/nuts-foundation/nuts-node/crypto"
```
reinkrul, line 29: why2?

crypto/interface.go
```diff
@@ -0,0 +1,41 @@
...
+// KidNamingFunc is a function passed to New() which generates the kid for the pub/priv key
+type KidNamingFunc func(key crypto.PublicKey) string
+
+// KeyStore defines the functions than can be called by a Cmd, Direct or via rest call.
+type KeyStore interface {
+	// New generates a keypair and returns the public key.
+	// the KidNamingFunc will provide the kid. priv/pub keys are appended with a postfix and stored
+	New(namingFunc KidNamingFunc) (crypto.PublicKey, error)
+	// GetPrivateKey returns the specified private key (for e.g. signing) in non-exportable form.
+	GetPrivateKey(kid string) (crypto.Signer, error)
```
reinkrul, line 34: Document the behaviour for when the key is missing (is there an error returned or not). Same for `GetPublicKey`

crypto/interface.go
```diff
@@ -0,0 +1,41 @@
...
+	// GetPublicKey returns the PublicKey
+	GetPublicKey(kid string) (crypto.PublicKey, error)
+	// SignJWT creates a signed JWT using the given key and map of claims (private key must be present).
+	SignJWT(claims map[string]interface{}, kid string) (string, error)
+	// PrivateKeyExists returns if the specified private key exists.
+	PrivateKeyExists(key string) bool
```
reinkrul, line 40: `key` -> `kid`
error can't occur?

crypto/jwx.go
```diff
@@ -0,0 +1,149 @@
...
+	"crypto/rsa"
+	"errors"
+
+	"github.com/lestrrat-go/jwx/jwa"
+	"github.com/lestrrat-go/jwx/jwk"
+	"github.com/lestrrat-go/jwx/jws"
+	"github.com/lestrrat-go/jwx/jwt"
+)
+
+// ErrUnsupportedSigningKey is returned when an unsupported private key is used to sign. Currently only ecdsa and rsa keys are supported
```
reinkrul, line 33: RSA too?

crypto/jwx.go
```diff
@@ -0,0 +1,149 @@
...
+
+// JWTKidAlg parses a JWT, does not validate it and returns the 'kid' and 'alg' headers
+func JWTKidAlg(tokenString string) (string, jwa.SignatureAlgorithm, error) {
+	j, err := jws.ParseString(tokenString)
+	if err != nil {
+		return "", "", err
+	}
+
+	if len(j.Signatures()) != 1 {
+		return "", "", errors.New("incorrect amount of signatures in JWT")
```
reinkrul, line 100: amount -> number

crypto/util/pem.go
```diff
@@ -0,0 +1,100 @@
...
+	"crypto/ed25519"
+	"crypto/rsa"
+	"crypto/x509"
+	"encoding/pem"
+)
+
+// PemToPublicKey converts a PEM encoded public key to a crypto.PublicKey
+func PemToPublicKey(pub []byte) (crypto.PublicKey, error) {
+	block, _ := pem.Decode(pub)
+	if block == nil || block.Type != "PUBLIC KEY" {
```
reinkrul, line 30: I think the header can also be specified with the type e.g. `RSA PUBLIC KEY` or `EC PUBLIC KEY` so it might not just be `public key`.

reinkrul (changes requested):

crypto/api/v1/api.go
woutslakhorst: I'll add an issue for now

crypto/crypto.go
woutslakhorst: yes, until we have a CLI binary (if we go that way)
It does use the global server mode instead of a crypto config mode now.

crypto/crypto.go
woutslakhorst: wrong docs. Changed it to: should be handled by storage and bad practise to re-use a kid.

crypto/engine/engine.go
woutslakhorst: default golib `crypto` package is also imported

crypto/interface.go
woutslakhorst: have left impl as it was.

crypto/util/pem.go
woutslakhorst: added other options as well.

woutslakhorst: although EC public key is always marshalled as ----BEGIN PUBLIC KEY------

reinkrul (approved):

PR #17, opened 2021-01-25
Generate config options documentation based on engine flags

reinkrul:
Like we did before in nuts-go.


stevenvegt (changes requested):
Nice! 👍 
Add this to our readme and makfile?

woutslakhorst (changes requested):
Add to readme generation via makefile

reinkrul:
already present in makefile, documented usage in readme

woutslakhorst (approved):

PR #26, opened 2021-01-28
Removed EngineCtl singleton to improve testability and isolation

reinkrul:
Replaced with System type which isn't a singleton.

### DIFF
--- a/core/config.go
+++ b/core/config.go
@@ -210,13 +210,13 @@ func (ngc *NutsGlobalConfig) bindFlag(fs *pflag.FlagSet, name string) error {
 }
 
 // PrintConfig outputs the current config to the logger on info level
-func (ngc *NutsGlobalConfig) PrintConfig(logger log.FieldLogger) {
+func (ngc *NutsGlobalConfig) PrintConfig(system *System, logger log.FieldLogger) {
 	title := "Config"
 	var longestKey = 10
 	var longestValue int
-	for _, e := range EngineCtl.Engines {
-		if e.FlagSet != nil {
-			e.FlagSet.VisitAll(func(flag *pflag.Flag) {
+	system.VisitEngines(func(engine *Engine) {
+		if engine.FlagSet != nil {
+			engine.FlagSet.VisitAll(func(flag *pflag.Flag) {
 				s := fmt.Sprintf("%v", ngc.v.Get(strings.ToLower(flag.Name)))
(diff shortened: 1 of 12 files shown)


core/config.go
```diff
@@ -87,7 +87,7 @@ const (
 	// GlobalServerMode is used for starting the application in server mode, running as Nuts node.
 	GlobalServerMode string = "server"
 	// GlobalCLIMode is used for starting the application in CLI mode, meaning it's used as CLI client administering
-	// for a remote Nuts node. Engines will start in client mode when this mode is specified.
+	// for a remote Nuts node. engines will start in client mode when this mode is specified.
```
woutslakhorst, line 90: no comment

core/engine.go
```diff
@@ -27,13 +27,85 @@ import (
...
+type System struct {
+	// engines is the slice of all registered engines
+	engines []*Engine
+}
+
+// Diagnostics returns the compound diagnostics for all engines.
+func (system *System) Diagnostics() []DiagnosticResult {
+	result := make([]DiagnosticResult, 0)
+	system.VisitEngines(func(engine *Engine) {
+		if engine.Diagnosable != nil {
```
woutslakhorst, line 45: we could also go to a system where you check if an engine can be cast to a Diagnosable? Removing the need to register it in the engine struct?

woutslakhorst (commented):
See other comment: what if we use interfaces: diagnosable, runnable, configurable and check if the engine supports that. Then remove func references from struct?

reinkrul:
> See other comment: what if we use interfaces: diagnosable, runnable, configurable and check if the engine supports that. Then remove func references from struct?

I started to rework exactly this; make Engine an interface which might implement these interfaces you mentioned. Got stuck with the 'engine instances'; since each engine is essentially composed of core functionality ("instance", e.g. `VDR` or `KeyStore`) and something we call an "engine" which attaches to it to configure and manage its lifecycle.

I'd like to merge this first and then we can see how to rework the `Engine` struct.

core/engine.go
reinkrul: See comment below

woutslakhorst (approved):
Please track engine struct -> interface refactor with an issue

reinkrul:
https://github.com/nuts-foundation/nuts-node/issues/29

PR #27, opened 2021-01-28
Remove engines modes

reinkrul:
Introduces the `server` command to start the server (instead of the root command). The root command now just prints the help. All other commands are client commands by default.

Based on https://github.com/nuts-foundation/nuts-node/pull/26 (must be merged first)
Fixes https://github.com/nuts-foundation/nuts-node/issues/22


woutslakhorst (changes requested):
This would be a nice moment to add something under the **Node & Network** doc chapter on how to use the executable (CLI/SERVER print help).

woutslakhorst (approved):

Issue #12, opened 2021-01-25
Add DID Doc Update validation


stevenvegt:
Update checks are not complete.

* [x] Version of a new doc does not start with 0
* [x] Version or updatedAt timestamp of a updated doc !> previous
* [x] New keys added should be added to the keystore
* [x] Removed keys should be removed from the keystore
* [x] Resolving a chain of controllers (more than 1)
* [x] Deactivation of ducuments
* [x] Timeline validation according to [RFC004 3.4.1](https://nuts-foundation.gitbook.io/drafts/rfc/rfc004-distributed-document-format#3-4-1-timeline-validation)
* [x] Full e2e test which will test the complete life cycle of a document:
  * [x] Creation of a new document A
  * [x] Update Document A with a new service
  * [x] Update Document A with a new authentication key, old one should be removed
  * [x] Creation of New Document B
  * [x] Update of Document B controller which transfers control to Document A
  * [x] Update of Document B with new service endpoints, signed by Document A
  * [ ] Creation of Document C
  * [ ] Update of Document C which makes Document B the controller
  * [ ] Deactivation of Document C
  * [x] Deactivation of Document A and thus also Document B

woutslakhorst:
Timeline validation is also to be removed: nuts-foundation/nuts-specification#30

PR #28, opened 2021-01-28
Vdr with network integration

stevenvegt:
Add checks to see if a DID document is valid and if so, add it to the did store.

Implements part of the checks listed in #12


vdr/ambassador.go
```diff
@@ -37,20 +49,245 @@ type Ambassador interface {
...
+	headerKey := document.SigningKey()
+	// Create key thumbprint
+	headerKeyThumbprint, err := headerKey.Thumbprint(thumbprintAlg)
+	if err != nil {
+		return fmt.Errorf("unable to generate network document signing key thumbprint: %w", err)
+	}
+
+	// Check if headerKey is one of the keys embedded in the authenticationMethod
+	didDocumentAuthKeys := nextDIDDocument.Authentication
+	if documentKey, err := n.findKeyByThumbprint(headerKeyThumbprint, didDocumentAuthKeys); documentKey == nil || err != nil {
```
reinkrul, line 111: Now all errors are mapped to the "key not found" error as well. Imo it should be written something along the lines of;

```
if documentKey, err := n.findKeyByThumbprint(headerKeyThumbprint, didDocumentAuthKeys); err != nil {
    return err
} else if documentKey == nil {
	return fmt.Errorf("key used to sign Network document must be be part of DID Document authentication")
}
```

vdr/ambassador.go
```diff
@@ -37,20 +49,245 @@ type Ambassador interface {
...
+		return fmt.Errorf("unable to unmarshall did document from network payload: %w", err)
+	}
+
+	if isUpdate(document) {
+		return n.updateDIDDocument(document, nextDIDDocument)
+	}
+	return n.createDIDDocument(document, nextDIDDocument)
+}
+
+func (n *ambassador) createDIDDocument(document dag.SubscriberDocument, nextDIDDocument did.Document) error {
```
reinkrul, line 98: maybe we should call this `handleCreateDIDDocument`? Because it processes/handles an incoming creation of a DID Document. At first I thought it was creating a new document (but it isn;t)

vdr/ambassador.go
```diff
@@ -37,20 +49,245 @@ type Ambassador interface {
...
+		Created:    document.SigningTime(),
+		Updated:    nil,
+		Version:    NewDocumentVersion,
+		TimelineID: document.Ref(),
+		Hash:       document.PayloadHash(),
+	}
+	return n.didStore.Write(nextDIDDocument, documentMetadata)
+}
+
+func (n *ambassador) updateDIDDocument(document dag.SubscriberDocument, nextDIDDocument did.Document) error {
```
reinkrul, line 128: `handleUpdateDIDDocument`?

vdr/ambassador.go
reinkrul, line 98: `nextDIDDocument` is a strange name is for this function. Maybe `createdDIDDocument`?

vdr/ambassador.go
reinkrul, line 128: nextDIDDocument  -> `updatedDIDDocument`

vdr/ambassador.go
```diff
@@ -37,20 +49,245 @@ type Ambassador interface {
...
+
+// thumbprintAlg is used for creating public key thumbprints
+var thumbprintAlg = crypto.SHA256
+
+// callback gets called when new DIDDocuments are received by the network. All checks on the signature are already performed.
+// This method will check the integrity of the DID document related to the public key used to sign the network document.
+// The rules are based on the Nuts RFC006
+// payload should be a json encoded did.document
+func (n *ambassador) callback(document dag.SubscriberDocument, payload []byte) error {
+	logging.Log().Info("Processing DID documents received from Nuts Network.", DIDDocumentType, document.Ref())
```
reinkrul, line 81: not sure how this logging is printed with the arguments just appended? I'd like a bit more formatting (something like `(type=%s,ref=%s)`). Also, maybe this should be a DEBUG level statement and add an INFO in the create/update handler? Or TRACE/DEBUG?

vdr/ambassador.go
```diff
@@ -37,20 +49,245 @@ type Ambassador interface {
...
+}
+
+func (n *ambassador) updateDIDDocument(document dag.SubscriberDocument, nextDIDDocument did.Document) error {
+	// Resolve current version of DID Document
+	resolverMetadata := &types.ResolveMetadata{
+		AllowDeactivated: false,
+	}
+	currentDIDDocument, currentDIDMeta, err := n.didStore.Resolve(nextDIDDocument.ID, resolverMetadata)
+	if err != nil {
+		return fmt.Errorf("unable to update did document: %s", err)
```
reinkrul, line 135: you generally wrap errors using `%w`

vdr/ambassador.go
```diff
@@ -37,20 +49,245 @@ type Ambassador interface {
...
+	}
+	// Check if the new document is actual newer by comparing timeline versions
+	if currentDIDMeta.Version >= document.TimelineVersion() {
+		return fmt.Errorf("unable to update did document: timeline version of current document is greater or equal to the new version")
+	}
+	if currentDIDMeta.TimelineID != document.TimelineID() {
```
reinkrul, line 141: `TimelineID` is a `SHA256Hash`? Not sure how the `!=` then works, but it has an `Equals` func

vdr/ambassador.go
```diff
@@ -37,20 +49,245 @@ type Ambassador interface {
...
+		return fmt.Errorf("timelineIDs of new and current DID documents must match")
+	}
+
+	// Resolve controllers of current version (could be the same document)
+	didControllers, err := n.resolveDIDControllers(currentDIDDocument)
+	logging.Log().Debug(didControllers)
```
reinkrul, line 147: leftover debugging? Maybe a TRACE level log, but at least it needs some more text to it

vdr/ambassador.go
```diff
@@ -37,20 +49,245 @@ type Ambassador interface {
...
+
+	var controllerVerificationRelationships = []did.VerificationRelationship{}
+	for _, didCtrl := range didControllers {
+		for _, auth := range didCtrl.Authentication {
+			controllerVerificationRelationships = append(controllerVerificationRelationships, auth)
+		}
+	}
+
+	// In an update, only the keyID is provided in te network document. Resolve the key from the key store
```
reinkrul, line 156: te -> the

vdr/ambassador.go
```diff
@@ -37,20 +49,245 @@ type Ambassador interface {
...
+	// This should succeed since the signature of the network document has already been verified.
+	pKey, err := n.keyResolver.GetPublicKey(document.SigningKeyID(), document.SigningTime())
+	if err != nil {
+		return fmt.Errorf("unable to resolve signingkey %w", err)
```
reinkrul, line 160: `: %w`

vdr/ambassador.go
```diff
@@ -37,20 +49,245 @@ type Ambassador interface {
...
+	}
+	headerKey, err := jwk.New(pKey)
+	if err != nil {
+		return fmt.Errorf("could not parse public key into jwk %w", err)
```
reinkrul, line 164: `: %w`

vdr/ambassador.go
```diff
@@ -37,20 +49,245 @@ type Ambassador interface {
...
+	}
+	// Create thumbprint
+	headerKeyThumbprint, err := headerKey.Thumbprint(thumbprintAlg)
+	if err != nil {
+		return fmt.Errorf("unable to generate network document signing key thumbprint")
```
reinkrul, line 169: wrap err

vdr/ambassador.go
```diff
@@ -37,20 +49,245 @@ type Ambassador interface {
...
+	}
+
+	// Check if the signingKey is listed as a valid authenticationMethod in one of the controllers
+	keyToSign, err := n.findKeyByThumbprint(headerKeyThumbprint, controllerVerificationRelationships)
+	if keyToSign == nil {
```
reinkrul, line 174: `err` is ignored

vdr/ambassador.go
```diff
@@ -37,20 +49,245 @@ type Ambassador interface {
...
+		return n.updateDIDDocument(document, nextDIDDocument)
+	}
+	return n.createDIDDocument(document, nextDIDDocument)
+}
+
+func (n *ambassador) createDIDDocument(document dag.SubscriberDocument, nextDIDDocument did.Document) error {
+	// First, check if the network document was signed by the same key as is embedded in the DID Document`s authenticationMethod:
+
+	// Take signing key from network document header
+	headerKey := document.SigningKey()
```
reinkrul, line 102: I'd name `headerKey` -> `signingKey`?

vdr/ambassador.go
reinkrul, line 111: Thumbprint is just a hash over the public key bytes over the key, I'd say the complete JWK must match exactly 100%? There could be other fields in the JWK than just the ones relevant for thumbprint calc?

vdr/ambassador.go
```diff
@@ -37,20 +49,245 @@ type Ambassador interface {
...
+
+// checkSubscriberDocumentIntegrity performs basic integrity checks on the SubscriberDocument fields:
+func checkSubscriberDocumentIntegrity(document dag.SubscriberDocument) error {
+	// check the payload type:
+	if document.PayloadType() != DIDDocumentType {
+		return fmt.Errorf("wrong payload type for this subscriber. Can handle: %s, got: %s", DIDDocumentType, document.PayloadType())
+	}
+
+	// PayloadHash must be set
+	if document.PayloadHash().Empty() {
```
reinkrul, line 212: this can't happen in practice, if it's something we'd like to check it should be in the network layer (not here). Your subscribers would also never be called when there's no payload (since it's called after payload is received).

vdr/ambassador.go
```diff
@@ -37,20 +49,245 @@ type Ambassador interface {
...
+		TimelineID: document.TimelineID(),
+		Hash:       document.PayloadHash(),
+	}
+	return n.didStore.Update(nextDIDDocument.ID, currentDIDMeta.Hash, nextDIDDocument, &documentMetadata)
+}
+
+// checkSubscriberDocumentIntegrity performs basic integrity checks on the SubscriberDocument fields:
+func checkSubscriberDocumentIntegrity(document dag.SubscriberDocument) error {
+	// check the payload type:
+	if document.PayloadType() != DIDDocumentType {
```
reinkrul, line 207: I think this is handled by just subscribing to that exact type?

vdr/ambassador.go
```diff
@@ -37,20 +49,245 @@ type Ambassador interface {
...
+	}
+
+	// PayloadHash must be set
+	if document.PayloadHash().Empty() {
+		return fmt.Errorf("payloadHash must be provider")
+	}
+
+	// Signing time should be set and lay in the past:
+	// allow for 2 seconds clock skew
+	if document.SigningTime().IsZero() || document.SigningTime().After(time.Now().Add(2*time.Second)) {
```
reinkrul, line 218: Not sure if this belongs here. It's also not specified in RFC006, so if we add it, it must be specified as well.

vdr/ambassador.go
```diff
@@ -37,20 +49,245 @@ type Ambassador interface {
...
+		return fmt.Errorf("signingTime must be set and in the past")
+	}
+
+	if isUpdate(document) {
+		// For a DID Document update TimelineID must be set
+		if document.TimelineID().Empty() {
```
reinkrul, line 224: Superfluous check since isUpdate checks exactly that;

```
func isUpdate(document dag.SubscriberDocument) bool {
	return !document.TimelineID().Empty()
}
```

vdr/ambassador.go
```diff
@@ -37,20 +49,245 @@ type Ambassador interface {
...
+			return fmt.Errorf("timelineID must be set for updates")
+		}
+
+		if document.TimelineVersion() <= NewDocumentVersion {
+			return fmt.Errorf("timelineVersion for updates must be greater than %d", NewDocumentVersion)
+		}
+	} else {
+		// For a new DID Document TimelineID must be nil
+		if !document.TimelineID().Empty() {
```
reinkrul, line 233: same for this

vdr/ambassador.go
```diff
@@ -37,20 +49,245 @@ type Ambassador interface {
...
+		Version:    document.TimelineVersion(),
+		TimelineID: document.TimelineID(),
+		Hash:       document.PayloadHash(),
+	}
+	return n.didStore.Update(nextDIDDocument.ID, currentDIDMeta.Hash, nextDIDDocument, &documentMetadata)
+}
+
+// checkSubscriberDocumentIntegrity performs basic integrity checks on the SubscriberDocument fields:
+func checkSubscriberDocumentIntegrity(document dag.SubscriberDocument) error {
+	// check the payload type:
```
reinkrul, line 206: I think this is missing the check that all keys having matching ID (matching tot the DID, so in the form of `<DID>#thumbprint`)

vdr/ambassador.go
```diff
@@ -37,20 +49,245 @@ type Ambassador interface {
...
+	}
+	if currentDIDMeta.TimelineID != document.TimelineID() {
+		return fmt.Errorf("timelineIDs of new and current DID documents must match")
+	}
+
+	// Resolve controllers of current version (could be the same document)
+	didControllers, err := n.resolveDIDControllers(currentDIDDocument)
+	logging.Log().Debug(didControllers)
+
+	var controllerVerificationRelationships = []did.VerificationRelationship{}
```
reinkrul, line 149: use `make` (at least that's what IntelliJ says)

vdr/ambassador.go
```diff
@@ -37,20 +49,245 @@ type Ambassador interface {
...
+		}
+		didControllers = append(didControllers, controllerDoc)
+	}
+	return didControllers, nil
+}
+
+// findKeyByThumbprint accepts a SHA256 generated thumbprint and tries to find it in a provided list of did.VerificationRelationship s.
+// Returns an error if it could not generate a thumbprint of one of the VerificationRelationship keys
+func (n ambassador) findKeyByThumbprint(thumbPrint []byte, didDocumentAuthKeys []did.VerificationRelationship) (jwk.Key, error) {
+	hashAlg := crypto.SHA256
```
reinkrul, line 277: I've seen this defined as const somewhere else in this package already?

vdr/ambassador.go
```diff
@@ -37,20 +49,245 @@ type Ambassador interface {
...
+
+	var documentKey jwk.Key
+	for _, key := range didDocumentAuthKeys {
+		// Create thumbprint
+		documentThumbprint, err := key.JWK().Thumbprint(hashAlg)
+		if err != nil {
+			return nil, fmt.Errorf("unable to generate did document signing key thumbprint")
```
reinkrul, line 284: wrap err

vdr/ambassador.go
```diff
@@ -37,20 +49,245 @@ type Ambassador interface {
...
+		}
+		// Compare thumbprints
+		if bytes.Equal(thumbPrint, documentThumbprint) {
+			documentKey = key.JWK()
```
reinkrul, line 288: `JWK()` called twice, maybe put it in a var so we don't ahve to do it twice

vdr/types/common.go
```diff
@@ -47,8 +47,8 @@ type DocumentMetadata struct {
 	Updated *time.Time `json:"updated,omitempty"`
 	// Version contains the semantic version of the DID document.
 	Version int `json:"version"`
-	// OriginJWSHash contains the hash of the JWS envelope of the first version of the DID document.
-	OriginJWSHash hash.SHA256Hash `json:"originJwsHash"`
+	// TimelineID contains the hash of the JWS envelope of the first version of the DID document.
+	TimelineID hash.SHA256Hash `json:"originJwsHash"`
```
reinkrul, line 51: maybe also change it in the REST API if we think it should be renamed to `TimelineID`?

vdr/vdr.go
```diff
@@ -127,7 +133,22 @@ func (r VDR) Resolve(id did.DID, metadata *types.ResolveMetadata) (*did.Document
 
 // Update updates a DID Document based on the DID and current hash
 func (r VDR) Update(id did.DID, current hash.SHA256Hash, next did.Document, metadata *types.DocumentMetadata) error {
-	return r.store.Update(id, current, next, metadata)
+	resolverMetada := &types.ResolveMetadata{
```
reinkrul, line 136: If I am correct we now first create the (updated) document on the store and then validate it when we receive it again. I'd like that we (also) validate it before storing it on the network (probably using the same validation rules) because that (hopefully) keeps the garbage on the network caused by errors (human/software) in updates to a minimum.

vdr/vdr.go
```diff
@@ -127,7 +133,22 @@ func (r VDR) Resolve(id did.DID, metadata *types.ResolveMetadata) (*did.Document
 
 // Update updates a DID Document based on the DID and current hash
 func (r VDR) Update(id did.DID, current hash.SHA256Hash, next did.Document, metadata *types.DocumentMetadata) error {
```
reinkrul, line 135: why is `metadata` a pointer? Also, why isn't it used? (drop it?)

vdr/vdr.go
```diff
@@ -127,7 +133,22 @@ func (r VDR) Resolve(id did.DID, metadata *types.ResolveMetadata) (*did.Document
...
+	}
+	_, meta, err := r.store.Resolve(id, resolverMetada)
+	if err != nil {
+		return err
+	}
+	payload, err := json.Marshal(next)
+	if err != nil {
+		return err
+	}
+	// TODO: look into the controller of the did for a signing key
```
reinkrul, line 148: make sure we don't forget this

vdr/vdr.go
```diff
@@ -127,7 +133,22 @@ func (r VDR) Resolve(id did.DID, metadata *types.ResolveMetadata) (*did.Document
...
+	keyID := next.Authentication[0].ID.String()
```
reinkrul, line 149: I think it must be signed by a key that was present in the _previous_ (current) version of the document, so from the `Resolve` action

reinkrul (changes requested):

network/dag/document.go
```diff
@@ -49,37 +50,72 @@ var invalidHeaderErrFmt = "invalid %s header"
...
+
 	// PayloadType returns the MIME-formatted type of the payload. It must contain the context and specific type of the
 	// payload, e.g. 'registry/endpoint'.
 	PayloadType() string
-	// Payload returns the hash of the payload of the document.
-	Payload() hash.SHA256Hash
+}
+
+// NetworkHeader groups methods for working with a network document header.
+type NetworkHeader interface {
```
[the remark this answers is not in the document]
stevenvegt: Perhaps, but these fields (`Previous` and `Version`) are exactly the ones I don't need in the `SubscriberDocument`.

network/dag/document.go
```diff
@@ -49,37 +50,72 @@ var invalidHeaderErrFmt = "invalid %s header"
...
 	// Previous returns the references of the previous documents this document points to.
 	Previous() []hash.SHA256Hash
 	// Version returns the version number of the distributed document format.
 	Version() Version
+}
+
+// SubscriberDocument contains a subset of the Document methods which are only relevant to subscribers.
+type SubscriberDocument interface {
```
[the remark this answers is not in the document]
stevenvegt: Well, we saved 3 fields in total, that is indeed not as much as I hoped for. Technically I can also do without  `SigningAlgorithm`. And if we combine the `SigningKey` and `SigningKeyID` in a preprossesing step into one jwk we can loose those the `SigningKeyID` as well.

vdr/ambassador.go
```diff
@@ -37,20 +49,245 @@ type Ambassador interface {
...
+
+	if isUpdate(document) {
+		return n.updateDIDDocument(document, nextDIDDocument)
+	}
+	return n.createDIDDocument(document, nextDIDDocument)
+}
+
+func (n *ambassador) createDIDDocument(document dag.SubscriberDocument, nextDIDDocument did.Document) error {
+	// First, check if the network document was signed by the same key as is embedded in the DID Document`s authenticationMethod:
+
```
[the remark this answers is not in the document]
stevenvegt: No, its the actual description what we will be doing here. I'll remove the First word.

vdr/ambassador.go
stevenvegt: I renamed it to ProposedDocument

vdr/ambassador.go
stevenvegt: Well, we want to be sure the public keys crypto matches. Not necessarily catch faults in metadata or marshalling/unmarshzlling errors. I think the thumbprint method is sufficient for that? A PublicKey.Equel is annouced in go: https://go-review.googlesource.com/c/go/+/223754/5

vdr/ambassador.go
stevenvegt: So I can convert it to a panic instead of error? :)

vdr/ambassador.go
stevenvegt: Yes, so I can convert it to a panic()?

vdr/ambassador.go
stevenvegt: Yes, but that is also abstracted away and that implementaion can change. This checks section covers all the preconditions in one place and there might be overlapp with checks happening in the network or in other functions. But these might change because they did not had the responsibility of an integrety check, this section has.

vdr/vdr.go
stevenvegt: It's part of the interface which is the same as the storage backend. Not sure yet what to do with it. But this implementation is not finished yet.

vdr/vdr.go
stevenvegt: I've added a list of checks to perform in an issue.

vdr/ambassador.go
```diff
@@ -37,20 +49,243 @@ type Ambassador interface {
...
+		return fmt.Errorf("payloadHash must be provider")
+	}
+
+	// Signing time should be set and lay in the past:
+	// allow for 2 seconds clock skew
+	if document.SigningTime().IsZero() || document.SigningTime().After(time.Now().Add(2*time.Second)) {
+		return fmt.Errorf("signingTime must be set and in the past")
+	}
+
+	if isUpdate(document) {
```
woutslakhorst, line 221: double isUpdate, 1 from the callback and 1 from integrityCheck. This check could be splitted in 2 functions removing the need for this check.

vdr/ambassador.go
```diff
@@ -37,20 +49,243 @@ type Ambassador interface {
...
+			return nil, fmt.Errorf("unable to resolve document controller: %w", err)
+		}
+		didControllers = append(didControllers, controllerDoc)
+	}
+	return didControllers, nil
+}
+
+// findKeyByThumbprint accepts a SHA256 generated thumbprint and tries to find it in a provided list of did.VerificationRelationship s.
+// Returns an error if it could not generate a thumbprint of one of the VerificationRelationship keys
+func (n ambassador) findKeyByThumbprint(thumbPrint []byte, didDocumentAuthKeys []did.VerificationRelationship) (jwk.Key, error) {
```
woutslakhorst, line 275: Doesn't do `<DID>#<thumbprint>`

vdr/vdr.go
```diff
@@ -76,14 +79,15 @@ func (r *VDR) Configure() error {
 
 	r.configOnce.Do(func() {
 		if r.networkAmbassador == nil {
-			r.networkAmbassador = NewAmbassador(r.network)
+			r.networkAmbassador = NewAmbassador(r.network, r.store, crypto.Instance())
```
woutslakhorst, line 82: it should forward the cryptoClient from the vdr struct, not create a new instance.

vdr/vdr.go
woutslakhorst: should be single line of code, right.....

woutslakhorst (changes requested):

vdr/ambassador.go
```diff
@@ -37,20 +49,243 @@ type Ambassador interface {
...
+		return fmt.Errorf("could not parse public key into jwk: %w", err)
+	}
+	// Create thumbprint
+	signingKeyThumbprint, err := signingKey.Thumbprint(thumbprintAlg)
+	if err != nil {
+		return fmt.Errorf("unable to generate network document signing key thumbprint: %w", err)
+	}
+
+	// Check if the signingKey is listed as a valid authenticationMethod in one of the controllers
+	keyToSign, err := n.findKeyByThumbprint(signingKeyThumbprint, controllerVerificationRelationships)
```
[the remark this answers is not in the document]
woutslakhorst: solved

vdr/ambassador.go
woutslakhorst: that would be one for the list of checks todo.

vdr/ambassador.go
stevenvegt: https://github.com/nuts-foundation/nuts-specification/issues/27

vdr/ambassador.go
stevenvegt: Added

vdr/vdr.go
stevenvegt: Will add a TODO for this.

vdr/ambassador.go
```diff
@@ -37,20 +49,243 @@ type Ambassador interface {
...
+	// Resolve current version of DID Document
+	resolverMetadata := &types.ResolveMetadata{
+		AllowDeactivated: false,
+	}
+	currentDIDDocument, currentDIDMeta, err := n.didStore.Resolve(proposedDIDDocument.ID, resolverMetadata)
+	if err != nil {
+		return fmt.Errorf("unable to update did document: %w", err)
+	}
+	// Check if the new document is actual newer by comparing timeline versions
+	if currentDIDMeta.Version >= document.TimelineVersion() {
```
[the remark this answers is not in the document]
stevenvegt: If we miss one, there might be inconsistancies in key material. But if we reject it, altough the rest of the document is valid, we might need a retry meganism?

vdr/ambassador.go
```diff
@@ -37,20 +49,243 @@ type Ambassador interface {
...
+		return fmt.Errorf("unable to update did document: timeline version of current document is greater or equal to the new version")
+	}
+	if !currentDIDMeta.TimelineID.Equals(document.TimelineID()) {
```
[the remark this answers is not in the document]
stevenvegt: Yes, it's an int which have the default uninitialized value of 0 in go.

vdr/ambassador.go
stevenvegt: That would be 3 functios than: general checks, checks for create and for updates. I see that it is technacally correct, but I don't think 3 seperate functions are better that 1 with a redundant check.

vdr/ambassador.go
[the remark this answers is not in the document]
woutslakhorst: It's actually allowed, but according to spec: 

- if version matches, apply lowest `iat` first (eg only update if `iat` is later)
- if `iat` matches, apply lowest hash first (eg only update if hash is bigger`)
- no error should be returned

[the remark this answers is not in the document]
stevenvegt: Valid point, updated the issue to fix that: https://github.com/nuts-foundation/nuts-node/issues/12

reinkrul (approved):

Issue #25, opened 2021-01-28
Verify signature of incoming Nuts Network Documents


reinkrul:
Still needs to be implemented, but we can do that now we have the KeyStore.

PR #30, opened 2021-01-29
Verify documents signatures upon receiving before adding them to the DAG

reinkrul:
Fixes https://github.com/nuts-foundation/nuts-node/issues/25
Based on https://github.com/nuts-foundation/nuts-node/pull/27 (must be merged first)

### DIFF
--- a/crypto/jwx.go
+++ b/crypto/jwx.go
@@ -54,28 +54,27 @@ func (client *Crypto) SignJWT(claims map[string]interface{}, kid string) (token 
 }
 
 // SignJWS creates a signed JWS (in compact form using) the given key (private key must be present), protected headers and payload.
-func (client *Crypto) SignJWS(payload []byte, protectedHeaders map[string]interface{}, kid string) (string, error) {
+func (client *Crypto) SignJWS(payload []byte, protectedHeaders map[string]interface{}, kid string) (string, jwa.SignatureAlgorithm, error) {
 	headers := jws.NewHeaders()
 	for key, value := range protectedHeaders {
 		if err := headers.Set(key, value); err != nil {
-			return "", fmt.Errorf("unable to set header %s: %w", key, err)
+			return "", "", fmt.Errorf("unable to set header %s: %w", key, err)
 		}
 	}
 	privateKey, err := client.Storage.GetPrivateKey(kid)
 	if err != nil {
-		return "", fmt.Errorf("error while signing JWS, can't get private key: %w", err)
+		return "", "", fmt.Errorf("error while signing JWS, can't get private key: %w", err)
 	}
 	privateKeyAsJWK, err := jwkKey(privateKey)
 	if err != nil {
-		return "", err
+		return "", "", err
 	}
 	algo := jwa.SignatureAlgorithm(privateKeyAsJWK.Algorithm())
-	protectedHeaders[jws.AlgorithmKey] = algo
 	data, err := jws.Sign(payload, algo, privateKey, jws.WithHeaders(headers))
 	if err != nil {
-		return "", fmt.Errorf("unable to sign JWS %w", err)
+		return "", "", fmt.Errorf("unable to sign JWS %w", err)
 	}
-	return string(data), nil
+	return string(data), algo, nil
 }
 
 func jwkKey(signer crypto.Signer) (key jwk.Key, err error) {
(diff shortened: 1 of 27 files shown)


network/dag/signing.go
```diff
@@ -110,5 +123,39 @@ func (d documentSigner) Sign(input UnsignedDocument, signingTime time.Time) (Doc
 	} else {
 		doc.signingKeyID = d.kid
 	}
+	doc.signingAlgorithm = algo
```
woutslakhorst, line 126: Together with SignJWS, this seems off. The algo should already be in the JWS header...

network/dag/signing.go
```diff
@@ -66,6 +72,13 @@ func (d documentSigner) Sign(input UnsignedDocument, signingTime time.Time) (Doc
 		if err != nil {
 			return nil, fmt.Errorf(errSigningDocumentFmt, err)
 		}
+		// The JWX library is fine with creating a JWK for a private key (including the private exponents), so
+		// we want to make sure the resolver didn't accidentally return a private key which just conforms to the
+		// crypto.PublicKey interface. Since that would lead to the node leaking its private key material in the
+		// resulting JWS which would be very, very bad.
+		if asSigner, ok := keyAsPublicKey.(crypto2.Signer); ok {
```
woutslakhorst, line 79: Can't we do this at another place? Getting a pub key and checking if it's a pub key feels off

network/dag/signing.go
```diff
@@ -17,10 +18,15 @@ type DocumentSigner interface {
 	Sign(input UnsignedDocument, signingTime time.Time) (Document, error)
 }
 
+// DocumentSignatureVerifier defines functions to verify document signatures.
+type DocumentSignatureVerifier interface {
```
woutslakhorst, line 22: The signer and verifier deserve their own files.

crypto/jwx.go
```diff
@@ -54,28 +54,27 @@ func (client *Crypto) SignJWT(claims map[string]interface{}, kid string) (token
 }
 
 // SignJWS creates a signed JWS (in compact form using) the given key (private key must be present), protected headers and payload.
-func (client *Crypto) SignJWS(payload []byte, protectedHeaders map[string]interface{}, kid string) (string, error) {
+func (client *Crypto) SignJWS(payload []byte, protectedHeaders map[string]interface{}, kid string) (string, jwa.SignatureAlgorithm, error) {
```
woutslakhorst, line 57: triple! return value.

network/dag/parser.go
```diff
@@ -108,15 +111,18 @@ func parseContentType(document *document, headers jws.Headers, _ *jws.Message) e
 	return nil
 }
 
-// parseSigningKey parses, validates and sets the document signing key (`jwk`) or key ID (`kid`).
-func parseSigningKey(document *document, headers jws.Headers, _ *jws.Message) error {
+// parseSignatureParams parses, validates and sets the document signing key (`jwk`) or key ID (`kid`).
+func parseSignatureParams(document *document, headers jws.Headers, _ *jws.Message) error {
```
woutslakhorst, line 115: it feels as if the `document` should wrap the JWS and provide functions instead of fields.

woutslakhorst (changes requested):
I can't fully decipher how the document signing and parsing works, but it looks like there's a layer too many. JWS headers that are copied are a symptom for that.

network/dag/signing.go
reinkrul: Moved it as a check to `crypto.signJWS`, it just refuses to sign when the `jwk` field contains a private key.

crypto/jwx.go
reinkrul: fixed

network/dag/signing.go
reinkrul: > I can't fully decipher how the document signing and parsing works, but it looks like there's a layer too many. JWS headers that are copied are a symptom for that.

It was a bit off, in my opinion especially because I returned the algorithm along with the JWS as string. That's fixed now (by just returning a new instance of the signed document, by parsing it). How it's structured;

* lestrrat/jws is used to actually sign JWS's
* Crypto signs generic JWS's using that library and the private keys from storage 
* network/dag/signing.go can network documents using that generic JWS signing using the rules specified by RFC004

reinkrul:
> I can't fully decipher how the document signing and parsing works, but it looks like there's a layer too many. JWS headers that are copied are a symptom for that.

It was a bit off, in my opinion especially because I returned the algorithm along with the JWS as string. That's fixed now (by just returning a new instance of the signed document, by parsing it). How it's structured;

* lestrrat/jws is used to actually sign JWS's
* Crypto signs generic JWS's using that library and the private keys from storage 
* network/dag/signing.go can network documents using that generic JWS signing using the rules specified by RFC004

woutslakhorst (approved):
Comment on dag/document.go is still valid but outside this PR.

PR #33, opened 2021-02-01
Replay network subscriptions on startup

### LINKED ISSUE #8: Improve network pub-sub
Applications (like the registry) can subscribe to specific document types on the network. Then the network engine will call the receiver func. when a document is received. The current implementation is very naive however;

- It's not persistent, meaning if the server crashes halfway some documents may not be received by the subscriber
- It's not ordered, so subscribers will receive documents out-of-order at some point
- It's not retryable, so error'd documents will never be retried by the application.

This needs a better solution.

reinkrul:
Since the current DIDStore is an in-memory implementation, the DID Documents stored in Nuts Network must be loaded into it (through the subscribers) every time the node is started. This actually makes things much simpler: subscriptions don't have to be persistent, meaning that we can replay the DAG every time a new subscription is added.

This PR alters the DAG subscription handling such that it uses an altered version of the Bread-First DAG walker, one which keeps track of which documents couldn't be processed at that point: e.g. when the payload is missing, meaning we can't call the subscribers for that particular document AND all subsequent documents on that branch. When we receive payload we can call the walker again, resuming at the documents we previously stopped.

Fixes #8


network/dag/publisher.go
```diff
@@ -0,0 +1,89 @@
...
+
+type replayingDAGPublisher struct {
+	subscribers  map[string]Receiver
+	algo         *bfsWalkerAlgorithm
+	payloadStore PayloadStore
+	dag          DAG
+}
+
+func (s *replayingDAGPublisher) PayloadWritten(_ interface{}) {
+	s.publish(hash.EmptyHash())
```
woutslakhorst, line 29: why the empty hash?

woutslakhorst (changes requested):
We also need an issue or conversation to keep track of the architecture of dagDocument to VDR/VCR

network/dag/publisher.go
reinkrul: Because when the node receives payload, we'd have to look up in the `payload hash -> document` index where we should resume (and check whether that's a resumation point) but that's not accessible from this type (it's an internal of the BBolt DAG implementation). So I chose to just resume from the resumation points regardless whether we can actually go forward with publishing to subscribers (it can only resume when the payload's there)

When we receive payload chances are pretty big that its from a resumation point, especially when your node is in sync with other nodes. We could optimize this later if it turns out to be a bottleneck.

reinkrul:
> We also need an issue or conversation to keep track of the architecture of dagDocument to VDR/VCR

I'll draw something up, made an issue just to be sure;

https://github.com/nuts-foundation/nuts-node/issues/47

woutslakhorst (approved):
Could you ignore/approve issues in codeclimate before merging?

PR #50, opened 2021-02-08
Improved Instance/Engine naming and instantation

reinkrul:
- `NetworkEngine` instance (which wasn't the actual engine) renamed to `Transactions`
- Made instance/network instantiation explicit in root command
- Removed last singletons and configOnce's
- Added previously missing tests


woutslakhorst (approved):
Ahhhh, less code 👍

PR #51, opened 2021-02-08
Refactor engine into interfaces

### LINKED ISSUE #29: Refactor Engine struct to interface
The Engine struct could be refactored to an interface which can be composed of the existing `Runnable`, `Configurable` and `Diagnosable` interfaces.

woutslakhorst:
Closes #29


woutslakhorst:
Increase in code is largely due to the extra interfaces, which results in more funcs and mocks

woutslakhorst:
and lots of files changed due to running gofmt/goimport

reinkrul (changes requested):
Although I understand why things are the way the are, what I don't like;

- The 'types' package per engine since packages should describe its functional purpose (we don't make a `util` or `consts` package) rather than what technicalities it contains
- Now the `Crypto`, `VDR` and `Network` structs are very much tied to `pflags` and `cobra` since the structs now contains functions which return them. It feels like we're mixing layers here (core functionality v.s. auxillary functionality to load the engine's configuration).

Any ideas?

reinkrul (approved):

PR #52, opened 2021-02-08
Separate CLI client config from server config

### LINKED ISSUE #45: Separate CLI config from server config
We now have 1 type of config (core/engine) which is loaded for both CLI and server, which creates ambiguity. This should be separated; the CLI should have specific flags (e.g. remote server address, time-outs) which are specific for the invocation of CLI and the server config shouldn't include these CLI-specific flags.

reinkrul:
Fixes #45 

> We now have 1 type of config (core/engine) which is loaded for both CLI and server, which creates ambiguity. This should be separated; the CLI should have specific flags (e.g. remote server address, time-outs) which are specific for the invocation of CLI and the server config shouldn't include these CLI-specific flags.

Changes:

- `NutsConfig` renamed to `ServerConfig` which is only loaded when running the `server` command
- Introduced `ClientConfig`, which is only loaded when running client commands
- Introduced global `timeout` config option for `ClientConfig`, replacing the `vdr.clientTimeout` config option
- Moved normalization (prepend address with `http://` if not present)  of remote server address (client config) from VDR to `ClientConfig`
- Moved server config loading to `server` command, so it doesn't get loaded when running client commands


woutslakhorst:
hmmm lots of conflicts with #51

stevenvegt (commented):
Can you add the generation op the options rst to the makefile?

core/config.go
```diff
@@ -1,158 +1,33 @@
-/*
```
stevenvegt, line 1: Why are these copyright messages removed?

reinkrul:
> Can you add the generation op the options rst to the makefile?

They already are, that's unchanged (only expanded)

core/config.go
reinkrul: Oops, renamed file than made a new one with the same name. Added them.

crypto/cmd/cmd.go
```diff
@@ -106,17 +104,18 @@ func Cmd() *cobra.Command {
 			cmd.Println(publicKeyAsPEM)
 		},
 	})
-
+	cmd.PersistentFlags().AddFlagSet(core.NewClientConfig().FlagSet())
```
woutslakhorst, line 107: Only the crypto command adds the flagset at this point. The vdr and network do not add it here. Why the difference?

core/client_config.go
```diff
@@ -0,0 +1,70 @@
...
+	}
+}
+
+// Load loads the client config from environment variables and commandline params.
+func (cfg *ClientConfig) Load() error {
+	return loadConfigIntoStruct(cfg.FlagSet(), cfg, koanf.New(defaultDelimiter))
+}
+
+// FlagSet returns the flags for configuring the client config.
+func (cfg *ClientConfig) FlagSet() *pflag.FlagSet {
```
woutslakhorst, line 56: doesn't need to be member func

README.rst
```diff
@@ -159,3 +156,15 @@ the options table must be regenerated using the Makefile:
...
+CLI options
+***********
+
+The following options can be supplied when running CLI commands:
+
+=======  ==============  ===================================================================================================================================
+Key      Default         Description
+=======  ==============  ===================================================================================================================================
+address  localhost:1323  Address of the remote node. Must contain at least host and port, URL scheme may be omitted. In that case it 'http://' is prepended.
+timeout  10s             Client time-out when performing remote operations.
```
woutslakhorst, line 168: my guess it uses unix style time strings, like 10s, 1m, etc? Would be good to add to description

woutslakhorst (changes requested):

crypto/cmd/cmd.go
reinkrul: good spot, leftover from reverted refactor

woutslakhorst (approved):

Issue #40, opened 2021-02-02
Consistent logging


woutslakhorst:

woutslakhorst:
- [ ] check if all engines create a logger
- [ ] set log level from config
- [ ] setup logging defaults, when and where to do debug, info, error, warn

woutslakhorst:
Proposal for logging:

## Error log
Use error logging at the 'edge' of the application: when an API returns a 500, at startup or when a asynchronous task fails and puts something in an dl-queue (eg: network receiving a transaction and not being able to process it). This will make sure each error is only logged once. Error logging should require immediate attention. System level events like out-of-memory also result in Error logging (and usual a crash of the application).

## Warn log
Use warn logging when an API returns a 40x code. Warn logging could also be used at places where 'undesired' behaviour is detected or when deprecated functionality is used. 
Warn log should act as input to change user behaviour or change the way connecting software works.

## Info log
Info logging may be used in the 'service' layer and API layer indicating an action has been completed with the result (like an identifier).

## Debug log
Debug logging may be used in the API and service layer when an action is started. It may also display the given parameters for the action. In general this would be the case for public methods.

## Trace logging
Trace logging may be added at each level of the application but only at the beginning of a function. Trace logging may also be added to private methods.

reinkrul:
Can we include not using `Fatal`, since that panics the application?


imo a panic should never be used where you can just return an error and let the call tree decide whether to exit the application or not.

woutslakhorst:
I saw a blog post somewhere where custom errors were used throughout the app. The custom errors contained a msg to the user but also a technical msg and trace for the administrator. We could use something like that to differentiate between return errors and logged errors.

reinkrul:
> I saw a blog post somewhere where custom errors were used throughout the app. The custom errors contained a msg to the user but also a technical msg and trace for the administrator. We could use something like that to differentiate between return errors and logged errors.

I always find that quite fiddly, because it depends on the interface through which the call is made, how technical the user is. So imo you can only decide on what to return to the caller in the interface layer. For instance; what message would you present to the user when an array index is out of bounds, or a piece of JSON can't be parsed? (answer: it depends...)

woutslakhorst:
When errors are replaced by another error: a file not found to a app specific error. That could be a good place to add some logging since the original message is gone. If it's unintended it's an `error` if it's user error it's a `warn`

reinkrul:
Stuff that's left to fix:

[...]

PR #54, opened 2021-02-09
Better authentication support for DID Documents

stevenvegt:
* Update keystore with new keys in an updated docuent.
* Revoke keys when keys are removed from the document
* Support a chain of controllers
* Add integration test for VDR
* Implement VDR.Deactivate


reinkrul:
Do you want to fix this one as well in this PR; https://github.com/nuts-foundation/nuts-node/issues/96

?

stevenvegt:
I'm not realy addressing services in this PR. Mostly relations and keys between docments.

vdr/doc-updater.go
```diff
@@ -0,0 +1,111 @@
...
+
+		if ecPKey.Curve == nil {
+			return "", errors.New("could not generate kid: empty key curve")
+		}
+
+		jwKey, err := jwk.New(pKey)
+		if err != nil {
+			return "", err
+		}
+		thumbprint, err := jwKey.Thumbprint(thumbprintAlg)
```
woutslakhorst, line 39: the func for creating DIDs uses `KeyID()` instead of thumbprint?

vdr/integration_test.go
```diff
@@ -0,0 +1,191 @@
...
+	"github.com/nuts-foundation/go-did"
+	log "github.com/sirupsen/logrus"
+	"github.com/stretchr/testify/assert"
+
+	"github.com/nuts-foundation/nuts-node/core"
+	crypto "github.com/nuts-foundation/nuts-node/crypto"
+	"github.com/nuts-foundation/nuts-node/network"
+)
+
+func TestVDRIntegration_Test(t *testing.T) {
```
woutslakhorst, line 19: This test is way too long and complicated. It feels like you're not trusting the unit tests for the intermediate results. What is the goal of this test? Test every step or make sure everything is integrated?

woutslakhorst: It also looks like multiple things are tested which can be broken up.

vdr/vdr.go
```diff
@@ -163,3 +171,50 @@ func (r *VDR) Deactivate(DID did.DID, current hash.SHA256Hash) {
 	logging.Log().Debugf("Deactivating DID Document: %s", DID)
 	panic("implement me")
 }
+
+func (r *VDR) resolveControllers(doc did.Document) ([]did.Document, error) {
+	return r._resolveControllers([]did.Document{doc})
+}
+
+func (r *VDR) _resolveControllers(input []did.Document) ([]did.Document, error) {
```
woutslakhorst, line 179: why the construction with a hidden private method that accepts multiple docs and one private that accepts a single one?

vdr/doc-updater.go
```diff
@@ -0,0 +1,111 @@
...
+		}
+
+		existingDID.Fragment = base64.RawURLEncoding.EncodeToString(thumbprint)
+
+		return existingDID.String(), nil
+	}
+}
+
+// AddNewAuthenticationMethodToDIDDocument creates a new VerificationMethod of type JsonWebKey2020 with a freshly generated key
+func (u NutsDocUpdater) AddNewAuthenticationMethodToDIDDocument(doc *did.Document) error {
```
woutslakhorst, line 51: Where is this used?

woutslakhorst (changes requested):

vdr/doc-updater.go
stevenvegt: Under water it's the same. Fixed it to a uniform way.

vdr/integration_test.go
stevenvegt: Its indeed an integrationtest. So instead of using mocks this test hits the complete stack. Its the lifetime of 2 documents. One controller of the other. It already caucht a few weird exceptions. Let me know what you think What do you mean with complicated? My goal was to make it readeable and logic to follow.

reinkrul: I don't think "syntactic sugar" (the newly introduced `NutsDocUpdater`) deserves a separate integration test. Since it just operates on the data structure and the VDR funcs, it can't do anything to that VDR can't do to cause a bug in the underlying network lib imo.

vdr/vdr.go
```diff
@@ -147,8 +147,16 @@ func (r VDR) Update(id did.DID, current hash.SHA256Hash, next did.Document, _ *t
...
+	if err != nil {
+		return fmt.Errorf("error while finding controllers for document: %w", err)
+	}
+	if len(controllers) == 0 {
+		return fmt.Errorf("could not find any contollers for document")
+	}
+
 	// TODO: look into the controller of the did for a signing key
-	keyID := currentDIDdocument.Authentication[0].ID.String()
+	keyID := controllers[0].Authentication[0].ID.String()
```
reinkrul, line 159: always the first?

vdr/vdr.go
reinkrul, line 179: I don't like stuff in Golang starting with an underscore. Feels a bit like "this is even more private than the one without the udnerscore, but Go doesn't support it". I'd say collaps the two and just call the one with the array. You can even make it vararg (`...`) so it works for both (recursive and initial call).

reinkrul (changes requested):

vdr/integration_test.go
woutslakhorst: over a 100 lines. It covers at least 2 different cases that could be split up in 2 tests?

vdr/vdr.go
stevenvegt: This is a good question. How to choose one? I think we should or just pick one, or specify which controller to use.

vdr/integration_test.go
stevenvegt: > I don't think NutsDocUpdater deserves a separate integration test.
Its an integrationtest for the whole VDR, not just the NutsDocUpdater.

vdr/doc-updater.go
stevenvegt: We should move this to another abstraction layer. Added references to #123

vdr/doc-updater.go
```diff
@@ -0,0 +1,111 @@
...
+	"errors"
+
+	"github.com/lestrrat-go/jwx/jwk"
+	"github.com/nuts-foundation/go-did"
+
+	nutsCrypto "github.com/nuts-foundation/nuts-node/crypto"
+)
+
+// NutsDocUpdater contains helper methods to update a nuts document.
+type NutsDocUpdater struct {
```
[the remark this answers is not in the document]
stevenvegt: See what you mean. But I also don't like a bunch of function littering the package namspace.

vdr/ambassador.go
```diff
@@ -213,6 +217,29 @@ func (n *ambassador) handleUpdateDIDDocument(document dag.SubscriberDocument, pr
...
+			return fmt.Errorf("could not parse public key from verificationMethod: %w", err)
+		}
+		if err := n.keyStore.AddPublicKey(vm.ID.String(), pKey, signingTime); err != nil {
+			return fmt.Errorf("unable to add new public key: %w", err)
+		}
+
+	}
+	for _, vm := range removedVMs {
+		if err := n.keyStore.RevokePublicKey(vm.ID.String(), signingTime); err != nil {
+			return fmt.Errorf("unable to revoke public key: %w", err)
```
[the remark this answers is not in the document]
stevenvegt: If that is the case, the key would be still in use and not in the removed list right?

vdr/doc-creator.go
```diff
@@ -67,56 +67,35 @@ func didKIDNamingFunc(pKey crypto.PublicKey) (string, error) {
...
-// jwkToVerificationMethod takes a jwk.Key and converts it to a DID VerificationMethod
-func keyToVerificationMethod(key crypto.PublicKey, keyID string) (*did.VerificationMethod, error) {
-	// make use of the jwk helper functions
-	publicKeyJWK, err := jwk.New(key)
-	if err != nil {
-		return nil, err
-	}
-	err = publicKeyJWK.Set(jwk.KeyIDKey, keyID)
+	// Build the AuthenticationMethod and add it to the document
+	verificationMethod, err := did.NewVerificationMethod(*keyID, did.JsonWebKey2020, did.DID{}, key)
```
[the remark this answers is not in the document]
stevenvegt: Lets fix that in the go-did lib than.

vdr/doc-updater.go
```diff
@@ -0,0 +1,125 @@
...
+// FIXME:This method is a bit too high level and should be moved as part of this issue:
+// https://github.com/nuts-foundation/nuts-node/issues/123
+func (u NutsDocUpdater) CreateNewAuthenticationMethodForDocument(doc *did.Document) error {
+	key, keyIDStr, err := u.keyCreator.New(newNamingFnForExistingDID(doc.ID))
+	if err != nil {
+		return err
+	}
+	keyID, err := did.ParseDID(keyIDStr)
+	if err != nil {
+		return err
```
reinkrul, line 77: this seems untestable to me because it can't occur in practice? Also a bit strange, we go from a DID (`doc.ID`) to a string, to a DID (`keyID`) which is actually a URI

vdr/doc-updater.go
```diff
@@ -0,0 +1,125 @@
...
+	return nil
+}
+
+// getVerificationMethodDiff is a helper function that makes a diff of verificationMethods between
+// a provided current and a proposedDocument. It returns a list with new and removed verificationMethods
+func getVerificationMethodDiff(currentDocument, proposedDocument did.Document) (new, removed []*did.VerificationMethod) {
+	for _, vmp := range proposedDocument.VerificationMethod {
+		for _, mpc := range currentDocument.VerificationMethod {
+			if vmp.ID.Equals(mpc.ID) {
+				goto nextProposedMethod
```
reinkrul, line 93: Although I used `goto` as well (in BFSWalker), I find this hard to read. Especially because the label which is jumped to is at the end of the for loop. Wouldn't it be better to use `found` bool in combination with a break/continue instead?

vdr/vdr.go
reinkrul: Or just return an error when there's more ("this is not implemented yet") so we don't accidentally use the wrong one.

I always prefer erroring out over silently doing the wrong thing.

vdr/vdr.go
```diff
@@ -132,32 +132,98 @@ func (r VDR) Resolve(id did.DID, metadata *types.ResolveMetadata) (*did.Document
...
+func (r *VDR) Deactivate(id did.DID, currentHash hash.SHA256Hash) error {
+	// A deactivated document is the original document stripped from a controller and keys.
+	// So, apart from the services, it is practically an empty document. Should we even keep the services?
+	emptyDoc := did.Document{ID: id}
+	return r.Update(id, currentHash, emptyDoc, nil)
+}
+
+// resolveControllers accepts a list of documents and finds their controllers
+// The resulting list are documents who control themselves
+func (r *VDR) resolveControllers(input []did.Document) ([]did.Document, error) {
```
reinkrul, line 183: This function is recursive so it supports second-, third-, Nth-generation controllers (A controls B controls C -> C controls A as well). Our specification isn't really clear on this point, just that "controllers can control the subject as well". Can you make the RFC more clear whether we do/don't support that delegated delegation (for lack of a better term)?

reinkrul, line 183: I found this function complex, so I ran its unit tests (with test coverage) to see whether it's completely tested. The unit test is missing a bit of branch coverage i.m.o., which is easy to test (since it's a simple algorithmic function with little dependencies).

vdr/vdr.go
```diff
@@ -132,32 +132,98 @@ func (r VDR) Resolve(id did.DID, metadata *types.ResolveMetadata) (*did.Document
...
+				// doc is its own controller
+				leaves = append(leaves, doc)
+			} else {
+				// add did to be resolved later
+				refsToResolve = append(refsToResolve, ctrlDID)
+			}
+		}
+	}
+	// resolve all unresolved docs
+	// TODO: check for loops or already resolved docs?
```
reinkrul, line 211: these are 2 TODO's;

- check for already resolved docs is simple to do, just loop the input array?
- check for loops is important; and should be clear from the spec/RFC

reinkrul (changes requested):

vdr/vdr.go
stevenvegt: https://github.com/nuts-foundation/nuts-specification/issues/39

vdr/ambassador.go
```diff
@@ -220,48 +224,75 @@ func (n *ambassador) handleUpdateDIDDocument(document dag.SubscriberTransaction,
...
+func (n ambassador) updateKeysInStore(currentDIDDocument, proposedDIDDocument did.Document, signingTime time.Time) error {
+	newVMs, removedVMs := getVerificationMethodDiff(currentDIDDocument, proposedDIDDocument)
+	for _, vm := range newVMs {
+		pKey, err := vm.PublicKey()
+		if err != nil {
+			return fmt.Errorf("could not parse public key from verificationMethod: %w", err)
+		}
+		logging.Log().Debugf("adding new public key: %s", vm.ID.String())
+		if err := n.keyStore.AddPublicKey(vm.ID.String(), pKey, signingTime); err != nil {
+			logging.Log().WithError(err).WithFields(map[string]interface{}{"keyID": vm.ID.String()}).Warn("could not add new public key")
```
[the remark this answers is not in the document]
woutslakhorst: see #40, for now I'd suggest to return the error.

vdr/ambassador.go
```diff
@@ -220,48 +224,75 @@ func (n *ambassador) handleUpdateDIDDocument(document dag.SubscriberTransaction,
...
 	// PayloadHash must be set
-	if document.PayloadHash().Empty() {
+	if transaction.PayloadHash().Empty() {
 		return fmt.Errorf("payloadHash must be provided")
 	}
 
 	// Signing time should be set and lay in the past:
 	// allow for 2 seconds clock skew
-	if document.SigningTime().IsZero() || document.SigningTime().After(time.Now().Add(2*time.Second)) {
+	if transaction.SigningTime().IsZero() || transaction.SigningTime().After(time.Now().Add(2*time.Second)) {
```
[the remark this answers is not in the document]
woutslakhorst: It's an issue: nuts-specification/issues#27 so a reference to that will be enough

vdr/doc-updater.go
```diff
@@ -0,0 +1,125 @@
...
+		return err
+	}
+	return nil
+}
+
+// CreateNewAuthenticationMethodForDocument creates a new VerificationMethod of type JsonWebKey2020 with a freshly generated key
+// and adds it to the provided document
+// FIXME:This method is a bit too high level and should be moved as part of this issue:
+// https://github.com/nuts-foundation/nuts-node/issues/123
+func (u NutsDocUpdater) CreateNewAuthenticationMethodForDocument(doc *did.Document) error {
```
woutslakhorst, line 70: better remove this entirely since a regular key rotation requires overlap of the keys for some period of time. Otherwise there'll be race conditions. Or it must just add a new Key not remove the old one.

vdr/vdr.go
```diff
@@ -132,32 +132,98 @@ func (r VDR) Resolve(id did.DID, metadata *types.ResolveMetadata) (*did.Document
...
 		Hash:             &current,
-		AllowDeactivated: false,
+		AllowDeactivated: true,
 	}
-	currentDIDdocument, meta, err := r.store.Resolve(id, resolverMetada)
+	currentDIDDocument, meta, err := r.store.Resolve(id, resolverMetadata)
 	if err != nil {
 		return err
 	}
+	if isDeactivated(currentDIDDocument) {
```
woutslakhorst, line 143: I'm a but puzzled why you chose to add additional functionality to the existing PR? This can easily go in a separate PR (even more since the discussion has just started)

woutslakhorst (changes requested):
The deactivation part really should not have been added to the existing PR.

vdr/vdr.go
stevenvegt: https://github.com/nuts-foundation/nuts-specification/issues/39

vdr/vdr.go
stevenvegt: Added the issue of recursive controllers to: https://github.com/nuts-foundation/nuts-specification/issues/39

vdr/doc-updater.go
stevenvegt: I removed the Rotation method. Kept the CreateKey since its quite a low level operation on the document.

vdr/doc-updater.go
```diff
@@ -0,0 +1,114 @@
...
+	// keyCreator is used for getting a fresh key and use it to generate the Nuts DID
+	keyCreator nutsCrypto.KeyCreator
+}
+
+// newNamingFnForExistingDID returns a KIDNamingFunc that can be used as param in the keyCreator.New function.
+// It wraps the KIDNamingFunc with the context of the DID of the document.
+// It returns a keyID in the form of the documents DID with the new keys thumbprint as fragment.
+func newNamingFnForExistingDID(existingDID did.DID) nutsCrypto.KIDNamingFunc {
+	return func(pKey crypto.PublicKey) (string, error) {
+		ecPKey, ok := pKey.(*ecdsa.PublicKey)
```
woutslakhorst, line 25: in general we still support multiple key types. the `jwk.New` from line 34 does these checks for you.
The two check on ecdsa keys can be removed.

woutslakhorst (changes requested):
1 small thingy: redundant test on specific key, see comment.

vdr/doc-updater.go
stevenvegt: Removed them and added extra test for empty key 👍

woutslakhorst (approved):

reinkrul (approved):

Issue #14, opened 2021-01-25
Decide whether we want HTTP bind addresses per engine or just use the global config


reinkrul:
Now each engine defines a HTTP bind address (or remote server address when running the client). The global config also defines the bind address. Since we don't plan to start just some modules per node (fine-grained microservices) but use the all-or-nothing approach, there's (imo) no need to specify a bind address per engine.

woutslakhorst:
Unless we want to bind internal API to a different interface. But that should be future work I think.

reinkrul:
In that case these alternative ports are determined by the core/global config anyways, not per engine I think. So we get a map with API categories -> ports;

`
binds:
  public: 0.0.0.0:1323
  admin: localhost:9999
  internal: nuts-node.lan:2000
`

or something

woutslakhorst:
This issue only focusses on using a single `address` config for each engine!

reinkrul:
Already fixed.

PR #56, opened 2021-02-09
Allow API calls to be bound to specific HTTP interfaces

### LINKED ISSUE #34: Bind different API security levels to different network interfaces
see also in discussion of #14

reinkrul:
Fixes #34


core/server_config.go
```diff
@@ -47,21 +47,40 @@ const loggerLevelFlag = "verbosity"
...
+	HTTPConfig `koanf:"default"`
+	// AltBinds contains binds for alternative HTTP interfaces. The key of the map is the first part of the path
+	// of the URL (e.g. `/internal/some-api` -> `internal`), the value is the HTTP interface it must be bound to.
+	AltBinds map[string]HTTPConfig `koanf:"alt"`
+}
+
+// HTTPConfig contains configuration for an HTTP interface, e.g. address.
+// It will probably contain security related properties in the future (TLS configuration, user/pwd requirements).
+type HTTPConfig struct {
+	// Address holds the interface address the HTTP service must be bound to, in the format of `host:port` (e.g. localhost:5555).
```
woutslakhorst, line 69: maybe also change `host:port` to `interface:port`

core/echo.go
```diff
@@ -0,0 +1,154 @@
...
+	group := getGroup(path)
+	groupAddress := c.groups[group]
+	if groupAddress != "" {
+		registerFn(c.interfaces[groupAddress])
+	} else {
+		registerFn(c.interfaces[c.groups[defaultEchoGroup]])
+	}
+}
+
+func (c MultiEcho) CONNECT(path string, h echo.HandlerFunc, m ...echo.MiddlewareFunc) *echo.Route {
```
woutslakhorst, line 83: or we could try to change oapicodegen with a PR, since https://github.com/labstack/echo/blob/master/echo.go#L401 forwards everything to the `add` method....

core/echo_test.go
```diff
@@ -0,0 +1,117 @@
...
+	})
+}
+
+func Test_MultiEcho_Routes(t *testing.T) {
+	t.Run("right HTTP methods are mapped", func(t *testing.T) {
+		ctrl := gomock.NewController(t)
+		defer ctrl.Finish()
+
+		s := NewMockEchoServer(ctrl)
+		s.EXPECT().PATCH("patch", nil)
```
woutslakhorst, line 29: you may want to split these up since switching 2 calls will still make this test succeed.

cmd/root_test.go
```diff
@@ -77,3 +125,49 @@ func Test_CreateSystem(t *testing.T) {
 	})
 	assert.Equal(t, 5, numEngines)
 }
+
+type stubEchoServer struct {
```
woutslakhorst, line 129: can you put it under `test/http`?

woutslakhorst (changes requested):
run `gofmt`

al those router methods are the result of the oapi-codegen tool. Echo itself uses `Add`, maybe we can change oapi-codegen?

core/echo.go
reinkrul: There's a template file which you can use to regenerate the generation code, but that would mean we'd have t fork it. I don't think this code is too bad as long as we test it?

reinkrul: It turned out the template is pretty small and we can simply override it. So I chose that route.

woutslakhorst (approved):
much cleaner this way

woutslakhorst:
closes #14

PR #61, opened 2021-02-11
Add method for deep copy of an entry

### LINKED ISSUE #16: Memory VDR store should return a deep copy of the DID document and metadata
Instead of pointer to the internal struct. Now you can alter the DID Document without going through `Update()`.

Suggestion: provide `Copy()` func on `DIDDocument`

gerardsn:
Fixes #16


vdr/store/memory.go
```diff
@@ -109,26 +109,41 @@ func (m *memory) Resolve(id did.DID, metadata *vdr.ResolveMetadata) (*did.Docume
 	}
 
 	// return a deep copy
-	doc := &did.Document{}
-	docMetadata := &vdr.DocumentMetadata{}
+	copyEntry, err := deepCopy(entry)
```
reinkrul, line 112: not sure if we want to have deepCopy() return an error? A memory copy should alway be able to succeed imo. I believe @woutslakhorst checked why/when marshal/unmarshal can fail

vdr/store/memory.go
```diff
@@ -109,26 +109,41 @@ func (m *memory) Resolve(id did.DID, metadata *vdr.ResolveMetadata) (*did.Docume
...
+	if err != nil {
+		return nil, nil, err
+	}
+
+	return &copyEntry.document, &copyEntry.metadata, nil
+}
 
+// deepCopy returns a deep copy of a memoryEntry
+func deepCopy(entry *memoryEntry) (*memoryEntry, error) {
```
reinkrul, line 121: I thought it was going to be a function in `go-did`?

reinkrul (changes requested):

vdr/store/memory.go
woutslakhorst: go-did can do a proper memory based deepCopy. For now it's fixed here. An issue still has to be made on go-did.

vdr/store/memory.go
woutslakhorst: when implemented in go-did yes. Now it's just on par with what it was but a bit more separated.

woutslakhorst:
Failing tests on circleCI!

go.mod
```diff
@@ -11,7 +11,7 @@ require (
 	github.com/labstack/echo/v4 v4.1.17
 	github.com/lestrrat-go/jwx v1.0.7
 	github.com/magiconair/properties v1.8.4
-	github.com/nuts-foundation/go-did v0.0.0-20210125091337-21da6ad69a82
```
stevenvegt, line 14: Why the update of go-did?

woutslakhorst (approved):

PR #63, opened 2021-02-12
Fixed starting multiple Echo servers

reinkrul:
`Echo.Start()` is blocking, so only the first (default) echo server was started. Now it's using goroutines.

### DIFF
--- a/docs/pages/production-configuration.rst
+++ b/docs/pages/production-configuration.rst
@@ -47,5 +47,5 @@ secure subnets. This is done through the `http` configuration:
         public:
           address: nuts.vendor.nl:443
         # The following binds all endpoints starting with `/status` to all interfaces on `:80`
-        public:
+        status:
           address: :80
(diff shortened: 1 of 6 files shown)


docs/pages/production-configuration.rst
```diff
@@ -46,3 +46,6 @@ secure subnets. This is done through the `http` configuration:
         # The following binds all endpoints starting with `/public` to `nuts.vendor.nl:443`
         public:
           address: nuts.vendor.nl:443
+        # The following binds all endpoints starting with `/status` to all interfaces on `:80`
```
woutslakhorst, line 49: err, this conflicts with previous lines?

woutslakhorst (changes requested):

docs/pages/production-configuration.rst
reinkrul: True, example should've been `status:`

woutslakhorst (approved):

Issue #66, opened 2021-02-15
Add generic VerifiableCredential support


woutslakhorst:
This is a tracking issue

- [x] implement concept mapping structure #69 
- [x] add NutsOrganizationCredential #100 
- [x] update documentation
- [x] search API #131 
- [x] implement issuing #100
- [x] implement revocation #130 
- [x] implement resolving #125
- [x] connect to network #127 
- [x] keep in line with RFC011 and RFC012
- [x] embed templates using go1.16 #100
- [x] add config params for trusted issuers #149 
- [x] CLI #173

PR #69, opened 2021-02-16
Add support for Verifiable Credentials

woutslakhorst:
part of #66


docs/_static/credential/v1.yaml
```diff
@@ -0,0 +1,246 @@
...
+            text/plain:
+              schema:
+                type: string
+        "500":
+          description: An error occurred while processing the request.
+          content:
+            text/plain:
+              schema:
+                type: string
+  /internal/credential/v1/{concept}:
```
reinkrul, line 108: Would it be viable to URL-encode it (`v1/{concept}?q={query}`)? Because then we can just make it a GET query, as expected for read operations for REST-y services

docs/_static/credential/v1.yaml
```diff
@@ -0,0 +1,246 @@
...
+          description: An error occurred while processing the request.
+          content:
+            text/plain:
+              schema:
+                type: string
+components:
+  schemas:
+    CredentialSubject:
+      type: object
+      description: Contents of a Verifiable Credential giving name and DID of organization.
```
reinkrul, line 160: I think this description makes assumptions of the VC subject regarding a specific concept/VC type, so I'd say something like "Subject of a Verifiable Credential identifying the holder and expressing claims"

docs/_static/credential/v1.yaml
```diff
@@ -0,0 +1,246 @@
...
+      type: object
+      description: used search params
+      required:
+        - key
+        - value
+      properties:
+        key:
+          description: >
+            Fields from VCs to search on. Concept specific keys must be prepended with the concept name and a '.'.
+            Default fields: issuer, subject, type
```
reinkrul, line 171: Default fields don't need to be prepended with concept names? Maybe clear that up little bit

docs/_static/credential/v1.yaml
```diff
@@ -0,0 +1,246 @@
...
+      description: A credential according to the W3C spec.
+      required:
+        - context
+        - type
+        - issuer
+        - issuanceDate
+        - credentialSubject
+        - proof
+      properties:
+        context:
```
reinkrul, line 206: context must be `@context`?

docs/_static/credential/v1.yaml
```diff
@@ -0,0 +1,246 @@
...
+            $ref: '#/components/schemas/KeyValuePair'
+        offset:
+          description: skips first x results, default 0
+          type: number
+        limit:
+          description: limit number of return values to x, default 10
+          type: number
+    VerifiableCredential:
+      type: object
+      description: A credential according to the W3C spec.
```
reinkrul, line 197: and our own, ofc

docs/_static/credential/v1.yaml
```diff
@@ -0,0 +1,246 @@
+openapi: "3.0.0"
+info:
+  title: Nuts Verifiable Credential API spec
+  description: API specification for Verifiable Credentials
+  version: 1.0.0
+  license:
+    name: GPLv3
+paths:
```
reinkrul, line 8: I think we're missing a call for verifying VCs?

Or do we only return verified VCs? If so maybe we should add a verificationDate to the VerificationOutcome and return it with the VC.

docs/_static/credential/v1.yaml
```diff
@@ -0,0 +1,246 @@
...
+          $ref: '#/components/schemas/CredentialSubject'
+        proof:
+          description: one or multiple cryptographic proofs
+          type: object
+    VerificationOutcome:
+      required:
+        - status
+      properties:
+        status:
+          description: if the credential is valid, revoked, untrusted or has some other status.
```
reinkrul, line 242: what is mapped to true and what to false? Not clear from the docs

docs/_static/credential/v1.yaml
```diff
@@ -0,0 +1,246 @@
...
+          type: boolean
+        statusReason:
```
reinkrul, line 244: maybe rename it to `invalidReason` or something, because `statusReason` it might contain a value for when the status is valid.

credential/concept/util.go
```diff
@@ -0,0 +1,37 @@
...
+import "strings"
+
+func max(x int, y int) int {
+	if x > y {
+		return x
+	}
+	return y
+}
+
+func path(currentPath string, additional string) string {
```
reinkrul, line 31: maybe `joinPath` so it somewhat conforms to stdlib's `path.Join`?

credential/config.go
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *
+ */
+
+package credential
+
+const (
+	moduleName = "Verifiable credential Store"
```
reinkrul, line 23: Credential

reinkrul, line 23: I like the similarity with 'VDR' (altough that should be acronymed to "Verifiable DID Registry" imo).

However if this is the module name, should we rename just the engine/type to 'VCS' instead of calling it "credential"? Now every time you name a variable "credential" (which is often, I'd expect) it collides with the package name. And I'd expect a `Credential` type to be, well, a credential. Not something that reads and writes credentials (because what would it be returning, credentials?)

credential/config.go
```diff
@@ -0,0 +1,33 @@
...
+	configKey  = "credential"
```
reinkrul, line 24: vcr?

credential/credential.go
```diff
@@ -0,0 +1,163 @@
...
+	return c.registry
+}
+
+func (c *credential) Reader() VCReader {
+	return c
+}
+
+func (c *credential) Configure(config core.ServerConfig) error {
+	var err error
+	fsPath := path.Join(config.Datadir, "credentials", "credentials.db")
```
reinkrul, line 58: if we rename `credential` to `vcr` (or vcs) I'd name the directory `vcr` as well rather than `credentials`, because that's how the other engines do it. The database file name is fine though since it really stores the credentials.

credential/config.go
reinkrul, line 23: I think it does more than storing, also issuing and revoking. Maybe call it "Verifiable Credential Registry"?

credential/concept/types.go
```diff
@@ -0,0 +1,53 @@
...
+import (
+	"errors"
+	"strings"
+)
+
+// ErrUnknownConcept is returned when an unknown concept is requested
+var ErrUnknownConcept = errors.New("unknown concept")
+
+// ErrNoType is returned when a template is loaded which doesn't have a type
+var ErrNoType = errors.New("no type found")
```
reinkrul, line 31: "template has no type"? (I'd add "template" to the message)

credential/concept/template.go
```diff
@@ -0,0 +1,319 @@
...
+
+func (cs templateString) isHardCoded() bool {
+	return !strings.Contains(string(cs), "<<")
+}
+
+func (cs templateString) hasIndex() bool {
+	return strings.Contains(string(cs), "@")
+}
+
+var regex, _ = regexp.Compile("((<<[a-zA-Z\\.]+>>)|([a-zA-Z\\.]+))(@({[1-9](_[1-9])?})(,{[1-9](_[1-9])?})*)?")
```
reinkrul, line 274: give the var a more descriptive name than just `regex`, which almost collides with the package name (what does it match?)

credential/concept/template.go
```diff
@@ -0,0 +1,319 @@
...
+// SubjectField defines the concept JSON path to a VC subject
+const SubjectField = "subject"
+
+// Template represents a json mapping template. In the template concept to json path mappings are stored.
+// Indices that need to be created by a DB are also created by the template.
+type Template interface {
+	// Indices returns a list of compound indices. Each entry contains the concept json path.
+	Indices() [][]string
+	// VCType returns the VC type.
+	VCType() string
```
reinkrul, line 52: Do we need to bind 'concepts' to VCs in code, or can we let it be generic ('Object'/'Entity'?)? Then we could use it for our DID store as well?

credential/concept/template.go
```diff
@@ -0,0 +1,319 @@
...
+	// key: value pairs that always need to be added to the backend query
+	fixedValues map[string]string
+}
+
+// templateString represents a value in a template.
+// it contains several convenience methods for parsing purposes
+type templateString string
+
+// ToVCPath returns the mapping of concept path to VC path
+func (ct *template) ToVCPath(conceptPath string) string {
```
reinkrul, line 74: unnecessary use of pointer

credential/concept/template.go
```diff
@@ -0,0 +1,319 @@
...
+	return ct.conceptIndexMapping[conceptPath]
+}
+
+func (ct *template) Indices() [][]string {
```
reinkrul, line 78: unnecessary use of pointer

credential/concept/template.go
```diff
@@ -0,0 +1,319 @@
...
+	return ct.indices
+}
+
+// VCType returns the VC type.
+func (ct *template) VCType() string {
```
reinkrul, line 83: unnecessary use of pointer

credential/concept/template.go
```diff
@@ -0,0 +1,319 @@
...
+
+type template struct {
+	raw string
+	// conceptIndexMapping contains mappings from concept name to VC JSON path
+	conceptIndexMapping map[string]string
+	// reverseIndexMapping contains mappings from VC JSON path to concept JSON path
+	reverseIndexMapping map[string]string
+	// indices contains a set of compound indices
+	indices [][]string
+	// key: value pairs that always need to be added to the backend query
```
reinkrul, line 65: incorrect godoc format

credential/concept/template.go
```diff
@@ -0,0 +1,319 @@
...
+	// uniq concepts
+	var cs = make([]string, 0)
+	for k := range cm {
+		cs = append(cs, k)
+	}
+
+	return cs
+}
+
+func (ct *template) mapR(jq *gojsonq.JSONQ, currentPath string) error {
```
reinkrul, line 133: what's `R`? Can we expand it for readability?

credential/concept/template.go
```diff
@@ -0,0 +1,319 @@
...
+	if jq.Error() != nil {
+		return jq.Error()
+	}
+
+	// slice
+	if _, ok := val.([]interface{}); ok {
+		return errors.New("json arrays are not supported")
+	}
+
+	if m, ok := val.(map[string]interface{}); ok {
```
reinkrul, line 145: if the above is called slice, this should be called map?

credential/concept/template.go
```diff
@@ -0,0 +1,319 @@
...
+		for k, v := range m {
+			nextPath := path(currentPath, k)
+
+			if sv, ok := v.(string); ok {
```
reinkrul, line 149: if this function is recursive, shouldn't this be at top level of the func, next to `slice` and `map`? Maybe it makes testing and reading easier (and lower complexity due to less nested control flows)

credential/concept/template.go
```diff
@@ -0,0 +1,319 @@
...
+}
+
+func (ct *template) processValue(value string, currentPath string) error {
+	processValue := templateString(value)
+
+	if !processValue.isValid() {
+		return fmt.Errorf("values has incorrect format at %s: %s", currentPath, value)
+	}
+
+	if processValue.isHardCoded() {
```
reinkrul, line 172: I think this is also called 'fixedValues'? Let's decide on 1 term

credential/concept/registry.go
```diff
@@ -0,0 +1,166 @@
...
+	"github.com/nuts-foundation/go-did"
+)
+
+// Registry defines the interface for accessing loaded concepts and using the templates
+// to generate queries and transform results.
+type Registry interface {
+	// ConceptTemplates returns a mapping of concept names to parsed templates.
+	ConceptTemplates() map[string][]Template
+	// Load template files from disk and parse them.
+	LoadTemplates() error
```
reinkrul, line 32: can't we do this in the `NewRegistry` function? Since it implies a specific implementation detail (concepts must be loaded, probably from disk/URL) and that ideally shouldn't be in a interface describing the contract for a `Registry` imo

credential/concept/registry.go
```diff
@@ -0,0 +1,166 @@
...
+	// Adds a template from a string source
+	AddFromString(concept string) error
```
reinkrul, line 34: can't we simply do `Add(tpl Template) error`?

Parsing a template should be imo a responsibility of Template, not of Registry

credential/concept/registry.go
```diff
@@ -0,0 +1,166 @@
...
+	// QueryFor creates a query for the given concept.
+	// The query is preloaded with required fixed values like the type.
```
reinkrul, line 36: 'Loading' a data type (preloading) sounds a bit strange here
`The query contains fixed values required for the concept`

credential/concept/query.go
```diff
@@ -0,0 +1,124 @@
...
+func (tq *TemplateQuery) Template() Template {
+	return tq.template
+}
+
+// VCType returns the VC type.
+func (tq *TemplateQuery) VCType() string {
+	return tq.template.VCType()
+}
+
+// Criteria abstracts different equality clauses, comparable to '=', '!=', 'between' and 'abc%' in SQL.
```
reinkrul, line 87: not completely true, since it's all mapped to `leiaquery.And`..

credential/concept/query.go
```diff
@@ -0,0 +1,124 @@
...
+type Criteria interface {
+	// Key returns the key to match against.
+	Key() string
+	// Seek returns the first matching value for this Criteria or "" if not applicable.
+	Seek() string
```
reinkrul, line 92: I had to look into `go-leia` to understand what's happening here. I feel like the `leia` query model leaks a bit into `credential` here? Maybe `Criteria` should be a struct containing these als regular fields, and `Eq` just returns an instance of that.

I consider it a bit like the `Matcher` interface.

credential/concept/query.go
```diff
@@ -0,0 +1,124 @@
...
+func (q query) Concept() string {
+	return q.concept
+}
+
+func (q query) Parts() []*TemplateQuery {
+	return q.parts
+}
+
+// AddTemplate adds a template to this query. It'll create a new TemplateQuery preconfigured with the template's hard coded values.
+func (q *query) AddTemplate(template *template) {
```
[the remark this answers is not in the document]
woutslakhorst: There's a template interface now, so it must have been an oversight.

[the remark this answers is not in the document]
woutslakhorst: actually it can be private.

credential/concept/query.go
```diff
@@ -0,0 +1,124 @@
...
+	return tq.template
+}
+
+// VCType returns the VC type.
+func (tq *TemplateQuery) VCType() string {
+	return tq.template.VCType()
+}
+
+// Criteria abstracts different equality clauses, comparable to '=', '!=', 'between' and 'abc%' in SQL.
+type Criteria interface {
```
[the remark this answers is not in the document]
woutslakhorst: or `Constraint`?

credential/interface.go
```diff
@@ -0,0 +1,73 @@
...
+	// Resolve returns a credential based on its ID. Returns an error when not found.
+	Resolve(ID string) (did.VerifiableCredential, error)
+	// Verify checks if a credential is valid and trusted at the given time.
+	Verify(vc did.VerifiableCredential, credentialSubject interface{}, at time.Time) (bool, error)
+	// Registry returns the concept registry
+	Registry() concept.Registry
+}
+
+// Credential is the interface that covers all functionality of the credential store.
+type Credential interface {
```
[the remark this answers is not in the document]
woutslakhorst: After some though (200ms), I'll merge all those interfaces since both the API and the network will need to read and write.

credential/interface.go
```diff
@@ -0,0 +1,73 @@
...
+}
+
+// VCReader allows for searching through public and private credentials, resolving and verifying VCs.
+type VCReader interface {
+	// Search for matching credentials based upon a query. It returns an empty list if no matches have been found.
+	Search(query concept.Query) ([]did.VerifiableCredential, error)
+	// Resolve returns a credential based on its ID. Returns an error when not found.
+	Resolve(ID string) (did.VerifiableCredential, error)
+	// Verify checks if a credential is valid and trusted at the given time.
+	Verify(vc did.VerifiableCredential, credentialSubject interface{}, at time.Time) (bool, error)
```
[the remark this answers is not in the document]
woutslakhorst: I'll add a todo, since Verify is not implemented yet and thus is also likely to change.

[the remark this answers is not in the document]
woutslakhorst: As of now, it's not implemented and not used. Better to split it when needed.

credential/credential.go
```diff
@@ -0,0 +1,163 @@
...
+	// init indices
+	if err = c.initIndices(); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (c *credential) initIndices() error {
+	for _, templates := range c.registry.ConceptTemplates() {
```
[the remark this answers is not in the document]
woutslakhorst: some parts can be abstracted to a more `store` like interface. This is surely the case when supporting more than 1 DB type. But in this PR I felt that showing how everything works is more important given the number of lines it would take to add the abstraction.

docs/_static/credential/v1.yaml
```diff
@@ -0,0 +1,246 @@
+openapi: "3.0.0"
+info:
+  title: Nuts Verifiable Credential API spec
+  description: API specification for Verifiable Credentials
+  version: 1.0.0
+  license:
+    name: GPLv3
+paths:
+  /internal/credential/v1/vc:
```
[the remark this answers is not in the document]
woutslakhorst: Will be added when implemented.

docs/_static/credential/v1.yaml
```diff
@@ -0,0 +1,246 @@
...
+      operationId: "revoke"
+      tags:
+        - credential
+      responses:
+        "202":
+          description: Revocation has been accepted.
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/VerificationOutcome"
```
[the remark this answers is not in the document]
woutslakhorst: renamed

docs/_static/credential/v1.yaml
woutslakhorst: It's subject to change when the API is implemented but the problem is that each concept has different query params. The idea is to accept the concept JSON as search body and create a query on given concept paths:

```
{
  "issuer": "X",
  "company": {
    "name": "query"
  }
}
```

docs/_static/credential/v1.yaml
```diff
@@ -0,0 +1,246 @@
...
+          - "did:nuts:B8PUHs2AUHbFF1xLLK4eZjgErEcMXHxs68FteY7NDtCY#90382475609238467"
+        schema:
+          type: string
+    post:
+      summary: "Revoke a credential"
+      operationId: "revoke"
+      tags:
+        - credential
+      responses:
+        "202":
```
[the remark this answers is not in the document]
woutslakhorst: It was in line with the "publish to network first" idea. Now I also changed the description so a revocation will already be processed locally before published.

docs/_static/credential/v1.yaml
woutslakhorst: Yes it's more generic now

docs/_static/credential/v1.yaml
woutslakhorst: I'm not sure if we need it. We'll see that is required when we integrate the node.

docs/_static/credential/v1.yaml
woutslakhorst: changed to string. This also related to RFC011

docs/_static/credential/v1.yaml
woutslakhorst: removed it for now since delete returns 201

credential/config.go
woutslakhorst: good suggestion

credential/config.go
woutslakhorst: `vcs`

credential/config.go
woutslakhorst: named it video2000 which was better than vhs......

credential/concept/template.go
woutslakhorst: Aren't DID documents more static than VCs?

credential/concept/template.go
woutslakhorst: This was a result of the json parsing. In the end I was able to remove the json library altogether since parsing to a `map[string]interface{}` works equally well. This has the side effect of placing the string check at the top level.

credential/concept/registry.go
woutslakhorst: Placed in in the `NewRegistry` function for now. It'll probably have to move to a sort of configure function in the future. Added a todo for that.

credential/concept/registry.go
woutslakhorst: The query params used are only valid within a certain concept. It's up to the DB query implementation to transform those to DB compatible queries. Therefor a query is created in the context of a concept.

credential/concept/registry.go
woutslakhorst: refactored it

woutslakhorst: You now need to do a `ParseTemplate(raw string)` and then Add it

credential/concept/query.go
woutslakhorst: for now yes, but when another DB implementation is added....

credential/concept/query.go
woutslakhorst: It totally depends on the DB used. It wouldn't surprise me if we will have to create these structs per DB type.
I'll add a notation for that at the top.

woutslakhorst:
@reinkrul there are some questions from me on some of the comments

woutslakhorst:
rebase required force push

credential/concept/query.go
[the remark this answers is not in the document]
reinkrul: Constraint feels like something that causes the process to fail or halt when it's not met, so not sure about that. Maybe 'Clause'?

vcr/interface.go
```diff
@@ -0,0 +1,68 @@
...
+
+import (
+	"errors"
+	"time"
+
+	"github.com/nuts-foundation/go-did"
+	"github.com/nuts-foundation/nuts-node/vcr/concept"
+)
+
+// ErrInvalidIssuer is returned when a vcr is issued by a DID that is unknown or when the private key is missing.
```
reinkrul, line 30: VCR -> VC (same goes for the message)

vcr/interface.go
```diff
@@ -0,0 +1,68 @@
...
+var ErrInvalidIssuer = errors.New("invalid vcr issuer")
+
+// ErrInvalidSubject is returned when a vcr is issued to a DID that is unknown or revoked.
```
reinkrul, line 33: VCR -> VC (same goes for the message)

vcr/interface.go
```diff
@@ -0,0 +1,68 @@
...
+var ErrInvalidSubject = errors.New("invalid vcr subject")
+
+// ErrNotFound is returned when a vcr can not be found based on its ID.
```
reinkrul, line 36: VCR -> VC (same goes for the message)

vcr/interface.go
```diff
@@ -0,0 +1,68 @@
...
+var ErrNotFound = errors.New("vcr not found")
+
+// Issuer can issue credentials for DIDs to DIDs.
+type Issuer interface {
+	// Issue creates and publishes a new vcr.
```
reinkrul, line 41: VC

vcr/interface.go
```diff
@@ -0,0 +1,68 @@
...
+}
+
+// VCR is the interface that covers all functionality of the vcr store.
+type VCR interface {
+	// Search for matching credentials based upon a query. It returns an empty list if no matches have been found.
+	Search(query concept.Query) ([]did.VerifiableCredential, error)
+	// Resolve returns a vcr based on its ID. Returns an error when not found.
+	// todo: not implemented yet and subject to change
+	Resolve(ID string) (did.VerifiableCredential, error)
+	// Verify checks if a vcr is valid and trusted at the given time.
```
reinkrul, line 60: VC

vcr/interface.go
```diff
@@ -0,0 +1,68 @@
...
+type Writer interface {
+	// Write writes a VC to storage.
+	Write(vc did.VerifiableCredential) error
+}
+
+// VCR is the interface that covers all functionality of the vcr store.
+type VCR interface {
+	// Search for matching credentials based upon a query. It returns an empty list if no matches have been found.
+	Search(query concept.Query) ([]did.VerifiableCredential, error)
+	// Resolve returns a vcr based on its ID. Returns an error when not found.
```
reinkrul, line 57: VC

vcr/test.go
```diff
@@ -0,0 +1,35 @@
...
+package vcr
+
+import (
+	"github.com/nuts-foundation/nuts-node/core"
+	"github.com/sirupsen/logrus"
+)
+
+// NewTestCredentialInstance returns a new vcr instance to be used for integration tests. Any data is stored in the
+// specified test directory.
+func NewTestCredentialInstance(testDirectory string) *vcr {
```
reinkrul, line 29: `NewTestVCR(Instance)`

reinkrul (changes requested):

credential/concept/query.go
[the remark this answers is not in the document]
woutslakhorst: `Clause` could work

vcr/interface.go
woutslakhorst: nasty find/replace

reinkrul (approved):

PR #70, opened 2021-02-16
No longer save pubKey in KeyCreator.New

stevenvegt:
Now the key should be added to the PublicKeyStore on later use e.g. when a valid network document is processed.

This has a few advantages:
* valid from period is connected to a purpose: the did document or whatever the key is used for
* The public key stores between the nodes are more equal since they reflect the accepted documents.


vdr/ambassador.go
```diff
@@ -223,6 +233,16 @@ func checkSubscriberDocumentIntegrity(document dag.SubscriberDocument) error {
...
+	// either a signingKey or a keyID must be set, not both, at least one.
+	if (document.SigningKeyID() == "" && document.SigningKey() == nil) || (document.SigningKeyID() != "" && document.SigningKey() != nil) {
+		return fmt.Errorf("either signingKey or keyID must be set")
+	}
+
+	// if the signingKey is set, it must contain a valid keyID
+	if document.SigningKey() != nil && document.SigningKey().KeyID() == "" {
+		return fmt.Errorf("the signingKey must contain a valid keyID")
+	}
+
```
woutslakhorst, line 245: Aren't these already checked on the network lvl?

woutslakhorst (changes requested):
Merge #72 first then rebase, now different stuff is in 1 PR

vdr/ambassador.go
stevenvegt: I couln't find it. @reinkrul can you point us in the right direction?

stevenvegt:
I merged #72 and rebased 👍

vdr/ambassador.go
stevenvegt: Found them in the dag/parser.  Added the keyID check.

woutslakhorst (approved):

PR #81, opened 2021-02-23
Added logs to VDR and flags to print only the document or metadata

reinkrul:
Pointed out while fixing end-to-end tests.

### DIFF
--- a/vdr/cmd/cmd.go
+++ b/vdr/cmd/cmd.go
@@ -65,10 +65,8 @@ func createCmd() *cobra.Command {
 			if err != nil {
 				return fmt.Errorf("unable to create new DID: %v", err)
 			}
-
 			bytes, _ := json.MarshalIndent(doc, "", "  ")
-
-			cmd.Printf("Created DID document: %s\n", string(bytes))
+			cmd.Println(string(bytes))
 			return nil
 		},
 	}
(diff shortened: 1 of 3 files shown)


vdr/cmd/cmd.go
```diff
@@ -127,14 +127,28 @@ func resolveCmd() *cobra.Command {
...
+			}
+			for _, o := range toPrint {
 				bytes, _ := json.MarshalIndent(o, "", "  ")
 				cmd.Printf("%s\n", string(bytes))
 			}
 
 			return nil
 		},
 	}
+	result.Flags().BoolVar(&printMetadata, "metadata", false, "Pass 'true' to print the metadata.")
```
woutslakhorst, line 149: I believe a boolean flag just requires the flag (value is optional)

woutslakhorst: defaults are incorrect. If both are false they will print....

woutslakhorst (changes requested):
- regenerate options.rst

vdr/cmd/cmd.go
reinkrul: > I believe a boolean flag just requires the flag (value is optional)

The var (pointer to `printMetadata`) is optional (you can also use `GetBool`).

> defaults are incorrect. If both are false they will print....

You're correct. I choose it this way because if you don't supply the flags the default (sane) behaviour it will just print both. If you specify one of them, it will only print that one. If you specify both, both are printed. I tried to describe this in the usage but stuff only got complicated, so I guess this is self-explanatory enough. Maybe not to the developer reading this code, but surely to the administrator.

reinkrul:
> * regenerate options.rst

There's no change to the options because it's a flag of a specific subcommand. Only top-level CLI option flags are included in the client options table (e.g. remote server address or client time-out).

vdr/cmd/cmd.go
woutslakhorst: I meant if you do `cmd --printMetadata` is equal to `cmd --printMetadata true`.
Maybe change the description to: **pass to only print X**

PR #82, opened 2021-02-23
Fix SSL offloading support for Nuts Network

reinkrul:
While fixing the end-to-end test for the new implementation, found out that SSL offloading (for incoming Nuts Network connections) was broken.

### DIFF
--- a/docs/pages/production-configuration.rst
+++ b/docs/pages/production-configuration.rst
@@ -1,4 +1,4 @@
-.. _prodruction-configuration:
+.. _production-configuration:
 
 Configuring for Production
 ##########################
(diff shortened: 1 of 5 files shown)


woutslakhorst (commented):
don't you need the truststore for the client certificate?

reinkrul:
> don't you need the truststore for the client certificate?

Nope, since it's only used for validating incoming TLS connections. For outgoing connections using the client certificate: the server appearantly sends enough chain (the truststore) that the client can select the right certificate to use. Might depend on the PKI used though, but I'd say it'll work fine with PKIoverheid.

woutslakhorst (approved):
After merge, I'll configure my own node to use nginx with ssl offloading...

reinkrul:
> After merge, I'll configure my own node to use nginx with ssl offloading...

We also have an end-to-end test that tests this, but it's always good to have an actual node running the topology 👍

PR #85, opened 2021-02-23
Rename Network Document to Transaction according to RFC004

reinkrul:


network/dag/bbolt_dag.go
```diff
@@ -94,7 +94,11 @@ func (d dataSizeStatistic) String() string {
 
 // NewBBoltDAG creates a etcd/bbolt backed DAG using the given database.
 func NewBBoltDAG(db *bbolt.DB) DAG {
-	return &bboltDAG{db: db}
+	result := &bboltDAG{db: db}
```
woutslakhorst, line 97: this is an additional feature in a rename branch?

woutslakhorst (changes requested):
Rename is OK, but the additional feature (migrate) needs to be in a separate PR

network/dag/bbolt_dag.go
reinkrul: Will rename the bucket in a separate PR

woutslakhorst (approved):

PR #86, opened 2021-02-23
Added uptime diagnostic to status engine

reinkrul:
Nice to have

### DIFF
--- a/core/status.go
+++ b/core/status.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"net/http"
 	"strings"
+	"time"
 
 	"github.com/labstack/echo/v4"
 )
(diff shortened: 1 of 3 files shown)


core/status.go
```diff
@@ -23,20 +23,23 @@ import (
...
+	system    *System
+	startTime time.Time
 }
 
 //NewStatusEngine creates a new Engine for viewing all engines
 func NewStatusEngine(system *System) Engine {
 	return &status{
-		system: system,
+		system:    system,
+		startTime: time.Now(),
```
woutslakhorst, line 42: I just read something about Monotonic clocks. It seems that the golang Time does both correct! 😱

woutslakhorst (approved):
Could also be a metric, but that would be quite boring

core/status.go
reinkrul: Nice, didnt realize that's a thing. Question is however, what 'correct' is; if your goal is to display a time relative to the absolute "world time" (e.g. signing time for a document), I'd use the wall clock. For this use case (server up for XXX seconds) I'd use the monotonic clock.

reinkrul:
> Could also be a metric, but that would be quite boring

Just found out it sort of already is;

```
# HELP process_start_time_seconds Start time of the process since unix epoch in seconds.
# TYPE process_start_time_seconds gauge
process_start_time_seconds 1.61407818789e+
```

Not as human-readable though.

PR #89, opened 2021-02-24
Update make instructions

gerardsn:
- Added `make install-tools` to install Go tools (should probably be updated to install specific version on Go >= 1.16)
- Moved python dependencies to requirements.txt and fixed the versions
- Updated instructions for generating code and docs


woutslakhorst (approved):
👍

reinkrul (approved):

PR #97, opened 2021-02-26
Added more unit tests for auth

reinkrul:


auth/services/irma/irmaconfig.go
```diff
@@ -26,98 +26,51 @@ var serverOnce = new(sync.Once)
...
+			SchemesPath:          validatorConfig.IrmaConfigPath,
+			DisableSchemesUpdate: !validatorConfig.AutoUpdateIrmaSchemas,
 		}
 
-		logging.Log().Info("Initializing IRMA library...")
-		logging.Log().Infof("irma baseurl: %s", config.URL)
+		logging.Log().Infof("Initializing IRMA library (baseURL=%s)...", config.URL)
 
 		irmaServer, err = irmaserver.New(config)
-		if err != nil {
```
woutslakhorst, line 85: the `err` is defined within this block

woutslakhorst (changes requested):
err is being ignored

auth/services/irma/irmaconfig.go
reinkrul: Good spotting. IntelliJ didn't notice. That's why named return values are so fiddly in Go. Fixed it.

woutslakhorst (approved):

PR #98, opened 2021-02-26
fix global flags for server command

### LINKED ISSUE #87: top level config flags broken
all top level flags for the server command seem to be broken. Only the config file works.

woutslakhorst:
fixes #87 

Flags need to be added to the cmd before the cmd is run so the help cmd knows about the flags.


woutslakhorst:
tests fail due to race condition. I fixed those in #99, so that one can be merged first.

reinkrul (approved):

PR #106, opened 2021-03-02
Discover network peers using DID service

### LINKED ISSUE #105: VDR/Network: discover new peers from DID service

reinkrul:
Fixes #105

### DIFF
--- a/core/echo_mock.go
+++ b/core/echo_mock.go
@@ -5,35 +5,36 @@
 package core
 
 import (
+	reflect "reflect"
+
 	gomock "github.com/golang/mock/gomock"
 	echo "github.com/labstack/echo/v4"
-	reflect "reflect"
 )
 
-// MockEchoServer is a mock of EchoServer interface
+// MockEchoServer is a mock of EchoServer interface.
 type MockEchoServer struct {
 	ctrl     *gomock.Controller
 	recorder *MockEchoServerMockRecorder
 }
 
-// MockEchoServerMockRecorder is the mock recorder for MockEchoServer
+// MockEchoServerMockRecorder is the mock recorder for MockEchoServer.
 type MockEchoServerMockRecorder struct {
 	mock *MockEchoServer
 }
 
-// NewMockEchoServer creates a new mock instance
+// NewMockEchoServer creates a new mock instance.
 func NewMockEchoServer(ctrl *gomock.Controller) *MockEchoServer {
 	mock := &MockEchoServer{ctrl: ctrl}
 	mock.recorder = &MockEchoServerMockRecorder{mock}
 	return mock
 }
 
-// EXPECT returns an object that allows the caller to indicate expected use
+// EXPECT returns an object that allows the caller to indicate expected use.
 func (m *MockEchoServer) EXPECT() *MockEchoServerMockRecorder {
 	return m.recorder
 }
 
-// Add mocks base method
+// Add mocks base method.
 func (m *MockEchoServer) Add(method, path string, handler echo.HandlerFunc, middleware ...echo.MiddlewareFunc) *echo.Route {
 	m.ctrl.T.Helper()
 	varargs := []interface{}{method, path, handler}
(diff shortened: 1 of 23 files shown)


core/echo_mock.go
```diff
@@ -5,35 +5,36 @@
...
 import (
+	reflect "reflect"
+
 	gomock "github.com/golang/mock/gomock"
 	echo "github.com/labstack/echo/v4"
-	reflect "reflect"
 )
 
-// MockEchoServer is a mock of EchoServer interface
+// MockEchoServer is a mock of EchoServer interface.
```
woutslakhorst, line 14: did you install the latest tools using the makefile? If so, what version is your mockgen on?

docs/pages/did-administration-cookbook.rst
```diff
@@ -0,0 +1,18 @@
+.. _did-administration-cookbook:
+
+DID Document Administration Cookbook
```
woutslakhorst, line 3: this page won't be added to the index.

network/interface.go
```diff
@@ -41,4 +41,7 @@ type Transactions interface {
 	CreateDocument(payloadType string, payload []byte, signingKeyID string, attachKey crypto2.PublicKey, timestamp time.Time, fieldsOpts ...dag.FieldOpt) (dag.Document, error)
 	// ListDocuments returns all documents known to this Network instance.
 	ListDocuments() ([]dag.Document, error)
+	// AddPeer instructs the P2P layer to try to connect to a new peer on the given address. It's safe to call it
+	// multiple times for the same address. If the P2P layer will attempt to connect to the address it returns `true`.
+	AddPeer(address string) bool
```
woutslakhorst, line 46: Is there also a remove peer? eg: when an endpoint is removed...

vdr/ambassador.go
```diff
@@ -335,3 +341,20 @@ func (n ambassador) findKeyByThumbprint(thumbPrint []byte, didDocumentAuthKeys [
 	}
 	return documentKey, nil
 }
+
+// discoverPeers tries to find Nuts Network services in the given DID Document to add them as Nuts Network peers.
+func (n ambassador) discoverPeers(document did.Document, _ types.DocumentMetadata) error {
+	for _, service := range document.Service {
+		if service.Type == "nuts-network-grpc" {
```
woutslakhorst, line 348: constant

woutslakhorst: our docs should reflect this feature and enable admins to change the endpoint. If part of future development, it must be stated in an issue.

vdr/types/common.go
```diff
@@ -70,3 +71,18 @@ type ResolveMetadata struct {
...
 	AllowDeactivated bool
 }
+
+// DocWriterDecorator wraps a DocWriter with a function that's called before the target writer is called.
+type DocWriterDecorator struct {
+	Target    DocWriter
+	Decorator func(document did.Document, metadata DocumentMetadata) error
+}
+
+func (d DocWriterDecorator) Write(document did.Document, metadata DocumentMetadata) error {
```
woutslakhorst, line 81: If the decorator fails, nothing is written. If the docWriter fails, connection will still be made....
The ordering and validity is very important and must be put in our specifications first.
If the decorator pattern is valid, depends on the specs.

woutslakhorst (changes requested):
This PR feels very quick and dirty. It's part of a bigger design including:

- service specification in DIDs, what happens if the endpoint is incorrect for the given type. This is mainly part of a bolt, but in this case it's not.
- what if endpoints are removed?
- what if nodes are down (and will never come back)
- documentation
- helper APIs? Since manipulating endpoints is quite hard right now.

An epic issue can be used to track the entire progress.

reinkrul:
> This PR feels very quick and dirty. It's part of a bigger design including:
> 
> * service specification in DIDs, what happens if the endpoint is incorrect for the given type. This is mainly part of a bolt, but in this case it's not.
> * what if endpoints are removed?
> * what if nodes are down (and will never come back)
> * documentation
> * helper APIs? Since manipulating endpoints is quite hard right now.
> 
> An epic issue can be used to track the entire progress.

I tried to spec it in https://github.com/nuts-foundation/nuts-specification/pull/34

But: since this I remember discussing this I thought we could just squeeze it in but it has more facets to it than I anticipated. Since we don't really need peer discovery at this point because there aren't that many peers to begin with (which can be configured perfectly fine as bootstrap nodes).

I'll keep the spec PR so we can work on it, but I'll close this one for now (for later work).

PR #110, opened 2021-03-03
VDR: Fixed DID Documents not resolveable after reboot

reinkrul:
Fixes 2 problems:

- VDR subscribed too late on network, when it had already replayed all transactions
- VDR tried to re-add the same public key over and over again on boot, which failed because crypto doesn't allow the same KID to be added twice


vdr/ambassador.go
```diff
@@ -121,8 +122,18 @@ func (n *ambassador) handleCreateDIDDocument(document dag.SubscriberDocument, pr
...
-	err = n.keyStore.AddPublicKey(document.SigningKey().KeyID(), rawKey, document.SigningTime())
-	if err != nil {
+	// Since we use an in-memory DID store, DIDs are re-registered on every startup. However, the keystore is persistent
+	// so we have to check whether we already have the public key to avoiding error conditions.
+	// TODO: this implementation is quite naive since the public key might actually differ or have a different validation
+	// period, which isn't taken into consideration. Should be fixed as part of https://github.com/nuts-foundation/nuts-node/issues/109
+	// since then we don't need to do this check any more.
+	_, err = n.keyStore.GetPublicKey(document.SigningKey().KeyID(), document.SigningTime())
+	if err != nil && errors.Is(err, nutsCrypto.ErrKeyNotFound) {
+		err = n.keyStore.AddPublicKey(document.SigningKey().KeyID(), rawKey, document.SigningTime())
```
stevenvegt, line 132: Could we just ignore an error of type `ErrKeyAlreadyExists`?

vdr/vdr.go
```diff
@@ -81,17 +81,13 @@ func (r *VDR) Config() interface{} {
 	return &r.config
 }
 
-// Start initiates the routines for auto-updating the data
-func (r *VDR) Start() error {
+// Configure configures the VDR engine.
+func (r *VDR) Configure(_ core.ServerConfig) error {
```
stevenvegt, line 85: I underdstand the fix, but I don't like that the network is replaying its events before all subscribers are ready. We should rewrite the subscriber meganism so that each subscription triggers its own replay from a given point.

vdr/vdr.go
```diff
@@ -81,17 +81,13 @@ func (r *VDR) Config() interface{} {
...
+	// Initiate the routines for auto-updating the data.
 	r.networkAmbassador.Start()
```
stevenvegt, line 87: Perhaps rename `Start` to `Configure`

vdr/ambassador.go
reinkrul: Much simpler, good idea!

vdr/vdr.go
reinkrul: I swear I though that was an existing issue to change it to something like NATS streaming, but can't find it.

For now: since it was decided to make the DID Store in-memory, we don't have much choice than just to replay the complete DAG. Also, we don't have other subscribers, so there's no-one requiring replays from a given point (although I see the feature for in the future :-p)

PR #111, opened 2021-03-03
Remove generated files from git

stevenvegt:
Generated files causes a lot of formatting errors and can cause errors when forgetting to regenerate on update. This should be part of the build process.
This PR removes the files and add build steps to the make file.


reinkrul:
Although I see the risk in definitions being updated but code not regenerated, building locally now became a bit weird, because when you pull changes you don't know whether the source of generated code has changed and thus code should be regenerated, e.g.;

```
go build .
```

```
docker build .
```

These failed locally unless I ran `make run-generators` once, and now risk being out-of-date locally when someone else changes definitions since the actual generated code isn't checked in. Running generators every time you build is also not a very good solution because it takes quite a bit of time (I see the `test` Make target doesn't even run all generators..) making local development cycle slower than necessary. 

Maybe we can keep checking it in but have the build server regenerate the code before building, so we make sure it builds/tests the right sources?

reinkrul (changes requested):
.

woutslakhorst:
`make install-tools` with the correct lib versions in `go.mod` is a better choice imho. Currently this isn't working correctly because the mock lib version isn't set.

stevenvegt:
Yes I agree about the discussion part. It started as a POC for me. Just to see what would be needed. The PR was for keeping track of the automated builds.
The motivation came after I got frustrated about codeclimate which kept complaining about fmt errors on generated files. I had encountered problems with different versions of oapi-gen and mockgen in the past. General wisdom says to not check in generated files. Also to make sure your build steps are still valid.

PR #113, opened 2021-03-03
Network: Show peer address in diagnostics

### LINKED ISSUE #107: Network: show IP addresses of peers in diagnostics

reinkrul:
Fixes #107

### DIFF
--- a/network/p2p/impl-peer.go
+++ b/network/p2p/impl-peer.go
@@ -35,9 +35,7 @@ type messageGate interface {
 
 // Peer represents a connected peer
 type peer struct {
-	id PeerID
-	// address holds the remote address of the node we're actually connected to
-	address string
+	Peer
 	// gate is used to send and receive messages
 	gate messageGate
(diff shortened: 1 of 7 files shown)


network/p2p/impl-peer.go
```diff
@@ -35,9 +35,7 @@ type messageGate interface {
 
 // Peer represents a connected peer
 type peer struct {
```
woutslakhorst, line 37: maybe a rename for this `peer` to distinguish it from `Peer`

woutslakhorst (changes requested):
The multiple `peer` structs are a bit confusing.

network/p2p/impl-peer.go
reinkrul: You asked for it :-p I renamed `peer` to `connection` which is a better name for the internal data struct in this layer. 'Peer' is how it's exposed to outside the P2P layer.

woutslakhorst (approved):

Issue #100, opened 2021-03-01
Create NutsOrganizationCredential


woutslakhorst:
via API, CLI and network

include documentation on how to use API/CLI based on configured concepts.

PR #119, opened 2021-03-05
Implement vcr.Verify

### LINKED ISSUE #117: VCR: Verify Credential
part of #66

When a VC is published over the network. It must be validated and we must make sure it's 100% correct.

When valid, it has to be added to the store.

woutslakhorst:
waiting for #100 
closes #117


vcr/test/vc.json
```diff
@@ -0,0 +1,27 @@
+{
+  "context": [
```
reinkrul, line 2: @context?

vcr/vcr.go
```diff
@@ -224,8 +227,62 @@ func (c *vcr) Resolve(ID string) (did.VerifiableCredential, error) {
 	panic("implement me")
 }
 
-func (c *vcr) Verify(vc did.VerifiableCredential, credentialSubject interface{}, at time.Time) (bool, error) {
-	panic("implement me")
+func (c *vcr) Verify(vc did.VerifiableCredential, at time.Time) error {
+	// it must have valid content
+	validator, _ := credential.FindValidatorAndBuilder(vc)
+	if validator == nil {
```
reinkrul, line 233: I think `Issue()` checks both validator and builder for `nil`? Let's decide on one style

vcr/vcr.go
```diff
@@ -224,8 +227,62 @@ func (c *vcr) Resolve(ID string) (did.VerifiableCredential, error) {
...
+	// create correct challenge for verification
+	payload, err := generateCredentialChallenge(vc)
+	if err != nil {
+		return fmt.Errorf("cannot generate challenge: %w", err)
+	}
+
+	// extract proof, can't fail already done in generateCredentialChallenge
+	var proofs = make([]JSONWebSignature2020Proof, 0)
+	_ = vc.UnmarshalProofValue(&proofs)
+	proof := proofs[0]
```
reinkrul, line 250: Could there be multiple? In that case I'd say return error when there's multiple

vcr/vcr.go
```diff
@@ -224,8 +227,62 @@ func (c *vcr) Resolve(ID string) (did.VerifiableCredential, error) {
...
+	}
+
+	// find key
+	pk, err := c.docResolver.ResolveSigningKey(proof.VerificationMethod.String(), &at)
+	if err != nil {
+		return err
+	}
+
+	// the proof must be correct
+	verifier, _ := jws.NewVerifier(jwa.ES256)
```
reinkrul, line 267: theoretically we support other keys in our DIDs than just RSA so should be check that?

vcr/vcr.go
```diff
@@ -224,8 +227,62 @@ func (c *vcr) Resolve(ID string) (did.VerifiableCredential, error) {
...
+	// the jws lib can't do this for us, so we concat hdr with payload for verification
+	challenge := fmt.Sprintf("%s.%s", splittedJws[0], base64.RawURLEncoding.EncodeToString(payload))
+	if err = verifier.Verify([]byte(challenge), sig, pk); err != nil {
+		return err
+	}
+
+	// next check timeRestrictions
+	if vc.IssuanceDate.After(at) {
```
reinkrul, line 275: on or after, I'd think?

vcr/vcr.go
```diff
@@ -224,8 +227,62 @@ func (c *vcr) Resolve(ID string) (did.VerifiableCredential, error) {
...
+		return errors.New("credential not valid yet at given time")
+	}
+	if vc.ExpirationDate != nil && (*vc.ExpirationDate).Before(at) {
+		return errors.New("credential not valid anymore at given time")
+	}
+
+	// check if issuer is trusted
+	// todo requires trusted config
```
reinkrul, line 283: add issue

reinkrul (changes requested):
I'm missing "it is added to the store" part as stated by the related issue?

vcr/test/vc.json
woutslakhorst: fixed by future PR

vcr/vcr.go
woutslakhorst: Only the validator is required here so no need to check the builder.

vcr/vcr.go
woutslakhorst: Covered by the validator.

vcr/vcr.go
woutslakhorst: already covered in #66

vcr/vcr.go
woutslakhorst: I think it should be valid at the same nanosecond. So an inclusive start but exclusive end.

vcr/vcr.go
woutslakhorst: added a `SigningAlgorithm` func to crypto/jwx to determine signing algo based on key.

woutslakhorst:
There's a later PR that connects the VCR to the network. No worries

stevenvegt (approved):

reinkrul (approved):

PR #120, opened 2021-03-05
Add simple docker build github action

stevenvegt:
Add muti arch builds for:
* linux/arm/v7
* linux/arm64
* linux/amd64

Build rules:
* Build every push to master, github tag and PR to master.
* PR builds are not pushed to docker hub.
* Master pushes are tagged with `master`
* Github tags in the form of `v1.2.3`  are tagged with `1.2.3`, `1.2` and `latest`


.github/workflows/build-images.yaml
```diff
@@ -0,0 +1,35 @@
...
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+
+      - name: Build and push
+        uses: docker/build-push-action@v2
+        with:
+          context: .
+          file: Dockerfile
+          platforms: linux/amd64,linux/arm/v7
+          push: true
+          tags: nutsfoundation/nuts-node:latest
```
woutslakhorst, line 35: does this automatically create tags for each architecture? eg: `armv7-latest`

gerardsn: All architectures are tagged under `latest`, see tag `arch-test` on [https://hub.docker.com/r/nutsfoundation/nuts-node/tags](https://hub.docker.com/r/nutsfoundation/nuts-node/tags). 

From [docker docs](https://docs.docker.com/docker-for-mac/multi-arch/): 
> Docker images can support multiple architectures, which means that a single image may contain variants for different architectures, and sometimes for different operating systems, such as Windows.
> 
> When running an image with multi-architecture support, docker will automatically select an image variant which matches your OS and architecture.

gerardsn:
Can we add a build for `linux/arm64`

woutslakhorst (commented):
Docker currently builds the `latest`, but also tagged versions. What is the plan? Mixed builds or migrate all?

stevenvegt:
I added a trigger and tagging for versions and PRs.
Note: the master build is now tagged `master` contrary to the `latest` tag that was used.
Tags are now tagged with ther version number and the latest version is also tagged `latest`.

woutslakhorst (approved):
Don't forget to turn off the github builder.

PR #124, opened 2021-03-08
Fix 121 address param

### LINKED ISSUE #121: CLI param address in client mode gets ignored
Does not work:
```
./nuts --address=192.168.178.1:1323 vdr create-document
```
Does work:
```
NUTS_ADDRESS=192.168.178.1:1323 ./nuts vdr create-document
```

stevenvegt:
fixes #121

### DIFF
--- a/core/client_config.go
+++ b/core/client_config.go
@@ -22,6 +22,7 @@ package core
 import (
 	"github.com/knadh/koanf"
 	"github.com/spf13/pflag"
+	"os"
 	"strings"
 	"time"
 )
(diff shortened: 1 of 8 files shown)


core/client_config.go
```diff
@@ -66,5 +67,6 @@ func ClientConfigFlags() *pflag.FlagSet {
 	flagSet := pflag.NewFlagSet("client", pflag.ContinueOnError)
 	flagSet.String(addressFlag, defaultAddress, "Address of the remote node. Must contain at least host and port, URL scheme may be omitted. In that case it 'http://' is prepended.")
 	flagSet.Duration(clientTimeoutFlag, defaultClientTimeout, "Client time-out when performing remote operations, such as '500ms' or '10s'. Refer to Golang's 'time.Duration' syntax for a more elaborate description of the syntax.")
+	flagSet.Parse(os.Args[1:])
```
woutslakhorst, line 70: This is one of the first things a `cobra.Cmd` does. Is the client flagset added before the command is executed?
The problem with a manual parse is that for Array/Slice types, entries can be duplicated.

stevenvegt: The `ClientConfigFlags()` is called in different places in the application. Se on each call the flagset is recreated. Perhaps that's the problem? Will look further into it.

stevenvegt: Fixed it by explicetly inserting the configFlags from the command which are parsed by cobra.

stevenvegt:
Codeclimate complains about duplicate code, but that is not really an issue since the duplicate sections each create seperate http clients. Code can be combined but that does not add to the maintainability imho.

woutslakhorst (approved):

PR #129, opened 2021-03-09
Updated network gRPC protocol to reflect spec (oneof for messages, re…

reinkrul:
- `oneof` for messages
- removed `header` from messages, will be replaced with a `version` header later on


stevenvegt:
Will this break current messages in the network?

stevenvegt:
So we probably need to to reset the store in the dev network? 
Can we use this change to gain experience on supporting older versions or is backwards compatibilty something for another time?

reinkrul:
> So we probably need to to reset the store in the dev network?
> Can we use this change to gain experience on supporting older versions or is backwards compatibilty something for another time?

No, the RFC004 transaction format is unchanged, it's just the network protocol (RFC005) that breaks. So nodes running the current version won't be able to exchange data with nodes running the version after this PR.

stevenvegt (approved):

PR #135, opened 2021-03-10
VCR: Add revocation

### LINKED ISSUE #130: VCR: revocation
- api
- publish
- process
- delete from store or invalidate in store?

part of #66

woutslakhorst:
Adds the possibility to revoke a VC via the API.
The revocation is also published over the network and stored in a db. This allows for revocation checking.
Closes #130


woutslakhorst:
might be some missing coverage, will fix

PR #139, opened 2021-03-12
VCR: Issue VC API must specify correct request body

reinkrul:


woutslakhorst (dismissed):

woutslakhorst (changes requested):
This is not true. As you can read here: https://nuts-node.readthedocs.io/en/latest/pages/development/vc.html

Creating a VC requires a simplified VC where some parts like ID are not required or even expcted.

stevenvegt:
Better introduce something like the `VerifiableCredentialCreateRequest` with a subset of the `VerifiableCredential` struct?

reinkrul:
@stevenvegt @woutslakhorst problem is mainly that we didn't have a `requestBody` at all, which lead to the client code generated by the OpenAPI generator unable to post a VC to the `create` function (since it didn't expect a body).

As of right now I see 2 options;

- accept a generic unstructured JSON object
- specify a type with only the fields that are relevant

What do you prefer?

woutslakhorst:
I would suggest a minimal body as described by the docs:

```
{
  "issuer": "did:nuts:...",
  "type": "NutsOrganizationCredential",
  "expirationDate": "optional",
  "credentialSubject": object
}
```

stevenvegt:
Looks good but make it a seperate component named something like `VerifiableCredentialCreateRequest`.

stevenvegt (changes requested):
Make it a component so the generator can make it a type.

reinkrul:
> Looks good but make it a seperate component named something like `VerifiableCredentialCreateRequest`.

The OpenAPI code generator made something like that; `CreateJSONRequestBody`. Why make it a separate, resuable component when it's never used in other places?

stevenvegt (approved):

docs/_static/vcr/v1.yaml
```diff
@@ -220,3 +226,27 @@ components:
...
           type: object
+    IssueVCRequest:
+      type: object
+      description: A request for issuing a new Verifiable Credential.
+      required:
+        - type
+        - issuer
+        - credentialSubject
+      properties:
+        type:
```
woutslakhorst, line 237: must be single type, not a list. According to spec we only accept one additional type.

vcr/api/v1/generated.go
```diff
@@ -18,6 +18,22 @@ import (
 	"github.com/labstack/echo/v4"
 )
 
+// IssueVCRequest defines model for IssueVCRequest.
+type IssueVCRequest struct {
```
woutslakhorst, line 22: You can add an exception to the generation and create an alias in `types.go`

woutslakhorst (changes requested):

vcr/vcr.go
```diff
@@ -191,12 +191,22 @@ func (c *vcr) Search(query concept.Query) ([]did.VerifiableCredential, error) {
...
 
-func (c *vcr) Issue(vc did.VerifiableCredential) (*did.VerifiableCredential, error) {
-	validator, builder := credential.FindValidatorAndBuilder(vc)
+func (c *vcr) Issue(template did.VerifiableCredential) (*did.VerifiableCredential, error) {
+	validator, builder := credential.FindValidatorAndBuilder(template)
 	if validator == nil || builder == nil {
 		return nil, errors.New("unknown credential type")
 	}
 
+	if len(template.Type) > 1 {
```
woutslakhorst, line 200: `builder.fill()` does this all for you at line 223

vcr/api/v1/api.go
```diff
@@ -89,13 +88,13 @@ func (w *Wrapper) Revoke(ctx echo.Context, id string) error {
 
 // Create a Verifiable credential
 func (w *Wrapper) Create(ctx echo.Context) error {
-	vc := did.VerifiableCredential{}
+	requestedVC := IssueVCRequest{}
```
woutslakhorst, line 91: Is this change needed since it is an alias?

woutslakhorst (changes requested):

vcr/api/v1/api.go
reinkrul: Nope, but I wanted to use the right type since that's what's defined in the YAML spec

vcr/vcr.go
reinkrul: I don't agree: it adds only adds the `VerifiableCredentialType` if not present. We clearly state "only 1 type is supported" and thus we should validate it, since we chose to reuse the `did.VerifiableCredential` type where `type` is a slice and thus it allows passing more than 1 type.

woutslakhorst (approved):

PR #146, opened 2021-03-15
ignore slices when transforming VCs to concepts

### LINKED ISSUE #145: VCR: Searching for VC yields "json arrays are not supported"
Using the following query:

```golang
response, err := client.Search(context.Background(), "organization", SearchJSONRequestBody{
	Params: []KeyValuePair{
		{
			Key:   "organization.city",
			Value: "Franeker",
		},
	},
})
```

Yields the following error: `json arrays are not supported `. I added the `currentPath` variable and it points out the problem is `@context`.

When I ignore the error (returning nil instead of returning an error), the search works properly.

woutslakhorst:
fixes #145

### DIFF
--- a/vcr/concept/template.go
+++ b/vcr/concept/template.go
@@ -234,7 +234,7 @@ func (ct *Template) transformRecursive(val interface{}, currentPath string, c Co
 
 	// we do not support mapped values within lists
 	if _, ok := val.([]interface{}); ok {
-		return errors.New("json arrays are not supported")
+		return nil
 	}
 
 	if m, ok := val.(map[string]interface{}); ok {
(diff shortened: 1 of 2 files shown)


stevenvegt (commented):
This feels a bit strange. Should it fail sailent? Than the test should probably "nested arrays are ignored" Otherwise perhaps fix this in another location? Who's fault is the nested array? Is it user input? Perhaps sanitise the user`s input?

reinkrul (approved):

woutslakhorst:
> This feels a bit strange. Should it fail sailent? Than the test should probably "nested arrays are ignored" Otherwise perhaps fix this in another location? Who's fault is the nested array? Is it user input? Perhaps sanitise the user`s input?

The error return was a copy-paste from the 'parseRecursive' method. If a concept is parsed, it must signal mapped arrays. But when transforming, it must just ignore them. Transformation only occurs when returning VCs via a search.

PR #150, opened 2021-03-15
Add cli and API DID deactivation

stevenvegt:
Adds the CLI and API `Deactivate` commands.
Adds the deactivated status to the resolvemetadata.

### DIFF
--- a/docs/_static/vdr/v1.yaml
+++ b/docs/_static/vdr/v1.yaml
@@ -109,13 +109,19 @@ paths:
         "200":
           description: DID document has been deactivated.
         "400":
-          description: DID document couldn't be updated because the did param was malformed or the document was already deactivated.
+          description: DID document couldn't be updated because the DID param was malformed.
           content:
             text/plain:
               schema:
                 type: string
         "404":
           description: Corresponding DID document could not be found.
+          content:
+            text/plain:
+              schema:
+                type: string
+        "409":
+          description: DID document couldn't be updated because the the document was already deactivated.
           content:
             text/plain:
               schema:
--- a/docs/_static/vdr/v1.yaml
+++ b/docs/_static/vdr/v1.yaml
@@ -83,7 +83,7 @@ paths:
               schema:
                  $ref: '#/components/schemas/DIDDocument'
         "400":
-          description: DID document couldn't be updated because the did param was malformed or the document was already deactivated.
+          description: DID document couldn't be updated because the did param was malformed or the document was deactivated.
           content:
             text/plain:
               schema:
@@ -109,7 +109,7 @@ paths:
         "200":
           description: DID document has been deactivated.
         "400":
-          description: DID couldn't be deactivated because the input conflicts or is malformed.
+          description: DID document couldn't be updated because the did param was malformed or the document was already deactivated.
           content:
             text/plain:
               schema:
--- a/docs/_static/vdr/v1.yaml
+++ b/docs/_static/vdr/v1.yaml
@@ -100,7 +100,32 @@ paths:
             text/plain:
               schema:
                 type: string
-
+    delete:
+      summary: Deactivatates a Nuts DID Document according to the specification.
+      operationId: "deactivateDID"
+      tags:
+        - DID
+      responses:
+        "200":
+          description: DID Document has been deactivated.
+        "400":
+          description: DID couldn't be deactivated because the input conflicts or is malformed.
+          content:
+            text/plain:
+              schema:
+                type: string
+        "404":
+          description: Corresponding DID document could not be found.
+          content:
+            text/plain:
+              schema:
+                type: string
+        "500":
+          description: An error occurred while processing the request.
+          content:
+            text/plain:
+              schema:
+                type: string
 components:
   schemas:
     DIDDocument:
--- a/docs/_static/vdr/v1.yaml
+++ b/docs/_static/vdr/v1.yaml
@@ -14,7 +14,7 @@ paths:
         - DID
       responses:
         "201":
-          description: "New DID has been created successfully. Returns the DID Document."
+          description: "New DID has been created successfully. Returns the DID document."
           content:
             application/json+did-document:
               schema:
@@ -36,7 +36,7 @@ paths:
         schema:
           type: string
     get:
-      summary: "Resolves a Nuts DID Document"
+      summary: "Resolves a Nuts DID document"
       operationId: "getDID"
       tags:
         - DID
@@ -65,7 +65,7 @@ paths:
               schema:
                 type: string
     put:
-      summary: Updates a Nuts DID Document.
+      summary: Updates a Nuts DID document.
       operationId: "updateDID"
       tags:
         - DID
@@ -77,13 +77,13 @@ paths:
               $ref: '#/components/schemas/DIDUpdateRequest'
       responses:
         "200":
-          description: DID Document has been updated.
+          description: DID document has been updated.
           content:
             application/json+did-document:
               schema:
                  $ref: '#/components/schemas/DIDDocument'
         "400":
-          description: DID couldn't be updated because the input conflicts or is malformed.
+          description: DID document couldn't be updated because the did param was malformed or the document was already deactivated.
           content:
             text/plain:
               schema:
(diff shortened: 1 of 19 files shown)


docs/_static/vdr/v1.yaml
```diff
@@ -100,7 +100,32 @@ paths:
             text/plain:
               schema:
                 type: string
-
+    delete:
+      summary: Deactivatates a Nuts DID Document according to the specification.
```
reinkrul, line 104: I'd almost think you're yoking about my speech disorder.... :-p

docs/_static/vdr/v1.yaml
```diff
@@ -100,7 +100,32 @@ paths:
...
+      operationId: "deactivateDID"
+      tags:
+        - DID
+      responses:
+        "200":
+          description: DID Document has been deactivated.
```
reinkrul, line 110: DID or DID Document?

When there's no content returned, consider returning HTTP `201` (No Content)

docs/_static/vdr/v1.yaml
```diff
@@ -100,7 +100,32 @@ paths:
...
+        "400":
+          description: DID couldn't be deactivated because the input conflicts or is malformed.
```
reinkrul, line 112: conflict = HTTP 409

docs/_static/vdr/v1.yaml
```diff
@@ -100,7 +100,32 @@ paths:
...
+          content:
+            text/plain:
+              schema:
+                type: string
+        "404":
+          description: Corresponding DID document could not be found.
```
reinkrul, line 118: can we decide on DID Document or DID document?

vdr/cmd/cmd.go
```diff
@@ -153,6 +155,24 @@ func resolveCmd() *cobra.Command {
 	return result
 }
 
+func deactivateCmd() *cobra.Command {
+	result := &cobra.Command{
+		Use:   "deactivate [DID]",
+		Short: "Deactivate a DID document based on its DID",
```
reinkrul, line 161: maybe mention this is a destructive operation and it can't be reverted

vdr/cmd/cmd_test.go
```diff
@@ -240,6 +240,47 @@ func TestEngine_Command(t *testing.T) {
...
 		})
 	})
+
+	t.Run("deactivate", func(t *testing.T) {
+		t.Run("ok", func(t *testing.T) {
+			cmd := Cmd()
+			s := httptest.NewServer(http2.Handler{StatusCode: http.StatusOK})
+			os.Setenv("NUTS_ADDRESS", s.URL)
+			defer os.Unsetenv("NUTS_ADDRESS")
+			core.NewServerConfig().Load(cmd)
```
reinkrul, line 250: I think you should be using ClientConfig?

vdr/vdr.go
```diff
@@ -170,11 +170,15 @@ func isDeactivated(document *did.Document) bool {
...
 // Deactivate updates the DID Document so it can no longer be updated
-func (r *VDR) Deactivate(id did.DID, currentHash hash.SHA256Hash) error {
+func (r *VDR) Deactivate(id did.DID) error {
+	_, meta, err := r.store.Resolve(id, &types.ResolveMetadata{AllowDeactivated: true})
+	if err != nil {
+		return err
+	}
 	// A deactivated document is the original document stripped from a controller and keys.
-	// So, apart from the services, it is practically an empty document. Should we even keep the services?
+	// So, apart from the services, it is practically an empty document.
```
reinkrul, line 179: It _is_ an empty document, even the services are removed (comment is incorrect)

reinkrul (approved):

reinkrul (changes requested):
s

docs/_static/vdr/v1.yaml
stevenvegt: 409 is rare and I would not recommend using it.

docs/_static/vdr/v1.yaml
stevenvegt: 201 is created. Perhaps you mean 204? I would like to keep the set of status codes small. 200 seems a legit reply code.

vdr/cmd/cmd.go
stevenvegt: Added a confirmation 👍

docs/_static/vdr/v1.yaml
reinkrul: I disagree but let's agree to disagree...? There's not tens of common HTTP codes for us to be just using 200, 301, 400 and 500...

vdr/cmd/cmd.go
```diff
@@ -153,6 +156,51 @@ func resolveCmd() *cobra.Command {
...
+		Args:  cobra.ExactArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if !askYesNo("This wil delete the did document, are you sure?", cmd) {
```
reinkrul, line 165: did -> DID
wiull

vdr/vdr.go
reinkrul: Comment is still incorrect. The document is updated to an empty document (except ID, ofc). Now I think of it, shouldn't we keep `@context` as well?

reinkrul (changes requested):

docs/_static/vdr/v1.yaml
```diff
@@ -77,13 +77,39 @@ paths:
...
+            text/plain:
+              schema:
+                type: string
+    delete:
+      summary: Deactivates a Nuts DID document according to the specification.
+      operationId: "deactivateDID"
+      tags:
+        - DID
+      responses:
+        "200":
```
woutslakhorst, line 109: Without any return body, a 202 is more suitable.

docs/_static/vdr/v1.yaml
```diff
@@ -77,13 +77,39 @@ paths:
...
         "200":
-          description: DID Document has been updated.
+          description: DID document has been updated.
           content:
             application/json+did-document:
               schema:
                  $ref: '#/components/schemas/DIDDocument'
         "400":
-          description: DID couldn't be updated because the input conflicts or is malformed.
+          description: DID document couldn't be updated because the did param was malformed or the document was deactivated.
```
woutslakhorst, line 86: I agree with Rein that a 409 should be supported. A 409 indicates you should not repeat the request even with other parameters where a 400 suggests that you can repeat is with other parameters.

vdr/cmd/cmd_test.go
```diff
@@ -240,6 +241,72 @@ func TestEngine_Command(t *testing.T) {
...
+	t.Run("deactivate", func(t *testing.T) {
+		t.Run("ok", func(t *testing.T) {
+			cmd := Cmd()
+			s := httptest.NewServer(http2.Handler{StatusCode: http.StatusOK})
+			os.Setenv("NUTS_ADDRESS", s.URL)
+			defer os.Unsetenv("NUTS_ADDRESS")
+			core.NewClientConfig().Load(cmd.Flags())
+			defer s.Close()
+
+			inBuf := new(bytes.Buffer)
```
woutslakhorst, line 254: lots of duplicate code for setting up the buffer. Could be placed in a func?

vdr/types/common.go
```diff
@@ -53,6 +53,8 @@ type DocumentMetadata struct {
 	TimelineID hash.SHA256Hash `json:"timelineID"`
 	// Hash of DID document bytes. Is equal to payloadHash in network layer.
 	Hash hash.SHA256Hash `json:"hash"`
+	// Deactivated indicates if the document is deactivated
+	Deactivated bool `json:"deactivated"`
```
woutslakhorst, line 57: This property is also stored which could make it deviate from the actual state. Maybe the memory store must set/unset this property which each load?

woutslakhorst (commented):
Some minor things that won't hurt if they are not changed.

docs/_static/vdr/v1.yaml
stevenvegt: 204, no content?

docs/_static/vdr/v1.yaml
stevenvegt: Ok, I 200 the 409 👍

vdr/types/common.go
stevenvegt: This is on purpose. The intention was to deactivate the document. If the form of a deactivated document changes in the future, this should not change the state of an already deactivated document.

vdr/cmd/cmd_test.go
stevenvegt: Fixed 👍

reinkrul (approved):

PR #151, opened 2021-03-15
Skip unimplemented VDR ambassador tests so we don't forget about them

reinkrul:


woutslakhorst (commented):
How does it help remember us? If we don't see the lines of code we won't notice.

reinkrul:
> How does it help remember us? If we don't see the lines of code we won't notice.

Fair point, I though `go test` showed them, but it's just IntelliJ that reports skipped tests...

@stevenvegt maybe we can just implement them?

stevenvegt:
I need to check if they are still relevant

PR #154, opened 2021-03-16
Provide CLI commands for querying node status and diagnostics

### LINKED ISSUE #38: 'diagnostics' CLI command doesn't work
Should return the diagnostics of the remote server

reinkrul:
Fixes #38

### DIFF
--- a/core/status/cmd.go
+++ b/core/status/cmd.go
@@ -1,0 +1,61 @@
+/*
+ * Nuts node
+ * Copyright (C) 2021. Nuts community
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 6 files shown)


core/status/cmd.go
```diff
@@ -0,0 +1,61 @@
...
+	cmd := &cobra.Command{
+		Use:   "status",
+		Short: "Shows the status of the Nuts Node.",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			config := core.NewClientConfig()
+			if err := config.Load(cmd.PersistentFlags()); err != nil {
+				return err
+			}
+			var targetURL string
+			if verbose {
```
woutslakhorst, line 42: a bit overkill since a status call will only return 200 OK and should never have a return body of any significance

woutslakhorst (changes requested):
I think returning diagnostics is enough. Status doesn't add anything,.

core/status/cmd.go
reinkrul: I agree

woutslakhorst (approved):

PR #159, opened 2021-03-17
all config keys to lowercase

### LINKED ISSUE #118: Config: cased config properties can't be passed using environment vars
E.g. the following does work, as expected;

```
NUTS_STRICTMODE=true
```
(`strictmode` is all-lowercase)

The following does not work because `bootstrapNodes` contains a capital `N`;

```
NUTS_NETWORK_BOOTSTRAPNODES=bootstrap:5555
```

When I change it to `bootstrapnodes` in the "koanf" tag, it does work.

woutslakhorst:
Closes #118 

snake_case won't work either. The only thing that works for making ENV variables work is to have all keys conform to `[a-z]`. The result is in this PR. We could decide to change some of the config keys to something different


reinkrul (changes requested):
I can live with this. There are some referenced to be fixed though:

- running-a-node.rst
- development/nuts.yaml

reinkrul (approved):

PR #160, opened 2021-03-17
auth: replaced dummy name resolver with vcr one

### LINKED ISSUE #90: Auth: Implement organization names based on VC for contracts
After https://github.com/nuts-foundation/nuts-node/pull/69 has been merged, we can resolve organzations names to include in Nuts Auth contracts instead of the dummy name resolver we currently have.

woutslakhorst:
although VC concepts are dynamic. The auth module depends on the *organization* concept.
The vcr is now also a `ConceptResolver`. The contracts could be refactored to contain the same concept setup as VCs. That way contracts can be connected to VCs!

Fixes #90


vcr/interface.go
```diff
@@ -54,6 +54,12 @@ var vcDocumentType = "application/vc+json"
 
 var revocationDocumentType = "application/vc+json;type=revocation"
 
+// ConceptFinder can resolve VC backed concepts for a DID.
+type ConceptFinder interface {
+	// Find returns the requested concept or ErrNotFound
+	Find(conceptName string, ID did.DID) (concept.Concept, error)
```
reinkrul, line 60: In our NutsOrganizationCredential ID is a DID, but it could be very well something else. We could make the API broader usable by accepting ID as string (also since it's converted to a string anyways before handing it to the database layer).

vcr/vcr.go
```diff
@@ -457,6 +457,27 @@ func (c *vcr) Untrust(credentialType did.URI, issuer did.URI) error {
...
+
+	q.AddClause(concept.Eq(concept.SubjectField, ID.String()))
+
+	vcs, err := c.Search(q)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(vcs) == 0 {
+		return nil, ErrNotFound
```
reinkrul, line 474: I recall that we decided that `Find` (similar to Search) operations generally return a list and more importantly don't return an error when not found? When returning a single entry and an error when it's not found it feels more like `Resolve` or `Get` (when you don't like resolve). `ConceptResolver`?

reinkrul (changes requested):

auth/services/contract/notary.go
```diff
@@ -58,15 +60,21 @@ func (s *contractNotaryService) DrawUpContract(template contract.Template, orgID
...
 		return nil, fmt.Errorf("could not draw up contract: organization is not managed by this node: %w", validator.ErrMissingOrganizationKey)
 	}
 
 	// DrawUpContract draws up a contract for a specific organisation from a template
-	orgName, err := s.nameResolver.Resolve(orgID)
+	result, err := s.conceptFinder.Find(concept.OrganizationConcept, orgID)
 	if err != nil {
 		return nil, fmt.Errorf("could not draw up contract: %w", err)
 	}
+	orgName, ok := result.GetValue(concept.OrganizationName).(string)
```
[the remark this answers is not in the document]
woutslakhorst: that's also inline with other frameworks. 👍

vcr/interface.go
woutslakhorst: actually it must be a URI: https://www.w3.org/TR/vc-data-model/#identifiers

vcr/vcr.go
woutslakhorst: It's defined on the vcr what makes naming a bit trickier

woutslakhorst: Named it to `Get`

reinkrul (approved):

Issue #161, opened 2021-03-17
Auth: renew IRMA test data


woutslakhorst:
test data still holds v1 contract signatures: `auth/test/contracts.go:95` (`ValidIrmaContract`).

Capture/debug signed contract (by phone) in the Nuts node.

woutslakhorst:
Do after legal base credential is implemented.

woutslakhorst:
won't fix

PR #162, opened 2021-03-17
removed deprecated services

woutslakhorst:
The old JWT tokens were only useable for a single Irma credential. Current contracts already require 2 credentials. So old stuff can be removed.

See #161 for some extra cleanup work


reinkrul (commented):
I'm a bit confused, what replaced the IRMA contract signing session?

woutslakhorst:
> I'm a bit confused, what replaced the IRMA contract signing session?

`CreateSignSession` in experimental. The removed one could only create a session for Irma, the new one does it for all means.

reinkrul (approved):

reinkrul:
> > I'm a bit confused, what replaced the IRMA contract signing session?
> 
> `CreateSignSession` in experimental. The removed one could only create a session for Irma, the new one does it for all means.

Which means the new experimental API should become v1?

woutslakhorst:
v0 and experimental should be merged to a single version. Which can still be v0 since this entire node is new.

reinkrul:
> v0 and experimental should be merged to a single version. Which can still be v0 since this entire node is new.

Although I'd say let's just give all APIs we intend to keep for v1, version v1?

PR #165, opened 2021-03-18
Bind echo to all network ifaces by default

### LINKED ISSUE #78: Echo server is not exposed to the outside world by default
Since it binds on `localhost:1323` you'd have to configure it to have any endpoints being available from the outside at all, which is not very convenient.

```
http:
  default:
    address: :1323
```

Maybe better: bind on `:1323` (all interfaces) by default so it's available from the outside (Docker container e.g.), and add some extra checks to strict mode, disallowing unsafe configuration. Maybe disallow this 'bind to all interfaces' in strict mode to enforce administrators to provide alternative bindings?

reinkrul:
Much more convenient, not much less safe.

Fixes #78

### DIFF
--- a/core/client_config.go
+++ b/core/client_config.go
@@ -29,6 +29,7 @@ import (
 
 const defaultClientTimeout = 10 * time.Second
 const clientTimeoutFlag = "timeout"
+const defaultAddress = "localhost" + defaultHTTPInterface
 
 // ClientConfig has CLI client settings.
 type ClientConfig struct {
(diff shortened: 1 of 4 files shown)


woutslakhorst (changes requested):
I'd like to see some*security considerations* in the docs somewhere when running production.

reinkrul:
> I'd like to see some_security considerations_ in the docs somewhere when running production.

Good idea. Added them.

woutslakhorst (approved):

PR #166, opened 2021-03-18
Introduced DID Manager to enable/disable Bolts

reinkrul:
@woutslakhorst @stevenvegt what do you think of this API?


woutslakhorst:
**didman** sounds ok. 
*Na na na na na na na na na na na na na na na na... DIDMAN!* (https://www.youtube.com/watch?v=EtoMN_xi-AM)

reinkrul:
> **didman** sounds ok.
> _Na na na na na na na na na na na na na na na na... DIDMAN!_ (https://www.youtube.com/watch?v=EtoMN_xi-AM)
> 
> the `map[string]...` is very generic. My guess is that every bolt would require its own implementation for setup or is there some way to make an abstraction?

I thought about this, having templates (a bit like the VCs concepts) which describe what to create (services/VCs) but as to not overengineer it right from the start, I think we could start simply with `Bolt` implementations which create the required services.

woutslakhorst:
For Bolts, we only need a set of services? What about a structure to add a service with endpoints:

- a service gets created with references to other services (the bolt service)
- services get created with endpoints
- assertion keys are checked
- VCs are checked << still a bit hardcoded.....

reinkrul:
Discussed with @woutslakhorst:

- Bolts only need DID services
- 2 parties are relevant; the subject (e.g. care organization) which's DID the Bolt is enabled for and the controller, which holds the actual endpoints. Depending on the deployment scenario (SaaS Provider with a single controller v.s. isolated DIDs) they can be the same DID or differ.
- Bolt services always consists of 2 service levels/types: a compound service that's created on the subject's DID which holds references to the concrete services containing the actual endpoints, on the controller's DID.
- We don't want the concept of "Bolt" hardcoded into the Nuts Node, because it hampers adoption for other domains . We also don't want the specific Bolts hardcoded, because one might be developing/testing a new Bolt in a development/test network, and you don't want it in the Nuts Node itself at that point. And you want to quickly iterate over changes to the definitions.

stevenvegt:
I like the `serviceTemplate` instead of the bolts 👍 Perhaps make it broader and call it a `useCaseConfig`.
Could this also be the place to list a set of trusted issuers for a specific VC?

reinkrul:
Usage:

```
nuts did svc-apply InzageBolt did:nuts:7ywyjKtbUtY5cQStmtKJHU86Lx4tJaySUENuN2jVNeXq  -pauthServerURL=http://nuts.nl
```
(controller and subject are equal)

```
nuts did svc-apply InzageBolt did:nuts:7ywyjKtbUtY5cQStmtKJHU86Lx4tJaySUENuN2jVNeXq did:nuts:8p8uD4sfJbN2xHZJGAHhWd9RUhdawAY2RSsYzrdREQTg -pauthServerURL=http://nuts.nl
```
(controller and subject differ)

reinkrul:
> I like the `serviceTemplate` instead of the bolts 👍 Perhaps make it broader and call it a `useCaseConfig`.
> Could this also be the place to list a set of trusted issuers for a specific VC?

Not sure why `useCaseConfig` is broader than `serviceTemplate`? If the intention is to manage more than just DID services we may rename it to something else, but if we foresee just services I think it 100% "deks de lading"?

reinkrul:
@stevenvegt @woutslakhorst it's become quite a bit of code; do we agree this is the right direction? Then I'll proceed to finish the PR with tests, documentation and fix TODO's and such.

woutslakhorst:
I'm missing some design decisions and or requirements here. For example: what part of a template can vary? If the structure is more or less static, then an API that accepts templates would be enough to start with? The input could be unmarshalled to a simple struct and no mustache parsing is required?

reinkrul:
Decided that it's too early for now

PR #170, opened 2021-03-19
Add CLI sort flag to network list

stevenvegt:
`nuts network list --sort=time`
`nuts network list --sort=type`

### DIFF
--- a/network/cmd/cmd.go
+++ b/network/cmd/cmd.go
@@ -20,14 +20,18 @@ package cmd
 
 import (
 	"fmt"
+	"sort"
+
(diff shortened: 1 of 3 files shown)


woutslakhorst (changes requested):
I would expect the `sort` flag to appear in the client_config command options documentation.

stevenvegt:
> I would expect the `sort` flag to appear in the client_config command options documentation.

It is a command specific option, not a client broad option. Just like `vdr resolve --document=true`.

woutslakhorst (approved):

PR #174, opened 2021-03-22
VDR: Add CLI and APIs for adding and removing verificationMethods

stevenvegt:
This PR adds `addNewVerificationMethod` and `removeVerificationMethod` to the VDR API
Implements these methods in the CLI.

In addition it:
* Ignores already revoked Keys in ambassador instead of throwing errors
* Adds new DocManipulator interface for higher level Document Manipulations
* Move higher level Deactivate from VDR to DocManipulator
* Remove KeyResolver method from VDR interface
* Add tests and improve coverage to touched code

TODO:
* [x] Wait for https://github.com/nuts-foundation/go-did/issues/33 and fix spec.


docs/_static/vdr/v1.yaml
```diff
@@ -150,6 +150,89 @@ paths:
...
+          - "did:nuts:1234"
+        schema:
+          type: string
+    post:
+      summary: Creates a new verificationMethod and adds it to the DID document
+      operationId: addNewVerificationMethod
+      tags:
+        - DID
+      responses:
+        "201":
```
woutslakhorst, line 169: a 201 links to the location where the newly created resource can be located. A 200 is more suitable

docs/_static/vdr/v1.yaml
```diff
@@ -150,6 +150,89 @@ paths:
...
+          description: "New verification method has been created and added successfully. Returns the DID document."
+          content:
+            application/json+did-document:
+              schema:
+                $ref: '#/components/schemas/VerificationMethod'
+        "403":
```
woutslakhorst, line 175: a 403 would indicate that credentials were wrong when trying to access the server. This is more generic so a 400?

docs/_static/vdr/v1.yaml
```diff
@@ -150,6 +150,89 @@ paths:
...
+      - name: did
+        in: path
+        description: URL encoded DID.
+        required: true
+        example:
+          - "did:nuts:1234"
+        schema:
+          type: string
+    post:
+      summary: Creates a new verificationMethod and adds it to the DID document
```
woutslakhorst, line 164: Does it also create a new private key?

docs/_static/vdr/v1.yaml
```diff
@@ -150,6 +150,89 @@ paths:
...
+        schema:
+          type: string
+    delete:
+      summary: Delete a specific verification method
+      description: Removes the verification method from the DID Document. Revokes the public key in the keystore.
+      operationId: deleteVerificationMethod
+      responses:
+        "204":
+          description: Verification Method was successfully deleted
+        "403":
```
woutslakhorst, line 218: same as above

docs/_static/vdr/v1.yaml
```diff
@@ -150,6 +150,89 @@ paths:
...
+        in: path
+        description: URL encoded DID identifying the verification method.
+        required: true
+        example:
+          - "did:nuts:1234#abc"
+        schema:
+          type: string
+    delete:
+      summary: Delete a specific verification method
+      description: Removes the verification method from the DID Document. Revokes the public key in the keystore.
```
woutslakhorst, line 213: *revoke* or just remove from DID document? What happens to the private key?

vdr/api/v1/client_test.go
```diff
@@ -167,6 +172,69 @@ func TestHTTPClient_Deactivate(t *testing.T) {
 	})
 }
 
+func TestHTTPClient_AddNewVerificationMethod(t *testing.T) {
+	didString := "did:nuts:1"
```
woutslakhorst, line 176: why not use the testDIDs?

vdr/api/v1/client_test.go
```diff
@@ -167,6 +172,69 @@ func TestHTTPClient_Deactivate(t *testing.T) {
...
+
+	t.Run("ok", func(t *testing.T) {
+		s := httptest.NewServer(http2.Handler{StatusCode: http.StatusCreated, ResponseData: string(methodJSON)})
+		c := HTTPClient{ServerAddress: s.URL, Timeout: time.Second}
+		// methodResponse, err := c.AddNewVerificationMethod(didString)
+		_, err := c.AddNewVerificationMethod(didString)
+		if !assert.NoError(t, err) {
+			return
+		}
+		// this fails because of https://github.com/nuts-foundation/go-did/issues/33
```
woutslakhorst, line 191: add issue or add reference from that issue to not forget to fix this.

vdr/cmd/cmd_test.go
```diff
@@ -280,7 +295,95 @@ func TestEngine_Command(t *testing.T) {
...
+
+	t.Run("addVerificationMethod", func(t *testing.T) {
+		id123, _ := did.ParseDID("did:nuts:123")
+		id123Method, _ := did.ParseDID("did:nuts:123#abc-method1")
+
+		pair, _ := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
+		verificationMethod, _ := did.NewVerificationMethod(*id123Method, ssi.JsonWebKey2020, *id123, pair.PublicKey)
+
+		t.Run("ok", func(t *testing.T) {
+			cmd := newCmd(t)
```
woutslakhorst, line 310: could you add the cool `t.Cleanup` to `newCmd` as well? With all the `Unsetenvs`?

vdr/doc-updater.go
```diff
@@ -8,15 +8,94 @@ import (
...
 )
 
-// NutsDocUpdater contains helper methods to update a nuts document.
+// NutsDocUpdater contains helper methods to update a Nuts DID document.
 type NutsDocUpdater struct {
-	// keyCreator is used for getting a fresh key and use it to generate the Nuts DID
-	keyCreator nutsCrypto.KeyCreator
+	// KeyCreator is used for getting a fresh key and use it to generate the Nuts DID
+	KeyCreator nutsCrypto.KeyCreator
+	VDR        types.VDR
```
woutslakhorst, line 18: Missing docs on VDR

vdr/doc-updater.go
```diff
@@ -8,15 +8,94 @@ import (
...
+// The key is not used yet and should be manually added to one of the VerificationRelationships
+func (u NutsDocUpdater) AddVerificationMethod(id did.DID) (*did.VerificationMethod, error) {
+	doc, meta, err := u.VDR.Resolve(id, &types.ResolveMetadata{AllowDeactivated: true})
+	if err != nil {
+		return nil, err
+	}
+	if meta.Deactivated {
+		return nil, types.ErrDeactivated
+	}
+	updater := NutsDocUpdater{KeyCreator: u.KeyCreator}
```
woutslakhorst, line 46: why do you need a new DocUpdater? You're already in one...

vdr/doc-updater.go
```diff
@@ -8,15 +8,94 @@ import (
...
+	emptyDoc := did.Document{
+		Context: []ssi.URI{did.DIDContextV1URI()},
+		ID:      id,
+	}
+	return u.VDR.Update(id, meta.Hash, emptyDoc, nil)
+}
+
+// AddVerificationMethod adds a new key as a VerificationMethod to the document.
+// The key is not used yet and should be manually added to one of the VerificationRelationships
+func (u NutsDocUpdater) AddVerificationMethod(id did.DID) (*did.VerificationMethod, error) {
```
woutslakhorst, line 38: Remove the `Nuts` prefix. The entire repo is *Nuts*

woutslakhorst (changes requested):
The wild growth of `Resolvers`, `interfaces` and indirections are becoming worrying. The number of arguments we're passing to `services` is becoming bigger and bigger. This is beginning to make everything less readable. Not that we need to fix it now, but this PR is adding some extra interfaces (some not needed imho) and we need to have a design discussion at some point.

maybe it's also time to discuss the removal of the publicKeystore and use the vdr for this?

vdr/types/interface.go
```diff
@@ -74,6 +74,28 @@ type DocDeactivator interface {
 	Deactivate(id did.DID) error
 }
 
+// DocKeyAdder is the interface that defies functionality to add new keys to a DID document.
+type DocKeyAdder interface {
```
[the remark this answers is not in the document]
stevenvegt: This mostly based on the https://golang.org/src/io/io.go which is often referred to when talking about good interface design. As you can see, many single method interfaces and combinations of those. I agree that io.go describes a very low level and generic set of interfaces and here in this DocKeyAdder we are not. Non the less, it gives a level of consistancy. Sometimes it is usefull to only require the dependancy on one function. Especially when the group interfaces are just grouped because they have a common theme and not because they depend on each other. e.g.: you need to resolve a key in auth. Do you inject the complete key store of only a KeyResolver? I would prefer the latter.

vdr/api/v1/client.go
```diff
@@ -117,15 +117,43 @@ func (hb HTTPClient) Deactivate(DID string) error {
...
 func (hb HTTPClient) withTimeout() (context.Context, context.CancelFunc) {
 	return context.WithTimeout(context.Background(), hb.Timeout)
 }
 
 func readDIDDocument(reader io.Reader) (*did.Document, error) {
-	var data []byte
-	var err error
-
-	if data, err = ioutil.ReadAll(reader); err != nil {
+	data, err := ioutil.ReadAll(reader)
```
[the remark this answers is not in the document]
stevenvegt: fixed it in the VDR module, created a issue for the rest: https://github.com/nuts-foundation/nuts-node/issues/199

vdr/resolvers.go
```diff
@@ -3,14 +3,42 @@ package vdr
...
-	doc, _, err := r.Resolve(holder, &types.ResolveMetadata{
+// NameResolver defines functions for resolving the name of an entity holding a DID.
+type NameResolver interface {
+	// Resolve resolves the name of a DID holder.
+	Resolve(input did.DID) (string, error)
+}
+
+// NewDummyNameResolver returns a NameResolver that generates a name based on the DID.
+// TODO: Remove this after implementing VCs (https://github.com/nuts-foundation/nuts-node/issues/90)
+func NewDummyNameResolver() NameResolver {
```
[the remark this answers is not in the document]
stevenvegt: ahh, rebase error, my bet.

auth/services/contract/notary.go
```diff
@@ -36,19 +36,19 @@ import (
...
 var timenow = time.Now
 
 // NewContractNotary accepts the registry and crypto Nuts engines and returns a ContractNotary
-func NewContractNotary(finder vcr.ConceptFinder, didResolver types.Resolver, keyStore crypto.PrivateKeyStore, contractValidity time.Duration) services.ContractNotary {
+func NewContractNotary(finder vcr.ConceptFinder, keyResolver types.KeyResolver, keyStore crypto.PrivateKeyStore, contractValidity time.Duration) services.ContractNotary {
 	return &contractNotaryService{
 		conceptFinder:    finder,
 		contractValidity: contractValidity,
-		didResolver:      didResolver,
+		keyResolver:      keyResolver,
```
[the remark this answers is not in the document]
stevenvegt: Yes! https://github.com/nuts-foundation/nuts-node/issues/201

docs/_static/vdr/v1.yaml
stevenvegt: Yes, will add it 👍

docs/_static/vdr/v1.yaml
stevenvegt: Every public key is stored under its key-id. It is theoretically possible to have multiple verification methods with different key ids for the same private key because of different DID prefixes. This specific public key will be revoked. The private key does not have any expiraton properties. The private key will not be altered.

docs/_static/vdr/v1.yaml
stevenvegt: As can read here: https://tools.ietf.org/html/rfc7231#section-6.5.3 
> a request might be forbidden for reasons unrelated to the credentials.

Which is here the case: your node does not have the rights to change the document. I'm in favour of the more specific 403 against the more general 400 status.

vdr/types/interface.go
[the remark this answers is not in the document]
woutslakhorst: One big difference: `io` is to be used as lib by other libs. Our stuff is used by us (an app), so why define interfaces that are not used?

vdr/api/v1/client.go
```diff
@@ -117,15 +116,43 @@ func (hb HTTPClient) Deactivate(DID string) error {
 	return nil
 }
 
+// AddNewVerificationMethod creates a new verificationMethod and adds it to the DID document
+// It expects a status 201 respond from the server, returns an error otherwise
```
woutslakhorst, line 120: wrong comment: 200

vdr/api/v1/client.go
```diff
@@ -117,15 +116,43 @@ func (hb HTTPClient) Deactivate(DID string) error {
...
+	}
+	if err := core.TestResponseCode(http.StatusOK, response); err != nil {
+		return nil, err
+	}
+
+	return readVerificationMethod(response.Body)
+}
+
+// DeleteVerificationMethod deletes a specified verificationMethod from the DID document
+// It expects a status 204 respond from the server, returns an error otherwise
```
woutslakhorst, line 137: comment typo

woutslakhorst (commented):
Some comment typos

vdr/types/interface.go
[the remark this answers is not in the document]
stevenvegt: Removed the interfaces 👍

woutslakhorst (approved):

PR #178, opened 2021-03-23
Allow CORS to be enabled on specific HTTP interfaces

### LINKED ISSUE #60: Decide which endpoints need CORS enabled
Currently CORS is enabled for all endpoints (because appearantly `nuts-auth` needs it?), but this could lead to security vulnerabilities. We need to decide which endpoints really need it and just enable it on those. Maybe using the new `HTTPConfig` configuration options.

reinkrul:
Required for development (and some deployment setups) on `auth` for IRMA flow

Fixes #60


stevenvegt (approved):

woutslakhorst (changes requested):
`cors: true` translates to `allow-origin: *` since that is the standard config for echo. That is a bit too much, you should be able to define the domains.

reinkrul:
> `cors: true` translates to `allow-origin: *` since that is the standard config for echo. That is a bit too much, you should be able to define the domains.

I thought about this too, since you have more options than just `AllowOrigin` (`allowed http methods` for instance). But I guess the `origin` parameter is so important that we should add it. I'd say for now, if you'd really like to specify the `http methods` parameter in your deployment scenario you'll just have to put a proxy in front of it.

Furthermore, we should disallow `origin = *` in strictmode.

woutslakhorst (approved):

PR #179, opened 2021-03-23
VDR/VCR: Support escaped DIDs in paths

reinkrul:
As the OpenAPI spec states the input DID must be escaped.

Based on https://github.com/nuts-foundation/nuts-node/pull/178


woutslakhorst:
I believe unescaping is handled by echo. When I tested it, I had to escape the `#` as `%23`. In the handler func it is given as `#`.
The colon doesn't require escaping?

woutslakhorst (commented):
I believe unescaping is handled by echo. When I tested it, I had to escape the # as %23. In the handler func it is given as #.
The colon doesn't require escaping?

stevenvegt (commented):
is this worth another top level package `net`? can we put it in core or something? Furthermore, i would like to see this as echo middleware if possible?

reinkrul:
> I believe unescaping is handled by echo. When I tested it, I had to escape the # as %23. In the handler func it is given as #.
> The colon doesn't require escaping?

> is this worth another top level package net? can we put it in core or something? Furthermore, i would like to see this as echo middleware if possible?

The NodeJS OpenAPI client (or Axios, not sure who decides to escape it) escapes the `:`: `did%3Anuts%3A8xC8kjWqT6VTLpseaK34aaFve7xZjk5QT1uEwh44Jfyu`. This is both how it is logged in NodeJS and how it is passed to the REST API function. You could call this a shortcoming of the OpenAPI Go generator (also see https://github.com/deepmap/oapi-codegen/issues/264).

The correct approach would be, in my opinion, to fix it there.

woutslakhorst:
Hmmm, As @stevenvegt said, some middleware would be preferable so every path param is properly unescaped before it reaches the handler.

reinkrul:
Did some more investigation path parameters are parsed by Echo and passed to the generated OpenAPI handlers. So if it were to be fixed it should be fixed inside Echo, but there seems to ahve been much discussion and changing back-and-forth, and not much incentive to make a choice and fix it (https://github.com/labstack/echo/issues/1117).

So for now we'll add middleware and decode all path params on our side.

woutslakhorst (approved):
OMG LMAO
LGTM

PR #181, opened 2021-03-23
Made logging consistent

### LINKED ISSUE #40: Consistent logging

reinkrul:
Fixes #40 

- No more `Fatal` (except for 1 single place where I couldn't trigger it)
- Correct usage of levels
- Tested setting level through `verbosity` config flag
- Consistent log format of parameters

### DIFF
--- a/network/dag/bbolt_dag.go
+++ b/network/dag/bbolt_dag.go
@@ -289,7 +289,7 @@ func (dag *bboltDAG) add(transaction Transaction) error {
 				return fmt.Errorf("unable to store forward reference %s->%s: %w", prev, ref, err)
 			}
 			if !exists(transactions, prev) {
-				log.Logger().Debugf("Transaction %s is referring to missing prev %s, marking it as missing", ref, prev)
+				log.Logger().Debugf("Transaction is referring to missing prev, marking it as missing (tx=%s, prev=%s)", ref, prev)
 				if err = missingTransactions.Put(prev.Slice(), []byte{1}); err != nil {
 					return fmt.Errorf("unable to register missing transaction %s: %w", prev, err)
 				}
--- a/network/network.go
+++ b/network/network.go
@@ -154,7 +154,7 @@ func (n *Network) ListTransactions() ([]dag.Transaction, error) {
 // If the key should be inside the transaction (instead of being referred to) `attachKey` should be true.
 func (n *Network) CreateTransaction(payloadType string, payload []byte, signingKeyID string, attachKey crypto2.PublicKey, timestamp time.Time, fieldOpts ...dag.FieldOpt) (dag.Transaction, error) {
 	payloadHash := hash.SHA256Sum(payload)
-	log.Logger().Infof("Creating transaction (payload hash=%s,type=%s,length=%d,signingKey=%s)", payloadHash, payloadType, len(payload), signingKeyID)
+	log.Logger().Debugf("Creating transaction (payload hash=%s,type=%s,length=%d,signingKey=%s)", payloadHash, payloadType, len(payload), signingKeyID)
 	// Create transaction
 	prevs := n.graph.Heads()
 	unsignedTransaction, err := dag.NewTransaction(payloadHash, payloadType, prevs, fieldOpts...)
@@ -180,6 +180,7 @@ func (n *Network) CreateTransaction(payloadType string, payload []byte, signingK
 	if err = n.payloadStore.WritePayload(payloadHash, payload); err != nil {
 		return nil, fmt.Errorf("unable to store payload of newly created transaction: %w", err)
 	}
+	log.Logger().Infof("Transaction created (ref=%s,type=%s,length=%d)", transaction.Ref(), payloadType, len(payload))
 	return transaction, nil
 }
 
--- a/network/p2p/impl-connection.go
+++ b/network/p2p/impl-connection.go
@@ -53,7 +53,7 @@ func (conn *connection) close() {
 	defer conn.closeMutex.Unlock()
 	if conn.grpcConn != nil {
 		if err := conn.grpcConn.Close(); err != nil {
-			log.Logger().Errorf("Unable to close client connection (peer=%s): %v", conn.Peer, err)
+			log.Logger().Warnf("Unable to close client connection (peer=%s): %v", conn.Peer, err)
 		}
 		conn.grpcConn = nil
 	}
@@ -73,7 +73,7 @@ func (conn *connection) send(message *transport.NetworkMessage) {
 func (conn connection) sendMessages() {
 	for message := range conn.outMessages {
 		if conn.gate.Send(message) != nil {
-			log.Logger().Errorf("Unable to broadcast message to peer (peer=%s)", conn.Peer)
+			log.Logger().Warnf("Unable to broadcast message to peer (peer=%s)", conn.Peer)
 		}
 	}
 }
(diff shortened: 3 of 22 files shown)


network/dag/bbolt_dag.go
```diff
@@ -289,7 +289,7 @@ func (dag *bboltDAG) add(transaction Transaction) error {
 				return fmt.Errorf("unable to store forward reference %s->%s: %w", prev, ref, err)
 			}
 			if !exists(transactions, prev) {
-				log.Logger().Debugf("Transaction %s is referring to missing prev %s, marking it as missing", ref, prev)
+				log.Logger().Debugf("Transaction is referring to missing prev, marking it as missing (tx=%s, prev=%s)", ref, prev)
```
stevenvegt, line 292: Perhaps use the `WithFields` as described here in the logrus docs;
https://github.com/sirupsen/logrus#fields

network/network.go
```diff
@@ -154,7 +154,7 @@ func (n *Network) ListTransactions() ([]dag.Transaction, error) {
 // If the key should be inside the transaction (instead of being referred to) `attachKey` should be true.
 func (n *Network) CreateTransaction(payloadType string, payload []byte, signingKeyID string, attachKey crypto2.PublicKey, timestamp time.Time, fieldOpts ...dag.FieldOpt) (dag.Transaction, error) {
 	payloadHash := hash.SHA256Sum(payload)
-	log.Logger().Infof("Creating transaction (payload hash=%s,type=%s,length=%d,signingKey=%s)", payloadHash, payloadType, len(payload), signingKeyID)
+	log.Logger().Debugf("Creating transaction (payload hash=%s,type=%s,length=%d,signingKey=%s)", payloadHash, payloadType, len(payload), signingKeyID)
```
stevenvegt, line 157: `WithFields`

network/network.go
```diff
@@ -180,6 +180,7 @@ func (n *Network) CreateTransaction(payloadType string, payload []byte, signingK
 	if err = n.payloadStore.WritePayload(payloadHash, payload); err != nil {
 		return nil, fmt.Errorf("unable to store payload of newly created transaction: %w", err)
 	}
+	log.Logger().Infof("Transaction created (ref=%s,type=%s,length=%d)", transaction.Ref(), payloadType, len(payload))
```
stevenvegt, line 183: `WithFields`

stevenvegt (changes requested):
I think the errors are better parsable when they are always the same and the values are seperately added using the `WithFields` option. Also very convenient when the logger type is of the JSON type.
https://github.com/sirupsen/logrus#fields

auth/api/v0/api.go
```diff
@@ -154,7 +154,7 @@ func (api *Wrapper) IntrospectAccessToken(ctx echo.Context) error {
 
 	claims, err := api.Auth.OAuthClient().IntrospectAccessToken(token)
 	if err != nil {
-		logging.Log().WithError(err).Debug("Error while inspecting access token")
+		logging.Log().WithError(err).Info("Error while inspecting access token")
```
woutslakhorst, line 157: Warn imo

auth/api/v0/api.go
```diff
@@ -185,13 +185,13 @@ const bearerPrefix = "bearer "
 // VerifyAccessToken verifies if a request contains a valid bearer token issued by this server
 func (api *Wrapper) VerifyAccessToken(ctx echo.Context, params VerifyAccessTokenParams) error {
 	if len(params.Authorization) == 0 {
-		logging.Log().Warn("No authorization header given")
+		logging.Log().Info("No authorization header given")
```
woutslakhorst, line 188: warn

auth/api/v0/api.go
```diff
@@ -185,13 +185,13 @@ const bearerPrefix = "bearer "
...
 		return ctx.NoContent(http.StatusForbidden)
 	}
 
 	index := strings.Index(strings.ToLower(params.Authorization), bearerPrefix)
 	if index != 0 {
-		logging.Log().Warn("Authorization does not contain bearer token")
+		logging.Log().Info("Authorization does not contain bearer token")
```
woutslakhorst, line 194: warn

network/p2p/impl-connection.go
```diff
@@ -73,7 +73,7 @@ func (conn *connection) send(message *transport.NetworkMessage) {
 func (conn connection) sendMessages() {
 	for message := range conn.outMessages {
 		if conn.gate.Send(message) != nil {
-			log.Logger().Errorf("Unable to broadcast message to peer (peer=%s)", conn.Peer)
+			log.Logger().Warnf("Unable to broadcast message to peer (peer=%s)", conn.Peer)
```
woutslakhorst, line 76: Why not error?

network/p2p/impl-connection.go
```diff
@@ -84,13 +84,13 @@ func receiveMessages(gate messageGate, peerId PeerID, receivedMsgQueue messageQu
 		msg, recvErr := gate.Recv()
 		if recvErr != nil {
 			if recvErr == io.EOF {
-				log.Logger().Infof("Peer closed connection: %s", peerId)
+				log.Logger().Infof("Peer closed connection (peer-id=%s)", peerId)
 			} else {
-				log.Logger().Errorf("Peer connection error (peer=%s): %v", peerId, recvErr)
+				log.Logger().Warnf("Peer connection error (peer-id=%s): %v", peerId, recvErr)
```
woutslakhorst, line 89: Also why not error?

network/p2p/impl.go
```diff
@@ -132,7 +132,7 @@ func (c *connector) connect(ownID PeerID, config *tls.Config) (*connection, erro
 	client := transport.NewNetworkClient(grpcConn)
 	gate, err := client.Connect(cxt)
 	if err != nil {
-		log.Logger().Errorf("Failed to set up stream (peer address=%s): %v", c.address, err)
+		log.Logger().Warnf("Failed to set up stream (peer address=%s): %v", c.address, err)
```
woutslakhorst, line 135: Also why?
And for the next 2 in this file: do we immediately have to act on these errors or not?

woutslakhorst (changes requested):
I do not agree with some of the downgrades from error to warn.
Also this PR doesn't totally solve the issue. The issue is also meant for adding log messages where there should be.

network/dag/bbolt_dag.go
reinkrul: I still don't agree; I'd use fields for key-value pairs that are applicable to all (or most) of the log entries, e.g. remote client IP address, authenticated user, module name (one of our current) since that would make it very useful for log analytics (e.g. using ELK). If you have a myriad of very different fields, it makes the log (and thus analytics) messy in my opinion.

reinkrul:
> I do not agree with some of the downgrades from error to warn.
> Also this PR doesn't totally solve the issue. The issue is also meant for adding log messages where there should be.

I added log messages at top-level services (VDR, VCR, Crypto, Network) at start (DEBUG) and end (INFO), unless it's a 'read' operation (in that case an INFO message isn't particulary useful in my experience).

network/p2p/impl.go
reinkrul: I classified network connection errors as WARN because it could very well be on the peer's end, or somewhere in between. In any case you should have multiple peers (it's dangerous to 100% rely on a single peer) so connection problems with 1 peer is not something that requires immediate intention of an operator.

Maybe it'd be better to have an ERROR when you can't connect to any of your bootstrap nodes?

woutslakhorst:
> > I do not agree with some of the downgrades from error to warn.
> > Also this PR doesn't totally solve the issue. The issue is also meant for adding log messages where there should be.
> 
> I added log messages at top-level services (VDR, VCR, Crypto, Network) at start (DEBUG) and end (INFO), unless it's a 'read' operation (in that case an INFO message isn't particularly useful in my experience).

Still doesn't solve #40. We haven't finished the logging setup in that issue I think.

reinkrul:
> > > I do not agree with some of the downgrades from error to warn.
> > > Also this PR doesn't totally solve the issue. The issue is also meant for adding log messages where there should be.
> > 
> > 
> > I added log messages at top-level services (VDR, VCR, Crypto, Network) at start (DEBUG) and end (INFO), unless it's a 'read' operation (in that case an INFO message isn't particularly useful in my experience).
> 
> Still doesn't solve #40. We haven't finished the logging setup in that issue I think.

What is missing in your opinion?

woutslakhorst:
> What is missing in your opinion?

trace logging per method, extended debug logging. A check if every API has error/warn logging in place.

stevenvegt:
So, bottom line: good PR, but it does not fix #40. This issue asks for a small project just like the better api errors. Following all the API paths engine by engine.

woutslakhorst (approved):

Issue #112, opened 2021-03-03
Implement RFC7807:  Problem Details for HTTP APIs


gerardsn:
Following #3 and #9 we propose to implement [RFC7807](https://tools.ietf.org/html/rfc7807) to standardize API error handling. More specifically, define "Title", "Status", and "detail" [members](https://tools.ietf.org/html/rfc7807#section-3.1) for each problem.

One of the current methods used is
```golang
func (a Wrapper) GetDID(ctx echo.Context, did string) error {
	d, err := did2.ParseDID(did)
	if err != nil {
		return ctx.String(http.StatusBadRequest, fmt.Sprintf("given DID could not be parsed: %s", err.Error()))
	}
       ...
}
```

which can be standardized using [https://github.com/mschneider82/problem](https://github.com/mschneider82/problem) to something like
```golang
func (a Wrapper) GetDID(ctx echo.Context, did string) error {
	d, err := did2.ParseDID(did)
	if err != nil {
		p := problem.Of(http.StatusBadRequest)
		p.Append(problem.Title("given DID could not be parsed"))
		p.Append(problem.Detail(err.Error()))
		return p
	}
       ...
}
```

This will require:
- [x] research on how to include external schemas in API generation #183 
- [x] add general return for status 500 #183 
- [x] add standard for error handling status >= 400 #183 

Implement standard in APIs:
- [x] auth #250 
- [x] crypto #183 
- [x]  network #273 
- [x] vcr #354 
- [x] vdr #354
- [x] update docs, add API error details #354

woutslakhorst:
@stevenvegt @reinkrul thoughts?

reinkrul:
I think this can really improve the consistency of our REST API and I'd really like our error responses to be JSON as well. We should take care we don't introduce a massive amount of code or complexity though (while implementing). Proposal looks good, though we should discussing collapsing the builder calls;

`return problem.Of(http.StatusBadRequest).Append(problem.Title("given DID could not be parsed")).Append(problem.Detail(err.Error()))` (something like that)

Also in this case, I'd say the title should identify the operation that was performed, and the detail what went wrong. Now they're a bit duplicate?

So title could be something like "can't resolve DID" and description the parser error? Or does the RFC state otherwise?

PR #183, opened 2021-03-24
Standardize API error return values

### LINKED ISSUE #3: Add error handling middleware to echo server
Which can handle error message format and logging.

See https://echo.labstack.com/guide/error-handling

gerardsn:
Adds general error handeling middleware.
Proposes standard for 400+ return values. Currently only for crypto, the rest is tracked in #112.
Closes #3 #9


gerardsn:
Increase max-width of read the docs 800 -> 1000px
Changed font swagger to 12px

core/echo.go
```diff
@@ -99,3 +101,26 @@ func getGroup(path string) string {
...
+	echoServer.HTTPErrorHandler = httpErrorHandler
+	// TODO: We might not want to enable CORS on all endpoints (https://github.com/nuts-foundation/nuts-node/issues/60)
+	echoServer.Use(middleware.CORS())
+	return echoServer
+}
+
+func httpErrorHandler(err error, ctx echo.Context) {
+	if prb, ok := err.(*problem.Problem); ok {
+		// todo: connect to module logger
+		if !ctx.Response().Committed {
```
reinkrul, line 118: what if the response already _is_ committed?

core/echo.go
```diff
@@ -99,3 +101,26 @@ func getGroup(path string) string {
...
+			if _, err := prb.WriteTo(ctx.Response()); err != nil {
+				//ctx.Echo().Logger.Error(err)
```
reinkrul, line 120: missing something?

crypto/api/v1/api.go
```diff
@@ -55,24 +56,27 @@ func (signRequest SignJwtRequest) validate() error {
...
+}
+
 // SignJwt handles api calls for signing a Jwt
 func (w *Wrapper) SignJwt(ctx echo.Context) error {
 	var signRequest = &SignJwtRequest{}
 	err := ctx.Bind(signRequest)
 	if err != nil {
 		log.Logger().Error(err.Error())
-		return echo.NewHTTPError(http.StatusBadRequest, err.Error())
+		return createProblem("SignJWT failed", http.StatusBadRequest, err.Error())
```
reinkrul, line 69: make consts of these titles

crypto/api/v1/api.go
```diff
@@ -88,25 +92,24 @@ func (w *Wrapper) PublicKey(ctx echo.Context, kid string, params PublicKeyParams
 	if params.At != nil {
 		at, err = time.Parse(time.RFC3339, *params.At)
 		if err != nil {
-			return ctx.String(http.StatusBadRequest, fmt.Sprintf("cannot parse '%s' as RFC3339 time format", *params.At))
+			return createProblem("Failed to get PublicKey", http.StatusBadRequest, fmt.Sprintf("cannot parse '%s' as RFC3339 time format", *params.At))
```
reinkrul, line 95: const

crypto/api/v1/api.go
```diff
@@ -88,25 +92,24 @@ func (w *Wrapper) PublicKey(ctx echo.Context, kid string, params PublicKeyParams
...
 		}
 	}
 
 	pubKey, err := w.C.GetPublicKey(kid, at)
 	if err != nil {
 		if errors.Is(err, crypto.ErrKeyNotFound) {
-			return ctx.NoContent(http.StatusNotFound)
+			return createProblem("Failed to get PublicKey", http.StatusNotFound, fmt.Sprintf("no pubkey found for %s", kid))
```
reinkrul, line 102: I'd fully write "public key"

reinkrul (changes requested):
I like the setup. Some small things.

core/echo.go
```diff
@@ -99,3 +101,26 @@ func getGroup(path string) string {
 	}
 	return ""
 }
+
+func creatorFn() EchoServer {
```
stevenvegt, line 105: This name is a bit generic.

crypto/api/v1/api_test.go
```diff
@@ -168,9 +185,16 @@ func TestWrapper_PublicKey(t *testing.T) {
 
 		ctx.echo.EXPECT().Request().Return(&http.Request{})
 		ctx.keyStore.EXPECT().GetPublicKey("kid", at).Return(nil, crypto.ErrKeyNotFound)
-		ctx.echo.EXPECT().NoContent(http.StatusNotFound)
+		//ctx.echo.EXPECT().NoContent(http.StatusNotFound)
```
stevenvegt, line 188: Remove comment

crypto/api/v1/api_test.go
```diff
@@ -181,18 +205,30 @@ func TestWrapper_PublicKey(t *testing.T) {
...
+		assert.Equal(t, 500, p.Status)
 	})
 
 	t.Run("error - 400 for incorrect time format", func(t *testing.T) {
 		ctx := newMockContext(t)
 		defer ctx.ctrl.Finish()
 
 		ctx.echo.EXPECT().Request().Return(&http.Request{})
-		ctx.echo.EXPECT().String(http.StatusBadRequest, "cannot parse 'b00m!' as RFC3339 time format")
+		//ctx.echo.EXPECT().String(http.StatusBadRequest, "cannot parse 'b00m!' as RFC3339 time format")
```
stevenvegt, line 221: Remove comment

core/test.go
```diff
@@ -45,3 +48,19 @@ func (i *TestEngine) FlagSet() *pflag.FlagSet {
...
+// Problem is a helper struct to Unmarshal problem.Problem
+type Problem struct {
+	Title  string `json:"title"`
+	Status int    `json:"status"`
+	Detail string `json:"detail"`
+}
+
+// ErrorToProblem returns a Problem generated from a problem.Problem
+// problem.Problem doesn't expose its fields
+func ErrorToProblem(err error) Problem {
```
stevenvegt, line 61: I don't like these poluting the global namespace. The `problem` can perhaps be made private?
Or a few helper methods:
* `TestAssertProblemErrStatus(err error, status int) bool`
* `TestAssertProblemErrTitle(err error, title string) bool`
* `TestAssertProblemErrMessage(err error, message string) bool`

core/echo.go
stevenvegt: Yes, if the write fails you want to see it in the logs.

core/echo_test.go
```diff
@@ -102,3 +106,33 @@ func Test_getGroup(t *testing.T) {
...
+	t.Run("Problem return", func(t *testing.T) {
+		f := func(c echo.Context) error {
+			return problem.New(problem.Status(http.StatusInternalServerError))
+		}
+		e.Add(http.MethodGet, "/problem", f)
+		req, _ := http.NewRequest(http.MethodGet, fmt.Sprintf("%s/problem", server.URL), nil)
+		resp, err := client.Do(req)
+
+		assert.NoError(t, err)
+		assert.Equal(t, http.StatusInternalServerError, resp.StatusCode)
```
stevenvegt, line 123: Test some more. We expect a json struct on the correct content type right? There should be a title and description in here too?

core/test.go
```diff
@@ -18,7 +18,10 @@
 
 package core
```
stevenvegt, line 19: Move these test functions to the test package.

stevenvegt (changes requested):
Nice first PR. Some minor points to fix 👍

core/echo.go
gerardsn: there is nothing we can do if the response is already committed (besides logging). The current approach replicates echo's `DefaultHTTPErrorHandler()`

core/echo.go
gerardsn: Failing to write a response is unlikely, so we can log this in the core logger

core/echo.go
gerardsn: changed to createEchoServer

core/test.go
gerardsn: These have been moved to `nuts-node/test`. Since the conversion from from `error` to `problem.Probelm` now takes place within the test method, I've added a check to make sure the error actually is a problem before verifying the field's specified value.

core/echo_test.go
gerardsn: added test to check content type, and response body.

crypto/api/v1/api.go
gerardsn: changed

core/echo_test.go
```diff
@@ -102,3 +108,45 @@ func Test_getGroup(t *testing.T) {
...
+		assert.Equal(t, problem.ContentTypeJSON, resp.Header.Get("Content-Type"))
+
+		// Validate response body with expected problem
+		prb := NewProblem("problem title", http.StatusInternalServerError, "problem detail")
```
stevenvegt, line 131: you already created this problem on line 120 right?

test/problem.go
```diff
@@ -0,0 +1,73 @@
...
+	return prb
+}
+
+// AssertErrIsProblem asserts err is a *problem.Problem
+func AssertErrIsProblem(t assert.TestingT, err error) bool {
+	return assert.IsType(t, &problem2.Problem{}, err, "err is not a *problem.Problem")
+}
+
+// AssertErrProblemTitle asserts err is a *problem.Problem with the specified title
+func AssertErrProblemTitle(t assert.TestingT, title string, err error) bool {
```
stevenvegt, line 49: Why the use of `asssert.TestingT` instead of golang's `testing.T`?

stevenvegt (commented):
Nice improvements. 2 minor things / questions.

woutslakhorst (approved):

PR #191, opened 2021-03-26
Auth: fixed schemeManager config flag

reinkrul:


woutslakhorst (approved):
LOL, I love these problems since it shows integration has begun

reinkrul:
> LOL, I love these problems since it shows integration has begun

It's been there a while (also in the `nuts-auth`) and it worked through `nuts.yaml`, but not so using env variables. It's a good thing we're now using the app (differently?) and finding the bugs indeed.

PR #192, opened 2021-03-26
Re-enable Irma contract validation

### LINKED ISSUE #84: Auth: Implement IRMA contract validator

woutslakhorst:
Closes #84

### DIFF
--- a/auth/services/irma/signer.go
+++ b/auth/services/irma/signer.go
@@ -26,13 +26,19 @@ import (
 	"strings"
 
 	"github.com/mdp/qrterminal/v3"
+	"github.com/nuts-foundation/go-did/did"
 	"github.com/nuts-foundation/nuts-node/auth/contract"
 	"github.com/nuts-foundation/nuts-node/auth/logging"
 	"github.com/nuts-foundation/nuts-node/auth/services"
+	"github.com/nuts-foundation/nuts-node/vcr/concept"
+	"github.com/nuts-foundation/nuts-node/vcr/credential"
 	"github.com/pkg/errors"
 	irmago "github.com/privacybydesign/irmago"
 	"github.com/privacybydesign/irmago/server"
 )
+
+// ErrLegalEntityNotFound indicates that the legalEntity is unknown by its name/city
+var ErrLegalEntityNotFound = errors.New("legalEntity not found")
 
 // SessionPtr should be made private when v0 is removed
 type SessionPtr struct {
@@ -136,6 +142,35 @@ func (v Service) SigningSessionStatus(sessionID string) (contract.SigningSession
 		return result, nil
 	}
 	return nil, services.ErrSessionNotFound
+}
+
+func (v Service) legalEntityFromContract(sic *SignedIrmaContract) (*did.DID, error) {
+	params := sic.contract.Params
+
+	q, err := v.VCResolver.Registry().QueryFor(concept.OrganizationConcept)
+	if err != nil {
+		return nil, err
+	}
+	q.AddClause(concept.Eq(concept.OrganizationName, params[contract.LegalEntityAttr]))
+	q.AddClause(concept.Eq(concept.OrganizationCity, params[contract.LegalEntityCityAttr]))
+
+	c, err := v.VCResolver.Search(q)
+	if err != nil {
+		return nil, err
+	}
+	if len(c) == 0 {
+		return nil, ErrLegalEntityNotFound
+	}
+
+	cs := make([]credential.BaseCredentialSubject, 0)
+	// can't go wrong: it comes from JSON
+	_ = c[0].UnmarshalCredentialSubject(&cs)
+
+	if len(cs) == 0 {
+		return nil, ErrLegalEntityNotFound
+	}
+
+	return did.ParseDID(cs[0].ID)
 }
 
 // SigningSessionResult implements the SigningSessionResult interface and contains the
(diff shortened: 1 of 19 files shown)


vcr/interface.go
```diff
@@ -68,34 +69,38 @@ type Writer interface {
 	StoreRevocation(r credential.Revocation) error
 }
 
+// Resolver binds all read type of operations into an interface
+type Resolver interface {
+	// Registry returns the concept registry
+	Registry() concept.Registry
```
reinkrul, line 75: Maybe better to have it expose (and delegate) `QueryFor`? Because now you have access to `ConceptTemplate`, `Add`, `Transform` which looks a bit strange to me given its a resolving (read-only).

vcr/interface.go
```diff
@@ -68,34 +69,38 @@ type Writer interface {
...
 	// Issue creates and publishes a new VC.
 	// An optional expirationDate can be given.
 	// VCs are stored when the network has successfully published them.
 	Issue(vcToIssue vc.VerifiableCredential) (*vc.VerifiableCredential, error)
-	// Search for matching credentials based upon a query. It returns an empty list if no matches have been found.
-	Search(query concept.Query) ([]vc.VerifiableCredential, error)
 	// Resolve returns a credential based on its ID.
 	// The credential will still be returned in the case of ErrRevoked and ErrUntrusted.
 	// For other errors, nil is returned
 	Resolve(ID ssi.URI) (*vc.VerifiableCredential, error)
```
reinkrul, line 93: Move `Resolve` to `Resolver`?

vcr/interface.go
```diff
@@ -68,34 +69,38 @@ type Writer interface {
...
+// Resolver binds all read type of operations into an interface
+type Resolver interface {
+	// Registry returns the concept registry
+	Registry() concept.Registry
+	// Search for matching credentials based upon a query. It returns an empty list if no matches have been found.
+	Search(query concept.Query) ([]vc.VerifiableCredential, error)
+	// Verify checks if a credential is valid and trusted at the given time.
+	// The time check is optional, so credentials can be issued that will become valid.
+	// If valid no error is returned.
+	Verify(vcToVerify vc.VerifiableCredential, at *time.Time) error
```
reinkrul, line 81: I feel like `Verify` shouldn't be a member of `Resolver`. Does something break when you move it back to `VCR`?

auth/services/irma/signer.go
```diff
@@ -138,6 +144,35 @@ func (v Service) SigningSessionStatus(sessionID string) (contract.SigningSession
...
+
+	c, err := v.VCResolver.Search(q)
+	if err != nil {
+		return nil, err
+	}
+	if len(c) == 0 {
+		return nil, ErrLegalEntityNotFound
+	}
+
+	cs := make([]credential.BaseCredentialSubject, 0)
```
reinkrul, line 165: I'd handle the case when there's multiple matches (= error) since it either means someone made a mistake or there's a deployment scenario we haven't anticipated (multiple vendors having the same care organizations). Either way I think it's better to fail loudly rather than silently do the (possibly) wrong thing.

Furthermore when there's multiple matches for the same DID (multiple VCs with the same subject to the same DID) they can be considered equal. Although that's such an edge case we might not want to support (yet).

auth/services/irma/signer.go
```diff
@@ -138,6 +144,35 @@ func (v Service) SigningSessionStatus(sessionID string) (contract.SigningSession
...
+	// can't go wrong: it comes from JSON
+	_ = c[0].UnmarshalCredentialSubject(&cs)
+
+	if len(cs) == 0 {
```
reinkrul, line 169: Could there be multiple subjects?

reinkrul (changes requested):

reinkrul:
Have to add `organization.city` to index `2` in `vcr/assets/NutsOrganizationCredential.json.template`

vcr/interface.go
woutslakhorst: missed one yes.

vcr/interface.go
woutslakhorst: I split the Registry interface int to 2 parts. Also transform doesn't transform anything internal, it transforms the given object.

vcr/interface.go
woutslakhorst: nope

auth/services/irma/signer.go
woutslakhorst: Multiple DIDs with the same name/city would require a rewrite of the session handling part. I'll add an issue for that: the DID will have to be stored or passed in the request so the `legalEntityFromContract` can match against the DID and not search it. #196

auth/services/irma/signer.go
woutslakhorst: Currently this is not supported. The concept mapping also doesn't support it.

woutslakhorst:
> Have to add `organization.city` to index `2` in `vcr/assets/NutsOrganizationCredential.json.template`

done by #185

auth/services/irma/signer.go
reinkrul: Then shouldn't we for now fail if there's multiple matches instead of picking a random one?

reinkrul (changes requested):

auth/services/irma/signer.go
woutslakhorst: I think it would hinder us now in dev/test

reinkrul (approved):

PR #194, opened 2021-03-27
Auth: remove very verbose log statement

reinkrul:


woutslakhorst (commented):
This statement is most likely used during debugging.... So a `Debug()` could come in its place?

reinkrul:
> This statement is most likely used during debugging.... So a `Debug()` could come in its place?

Changed it a bit and made it debug.

woutslakhorst (approved):

PR #195, opened 2021-03-27
Fixes so Demo EHR can work with the development setup

reinkrul:


woutslakhorst (approved):
so the demo runs on 8080?

reinkrul:
> so the demo runs on 8080?

It runs on 8000, 8001 and 8002. I changed it previously so it works as-is but I figured it would be more fitting to change the `demo-ehr` configuration and just let the ports in the development config start at 8080. So I changed it back to `8080`. I removed the `bootstrap` node though, since it isn't a very defined role and actually "just another node" .

PR #197, opened 2021-03-29
Removed signing of nuts_auth_token when getting session status from Irma

### LINKED ISSUE #196: Auth: match entity from contract to actor DID.
When a contract signing session is retrieved, the DID is reverse engineered from the name/city of the contract. This should be done using the Actor DID that was used in the initial request. Otherwise only 1 DID can have a specific name/city.

woutslakhorst:
This was old code, now an extra step is required: createJwtBearerToken which does all the checks

Fixes #196


reinkrul (approved):
I like less code.

PR #204, opened 2021-03-31
Add generic way of checking relationship between node and DID

stevenvegt:
This is currently a draft of a proof of concept.

It introdces a VDR.OwnershipChecker which can be used in every api or service which needs to check if a DID is owned by this node.
I applied this pattern already in auth/api and vdr/api so you can see how this looks like. If you agree this PR will apply it on every relevant api handler.

### DIFF
--- a/crypto/interface.go
+++ b/crypto/interface.go
@@ -73,12 +73,15 @@ type PublicKeyStore interface {
 	RevokePublicKey(kid string, validTo time.Time) error
 }
 
-// PrivateKeyStore defines the functions for working with private keys.
-type PrivateKeyStore interface {
+type PrivateKeyChecker interface {
 	// PrivateKeyExists returns if the specified private key exists.
 	// If an error occurs, false is also returned
 	PrivateKeyExists(kid string) bool
+}
 
+// PrivateKeyStore defines the functions for working with private keys.
+type PrivateKeyStore interface {
+	PrivateKeyChecker
 	KeyCreator
 	JWSSigner
 	JWTSigner
--- a/vdr/api/v1/api.go
+++ b/vdr/api/v1/api.go
@@ -33,9 +33,15 @@ import (
 
 // Wrapper is needed to connect the implementation to the echo ServiceWrapper
 type Wrapper struct {
-	VDR            types.VDR
-	DocManipulator types.DocManipulator
+	VDR              types.VDR
+	DocManipulator   types.DocManipulator
+	OwnershipChecker types.OwnershipChecker
 }
+
+const (
+	deactivateErrTemplate = "could not deactivate document :%s"
+	updateErrTemplate     = "could not update document: %s"
+)
 
 // DeleteVerificationMethod accepts a DID and a KeyIdentifier of a verificationMethod and calls the DocManipulator
 // to remove the verificationMethod from the given document.
@@ -43,6 +49,11 @@ func (a *Wrapper) DeleteVerificationMethod(ctx echo.Context, didStr string, kidS
 	id, err := did.ParseDID(didStr)
 	if err != nil {
 		return ctx.String(http.StatusBadRequest, fmt.Sprintf("given DID could not be parsed: %s", err.Error()))
+	}
+	// Check if org is managed by this node
+	err = a.OwnershipChecker.OwnedByThisNode(*id)
+	if err != nil {
+		return err
 	}
 
 	kid, err := did.ParseDID(kidStr)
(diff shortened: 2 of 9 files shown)


crypto/interface.go
```diff
@@ -73,12 +73,15 @@ type PublicKeyStore interface {
 	RevokePublicKey(kid string, validTo time.Time) error
 }
 
-// PrivateKeyStore defines the functions for working with private keys.
-type PrivateKeyStore interface {
+type PrivateKeyChecker interface {
```
reinkrul, line 76: Not entirely sure whether this interface adds something or just increases complexity?

vdr/api/v1/api.go
```diff
@@ -128,23 +149,28 @@ func (a Wrapper) UpdateDID(ctx echo.Context, targetDID string) error {
...
 // It returns a 200 and an empty body if the deactivation was successful.
-func (a *Wrapper) DeactivateDID(ctx echo.Context, targetDID string) error {
-	id, err := did.ParseDID(targetDID)
+func (a *Wrapper) DeactivateDID(ctx echo.Context, didStr string) error {
+	id, err := did.ParseDID(didStr)
 	if err != nil {
 		return ctx.String(http.StatusBadRequest, fmt.Sprintf("given DID could not be parsed: %s", err.Error()))
 	}
+	// Check if org is managed by this node
+	err = a.OwnershipChecker.OwnedByThisNode(*id)
```
reinkrul, line 167: this validation step should be performed by VDR service implementation, because now you might have to call it from multiple locations before deactivating/updating.

vdr/ownership-checker.go
```diff
@@ -0,0 +1,25 @@
...
+	"github.com/nuts-foundation/nuts-node/crypto"
+	"github.com/nuts-foundation/nuts-node/vdr/types"
+)
+
+type OwnershipChecker struct {
+	KeyResolver KeyResolver
+	KeyChecker  crypto.PrivateKeyChecker
+}
+
+func (c OwnershipChecker) OwnedByThisNode(id did.DID) (error) {
```
reinkrul, line 15: Just let VDR implement an OwnershipChecker interface?

And although it's a high level function,  I think the name is extremely high level (`node ownership`), possibly too high level for `VDR`. We planned to have `didman` for very high level stuff imo.

`Assertable` (Wout's suggestion) sounds maybe a bit (too) technical to me? Maybe `CanManage`/`CanUpdate`?

reinkrul (changes requested):

crypto/interface.go
stevenvegt: Single responsibility. I dont need the jungle and the monkey, just the banana: checking the existence of a private key, so the PrivateKeyStore is clearly way to convoluted with a keyCreator and even several JWS signing methods.

vdr/api/v1/api.go
stevenvegt: Yes, that is the idea. I need to call it from API's so the application can easily check if an operation is permitted. To make everything dependant on VDR seems wrong, since that is manly for working with documents from the data registry: Reesolve, Create and Update.
This check is not only needed by VDR but also by auth/notary. Therfore it is put in a seperate function.

vdr/ownership-checker.go
stevenvegt: Imo this is a combined functionality over the keystore and the document store. So neither one of those should implement it. It is meant to be an Authorization cheker if a user should be able to update a DID Document or can use a key to sign a token or create a VC etc.

vdr/api/v1/api.go
reinkrul: If Auth/Notary needs `OwnershipChecker` which in turns needs `VDR` (through `KeyResolver`), Auth/Notary still transitively depend on VDR. So I wouldn't let it depend on VDR, but on the `OwnershipChecker` (to be renamed?) interface which is implemented by VDR?

reinkrul: And I still think validating business rules in the API layer which aren't validated in the service layer is bad practice. So whether `OwnershipChecker` is implemented by `VDR` or not, `VDR` itself should call `OwnershipChecker`. API layer is a presentation layer, only responsible for translating between HTTP/REST and Go services.

woutslakhorst: The check is input validation, the business logic doesn't care if this check succeeds or not. It prevents a more annoying error in the next API call. Placing it in the business logic creates more spaghetti code than placing it here.

woutslakhorst:
We'll remove the public keystore first and see if it cleans some stuff up.

PR #205, opened 2021-04-01
Differentiate between SignJwt error returns

### LINKED ISSUE #182: wrong status code for SignJWT api error returns
It returns a 400 for all signJwt errors. It should return 400 for `ErrKeyNotFound` but a 500 in all other cases.

gerardsn:
Now returns 400 for `ErrKeyNotFound` and 500 in all other cases
closes #182


reinkrul (approved):
LGTM

PR #206, opened 2021-04-01
added git tag, branch and hash. Also added Os/Arch to diagnostics

### LINKED ISSUE #62: Log version at startup and include it in diagnostics page
Now there's no way to determine what version a running node has. It would be convenient if it was displayed on the status page and logged on node startup;

- Git branch that the build was made from
- Git commit ID
- Build time(?)
- Git tag (for identifying releases?)

woutslakhorst:
Closes #62 

- [x] github actions also injects stuff

### DIFF
--- a/.github/workflows/build-images.yaml
+++ b/.github/workflows/build-images.yaml
@@ -57,9 +57,9 @@ jobs:
       - name: Build and push
         uses: docker/build-push-action@v2
         env:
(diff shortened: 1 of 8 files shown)


reinkrul (approved):

woutslakhorst:
forgot some tests..

PR #213, opened 2021-04-06
use compound service for oauth endpoint resolution

### LINKED ISSUE #83: Auth: implement oauth endpoint selection by bolt
Endpoints are specified in the DID Documents. The RFC describes the use of compound endpoints to use endpoints from another DID document. This needs to be supported by Nuts Auth.

woutslakhorst:
Closes #83 

Waiting for nuts-foundation/nuts-specification#51

### DIFF
--- a/auth/services/util.go
+++ b/auth/services/util.go
@@ -1,10 +1,10 @@
 package services
 
 import (
+	"fmt"
 	ssi "github.com/nuts-foundation/go-did"
 	"github.com/nuts-foundation/go-did/did"
 	"github.com/nuts-foundation/nuts-node/vdr/types"
-	"fmt"
 	"time"
 )
 
--- a/auth/services/util.go
+++ b/auth/services/util.go
@@ -2,20 +2,22 @@ package services
 
 import (
 	"fmt"
+	"time"
+
 	ssi "github.com/nuts-foundation/go-did"
 	"github.com/nuts-foundation/go-did/did"
 	"github.com/nuts-foundation/nuts-node/vdr/types"
-	"time"
 )
 
-// ResolveServiceURL finds the endpoint with the given type of the given holder referenced from the given service, and unmarshals it as single URL.
-// It returns the endpoint ID and URL, or an error if anything went wrong;
+// ResolveCompoundServiceURL finds an endpoint for a compound service. A compound service describes a use-case for which multiple absolute endpoints are required.
+// ResolveCompoundServiceURL will find the compound service. It'll check if the compound service has an absolute endpoint for the given type.
+// If it exists, it will resolve that endpoint and return the serviceID of the absolute endpoint and the URL as string.
 // - holder document can't be resolved,
 // - service with given type doesn't exist,
 // - multiple services match,
 // - referenced service doesn't resolve.
 // - referenced service not unique.
-func ResolveServiceURL(resolver types.DocResolver, holder did.DID, serviceType string, endpointType string, validAt *time.Time) (ssi.URI, string, error) {
+func ResolveCompoundServiceURL(resolver types.DocResolver, holder did.DID, serviceType string, endpointType string, validAt *time.Time) (ssi.URI, string, error) {
 	doc, _, err := resolver.Resolve(holder, &types.ResolveMetadata{ResolveTime: validAt})
 	if err != nil {
 		return ssi.URI{}, "", err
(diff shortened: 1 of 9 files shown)


woutslakhorst:
not sure about the `scope` to `service` rename. We use `scope` at some other points as well, but in our current way of thinking this relates to bolts...

reinkrul:
> not sure about the `scope` to `service` rename. We use `scope` at some other points as well, but in our current way of thinking this should relate to bolts...

I think an alternative name for `scope` is appropriately, since OAuth has a concept of "scope" that limits the resources accessible with the authentication token. Which might map to the concept we need, but OAuth 2 expresses `scope` in the request URL, not in the authentication token. Because the concepts sort of map but the technical representation differs I think we should rename it to something else.

`service` is closer to the DID language. I think we decided to keep the term "bolt" out of the implementation (if we stick to something closer to "bolt" we could consider "use case"?). Maybe "service" is fine for now.

woutslakhorst:
Added #218 for that.

auth/services/util.go
```diff
@@ -1,22 +1,87 @@
...
 import (
+	"fmt"
 	ssi "github.com/nuts-foundation/go-did"
 	"github.com/nuts-foundation/go-did/did"
 	"github.com/nuts-foundation/nuts-node/vdr/types"
 	"time"
 )
 
-// ResolveEndpointURL finds the endpoint with the given type of the given holder and unmarshals it as single URL.
+// ResolveServiceURL finds the endpoint with the given type of the given holder referenced from the given service, and unmarshals it as single URL.
```
reinkrul, line 11: I find it hard to wrap my head around this comment, especially the "referenced from the given service".

Probably because the name and signature of this function confuses me: it contains a `serviceType` and `endpointType` which are from DID spec perspective, the same thing. A `service` has a `type` property, and a `serviceEndpoint` which can be whatever is spec'd. I'd make the naming more explicit/elaborate;

`ResolveCompoundServiceURL(resolver, holder, compoundServiceType, serviceType, validAt)`

auth/services/util.go
```diff
@@ -1,22 +1,87 @@
...
+
+	uri, url, err := resolveEndpointURL(resolver, *refDID, validAt)
+	if err != nil {
+		return ssi.URI{}, "", fmt.Errorf("failed to resolve endpoint from service (did=%s, service=%s, type=%s)", doc.ID, endpointReference, endpointType)
+	}
+
+	return uri, url, nil
+}
+
+type compoundServiceType map[string]string
```
reinkrul, line 46: I'd rather not have types postfix `Type` (just like in C# `SomeInterface`). Maybe `compoundServiceEndpoint`?

auth/services/util.go
```diff
@@ -1,22 +1,87 @@
...
+
+func resolveCompoundService(doc did.Document, serviceType string) (service compoundServiceType, err error) {
```
reinkrul, line 48: why named parameters while you don't use them?

auth/services/util.go
```diff
@@ -1,22 +1,87 @@
...
+	cs := compoundServiceType{}
+	var services []did.Service
+	for _, service := range doc.Service {
+		if service.Type == serviceType {
+			services = append(services, service)
+		}
+	}
+	if len(services) != 1 {
+		return cs, fmt.Errorf("incorrect number of services (did=%s, service=%s)", doc.ID, serviceType)
```
reinkrul, line 57: 'incorrect number of services' doesn't tell the operator what was expected and/or how many were found. I'd at least rewrite it to; "expected 1 service" (and maybe add; "found %d")

auth/services/util.go
```diff
@@ -1,22 +1,87 @@
...
+	if err != nil {
+		return ssi.URI{}, "", err
+	}
+
+	typeQuery, ok := serviceURI.Query()["type"]
+	if !ok {
+		return ssi.URI{}, "", fmt.Errorf("incorrect type selector in service reference: %s", serviceURI.String())
+	}
+	if len(typeQuery) != 1 {
+		return ssi.URI{}, "", fmt.Errorf("incorrect type selector in service reference: %s", serviceURI.String())
```
reinkrul, line 78: "expected one"?

docs/_static/auth/v1.yaml
```diff
@@ -453,17 +453,21 @@ components:
           type: string
           description: |
             Base64 encoded IRMA contract conaining the identity of the performer
-        scope:
+        service:
           type: string
-          description: Space-delimited list of strings. For what kind of operations can the access token be used? Scopes will be specified for each use-case
-          example: nuts-sso nuts-patient-transfer
+          description: For which service can the access token be used? The right endpoint is selected based on the service.
```
reinkrul, line 458: Question mark looks a bit strange in the description (it isn't a question)

reinkrul (changes requested):

auth/services/util.go
woutslakhorst: Renamed the service and added some more lines of documentation.

auth/services/util.go
woutslakhorst: intellij keeps suggesting `type` postfixes

auth/services/util.go
woutslakhorst: copy-feest

docs/_static/auth/v1.yaml
woutslakhorst: rewrote

reinkrul (approved):

PR #217, opened 2021-04-07
Validate aud when creating accesstoken

### LINKED ISSUE #215: Auth: implement audience (aud) validation when getting access token
according to RFC003 §5.2.1.6

the todo @ auth/services/oauth/oauth.go#132

woutslakhorst:
Closes #215 

After merge fix oauth e2e test

### DIFF
--- a/auth/services/oauth/oauth.go
+++ b/auth/services/oauth/oauth.go
@@ -127,9 +127,9 @@ func (s *service) CreateAccessToken(request services.CreateAccessTokenRequest) (
 	}
 
 	// validate the endpoint in aud, according to RFC003 §5.2.1.6
-	// the aud field must have the identifier of the endpoint registered by the vendor of this node!
-	// this is needed to prevent relay attacks.
-	// todo: implement when services and endpoints in registry have been implemented (https://github.com/nuts-foundation/nuts-registry/issues/156)
+	if err := s.validateAud(&context); err != nil {
+		return nil, err
+	}
 
 	// validate the legal base, according to RFC003 §5.2.1.7 if sid is present
 	if err = s.validateLegalBase(context.jwtBearerTokenClaims); err != nil {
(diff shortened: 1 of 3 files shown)


auth/services/oauth/oauth.go
```diff
@@ -155,6 +155,29 @@ func (s *service) validateActor(context *validationContext) error {
 	return nil
 }
 
+// check if the aud service identifier matches the oauth endpoint of the requested scope
+func (s *service) validateAud(context *validationContext) error {
```
reinkrul, line 159: rest of the func names are writtng in full, so let's just say `validateAudience`

auth/services/oauth/oauth.go
```diff
@@ -155,6 +155,29 @@ func (s *service) validateActor(context *validationContext) error {
...
+	if len(context.jwtBearerToken.Audience()) != 1 {
+		return errors.New("aud does not contain a single URI")
+	}
+	audience := context.jwtBearerToken.Audience()[0]
+	scope := context.jwtBearerTokenClaims.Scope
+	// parsing is already done in a previous check
+	subject, _ := did.ParseDID(context.jwtBearerToken.Subject())
+	iat := context.jwtBearerToken.IssuedAt()
+
+	uri, _, err := services.ResolveCompoundServiceURL(s.didResolver, *subject, scope, services.OAuthEndpointType, &iat)
```
reinkrul, line 169: when we start supporting multiple current version, should this fail with an error? If so, all is OK, if not we might want to add a TODO here

reinkrul (changes requested):

auth/services/oauth/oauth.go
woutslakhorst: `ResolveCompoundServiceURL` already returns an error if `len(services) != 1`

reinkrul (approved):

PR #227, opened 2021-04-13
Move KeyResolver to vdr

### LINKED ISSUE #222: Resolve public keys from VDR
remove the public key store

woutslakhorst:
closes #222 

The store had to be extracted, since that's the common component used by both vdr and network


reinkrul (approved):

vdr/doc-updater.go
```diff
@@ -17,12 +17,14 @@ type DocUpdater struct {
 	KeyCreator nutsCrypto.KeyCreator
 	// VDR is used for resolving before and updating DID documents after the operation has been performed
```
reinkrul, line 18: not for resolving, since that's what `DocResolver` now is for ("what's up, doc?")

vdr/doc-updater.go
```diff
@@ -17,12 +17,14 @@ type DocUpdater struct {
...
 	VDR types.VDR
+	// DocResolver is used for resolving DID Documents
+	DocResolver types.DocResolver
```
reinkrul, line 21: just "Resolver"?

vdr/doc/resolvers.go
```diff
@@ -0,0 +1,125 @@
...
+	"fmt"
+	"time"
+
+	ssi "github.com/nuts-foundation/go-did"
+	"github.com/nuts-foundation/go-did/did"
+	"github.com/nuts-foundation/nuts-node/vdr/types"
+)
+
+// Resolver implements the DocResolver interface with a types.Store as backend
+type Resolver struct {
```
reinkrul, line 35: I find the naming of these types a bit odd, partly because we abbreviate `Document` to `Doc` in `DocResolver` which we never somewhere else, and because the type name of the implementation of the interface is shorter and less-specific than the interface. It might be my Java background, but I'd think of one of the options below;

- StoreResolver implements Resolver
- StoreDocResolver implements DocResolver
- StoreDocumentResolver implements DocumentResolver

network/dag/sigverifier.go
```diff
@@ -6,7 +6,7 @@ import (
 
 	"github.com/lestrrat-go/jwx/jwa"
 	"github.com/lestrrat-go/jwx/jws"
-	"github.com/nuts-foundation/nuts-node/crypto"
+	"github.com/nuts-foundation/nuts-node/vdr/types"
```
reinkrul, line 9: I still find the "types" package an odd thing, especially because not all engines have it (some manage to be a dependency to other engines without it).

"packages should describe the functionality they provide, not the sort of thing they contain" (it's the same as "no util package" imo)

reinkrul (changes requested):

vdr/doc/resolvers.go
woutslakhorst: There's more to do in the VDR package to clean stuff up. 
The implementation is missing the `Doc` prefix because of the package name, it's used as `doc.Resolver`.
The biggest problem is that both the `Store` and `Resolver` implement the same interface. Refactoring that is beyond the scope of this PR.

network/dag/sigverifier.go
woutslakhorst: beyond the scope of this PR.

reinkrul (approved):

PR #229, opened 2021-04-14
Iss221 remove controller recursion

### LINKED ISSUE #221: Update document controller check according to spec
https://github.com/nuts-foundation/nuts-specification/pull/46

- [ ] remove recursion for finding controllers #229 
- [ ] use correct key from document #229

woutslakhorst:
Closes #221

There was duplication of controller resolution between VDR and ambassador. Used #227 as basis since it resolves part of the requirements.

### DIFF
--- a/vdr/ambassador.go
+++ b/vdr/ambassador.go
@@ -28,6 +28,7 @@ import (
 
 	ssi "github.com/nuts-foundation/go-did"
 	"github.com/nuts-foundation/nuts-node/vdr/doc"
+	"github.com/nuts-foundation/nuts-node/vdr/store"
 
 	"github.com/lestrrat-go/jwx/jwk"
 	"github.com/nuts-foundation/go-did/did"
@@ -118,7 +119,7 @@ func (n *ambassador) handleCreateDIDDocument(transaction dag.SubscriberTransacti
 		return fmt.Errorf("unable to generate network transaction signing key thumbprint: %w", err)
 	}
 
-	// Check if signingKey is one of the keys embedded in the authenticationMethod
+	// Check if signingKey is one of the keys embedded in the CapabilityInvocation
 	didDocumentAuthKeys := proposedDIDDocument.CapabilityInvocation
 	if documentKey, err := n.findKeyByThumbprint(signingKeyThumbprint, didDocumentAuthKeys); documentKey == nil || err != nil {
 		if err != nil {
@@ -148,12 +149,12 @@ func (n *ambassador) handleUpdateDIDDocument(document dag.SubscriberTransaction,
 	}
 
 	// Resolve controllers of current version (could be the same document)
-	didControllers, err := n.docResolver.ResolveControllers([]did.Document{*currentDIDDocument})
+	didControllers, err := n.docResolver.ResolveControllers(*currentDIDDocument)
 
 	var controllerVerificationRelationships []did.VerificationRelationship
 	for _, didCtrl := range didControllers {
-		for _, auth := range didCtrl.CapabilityInvocation {
-			controllerVerificationRelationships = append(controllerVerificationRelationships, auth)
+		for _, capInv := range didCtrl.CapabilityInvocation {
+			controllerVerificationRelationships = append(controllerVerificationRelationships, capInv)
 		}
 	}
 
@@ -203,7 +204,7 @@ func (n *ambassador) handleUpdateDIDDocument(document dag.SubscriberTransaction,
 		Created:     currentDIDMeta.Created,
 		Updated:     &updatedAt,
 		Hash:        document.PayloadHash(),
-		Deactivated: isDeactivated(&proposedDIDDocument),
+		Deactivated: store.IsDeactivated(proposedDIDDocument),
 	}
 	return n.didStore.Update(proposedDIDDocument.ID, currentDIDMeta.Hash, proposedDIDDocument, &documentMetadata)
 }
--- a/vdr/ambassador.go
+++ b/vdr/ambassador.go
@@ -51,6 +51,7 @@ type ambassador struct {
 	networkClient network.Transactions
 	didStore      types.Store
 	keyResolver   types.KeyResolver
+	docResolver   types.DocResolver
 }
 
 // NewAmbassador creates a new Ambassador,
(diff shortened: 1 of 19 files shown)


vdr/doc/resolvers.go
```diff
@@ -40,6 +40,60 @@ func (d Resolver) Resolve(id did.DID, metadata *types.ResolveMetadata) (*did.Doc
...
+			return nil, fmt.Errorf("unable to resolve controllers: %w", err)
+		}
+		leaves = append(leaves, *node)
+	}
+
+	// filter deactivated
+	j := 0
+	for i, doc := range leaves {
+		if !isDeactivated(doc) {
+			leaves[i] = doc
```
reinkrul, line 85: This way of filtering in a memory-optimized way is too smart for its own good and reeks of premature optimization.

`leaves[i] = doc` should be `leaves[j] = doc`

(see committed unit test which now fails)

vdr/types/interface.go
```diff
@@ -35,6 +35,9 @@ type DocResolver interface {
 	// If metadata is provided then the result is filtered or scoped on that metadata.
 	// It returns ErrNotFound if there are no corresponding DID documents or when the DID Documents are disjoint with the provided ResolveMetadata
 	Resolve(id did.DID, metadata *ResolveMetadata) (*did.Document, *DocumentMetadata, error)
+	// ResolveControllers accepts a list of documents and finds their controllers
+	// The resulting list are documents who control themselves
```
reinkrul, line 39: The addition of "The resulting list are documents who control themselves" can now be removed I think.

vdr/types/interface.go
```diff
@@ -86,7 +89,11 @@ type KeyResolver interface {
 
 // Store is the interface that groups all low level VDR DID storage operations.
 type Store interface {
-	DocResolver
+	// Resolve returns a DID Document for the provided DID.
```
reinkrul, line 92: "a DID Document" -> "the DID Document" (not just any, it's _the_ document)

vdr/vdr.go
```diff
@@ -150,7 +152,7 @@ func (r VDR) Update(id did.DID, current hash.SHA256Hash, next did.Document, _ *t
 		return err
 	}
 
-	keyID := controllers[0].Authentication[0].ID.String()
+	keyID := controllers[0].CapabilityInvocation[0].ID.String()
```
reinkrul, line 155: should the alterating controller be specified in the future? Maybe we're missing a TODO/issue here.

The controller might not be controlled by the local Nuts node (edge case though).

vdr/vdr.go
```diff
@@ -165,58 +167,5 @@ func (r VDR) Update(id did.DID, current hash.SHA256Hash, next did.Document, _ *t
...
-	if err != nil {
-		return nil, err
-	}
-	// Merge local leaves and new found leaves
-	for _, leave := range newLeaves {
-		leaves = append(leaves, leave)
-	}
-
-	return leaves, nil
+	return len(document.Controller) == 0 && len(document.CapabilityInvocation) == 0
```
reinkrul, line 170: duplicate func? (also in memoryStore)

reinkrul (changes requested):

vdr/ambassador.go
```diff
@@ -59,6 +60,7 @@ func NewAmbassador(networkClient network.Transactions, didStore types.Store) Amb
 		networkClient: networkClient,
 		didStore:      didStore,
 		keyResolver:   doc.KeyResolver{Store: didStore},
+		docResolver:   doc.Resolver{Store: didStore},
```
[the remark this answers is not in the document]
woutslakhorst: That would add 2 extra params (not needed). The resolvers are also nothing more than a wrapper. They are not real 'services'. It must also be clear that the `store` can be passed along without creating a circular dependency.

vdr/ambassador.go
```diff
@@ -117,7 +119,7 @@ func (n *ambassador) handleCreateDIDDocument(transaction dag.SubscriberTransacti
 	}
 
 	// Check if signingKey is one of the keys embedded in the authenticationMethod
-	didDocumentAuthKeys := proposedDIDDocument.Authentication
+	didDocumentAuthKeys := proposedDIDDocument.CapabilityInvocation
```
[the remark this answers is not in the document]
woutslakhorst: indeed...
done

vdr/doc/creator.go
```diff
@@ -112,7 +112,7 @@ func (n Creator) Create() (*did.Document, error) {
 		ID:      didID,
 	}
 
-	doc.AddAuthenticationMethod(verificationMethod)
+	doc.AddCapabilityInvocation(verificationMethod)
```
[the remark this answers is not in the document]
woutslakhorst: already done, latest merge on e2e tests was based upon this PR

vdr/doc/resolvers.go
```diff
@@ -40,6 +40,60 @@ func (d Resolver) Resolve(id did.DID, metadata *types.ResolveMetadata) (*did.Doc
 	return d.Store.Resolve(id, metadata)
 }
 
+// ResolveControllers accepts a list of documents and finds their controllers
+func (d Resolver) ResolveControllers(input []did.Document) ([]did.Document, error) {
```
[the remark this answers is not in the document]
woutslakhorst: Indeed will change to 1

vdr/doc/resolvers.go
woutslakhorst: it indeed failed, but is a very common thing within Go

vdr/vdr.go
woutslakhorst: the `OwnedByThisNode` helper should fix that

vdr/vdr.go
woutslakhorst: Yes, dependencies are a bit of a problem here. We could remove the check from the store and only do it in the VDR?

woutslakhorst: moved it to a func in `store`

vdr/doc/resolvers.go
```diff
@@ -40,6 +40,53 @@ func (d Resolver) Resolve(id did.DID, metadata *types.ResolveMetadata) (*did.Doc
 	return d.Store.Resolve(id, metadata)
 }
 
+// ResolveControllers accepts a list of documents and finds their controllers
```
reinkrul, line 43: godoc still incorrect (list of documents)

reinkrul (approved):

vdr/doc/resolvers.go
woutslakhorst: removed it, since the doc is correct on the interface

PR #233, opened 2021-04-16
Iss232 fix service in token

### LINKED ISSUE #232: auth: service scope not set in bearer token
This prevents the custodian from checking the correct aud

woutslakhorst:
Fixes #232


auth/services/oauth/oauth.go
```diff
@@ -291,6 +291,8 @@ func claimsFromRequest(request services.CreateJwtBearerTokenRequest, audience st
 	result[jwt.IssuerKey] = request.Actor
 	result[jwt.NotBeforeKey] = 0
 	result[jwt.SubjectKey] = request.Custodian
+	result[jwt.SubjectKey] = request.Custodian
```
reinkrul, line 294: duplicate?

reinkrul (changes requested):

auth/services/oauth/oauth.go
woutslakhorst: incorrect merge I guess

reinkrul (approved):

Issue #231, opened 2021-04-15
VDR: implement conflict resolution


woutslakhorst:
With the latest update of RFC006, conflict detection and resolvement for DID Documents can be implemented. 

- did merger
- conflict detection
- conflict diagnostics
- change `tx.prevs` for did updates.

Issue #109, opened 2021-03-03
VDR: Make in-memory DID Document store persistent


reinkrul:

stevenvegt:
Or, make keystore also in memory?

woutslakhorst:
It'll have to be persistent at some point otherwise node startup will become slower and slower and slower

reinkrul:
Or store it in bbolt which memory-maps a file, best of both worlds :-D

woutslakhorst:
Storage does need to take processed vs unprocessed into account.
The storage layer will have to comply with ACID properties.

reinkrul:
> It does need to play well with the reloading of the DAG. Prevent duplicate processing and such.
> When replaying the DAG, how would we know a Tx is new or already processed?

I see no other way other than manually administering it in the subscriber or introducing a message broker (like Nats Streaming) that keeps a cursor for every subscriber.

woutslakhorst:
When storage is built, change `vdr.Diagnostics()` so it doesn't iterate over all documents.

reinkrul:
Still needs to be integrated (its there, but not used)

dmeijboom:
> Still needs to be integrated (its there, but not used)

Yeah indeed. It was automatically closed (by accident) as I didn't remove the `Fixes: #x` in the description.

woutslakhorst:
Pickup when **storage** is better defined.

PR #236, opened 2021-04-16
add conflict detection for did documents

woutslakhorst:
first part of #231 to detect conflicted docs.


vdr/types/common.go
```diff
@@ -52,6 +52,8 @@ type DocumentMetadata struct {
 	Updated *time.Time `json:"updated,omitempty"`
 	// Hash of DID document bytes. Is equal to payloadHash in network layer.
 	Hash hash.SHA256Hash `json:"hash"`
+	// Transactions points to the transaction(s) that created this DID Document. If multiple transactions are listed, the DID Document is conflicted
```
reinkrul, line 55: "that created this document" is only true for the initial transaction?

Isn't it more like "points to the previous versions of the DID document, or the current version in case of a new DID document"?

vdr/types/common.go
```diff
@@ -52,6 +52,8 @@ type DocumentMetadata struct {
...
+	Transactions []hash.SHA256Hash `json:"txs"`
```
reinkrul, line 56: I find "Transactions" very non-descriptive; the name doesn't explain what the meaning of the field is, just what it contains.

vdr/vdr.go
```diff
@@ -91,7 +91,21 @@ func (r *VDR) Configure(_ core.ServerConfig) error {
 
 // Diagnostics returns the diagnostics for this engine
 func (r *VDR) Diagnostics() []core.DiagnosticResult {
-	return []core.DiagnosticResult{}
+	// return # conflicted docs
+	count := 0
+	r.store.Iterate(func(doc did.Document, metadata types.DocumentMetadata) error {
+		if metadata.IsConflicted() {
```
reinkrul, line 97: It looks like this is going to be expensive when there are many documents?

vdr/vdr.go
```diff
@@ -91,7 +91,21 @@ func (r *VDR) Configure(_ core.ServerConfig) error {
 
 // Diagnostics returns the diagnostics for this engine
 func (r *VDR) Diagnostics() []core.DiagnosticResult {
-	return []core.DiagnosticResult{}
+	// return # conflicted docs
+	count := 0
+	r.store.Iterate(func(doc did.Document, metadata types.DocumentMetadata) error {
```
reinkrul, line 96: `doc` is unused -> `_`

vdr/vdr.go
```diff
@@ -91,7 +91,21 @@ func (r *VDR) Configure(_ core.ServerConfig) error {
 
 // Diagnostics returns the diagnostics for this engine
 func (r *VDR) Diagnostics() []core.DiagnosticResult {
```
reinkrul, line 93: Should we already add diagnostics on one's own conflicted documents, which can then be monitored by an administrator (or preferably, tooling which alerts)

reinkrul (changes requested):

vdr/types/common.go
woutslakhorst: "that created the current version of this DID Document."

vdr/types/common.go
woutslakhorst: Called it `SourceTransactions`

vdr/vdr.go
woutslakhorst: I'll add something to #109, since you'll require an index and search functionality. It's better to do this when storage changes.

vdr/vdr.go
woutslakhorst: Issue added, since multiple PRs are required first: #241

reinkrul (approved):

PR #237, opened 2021-04-16
Add DID Document merge functionality

woutslakhorst:
part of #231


woutslakhorst:
@reinkrul g2g

reinkrul:
Can we have a more describing title for this PR?

vdr/doc/merge.go
```diff
@@ -0,0 +1,209 @@
...
+		for _, context := range doc.Context {
+			contexts[context.String()] = context
+		}
+	}
+	for _, context := range contexts {
+		result.Context = append(result.Context, context)
+	}
+}
+
+// mergeServices merges keys based upon their ID. The ID is derived from the public key.
```
reinkrul, line 76: incorrect godoc

vdr/doc/merge.go
```diff
@@ -0,0 +1,209 @@
...
+// MergeConflicted merges two DID Documents that share the same ID
+func MergeConflicted(docA did.Document, docB did.Document) (*did.Document, error) {
+	result := &did.Document{}
+	docs := []did.Document{docA, docB}
+
+	if !docA.ID.Equals(docB.ID) {
+		return nil, ErrDiffID
+	}
+
+	mergeBasics(docs, result)
```
reinkrul, line 43: I'm thinking, everything (except `ID`) are basically sets that are merged into 1. Can't we simply unmarshal into a `map[string]interface{}` and merge those 2 maps? (you might still want/have to sort afterwards though).

vdr/doc/merge.go
```diff
@@ -0,0 +1,209 @@
...
+
+	ssi "github.com/nuts-foundation/go-did"
+	"github.com/nuts-foundation/go-did/did"
+)
+
+// ErrDiffID is returned when a merge is done on documents that do not share the same ID
+var ErrDiffID = errors.New("no matching IDs")
+
+// MergeConflicted merges two DID Documents that share the same ID
+func MergeConflicted(docA did.Document, docB did.Document) (*did.Document, error) {
```
reinkrul, line 35: I'd not call the func "Conflicted" because that dictates the usage/context of this function. Isn't it simply `MergeDocuments`? (also; we don't have other document merges so there won't be any confusion)

It's like reading a byte stream which happens to be from disk and calling it `ReadBytesFromDisk(stream)`

reinkrul (changes requested):

vdr/doc/merge.go
woutslakhorst: I was already looking for a way to reduce the similar code. This might be doable.

woutslakhorst: This does more harm than good. Each method can contain an array with a map[string]interface{} or a reference which is just a string.
Unmarshalling in `map[string]interface{}` would also result in several casts on `interface{}` and we should also be aware for a combination of usages and different usages per doc. The resulting codes has less duplication but is way harder to read and thus to maintain.

reinkrul (approved):

PR #243, opened 2021-04-20
Complete did.Document conflict resolution integration

### LINKED ISSUE #231: VDR: implement conflict resolution

woutslakhorst:
Closes #231

### DIFF
--- a/network/dag/transaction.go
+++ b/network/dag/transaction.go
@@ -110,7 +110,7 @@ type Transaction interface {
 }
 
 // NewTransaction creates a new unsigned transaction. Parameters payload and payloadType can't be empty, but prevs is optional.
-// Prevs must not contain empty or invalid hashes.
+// Prevs must not contain empty or invalid hashes. Duplicate prevs will be removed when given.
 func NewTransaction(payload hash.SHA256Hash, payloadType string, prevs []hash.SHA256Hash) (UnsignedTransaction, error) {
 	if !ValidatePayloadType(payloadType) {
 		return nil, errInvalidPayloadType
@@ -120,13 +120,29 @@ func NewTransaction(payload hash.SHA256Hash, payloadType string, prevs []hash.SH
 			return nil, errInvalidPrevs
 		}
 	}
+
+	// deduplicate prevs
+	deduplicated := make([]hash.SHA256Hash, 0)
+	for _, prev := range prevs {
+		found := false
+		for _, dd := range deduplicated {
+			if dd.Equals(prev) {
+				found = true
+				break
+			}
+		}
+		if !found {
+			deduplicated = append(deduplicated, prev)
+		}
+	}
+
 	result := transaction{
 		payload:     payload,
 		payloadType: payloadType,
 		version:     currentVersion,
 	}
-	if len(prevs) > 0 {
-		result.prevs = append(prevs)
+	if len(deduplicated) > 0 {
+		result.prevs = append(deduplicated)
 	}
 	return &result, nil
 }
--- a/network/dag/transaction.go
+++ b/network/dag/transaction.go
@@ -77,6 +77,7 @@ type SubscriberTransaction interface {
 	Signable
 	Referenceable
 	PayloadReferencer
+	NetworkHeader
 }
 
 // Signable groups a set of functions to access information about a implementors signature.
@@ -110,7 +111,7 @@ type Transaction interface {
 
 // NewTransaction creates a new unsigned transaction. Parameters payload and payloadType can't be empty, but prevs is optional.
 // Prevs must not contain empty or invalid hashes.
-func NewTransaction(payload hash.SHA256Hash, payloadType string, prevs []hash.SHA256Hash, additionalFields ...FieldOpt) (UnsignedTransaction, error) {
+func NewTransaction(payload hash.SHA256Hash, payloadType string, prevs []hash.SHA256Hash) (UnsignedTransaction, error) {
 	if !ValidatePayloadType(payloadType) {
 		return nil, errInvalidPayloadType
 	}
@@ -127,14 +128,8 @@ func NewTransaction(payload hash.SHA256Hash, payloadType string, prevs []hash.SH
 	if len(prevs) > 0 {
 		result.prevs = append(prevs)
 	}
-	for _, field := range additionalFields {
-		field(&result)
-	}
 	return &result, nil
 }
-
-// FieldOpt defines a function for specifying fields on a transaction.
-type FieldOpt func(target *transaction)
 
 // ValidatePayloadType checks whether the payload type is valid according to RFC004.
 func ValidatePayloadType(payloadType string) bool {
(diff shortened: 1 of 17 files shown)


woutslakhorst:
ready for review

network/dag/transaction.go
```diff
@@ -77,6 +77,7 @@ type SubscriberTransaction interface {
 	Signable
 	Referenceable
 	PayloadReferencer
+	NetworkHeader
```
reinkrul, line 80: Thanks for making me feel clairvoyant when we discussed introducing this type :-p

network/network.go
```diff
@@ -156,12 +156,16 @@ func (n *Network) ListTransactions() ([]dag.Transaction, error) {
...
-func (n *Network) CreateTransaction(payloadType string, payload []byte, signingKeyID string, attachKey crypto2.PublicKey, timestamp time.Time, fieldOpts ...dag.FieldOpt) (dag.Transaction, error) {
+func (n *Network) CreateTransaction(payloadType string, payload []byte, signingKeyID string, attachKey crypto2.PublicKey, timestamp time.Time, additionalPrevs []hash.SHA256Hash) (dag.Transaction, error) {
 	payloadHash := hash.SHA256Sum(payload)
 	log.Logger().Debugf("Creating transaction (payload hash=%s,type=%s,length=%d,signingKey=%s)", payloadHash, payloadType, len(payload), signingKeyID)
+
 	// Create transaction
 	prevs := n.graph.Heads()
-	unsignedTransaction, err := dag.NewTransaction(payloadHash, payloadType, prevs, fieldOpts...)
+	for _, addPrev := range additionalPrevs {
+		prevs = append(prevs, addPrev)
```
reinkrul, line 166: let's make sure there aren't any duplicates, since I think that wouldn't be according to spec (didn't check though). Or let `NewTransaction` take care of deduplication

network/network_test.go
```diff
@@ -196,7 +196,7 @@ func TestNetwork_CreateTransaction(t *testing.T) {
 		if !assert.NoError(t, err) {
 			return
 		}
-		_, err = cxt.network.CreateTransaction(payloadType, payload, "signing-key", privateKey.PublicKey, time.Now())
+		_, err = cxt.network.CreateTransaction(payloadType, payload, "signing-key", privateKey.PublicKey, time.Now(), []hash.SHA256Hash{})
```
reinkrul, line 199: test with `additionalPrevs`?

vdr/ambassador_test.go
```diff
@@ -791,6 +852,40 @@ func Test_checkDIDDocumentIntegrity(t *testing.T) {
...
+	h2 := hash.SHA256Sum([]byte("hash2"))
+	h3 := hash.SHA256Sum([]byte("hash3"))
+
+	t.Run("non-conflicted updated as expected", func(t *testing.T) {
+		current := []hash.SHA256Hash{h1}
+		incoming := []hash.SHA256Hash{h1, h2}
+
+		diff := missingTransactions(current, incoming)
+
+		assert.Len(t, diff, 0)
```
reinkrul, line 866: trivial; you can use `assert.IsEmpty(t, diff)`

vdr/ambassador_test.go
```diff
@@ -791,6 +852,40 @@ func Test_checkDIDDocumentIntegrity(t *testing.T) {
...
+		assert.Equal(t, current, diff)
+	})
+
+	t.Run("conflicted resolved", func(t *testing.T) {
+		current := []hash.SHA256Hash{h1, h2}
+		incoming := []hash.SHA256Hash{h1, h2, h3}
+
+		diff := missingTransactions(current, incoming)
+
+		assert.Len(t, diff, 0)
```
reinkrul, line 885: idem

vdr/vdr.go
```diff
@@ -147,7 +147,7 @@ func (r VDR) Update(id did.DID, current hash.SHA256Hash, next did.Document, _ *t
 		Hash:             &current,
 		AllowDeactivated: true,
 	}
-	currentDIDDocument, _, err := r.store.Resolve(id, resolverMetadata)
+	currentDIDDocument, currentMeta, err := r.store.Resolve(id, resolverMetadata)
```
reinkrul, line 150: I have mixed feelings about this; updating one's DID Document automatically "fixes" any conflicts because it refers to all source transactions, whether that's a single one or multiple.

Should we protect administrators from accidentally overwriting a conflicting change, since the new document might not include changes from both/all documents since the administrator didn't know about it at time of drawing up the document? For instance by checking the number of source transactions, requiring a flag to be passed when there's multiple?

If it's too intrusive here, we could add it just in the REST API and CLI.

reinkrul, line 150: in context of the `current` parameter, what if there are multiple? Just pick one?

vdr/ambassador.go
```diff
@@ -221,6 +231,26 @@ func (n *ambassador) handleUpdateDIDDocument(transaction dag.SubscriberTransacti
...
+	for _, h := range current {
+		found := false
+		for _, h2 := range incoming {
+			if h.Equals(h2) {
+				found = true
+				break
+			}
+		}
+		if !found {
+			current[j] = h
```
reinkrul, line 246: slices are always passed by reference, so this mutates the `current` slice. Intended (potentially dangerous) side-effect?

Now I think of it, might a case where a deep copy from the in-memory DID store protects data corruption, if applicable?

network/interface.go
```diff
@@ -38,7 +38,7 @@ type Transactions interface {
 	GetTransaction(transactionRef hash.SHA256Hash) (dag.Transaction, error)
 	// CreateTransaction creates a new transaction with the specified payload, and signs it using the specified key.
 	// If the key should be inside the transaction (instead of being referred to) `attachKey` should be true.
```
reinkrul, line 40: maybe document the behaviour in relation to `additionalPrevs`

reinkrul (changes requested):
Accidentally approved

vdr/vdr.go
woutslakhorst: That's why we introduced a `hash` in the API that must match the current state. That way we're sure the admin has seen the current state.

woutslakhorst: the current hash refers to the hash of the current (merged) doc. There's no multiple, just one

vdr/ambassador.go
woutslakhorst: document and metadata are already a deepCopy when returned from the DB

network/network.go
woutslakhorst: I moved it to NewTransaction

network/interface.go
woutslakhorst: added some info

reinkrul (approved):
You have my blessing. Codeclimate still complains though.

PR #249, opened 2021-04-22
Network: expose graphviz DAG visitor in API for visualization/analytical purposes

reinkrul:
Goal is to visualize it in Nuts Explorer.

### DIFF
--- a/docs/_static/network/v1.yaml
+++ b/docs/_static/network/v1.yaml
@@ -70,3 +70,19 @@ paths:
               example:
         "404":
           description: "Transaction (or payload) wasn't found"
+  /internal/network/v1/diagnostics/graph:
+    get:
+      summary: "Visualizes the DAG as a graph"
+      description: >
+        Walks the DAG as subscribers of the DAG do, rendering it as graph. By default it renders in Graphviz format,
+        which can be rendered to an image using `dot`.
+      operationId: "renderGraph"
+      tags:
+        - transactions
+      responses:
+        "200":
+          description: "Graph successfully rendered"
+          content:
+            text/vnd.graphviz:
+              schema:
+                type: string
--- a/network/api/v1/api.go
+++ b/network/api/v1/api.go
@@ -19,6 +19,7 @@
 package v1
 
 import (
+	"github.com/nuts-foundation/nuts-node/network/dag"
 	"net/http"
 
 	"github.com/labstack/echo/v4"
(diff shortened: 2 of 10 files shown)


docs/_static/network/v1.yaml
```diff
@@ -69,4 +69,20 @@ paths:
             application/octet-stream:
               example:
         "404":
-          description: "Transaction (or payload) wasn't found"
\ No newline at end of file
+          description: "Transaction (or payload) wasn't found"
+  /api/graph:
```
woutslakhorst, line 73: `/internal/network/v1/diagnostics/graph`

network/api/v1/api.go
```diff
@@ -89,3 +90,15 @@ func (a Wrapper) GetTransactionPayload(ctx echo.Context, hashAsString string) er
...
 	return err
 }
+
+// RenderGraph visualizes the DAG as Graphviz/dot graph
+func (a Wrapper) RenderGraph(ctx echo.Context) error {
+	visitor := dag.NewDotGraphVisitor(dag.ShowShortRefLabelStyle)
+	err := a.Service.Walk(visitor.Accept)
+	if err != nil {
+		log.Logger().Errorf("Error while rendering graph: %v", err)
+		return ctx.String(http.StatusInternalServerError, err.Error())
```
woutslakhorst, line 100: `core.NewProblem` is the new way of returning errors.

docs/_static/network/v1.yaml
```diff
@@ -69,4 +69,20 @@ paths:
...
+    get:
+      summary: "Visualizes the DAG as a graph"
+      description: >
+        Walks the DAG as subscribers of the DAG do, rendering it as graph. By default it renders in Graphviz format,
+        which can be rendered to an image using `dot`.
+      operationId: "renderGraph"
+      tags:
+        - transactions
+      responses:
```
woutslakhorst, line 82: add errors responses. May reference to generic file. See crypto api as example

woutslakhorst (changes requested):

network/api/v1/api.go
reinkrul: I'll do that in one go (no pun intended) for all of `network`, instead of bit by bit

docs/_static/network/v1.yaml
reinkrul: idem dito

woutslakhorst (changes requested):
Could you add some development/monitoring documentation to the docs so we know how to use it?

woutslakhorst (approved):

PR #260, opened 2021-04-28
Add creation options to DID creation

### LINKED ISSUE #255: Add creationOptions to DID creation
A creationOptions object defines the following:

```
{
  "controllers": [], // list of controller DIDs
  "isController": true/false, // is the newly generated DID in control of its own document
  "hasCapablilityInvocation": true/false, // is the generated key used for capabilitInvocations
  "hasAssertionMethod": true/false, // key used for assertion?
  "hasCapablilityDelegation": true/false, // key used for capablilityDelegation?
  "hasAuthentication": true/false, // key used for authentication?
  "hasKeyAgreement": true/false // key used for keyAgreement?
}
```
- `isController == false` and `hasCapablilityInvocation == true` is a valid state since it will allow a DID Dubject to control DID Documents of its children.
- if `isController == true`, then the generated DID will be added to the list of controllers.
- `isController == true` and `hasCapabilityInvocation == false` is not allowed.
- If `isController == false` and `hasCapabilityInvocation == false` then the private key must not be persisted and wiped after the transaction has been published. The `controller` MUST then have at least one entry.

woutslakhorst:
closes #255


docs/_static/vdr/v1.yaml
```diff
@@ -362,4 +377,40 @@ components:
...
+            List of DIDs that can control the new DID Document. If selfControl = true and controllers is not empty.
+            The newly generated DID will be added to the list of controllers.
+          type: array
+          items:
+            type: string
+            description: DID according to Nuts specification
+            example: "did:nuts:128903fjgfslcnmgpe84"
+        assertionMethod:
+          type: boolean
+          description: indicates if the generated key pair can be used for assertions. Defaults to true.
```
stevenvegt, line 394: It's unclear to me if `the generated key pair` refers to one pair per document used for multiple purposes or that every boolean set to true creates an extra key pair.

network/network_integration_test.go
```diff
@@ -120,64 +113,12 @@ func TestNetworkIntegration_HappyFlow(t *testing.T) {
 	fmt.Printf("%v\n", node2.Diagnostics())
 }
 
-func TestNetworkIntegration_SignatureIncorrect(t *testing.T) {
```
stevenvegt, line 123: Why is this removed?

stevenvegt (commented):
PR is a bit difficult to read with all the clean up. But I like how much simpler the code becomes.
Do we need to create new Key pairs of each assertionMethod or reuse keys for multiple purposes? Or should this also be an option?

crypto/ephemeral.go
```diff
@@ -0,0 +1,38 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *
+ */
+
+package crypto
+
+// NewEphemeralKey returns a KeySelector with a single key for single use.
+func NewEphemeralKey(namingFunc KIDNamingFunc) (KeySelector, error) {
```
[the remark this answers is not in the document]
woutslakhorst: yes, that saved 2 lines of code.

docs/_static/vdr/v1.yaml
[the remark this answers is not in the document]
woutslakhorst: there is, added default values

woutslakhorst: a single, added it to the POST description.

network/network_integration_test.go
woutslakhorst: because it didn't really test anything. For one node it manipulated the internals and for the other it didn't. Not really an integration test if internals are manipulated. Also with the new setup it's way harder to manipulate the used private key. You can now only do that from unit test context.

woutslakhorst:
- [ ] API client

docs/_static/vdr/v1.yaml
```diff
@@ -362,4 +378,44 @@ components:
...
           description: The type of the key.
           example: "JsonWebKey2020"
-          type: string
\ No newline at end of file
+          type: string
+    DIDCreateRequest:
+      properties:
+        controllers:
+          description: |
+            List of DIDs that can control the new DID Document. If selfControl = true and controllers is not empty.
```
reinkrul, line 386: second sentence ends with a dot but should probably end with a comma?

docs/_static/vdr/v1.yaml
```diff
@@ -362,4 +378,44 @@ components:
...
+          type: boolean
+          description: indicates if the generated key pair can be used for capability delegations.
+          default: false
+        keyAgreement:
+          type: boolean
+          description: indicates if the generated key pair can be used for Key agreements.
+          default: false
+        selfControl:
+          type: boolean
+          description: wether the generated DID Document can be altered with its own capabilityInvocation key.
```
reinkrul, line 418: whether

reinkrul, line 418: what if `selfControl = false` and `capabilityInvocation = true`?

vcr/vcr.go
```diff
@@ -643,7 +653,7 @@ func (c *vcr) convert(query concept.Query) map[string]leia.Query {
 	return qs
 }
 
-func (c *vcr) generateProof(credential *vc.VerifiableCredential, kid ssi.URI) error {
+func (c *vcr) generateProof(credential *vc.VerifiableCredential, kid ssi.URI, key crypto.KeySelector) error {
```
reinkrul, line 656: `KeySelector` has `KID` property which can be used instead of passing `kid`?

vcr/vcr.go
```diff
@@ -425,11 +430,16 @@ func (c *vcr) Revoke(ID ssi.URI) (*credential.Revocation, error) {
...
+	if err != nil {
+		return nil, fmt.Errorf("could not resolve kid: %w", err)
+	}
+
 	// set defaults
 	r := credential.BuildRevocation(target)
 
 	// sign
-	if err = c.generateRevocationProof(&r, kid); err != nil {
+	if err = c.generateRevocationProof(&r, kid, key); err != nil {
```
reinkrul, line 442: why pass both `key` and `kid`?

vdr/api/v1/api.go
```diff
@@ -79,9 +80,47 @@ func (a *Wrapper) Routes(router core.EchoRouter) {
 
 // CreateDID creates a new DID Document and returns it.
 func (a Wrapper) CreateDID(ctx echo.Context) error {
-	doc, err := a.VDR.Create()
+	req := DIDCreateRequest{}
+	if err := ctx.Bind(&req); err != nil {
+		return ctx.String(http.StatusBadRequest, fmt.Sprintf("create request could not be parsed: %s", err.Error()))
+	}
+
+	options := doc2.DefaultCreationOptions()
```
reinkrul, line 88: `doc2` is a bit ugly and non-descriptive

vdr/cmd/cmd.go
```diff
@@ -64,12 +64,26 @@ func Cmd() *cobra.Command {
 }
 
 func createCmd() *cobra.Command {
-	return &cobra.Command{
+	// needs to be initialized for pflags, values will be overwritten with defaults from pflag
+	trueVal := true
+	falseVal := false
+	var sliceVar []string
+	var createRequest = api.DIDCreateRequest{
```
reinkrul, line 71: does pflag assign a new ptr bool var or does it just change the value? If it's the last, if the last `falseVal` is false, all are false? (and vice versa for `trueVal`). Because you pass the same var for many properties.

vdr/api/v1/api.go
```diff
@@ -79,9 +80,47 @@ func (a *Wrapper) Routes(router core.EchoRouter) {
...
+		for _, c := range *req.Controllers {
+			id, err := did.ParseDID(c)
+			if err != nil {
+				return ctx.String(http.StatusBadRequest, fmt.Sprintf("controller entry (%s) could not be parsed: %s", c, err.Error()))
+			}
+			options.Controllers = append(options.Controllers, *id)
+		}
+	}
+
+	if req.Authentication != nil && *req.Authentication {
```
reinkrul, line 99: The functioning of this is now dependant on what the defaults are which is dangerous since defaults may change (and we forget to change these if's). Wouldn't it be both simpler and more adaptive to just do;

```
if req.Authentication != nil {
    options.Authentication = *req.Authentication
}
```

Since if it's set, you can override the default value.

Same ofc goes for the `if`s below.

vdr/types/interface.go
```diff
@@ -45,7 +45,7 @@ type DocCreator interface {
 	// The ID in the provided DID document will be ignored and a new one will be generated.
```
reinkrul, line 45: Not part of this PR, but still; `The ID in the provided DID document will be ignored and a new one will be generated.`?

vdr/vdr.go
```diff
@@ -167,8 +162,15 @@ func (r VDR) Update(id did.DID, current hash.SHA256Hash, next did.Document, _ *t
 		return err
 	}
 
-	keyID := controllers[0].CapabilityInvocation[0].ID.String()
-	_, err = r.network.CreateTransaction(didDocumentType, payload, keyID, nil, time.Now(), currentMeta.SourceTransactions)
+	key, err := r.keyStore.Resolve(controllers[0].CapabilityInvocation[0].ID.String())
```
reinkrul, line 165: is the first controller guaranteed to have a `capabilityInvocation` key? Otherwise this'll lead to an out of bounds error

vdr/vdr.go
```diff
@@ -167,8 +162,15 @@ func (r VDR) Update(id did.DID, current hash.SHA256Hash, next did.Document, _ *t
...
+	if err != nil {
+		if errors.Is(err, crypto.ErrKeyNotFound) {
+			return types.ErrDIDNotManagedByThisNode
+		}
+		return fmt.Errorf("could not find key from controller: %w", err)
```
reinkrul, line 170: `could not find capabilityInvocation key for controller`

reinkrul (changes requested):

crypto/crypto.go
```diff
@@ -117,3 +127,34 @@ func generateECKeyPair() (*ecdsa.PrivateKey, error) {
 func (client *Crypto) PrivateKeyExists(kid string) bool {
 	return client.Storage.PrivateKeyExists(kid)
 }
+
+func (client *Crypto) Resolve(kid string) (KeySelector, error) {
```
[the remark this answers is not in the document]
woutslakhorst: maybe `Key` would be the best fit.

crypto/interface.go
```diff
@@ -40,31 +33,34 @@ type KIDNamingFunc func(key crypto.PublicKey) (string, error)
 type KeyCreator interface {
 	// New generates a keypair and returns the public key.
 	// the KIDNamingFunc will provide the kid. priv/pub keys are appended with a postfix and stored
```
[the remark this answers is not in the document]
woutslakhorst: that is old doc indeed

crypto/interface.go
```diff
@@ -40,31 +33,34 @@ type KIDNamingFunc func(key crypto.PublicKey) (string, error)
...
+	New(namingFunc KIDNamingFunc) (KeySelector, error)
 }
 
 // KeyStore defines the functions for working with private keys.
 type KeyStore interface {
 	// PrivateKeyExists returns if the specified private key exists.
 	// If an error occurs, false is also returned
 	PrivateKeyExists(kid string) bool
+	// Resolve returns a KeySelector for the given KID. ErrKeyNotFound is returned for an unknown KID.
+	Resolve(kid string) (KeySelector, error)
```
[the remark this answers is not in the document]
woutslakhorst: I'll rename `PrivateKeyExists` to `Exists`

docs/_static/vdr/v1.yaml
woutslakhorst: indeed

docs/_static/vdr/v1.yaml
woutslakhorst: Than that DID Document can control sub-documents.

vcr/vcr.go
woutslakhorst: different types: `ssi.URI` vs `string`

vcr/vcr.go
woutslakhorst: different types: `ssi.URI` vs `string`

vdr/api/v1/api.go
woutslakhorst: changed to `vdrDoc`

vdr/cmd/cmd.go
woutslakhorst: changed to `new(bool)`

vdr/types/interface.go
woutslakhorst: Since you can't set your own DID

vdr/vdr.go
woutslakhorst: nope, this should be fixed by the `canControl` functionality that was started on. I'll make sure there's an issue for that.

woutslakhorst: #266

stevenvegt (approved):

woutslakhorst:
@reinkrul your approval is required before everything turns green

reinkrul (approved):

PR #262, opened 2021-04-29
altered DID Document check according to RFC006

### LINKED ISSUE #261: Change DID creation validation
A verificationMethod is no longer required. The only check that needs to be done is that the JWK in the transaction must match the fingerprint/DID of the new document.

woutslakhorst:
closes #261

### DIFF
--- a/vdr/doc/creator.go
+++ b/vdr/doc/creator.go
@@ -21,10 +21,6 @@ package doc
 
 import (
 	"crypto"
-	"crypto/ecdsa"
-	"crypto/elliptic"
-	"crypto/sha256"
-	"errors"
 	"fmt"
 
 	ssi "github.com/nuts-foundation/go-did"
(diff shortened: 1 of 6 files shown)


vdr/doc/creator.go
```diff
@@ -47,28 +43,22 @@ type Creator struct {
...
 
 	// generate idString
-	pkBytes := elliptic.Marshal(ecPKey.Curve, ecPKey.X, ecPKey.Y)
-	pkHash := sha256.Sum256(pkBytes)
-	idString := base58.Encode(pkHash[:], base58.BitcoinAlphabet)
-
-	// generate kid fragment
 	jwKey, err := jwk.New(pKey)
+	if err != nil {
+		return "", fmt.Errorf("could not generate kid: invalid key type")
```
reinkrul, line 52: why not just wrap `err`? It's now thrown away

vdr/doc/creator.go
```diff
@@ -47,28 +43,22 @@ type Creator struct {
...
+	}
+
+	pkHash, err := jwKey.Thumbprint(crypto.SHA256)
 	if err != nil {
 		return "", err
 	}
+	idString := base58.Encode(pkHash[:], base58.BitcoinAlphabet)
```
reinkrul, line 59: make a const of `BitcoinAlphabet` since it's now used twice as JWK thumbprint encoding scheme (or make a util func)

reinkrul (changes requested):
Ouch, there go all our development network DIDs :-p

vdr/doc/creator.go
woutslakhorst: moved Thumbprint to crypto and added test that matches example from RFC

reinkrul (approved):
Seal of approval

```
          -~~~~-
         |O __ O|
      -__|_\__/_|__-
      ---(__/\__)---
        /||~--~||\
       / ||    || \
      |  `'    `'  |
    _/ /\__    __/\ \_
  ~~~~~    ~~~~    ~~~~~
```

PR #272, opened 2021-05-03
Network: validate transactions loaded from disk (at startup)

### LINKED ISSUE #267: Network: validate transactions on startup (loaded from disk)
In case of e.g. corrupt storage or a new version of the nuts node with a bug due to changed business rules, which don't comply with the transactions already in storage. When encountering invalid transaction on startup, it should refuse to start.

Should validate:

- TX validity (semantic, cryptographic)
- DAG integrity
  - All prevs are present

### LINKED ISSUE #268: Network: disallow DAG transactions with missing previous'
It's currently allowed to have transactions on the DAG which's previous transactions are missing (for historic reasons). This complicates things for head calculation and DAG walking (transaction looks to be present but is never reached by the DAG walker) and should be disallowed.

reinkrul:
Fixes #267
Fixes #268
Implements https://github.com/nuts-foundation/nuts-specification/pull/65

Refactors transaction validation to `dag.Verifier`s which are used for 2 purposes:

- Validating transactions in persistence when starting the node (boot time)
- Validating transactions newly received from the network (runtime)

After this PR transactions with `prev` transactions that are missing are disallowed, which might break in conjunction with the protocol implementation which doesn't care about transaction order when responding to a transaction query or processing it (the response). Thus if it'd happen to send the transactions in an order other than how the DAG walker would visit them it now refuses the transaction.

Thus, guaranteeing a predictable transaction order (received from the network) should be fixed first.

PR https://github.com/nuts-foundation/nuts-node/pull/274 must be merged first to guarantee transactions are received in proper order.


woutslakhorst (commented):
> Thus, guaranteeing a predictable transaction order (received from the network) should be fixes first.

is blocking for this PR. Why not change: `dag.FindBetween9begin, end)` to use a walker or to sort the resulting array? That way the current issues for the DB network are resolved.

In any case we do need a better heads->query->list specification.

reinkrul:
> > Thus, guaranteeing a predictable transaction order (received from the network) should be fixes first.
> 
> is blocking for this PR. Why not change: `dag.FindBetween9begin, end)` to use a walker or to sort the resulting array? That way the current issues for the DB network are resolved.
> 
> In any case we do need a better heads->query->list specification.

that's the idea! 👍

network/dag/verifier.go
```diff
@@ -0,0 +1,67 @@
...
+	_, err := jws.Verify(input.Data(), jwa.SignatureAlgorithm(input.SigningAlgorithm()), signingKey)
+	return err
+}
+
+// NewPrevTransactionsVerifier creates a transaction verifier that asserts that all previous transactions are known.
+func NewPrevTransactionsVerifier() Verifier {
+	return &prevTransactionsVerifier{}
+}
+
+type prevTransactionsVerifier struct {
```
stevenvegt, line 53: If a verifier is a single function it does not have to be a struct?

stevenvegt (approved):
Perhaps refactor the verifier struct into a single func?

PR #273, opened 2021-05-03
Network: return JSON problems instead of text/plain errors

reinkrul:


gerardsn:
part of #112

docs/_static/network/v1.yaml
```diff
@@ -48,6 +50,8 @@ paths:
                 type: string
         "404":
           description: "Transaction wasn't found in the transaction log"
+        default:
+          $ref: '../common/error_response.yaml'
```
gerardsn, line 54: default is for all 400+ status codes. In other apis the description for error returns is documented per path under path.get.description

docs/_static/network/v1.yaml
```diff
@@ -70,6 +74,8 @@ paths:
               example:
         "404":
           description: "Transaction (or payload) wasn't found"
+        default:
+          $ref: '../common/error_response.yaml'
```
gerardsn, line 78: default is for all 400+ status codes. In other apis the description for error returns is documented per path under path.get.description

network/api/v1/api.go
```diff
@@ -76,14 +81,14 @@ func (a Wrapper) GetTransaction(ctx echo.Context, hashAsString string) error {
...
 	if err != nil {
-		return ctx.String(http.StatusBadRequest, err.Error())
+		return core.NewProblem(problemTitleGetTransactionPayload, http.StatusBadRequest, err.Error())
 	}
 	data, err := a.Service.GetTransactionPayload(hash)
 	if err != nil {
-		return ctx.String(http.StatusInternalServerError, err.Error())
+		return core.NewProblem(problemTitleGetTransactionPayload, http.StatusInternalServerError, err.Error())
 	}
 	if data == nil {
```
gerardsn, line 90: Logging Is not part of the PR, but shouldn't an internal server error be logged?

gerardsn (commented):
I've added some comment, but otherwise looks good to me. I will leave approval to @stevenvegt

network/api/v1/api.go
reinkrul: Agreed. I must admit I spotted this as well but was kinda lazy. Will add a log statement.

stevenvegt (approved):

PR #275, opened 2021-05-06
Network: add logging for diverging block dates

reinkrul:


woutslakhorst (commented):
I think this will occur every day due to clock skew plus transmission delays. So it would be normal behaviour.

Issue #280, opened 2021-05-10
DIDMan: provide API for creating compound service


reinkrul:
Provide an API that takes:

- target DID
- service endpoint, which is a map containing references to concrete URL services (according to RFC006)
- service type

And creates the services. It should validate that all referenced concrete URL services can be resolved.

PR #287, opened 2021-05-11
DIDMan: API for adding a compound service

reinkrul:
Fixes https://github.com/nuts-foundation/nuts-node/issues/280

### DIFF
--- a/didman/api/v1/api.go
+++ b/didman/api/v1/api.go
@@ -103,20 +103,28 @@ func (w *Wrapper) AddCompoundService(ctx echo.Context, didStr string) error {
 	}
 
 	references := make(map[string]ssi.URI, 0)
-	for key, refAsStr := range request.Endpoint {
-		refAsURI, err := ssi.ParseURI(fmt.Sprintf("%s", refAsStr))
+	for key, value := range request.Endpoint {
+		uri, err := interfaceToURI(value)
 		if err != nil {
 			err = fmt.Errorf("invalid reference for service '%s': %w", key, err)
 			logging.Log().WithError(err).Warn(problemTitleAddCompoundService)
 			return core.NewProblem(problemTitleAddCompoundService, http.StatusBadRequest, err.Error())
 		}
-		references[key] = *refAsURI
+		references[key] = *uri
 	}
 	if err = w.Didman.AddCompoundService(*id, request.Type, references); err != nil {
 		logging.Log().WithError(err).Warn(problemTitleAddCompoundService)
 		return core.NewProblem(problemTitleAddCompoundService, getStatusCode(err), err.Error())
 	}
 	return ctx.NoContent(http.StatusNoContent)
+}
+
+func interfaceToURI(input interface{}) (*ssi.URI, error) {
+	str, ok := input.(string)
+	if !ok {
+		return nil, errors.New("not a string")
+	}
+	return ssi.ParseURI(str)
 }
 
 // DeleteService handles calls to delete a service. It only checks params and sets the correct return status code.
--- a/didman/api/v1/api.go
+++ b/didman/api/v1/api.go
@@ -86,6 +86,12 @@ func (w *Wrapper) AddEndpoint(ctx echo.Context, didStr string) error {
 	return ctx.NoContent(http.StatusNoContent)
 }
 
+// AddCompoundService handles calls to add a compound service. It only checks params and sets the correct return status code.
+// didman.AddCompoundService does the heavy lifting.
+func (w *Wrapper) AddCompoundService(ctx echo.Context, didStr string) error {
+	panic("implement me")
+}
+
 // DeleteService handles calls to delete a service. It only checks params and sets the correct return status code.
 // didman.DeleteService does the heavy lifting.
 func (w *Wrapper) DeleteService(ctx echo.Context, uriStr string) error {
--- a/didman/api/v1/api.go
+++ b/didman/api/v1/api.go
@@ -36,6 +36,7 @@ import (
 )
 
 const problemTitleAddEndpoint = "Adding Endpoint failed"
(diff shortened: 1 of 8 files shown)


reinkrul:
First round of reviews before implementing API spec

didman/didman.go
```diff
@@ -66,25 +73,49 @@ func (d *didman) Name() string {
...
+			}
+			documents[referencedDIDStr.String()] = document
+		}
+		queriedServiceType := serviceRef.Query().Get("type")
+		if len(queriedServiceType) == 0 {
+			return ErrInvalidServiceQuery
+		}
+		_, _, err := document.ResolveEndpointURL(queriedServiceType)
+		if err != nil {
+			return errors2.WithMessage(ErrReferencedServiceNotAnEndpoint, err.Error())
```
woutslakhorst, line 112: We have a mix of `github.com/pkg/errors` and `errors` inour code. `github.com/pkg/errors` had its uses in older versions of Go. I think it would be best to stick to `errors` and `fmt.Errorf()`. That way `errors.Is` can be used.

In the case of this error, a distinction needs to be made between a technical error or a ErrNotFound/ErrDupl.

didman/didman.go
```diff
@@ -195,17 +212,34 @@ func filterContactInfoServices(doc *did.Document) []did.Service {
 	return contactServices
 }
 
-func constructService(id did.DID, serviceType string, u url.URL) did.Service {
+func (d *didman) addService(id did.DID, serviceType string, serviceEndpoint interface{}, updater func(*did.Document)) error {
```
woutslakhorst, line 215: it's not really an `updater`, more a `preprocessor`

didman/didman.go
```diff
@@ -66,25 +73,49 @@ func (d *didman) Name() string {
...
-		if s.Type == serviceType {
-			return ErrDuplicateService
+func (d *didman) AddCompoundService(id did.DID, serviceType string, references map[string]ssi.URI) error {
+	logging.Log().Debugf("Adding compound service (did: %s, type: %s, references: %v)", id.String(), serviceType, references)
+
+	// Resolve and validate all service references: they must all resolve to a single endpoint URL
+	// Cache resolved DID documents because most of the time a compound service will refer the same DID document in all service references.
+	documents := make(map[string]*did.Document)
+	var err error
+	for _, serviceRef := range references {
```
woutslakhorst, line 90: Resolving the referenced documents can have its own function, reduces complexity.

docs/_static/didman/v1.yaml
```diff
@@ -87,6 +87,44 @@ paths:
...
+        The endpoints may be in the same or in another DID Document, but they must be resolvable when the service is created.
+        The references must follow the format for service references as specified by Nuts RFC006.
+
+        This API will also check if an endpoint with the same type already exists. This API is not meant to add endpoints.
+
+        error returns:
+        * 400 - incorrect input
+        * 404 - unknown DID
+        * 409 - a service with the same type already exists
+        * 422 - one or more of the referenced services can't be resolved
```
woutslakhorst, line 113: To keep things simpler, a 400 would be acceptable

woutslakhorst (changes requested):

didman/didman.go
reinkrul: Fixed it. About the distinction I don't agree: it's up to the caller to interpret the particular error and call it again after fixing the input. And since the caller will (most probably) be a REST client it doesn't really add value to specialize these errors any further?

woutslakhorst: I see it's not possible to distinguish between different errors and that no technical error can occur.

didman/api/v1/api.go
```diff
@@ -86,6 +87,38 @@ func (w *Wrapper) AddEndpoint(ctx echo.Context, didStr string) error {
...
+	}
+
+	id, err := parseDID(didStr, problemTitleAddCompoundService)
+	if err != nil {
+		return err
+	}
+
+	references := make(map[string]ssi.URI, 0)
+	for key, refAsStr := range request.Endpoint {
+		refAsURI, err := ssi.ParseURI(fmt.Sprintf("%s", refAsStr))
```
woutslakhorst, line 107: shouldn't the refAsStr be cast to string first (and check the cast)? Otherwise a slice might end up being parsed as URI? Which is probably a mem address but also a valid URI?

didman/api/v1/api_test.go
```diff
@@ -232,6 +232,112 @@ func TestWrapper_AddEndpoint(t *testing.T) {
...
+		test.AssertErrProblemStatusCode(t, http.StatusInternalServerError, err)
+		test.AssertErrProblemDetail(t,  "failed", err)
+		test.AssertErrProblemTitle(t, problemTitleAddCompoundService, err)
+	})
+
+	t.Run("error - incorrect endpoint", func(t *testing.T) {
+		ctx := newMockContext(t)
+		ctx.echo.EXPECT().Bind(gomock.Any()).DoAndReturn(func(f interface{}) error {
+			p := f.(*CompoundServiceCreateRequest)
+			*p = CompoundServiceCreateRequest{Type: "type", Endpoint: map[string]interface{}{"foo": ":"}}
```
woutslakhorst, line 297: what happens if you enter an object or array instead of ":"?

woutslakhorst (changes requested):
small stuff

didman/api/v1/api.go
reinkrul: Yeah you're right, and `nil` will be formatted as "nil", which is a valid URI and thus happily be passed.

didman/api/v1/api_test.go
reinkrul: Now fixed & tested

woutslakhorst (approved):

PR #290, opened 2021-05-11
Network: solve desync of historic block and omnihashes being all zeroed

### LINKED ISSUE #289: Network: omnihashes are empty
Sometimes the peer's omnihashes appear to be empty;

```
[Protocol] Peer omnihashes: e2f55dca6e4b876de7c9904f0232db5c9eef559af502e223bf28cf2ad84e182a={d87529e2-b0ab-467f-a9dd-40f86476117e, 6c50f090-2963-437a-8b97-7639cd2c9d43, 2fede59f-6568-4458-9201-5e3f28e88614}, 0000000000000000000000000000000000000000000000000000000000000000={8d4ab8b1-ee6b-47a6-bcd6-9ad2d0040bf9, 50fc6af2-93e8-4e06-851f-6682cd9f5fa9}
```

Restarting looks to be fixing it. We need to find out what's happening.

reinkrul:
Fixes #289 

- Calculate peer omnihash even when historic block differs (solves all-zeroes omnihashes)
- Reject transactions with a signing time further than 1 day in the future (makes determining heads much easier, if we keep it we should put it in the spec)
- Fix marking new transaction as head instead of the previous transaction if they're in the same block

The time-based blocks are complicated and I discovered more issues with it (next TX being in an older block). Should probably be rethought. (even if it's fixed for now)

### DIFF
--- a/network/dag/verifier.go
+++ b/network/dag/verifier.go
@@ -7,6 +7,7 @@ import (
 	"github.com/lestrrat-go/jwx/jwa"
 	"github.com/lestrrat-go/jwx/jws"
 	"github.com/nuts-foundation/nuts-node/vdr/types"
+	"time"
 )
 
 // ErrPreviousTransactionMissing indicates one or more of the previous transactions (which the transaction refers to)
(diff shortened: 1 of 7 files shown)


woutslakhorst (commented):
How does it fix the historic block synching again? Was that a problem due to incorrect head calculation?

reinkrul:
> How does it fix the historic block synching again? Was that a problem due to incorrect head calculation?

Correct. It also was semi-random, I sometimes got different (incorrect) historic heads after restart.

They never sync'd because if the historic heads were different, they were queried first, but they could never get sync'd.

woutslakhorst (approved):

PR #291, opened 2021-05-12
added CRM getting started guide

woutslakhorst:
it explains how to register a vendor (Saas provider) and its customers.


docs/pages/getting-started/4-connecting-crm.rst
```diff
@@ -0,0 +1,216 @@
...
+Getting Started on customer integration
+#######################################
+
+This getting started manual assumes a vendor sells services to its customers.
+The vendor manages the presence of those customers on the Nuts network through the Nuts registry.
+It's very likely the vendor has software to manage customer environments. We'll call it a *CRM* where the *customers* part relates to organizations and not people.
+This does not mean that a stand-alone installation isn't supported. In that case the vendor and organization are the same.
+
+The Nuts registry enables service discovery for organizations. The registry identifies organizations through their :ref:`DID <did>`.
+The DIDs are new and do not match the internal identifiers known to the CRM.
```
reinkrul, line 12: "New" and "internal identifiers" is a bit vague imo. "Connection" makes me think of network connections, instead of mapping something from an external domain. Why not make it explicit, something like:

> The DIDs are unique, random identifiers which are generated when the organization is registered in the Nuts registry. After creation of the DID the CRM should store and map it to its customer record, so it can refer to it when updating the customer's DID Document and issue Verifiable Credentials.

docs/pages/getting-started/4-connecting-crm.rst
```diff
@@ -0,0 +1,216 @@
...
+As a vendor, you're in power of:
+
+- running a Nuts node
+- handling organization key material
+- updating the organization DID Document
+- defining service endpoints
+- issuing name credentials for organizations
+- trusting other vendors
+
+The last three points require a setup where a vendor DID is created. This DID will act as the controller of all organization DID Documents.
```
reinkrul, line 30: "issuing of verifiable credentials" > "issuing verifiable credentials" or ("issuance of verifiable credentials")

docs/pages/getting-started/4-connecting-crm.rst
```diff
@@ -0,0 +1,216 @@
...
+This will allow for reuse of service endpoints and issuing of verifiable credentials. It's easier for other vendors to trust a single issuer DID.
```
reinkrul, line 31: consistency: Verifiable Credentials vs verfiable credentials

reinkrul, line 31: > It's easier for other vendors to trust a single issuer DID.

Sounds like it's missing something. Something like "Since every DID issuing Verifiable Credentials must be trusted individually, it's easier for other vendors when the vendor uses a single DID for issuing credentials."

docs/pages/getting-started/4-connecting-crm.rst
```diff
@@ -0,0 +1,216 @@
...
+
+Create and store a vendor DID
+=============================
+
+You'll have to prepare your CRM to store a DID. A DID is a string similar to:
```
reinkrul, line 36: "You'll have to prepare your CRM" sounds a bit like textual fuzz, while it's missing some info on integrators on how it should be stored. Something like "Your CRM must store the DIDs created for your vendor and your customers. It's a string with a maximum length of ...".

docs/pages/getting-started/4-connecting-crm.rst
```diff
@@ -0,0 +1,216 @@
...
+.. code-block:: text
+
+    POST <internal-node-address>/internal/vdr/v1/did
+    {
+        "selfControl": true,
+        "assertionMethod": true,
+        "capabilityInvocation": true
+    }
+
+The request above instruct the node to create a new DID and DID Document. The DID Document will be published to all other nodes.
```
reinkrul, line 55: instructs

docs/pages/getting-started/4-connecting-crm.rst
```diff
@@ -0,0 +1,216 @@
...
+The DID we're about to create is your *vendor DID*. It will be used in the all of the next steps.
+For the API calls that will need to be made to the Nuts node, we'll use ``<internal-node-address>`` as the address where the internal API's are exposed.
+Consult the :ref:`configuration reference <nuts-node-config>` on how to configure the node address.
+
+.. code-block:: text
+
+    POST <internal-node-address>/internal/vdr/v1/did
+    {
+        "selfControl": true,
+        "assertionMethod": true,
```
reinkrul, line 51: assertionMethod and capabilityInvocation are `true` by default, I think?

docs/pages/getting-started/4-connecting-crm.rst
```diff
@@ -0,0 +1,216 @@
...
+==================================
+
+Things can go wrong: a node is misbehaving or a DID Document is conflicted.
+If the node operator is not resolving the problem it's extremely convenient if others can contact the node operator and relay the problem.
+For this use-case, Nuts supports the registration of node contact information. The contact information will be added to a DID Document as a service.
+A convenience API is available to add the contact information to a DID Document. The vendor DID should be used for this.
+
+.. code-block:: text
+
+    PUT <internal-node-address>/internal/didman/v1/did/{did}/contactinfo
```
reinkrul, line 99: maybe we can use 1 style of parameterization of the URL ( <...> vs {...} )

docs/pages/getting-started/4-connecting-crm.rst
```diff
@@ -0,0 +1,216 @@
...
+
+.. code-block:: text
+
+    POST <internal-node-address>/internal/didman/v1/did/{did}/endpoint
+    {
+        "type": "example-production-api",
+        "endpoint": "https://api.example.com"
+    }
+
+Where ``{did}`` must be replaced with the vendor DID. The ``type`` may be freely chosen and is used as reference in the organization services.
```
reinkrul, line 123: why must endpoint be a fully qualified domain name? It entirely depends on the type of endpoint?

reinkrul (changes requested):
Not fully reviewed yet

docs/pages/getting-started/4-connecting-crm.rst
woutslakhorst: We haven't specified a maximum length, did we?

docs/pages/getting-started/4-connecting-crm.rst
woutslakhorst: I didn't want to bother here with what's default and what not, doesn't make thing more clear.

docs/pages/getting-started/4-connecting-crm.rst
woutslakhorst: using <> now

docs/pages/getting-started/4-connecting-crm.rst
woutslakhorst: changed it

docs/pages/getting-started/4-connecting-crm.rst
```diff
@@ -0,0 +1,216 @@
...
+================================
+
+A node operator must not blindly trust all the data is published over the network. Before credentials can be found, the issuer has to be trusted.
+By default, no issuers are trusted. A list of untrusted issuers can be obtained from the node through:
+
+.. code-block:: text
+
+    GET <internal-node-address>/internal/vcr/v1/NutsOrganizationCredential/untrusted
+
+This will return a list of all DIDs that are currently not trusted. If a DID is to be trusted should be validated out-of-band.
```
reinkrul, line 192: it should be validated out-of-band (missing "it")

Does "everyone" know what out-of-band is or do we need to provide an example?

docs/pages/getting-started/4-connecting-crm.rst
```diff
@@ -0,0 +1,216 @@
...
+
+    PUT <internal-node-address>/internal/didman/v1/did/{did}/contactinfo
+    {
+        "name": "vendor X",
+        "phone": "06-12345678",
+        "email": "info@example.com",
+        "website": "https://example.com"
+    }
+
+Where ``{did}`` must be replaced with the vendor DID.
```
reinkrul, line 107: Should we mention the contact info isn't authenticated in any way and thus (e.g. the name) shouldn't be trusted unless verified in some way?

docs/pages/getting-started/4-connecting-crm.rst
```diff
@@ -0,0 +1,216 @@
...
+
+Where ``{did}`` must be replaced with the vendor DID. The ``type`` may be freely chosen and is used as reference in the organization services.
+The ``endpoint`` must be a fully qualified domain name. For some services this could be a base-url. If this is the case, the bolt description will note this.
+
+Organization integration
+************************
+
+Each organization (or customer) must be registered with its own DID and DID Document.
+The vendor CRM should make it possible to store a DID for each organization.
+Requests that are made in the context of the organization will use the private-key of the organization.
```
reinkrul, line 131: private key ("koppelstreepjes" are almost never used in English texts)

docs/pages/getting-started/4-connecting-crm.rst
```diff
@@ -0,0 +1,216 @@
...
+This getting started manual assumes a vendor sells services to its customers.
+The vendor manages the presence of those customers on the Nuts network through the Nuts registry.
+It's very likely the vendor has software to manage customer environments. We'll call it a *CRM* where the *customers* part relates to organizations and not people.
+This does not mean that a stand-alone installation isn't supported. In that case the vendor and organization are the same.
+
+The Nuts registry enables service discovery for organizations. The registry identifies organizations through their :ref:`DID <did>`.
+The DIDs are new and do not match the internal identifiers known to the CRM.
+This connection has to be made in order for the CRM to make adjustments to DID Documents and issue Verifiable Credentials.
+
+All APIs used in the following chapters are documented at the :ref:`API <nuts-node-api>` page.
```
reinkrul, line 15: Maybe we should mention that integrators shouldn't store the DID Document itself in their software, but only the DID (because otherwise it can become stale when compared to the actual version on the network). So they don't need to store the result of the create operation, just its ID.

docs/pages/getting-started/4-connecting-crm.rst
```diff
@@ -0,0 +1,216 @@
...
+    }
+
+Where ``{did}`` must be replaced with the validated DID.
+It's also possible to update the ``vcr/trusted_issuers.yaml`` file located in the data directory (configured via the ``datadir`` property).
+After a vendor has been trusted, any of its registered organizations should be searchable by name.
+
+.. note::
+
+    Future development will see new cryptographic means. These means could enable the organization to self-register its name.
+    The network should then transfer to a trust model where the issuer of those means is trusted instead of the different vendors.
```
reinkrul, line 211: transfer -> migrate/move

reinkrul (changes requested):

docs/pages/getting-started/4-connecting-crm.rst
woutslakhorst: "it" isn't needed. I added an example.

docs/pages/getting-started/4-connecting-crm.rst
woutslakhorst: yes

docs/pages/getting-started/4-connecting-crm.rst
woutslakhorst: changed to migrate

reinkrul (approved):

PR #294, opened 2021-05-17
Network: reverted accidental struct rename

reinkrul:
Funny nobody noticed :-)


woutslakhorst (approved):
how did we miss that?

PR #295, opened 2021-05-17
Network: fix racy behaviour while broadcasting to a closing connection

reinkrul:
P2P layer used connection out channel directly instead of using `send()` which is protected by a mutex.


woutslakhorst (approved):
lg2m

PR #296, opened 2021-05-18
Refactored DIDMan API to auto-generate problem titles and derive status codes from occurred errors

reinkrul:
Proposal for improved error handling in API layers:
- We're creating lots of almost-duplicate code (create error, log, report problem).
- Derives the problem title from the API function name, which is determined by the OpenAPI operation name in the YAML spec. This makes problem construction less error-prone (since you can't provide the wrong problem title)

If agreed upon the new helper funcs and types should probably be moved to its own, shared package.

We could even move `reportProblem` to an Echo filter.


reinkrul:
Is this something to like?

woutslakhorst:
It makes the APIs cleaner to look at, reduces complexity and reduces the number of asserts needed.
👍

gerardsn:
I think it is something to like.
I was looking at this before but could not figure out how to get `reportProblem()` to  use the correct logger.

reinkrul:
> I think it is something to like.
> I was looking at this before but wasn't sure how I could get `reportProblem()` to use the correct engine specific logger.

The not-so-nice solution is to simply pass it to `reportProblem`. A bit nicer (but slightly overengineered) might be to make a `ProblemReporter` struct that holds the target logger as field. Each API `Wrapper` then holds an instance of a `problemReporter`.

On the other hand, maybe this doesn't need to be called by each engine by itself and we should make it a middleware function?

To avoid an ugly dependency (middleware having a direct dependency on each engine's domain-specific errors), each engine API wrapper could register the errors they're expecting with the wanted HTTP response code (e.g. in a static initializer in the API package).

woutslakhorst:
An option would be to derive the module from the http path. Currently this is a 1-on-1 mapping. Then echo middleware can add the module to the logging.

stevenvegt:
Looks good. I prefer the middleware option as well.
inputError -> ArgumentError? Or RequestFormatError?
Perhaps we can add the API operationID  to the echo context using some custom api generator logic? -> https://github.com/deepmap/oapi-codegen/tree/master/pkg/codegen/templates

woutslakhorst:
@reinkrul shall we close this PR or mark it as blocked until the middleware option has been worked out?

reinkrul:
> @reinkrul shall we close this PR or mark it as blocked until the middleware option has been worked out?

I was trying to work it out in middleware between tasks, but got stuck refactoring. Will report my findings here.

woutslakhorst:
> Edit: I'd say merge this and make an issue to solve it in a broader sense?

Can you link the issue here?

woutslakhorst (approved):

reinkrul:
Moved the error handling to the middleware.

API wrappers now can implement the `ErrorStatusCodeMapper` to provide custom error -> http status code mapping. If a mapping can be found, the problem is logged as warning. If not an error 500 is returned and it's logged as error.

I had to alter the OpenAPI template (`register.tmpl`) to provide the middleware info on the operation that was executed (for the problem title) and the status code mapping provided by the API wrapped.

After https://github.com/nuts-foundation/go-did/pull/43 is merged errors returned from DID parsing can also be mapped to http status codes.

@woutslakhorst @stevenvegt @gerardsn can you give feedback on this solution before I take the effort of satisfying codeclimate and fix/write all tests?

core/echo_errors.go
```diff
@@ -0,0 +1,121 @@
...
+			if _, err := result.WriteTo(ctx.Response()); err != nil {
+				ctx.Echo().Logger.Error(err)
+			}
+		} else {
+			ctx.Echo().Logger.Warnf("Unable to send error back to client, response already committed: %v", err)
+		}
+	}
+}
+
+func NotFoundError(errStr string, args ...interface{}) HTTPStatusCodeError {
```
woutslakhorst, line 40: you could have these funcs return the `error` interface and make `HTTPStatusCodeError` private. That would make it less confusing for others to use it. (or misuse it)

network/api/v1/api.go
```diff
@@ -48,7 +43,7 @@ func (a Wrapper) ListTransactions(ctx echo.Context) error {
 	transactions, err := a.Service.ListTransactions()
 	if err != nil {
 		log.Logger().Errorf("Error while listing transactions: %v", err)
```
woutslakhorst, line 45: example of redundant logger

woutslakhorst (commented):
Lots of loggers can be removed from the API classes.

without changes to generated code, it would remove lines of code, so that's a plus.

network/api/v1/api.go
reinkrul: You're right, I missed those

crypto/api/v1/api_test.go
```diff
@@ -48,10 +47,7 @@ func TestWrapper_SignJwt(t *testing.T) {
 
 		err := ctx.client.SignJwt(ctx.echo)
 
-		test2.AssertErrProblemTitle(t, problemTitleSignJwt, err)
-		test2.AssertErrProblemStatusCode(t, http.StatusBadRequest, err)
-		test2.AssertErrProblemDetail(t, "missing claims", err)
-
+		assert.EqualError(t, err, "invalid sign request: missing claims")
```
woutslakhorst, line 50: same here for testing

test/error.go
```diff
@@ -0,0 +1,34 @@
...
+package test
+
+import (
+	"errors"
+	"github.com/stretchr/testify/assert"
+	"testing"
+)
+
+// AssertIsError asserts that expected is, or is the cause of the given actual error (according to errors.Is()).
+func AssertIsError(t *testing.T, actual error, expected error) bool {
```
woutslakhorst, line 28: How does this differ from : https://github.com/stretchr/testify/blob/dc5c261377d971ad796211563f0e3e73954af5e6/assert/assertions.go#L1734

woutslakhorst (changes requested):
Happy with the cleanup 👍 

some small stuff left

test/error.go
reinkrul: Not very much. I upgraded the lib and now we can use it 👍  (finally it's there)

auth/api/v1/api_test.go
```diff
@@ -155,9 +156,7 @@ func TestWrapper_GetSignSessionStatus(t *testing.T) {
 
 		err := ctx.wrapper.GetSignSessionStatus(ctx.echoMock, signingSessionID)
 
-		test2.AssertErrProblemTitle(t, problemTitleSignSessionStatus, err)
-		test2.AssertErrProblemStatusCode(t, http.StatusNotFound, err)
-		test2.AssertErrProblemDetail(t, "failed to get session status for 123, reason: session not found", err)
+		test2.AssertIsError(t, err, services.ErrSessionNotFound)
```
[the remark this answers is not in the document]
reinkrul: Not entirely sure about that. The test of `ErrorStatusCodes()` will literally assert the contents of the hardcoded map? Not sure how to test it properly, though. Without making integration tests of course.

Question is also, what would be the impact if it were to break at some point. It'll only return a HTTP 500 instead of the mapped error code (e.g. 400, 403, etc) in case the error occurs. The error response is still returned. So I don't think it's worth it.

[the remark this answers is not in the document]
woutslakhorst: But now there's no test if an error is actually correctly mapped. I can change a mapping and an API will return a status code which is not according to spec and no test will fail. In short we can break stuff without tests failing.

reinkrul:
@woutslakhorst ready for re-review

vcr/api/v1/api_test.go
```diff
@@ -91,43 +89,16 @@ func TestWrapper_CreateDID(t *testing.T) {
...
 		ctx.echo.EXPECT().Bind(gomock.Any())
-		ctx.echo.EXPECT().String(http.StatusBadRequest, gomock.Any())
 		ctx.vcr.EXPECT().Issue(gomock.Any()).Return(nil, credential.ErrValidation)
 
 		err := ctx.client.Create(ctx.echo)
 
-		assert.NoError(t, err)
-	})
-
-	t.Run("error - DID not found", func(t *testing.T) {
```
woutslakhorst, line 107: obsolete tests?

vcr/api/v1/api_test.go
```diff
@@ -91,43 +89,16 @@ func TestWrapper_CreateDID(t *testing.T) {
...
-		defer ctx.ctrl.Finish()
-
-		ctx.echo.EXPECT().Bind(gomock.Any())
-		ctx.echo.EXPECT().String(http.StatusBadRequest, gomock.Any())
-		ctx.vcr.EXPECT().Issue(gomock.Any()).Return(nil, fmt.Errorf("wrapped error: %w", types.ErrKeyNotFound))
-
-		err := ctx.client.Create(ctx.echo)
-
-		assert.NoError(t, err)
+		assert.ErrorIs(t, err, credential.ErrValidation)
```
woutslakhorst, line 101: missing status code checks for this file.

woutslakhorst (changes requested):
some leftover minor stuff.

vcr/api/v1/api_test.go
reinkrul: Yes they all tested the same: behaviour when the `vcr.Issuer()` returns an error

vcr/api/v1/api_test.go
reinkrul: Ah missed those, good find

PR #297, opened 2021-05-18
DIDMan: added client API calls for AddEndpoint and AddCompoundService

reinkrul:


woutslakhorst (commented):
This would be code purely for the registry admin demo backend?

Then this might not be the correct place. The client code can be defined in the registry admin demo....

stevenvegt (approved):

Issue #266, opened 2021-04-30
Add canControl func


woutslakhorst:
At several places the first controller + key is taken/checked to change a DID Document or sign a transaction.

we need a function that checks if a private key exists (and which key/controller that is) that can alter a DID Document.

woutslakhorst:
such a function is needed per key type.

woutslakhorst:
Only the update function checks for a controller key, so closing issue for now.

PR #299, opened 2021-05-20
Search all controllers for CapabilityInvocation key when updating DID Document

woutslakhorst:
What started out as a fix for #266 and a refactor, resulted in this minimum change.

The issue stated that controller checking is required at multiple places....
Right now it's only done at `vdr.Update`.
All VDR APIs rely on Update failing to return the correct status code.

There's quite some logic in auth that could benefit from a refactor, but it isn't broken. The selection of assertion keys in auth does a check via the DID Document through public key resolving and then resolves the private key in the keystore. This leads to quite some duplicate error checking. But refactoring this is quite a challenge. It could theoretically be possible to use an old assertion key for signing. It won't be usable since the other node will reject it based on assertion key validity in the DID Document.


woutslakhorst:
also ran gofmt

vdr/vdr.go
```diff
@@ -162,7 +162,17 @@ func (r VDR) Update(id did.DID, current hash.SHA256Hash, next did.Document, _ *t
...
+			key, err = r.keyStore.Resolve(cik.ID.String())
+			if err == nil {
+				break outer
+			}
+		}
+	}
+
 	if err != nil {
 		if errors.Is(err, crypto.ErrKeyNotFound) {
 			return types.ErrDIDNotManagedByThisNode
```
reinkrul, line 178: The error on line 180 is now incorrect because it insinuates a single controller, but all controllers are checked.

`return fmt.Errorf("could not find capabilityInvocation key for controller: %w", err)`

This should be something along the lines of "could not find capabilityInvocation key for updating the DID document"

vdr/vdr.go
```diff
@@ -162,7 +162,17 @@ func (r VDR) Update(id did.DID, current hash.SHA256Hash, next did.Document, _ *t
 		return err
 	}
 
-	key, err := r.keyStore.Resolve(controllers[0].CapabilityInvocation[0].ID.String())
+	var key crypto.Key
+outer:
+	for _, c := range controllers {
```
reinkrul, line 167: I checked what `ResolveControllers` actually does and this function's (`Update()`) behaviour was incorrect and still is incorrect when you read RFC006, which states:

> - a key referenced from the `capabilityInvocation` section of the latest DID document version.
- a key referenced from the `capabilityInvocation` section of a controller. One of the entries in the `controller` field MAY refer to a different DID Document.
  The `capabilityInvocation` keys of the latest version of that DID Document can be used as authorized key.

But `ResolveControllers()` returns _either_ the document itself as controller when is has no controller, but has capabilityInvocation keys, _or_ the controllers, not being the document itself. So according to RFC006 when the document has controllers (not including the subject DID itself) and a capabiltyInvocation key it can be used to upadte the DID document, but this `Update()` function won't use it (due to the implementation of `ResolveControllers`).

reinkrul (changes requested):
I had to read the code to understand what this PR is actually about, the summary doesn't tell what has improved and the description doesn't mention that it now uses all controllers to determine the capInv key to be used. I'd expect something like "Search all controllers for capabilityInvocation key when updating, instead of only the first one", or something

vdr/vdr.go
woutslakhorst: Then the RFC is dodgy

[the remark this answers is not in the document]
woutslakhorst: Yes, I looked too much at coverage.

woutslakhorst: see nuts-foundation/nuts-specification#72

reinkrul:
> I had to read the code to understand what this PR is actually about, the summary doesn't tell what has improved and the description doesn't mention that it now uses all controllers to determine the capInv key to be used. I'd expect something like "Search all controllers for capabilityInvocation key when updating, instead of only the first one", or something

fixed the title for you :-p

reinkrul (approved):

Issue #242, opened 2021-04-19
VDR: ambassador is missing checks when updating DID Documents


woutslakhorst:
lots of comments and not implemented tests.

reinkrul:
TODO to fix:

```
// TODO: perform all these tests:
// Take capabilityInvocation keys from the controllers
// Check if network signingKeyID is one of capabilityInvocations of the controller
//
// For each verificationMethod in the next version document
// 		check if the provided key thumbprint matches the corresponding thumbprint in the key store
// Take diff of verificationMethods between next and current versions:
// if new verificationMethod is added:
// 		Add public key to key store
// if verificationMethod is removed:
//		Mark keyID as expired since the updatedAt time from new DID document
// make a diff of the controllers
// 	if controller is added
//		check if it is known.
// check if the transactions contains all SourceTransactions
```

PR #310, opened 2021-05-27
VDR: Implemented missing checks

reinkrul:
Fixes https://github.com/nuts-foundation/nuts-node/issues/242


vdr/validators.go
```diff
@@ -24,10 +41,15 @@ type serviceValidator struct{}
 
 func (s serviceValidator) Validate(document did.Document) error {
 	knownServiceIDs := make(map[string]bool, 0)
+	knownServiceTypes := make(map[string]bool, 0)
 	for _, method := range document.Service {
 		if err := verifyDocumentEntryID(document.ID, method.ID, knownServiceIDs); err != nil {
 			return fmt.Errorf("invalid service: %w", err)
 		}
+		if knownServiceTypes[method.Type] {
+			return errors.New("invalid service: duplicate service type")
```
woutslakhorst, line 50: make it a `ErrDuplicateService` so you can return a 400 when called from API?

woutslakhorst: also relates to other PR

woutslakhorst (changes requested):

vdr/validators.go
reinkrul: Will fix this in https://github.com/nuts-foundation/nuts-node/pull/313 since API call if only possible through that PR

woutslakhorst (approved):

PR #315, opened 2021-05-27
DIDMan administer compound services

stevenvegt:
Add a Get CompoundServices endpoint
Return newly created Service from endpoint call.


didman/didman.go
```diff
@@ -67,19 +67,47 @@ func (d *didman) Name() string {
...
+	doc, _, err := d.docResolver.Resolve(id, nil)
+	if err != nil {
+		return err
+	}
+
+	var serviceID *ssi.URI
+	for _, s := range doc.Service {
+		if s.Type == serviceType {
+			serviceID = &s.ID
+			break
```
woutslakhorst, line 89: duplicate services possible when conflicted

woutslakhorst (changes requested):
Are your tools up2date? The changes to the mock files shouldn't happen.

didman/didman.go
```diff
@@ -223,6 +252,19 @@ func (d *didman) validateCompoundServiceEndpoint(references map[string]ssi.URI)
 	return nil
 }
 
+func filterCompoundServices(doc *did.Document) []did.Service {
+	var compoundServices []did.Service
+	for _, service := range doc.Service {
+		if service.Type == ContactInformationServiceType {
+			continue
+		}
+		if _, ok := service.ServiceEndpoint.(map[string]interface{}); ok {
```
reinkrul, line 261: I think this is rather naive. Don't we have a more concise way to determine whether an endpoint is (_probably_) a compound service?

didman/didman.go
stevenvegt: Yes, so, delete all of them?

didman/didman.go
stevenvegt: The `ServiceEndpoint` of `Endpoint` and `EndpointRef` are of type `string`, compound is type `map[string]interface{}`. I don't really see another way?

didman/types.go
```diff
@@ -34,18 +34,27 @@ type Didman interface {
 	// AddEndpoint adds a service to a DID Document. The serviceEndpoint is set to the given URL.
 	// It returns ErrDuplicateService if a service with the given type already exists.
 	// It can also return various errors from DocResolver.Resolve and VDR.Update
-	AddEndpoint(id did.DID, serviceType string, u url.URL) error
+	AddEndpoint(id did.DID, serviceType string, u url.URL) (*did.Service, error)
+
+	// DeleteEndpoint takes an did and type and removes the endpoint with the type from the DID Document.
```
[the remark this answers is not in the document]
stevenvegt: We can also delete them all when more are matching. It feels weird to work with types on resolution and creation but work with ids when deleting.

stevenvegt:
I'm using v 1.5.0, should be te latest.

didman/api/v1/api.go
```diff
@@ -103,17 +143,30 @@ func (w *Wrapper) AddCompoundService(ctx echo.Context, didStr string) error {
...
-	for key, value := range request.Endpoint {
+	for key, value := range request.ServiceEndpoint {
 		uri, err := interfaceToURI(value)
 		if err != nil {
 			return core.InvalidInputError("invalid reference for service '%s': %v", key, err)
 		}
 		references[key] = *uri
 	}
-	if err = w.Didman.AddCompoundService(*id, request.Type, references); err != nil {
+	service, err := w.Didman.AddCompoundService(*id, request.Type, references)
```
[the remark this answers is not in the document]
stevenvegt: I don't like copying types. The risk is changing the spec but not seeing that change back in the codebase.

woutslakhorst:
> I'm using v 1.5.0, should be te latest.

toolchain is connected to go.mod file. The makefile target `install-tools` installs the correct versions. Your mockgen is newer, so also update the go.mod file

docs/_static/didman/v1.yaml
```diff
@@ -81,10 +83,43 @@ paths:
...
-          description: The given type and URL have been added as service to the DID Document
+        "200":
+          description: The service which has been added to the DID Document
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Endpoint'
+        default:
+          $ref: '../common/error_response.yaml'
+  /internal/didman/v1/did/{did}/endpoint/{type}:
```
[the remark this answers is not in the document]
stevenvegt: Renamed to deleteEndpointsByType

docs/_static/didman/v1.yaml
```diff
@@ -61,30 +61,67 @@ paths:
...
+        schema:
+          type: string
+    delete:
+      operationId: deleteEndpointsByType
+      description: Delete all endpoints with the provided type from the DID Document.
+      responses:
+        204:
+          description: |
+            All existing endpoints with with this type were succesfully deleted.
+        400:
```
woutslakhorst, line 121: error responses should go in the description. A `default` is added with a schema, the current setup would suggest `400` and `404` do not return a error return body.

docs/_static/didman/v1.yaml
```diff
@@ -97,6 +134,20 @@ paths:
           - "did:nuts:1234"
         schema:
           type: string
+    get:
+      operationId: getCompoundServices
+      summary: Get a list of compound services for a DID document.
+      responses:
+        "200":
```
woutslakhorst, line 141: missing default return (for 404/500)

didman/didman.go
```diff
@@ -67,19 +67,48 @@ func (d *didman) Name() string {
...
+
+	found := false
+	for _, s := range doc.Service {
+		if s.Type == serviceType {
+			if err = d.DeleteService(s.ID); err != nil {
+				return err
+			}
+		}
+	}
+	if !found {
```
woutslakhorst, line 93: found is never set to true

woutslakhorst (changes requested):

docs/_static/didman/v1.yaml
```diff
@@ -174,6 +245,39 @@ components:
           description: type of the endpoint. May be freely choosen.
           type: string
         endpoint:
+          description: An endpoint URL or a reference to another service.
+          type: string
+          example:
+            - ref://did:nuts:123/serviceEndpoint?type=eOverdracht-fhir
```
woutslakhorst, line 251: wrong ref URI

docs/_static/didman/v1.yaml
```diff
@@ -150,22 +215,28 @@ paths:
...
         type:
           description: type of the endpoint. May be freely choosen.
           type: string
         endpoint:
-          description: A URL.
+          description: An endpoint URL or a reference to another service.
           type: string
-    CompoundServiceCreateRequest:
+          example:
+            - ref://did:nuts:123/serviceEndpoint?type=eOverdracht-fhir
```
woutslakhorst, line 235: wrong ref URI

woutslakhorst (changes requested):
only some wrong examples left. Could you also link an issue for nuts-node where the ref URI's resolvement is implemented?

woutslakhorst (approved):

Issue #308, opened 2021-05-27
Network: implement diagnostics querying


reinkrul:
Recently added to RFC005

PR #316, opened 2021-05-31
Allow nested diagnostics and let REST clients retrieve them as JSON

reinkrul:
In preparation of https://github.com/nuts-foundation/nuts-node/issues/308 to allow nested diagnostics, so they can be structured per-peer.


reinkrul:
Not a good change

PR #317, opened 2021-05-31
Network: broadcast diagnostics to peers and expose it through API

reinkrul:


network/proto/impl.go
```diff
@@ -101,31 +127,42 @@ func (p protocol) startAdvertingHashes() {
 	}
 }
 
+func (p protocol) startQueryingPeerDiagnostics() {
```
woutslakhorst, line 130: how do these routines stop nicely?

network/proto/interface.go
```diff
@@ -42,11 +42,14 @@ var ErrUnsupportedProtocolVersion = errors.New("unsupported protocol version")
 type Protocol interface {
 	core.Diagnosable
 	// Configure configures the Protocol. Must be called before Start().
-	Configure(p2pNetwork p2p.Adapter, graph dag.DAG, publisher dag.Publisher, payloadStore dag.PayloadStore, advertHashesInterval time.Duration, peerID p2p.PeerID)
+	Configure(p2pNetwork p2p.Adapter, graph dag.DAG, publisher dag.Publisher, payloadStore dag.PayloadStore, diagnosticsProvider func() Diagnostics,
```
woutslakhorst, line 45: the `diagnosticsProvider` could also just be an interface.
Also: too many arguments....

woutslakhorst (changes requested):
What about a CLI command?

network/network.go
```diff
@@ -92,9 +94,12 @@ func (n *Network) Configure(config core.ServerConfig) error {
 	n.graph = dag.NewBBoltDAG(db, dag.NewSigningTimeVerifier(), dag.NewPrevTransactionsVerifier(), dag.NewTransactionSignatureVerifier(n.keyResolver))
 	n.payloadStore = dag.NewBBoltPayloadStore(db)
 	n.publisher = dag.NewReplayingDAGPublisher(n.payloadStore, n.graph)
-	peerID := p2p.PeerID(uuid.New().String())
-	n.protocol.Configure(n.p2pNetwork, n.graph, n.publisher, n.payloadStore, time.Duration(n.config.AdvertHashesInterval)*time.Millisecond, peerID)
-	networkConfig, p2pErr := n.buildP2PConfig(peerID)
+	n.peerID = p2p.PeerID(uuid.New().String())
+	n.protocol.Configure(n.p2pNetwork, n.graph, n.publisher, n.payloadStore, n.collectDiagnostics,
```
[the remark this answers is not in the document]
reinkrul: Yeah I noticed... Could indicate the type is doing too much. I think fixing that would mean quite a big refactor of how the protocol is implemented (it's really no-nonsense, straight coding right now). Might have to look how other systems implement their protocol handling to avoid a God-class like we're creating now. Without DI of course.

network/config.go
```diff
@@ -18,17 +18,20 @@ type Config struct {
...
+	QueryPeerDiagnosticsInterval int
 }
 
 // DefaultConfig returns the default NetworkEngine configuration.
 func DefaultConfig() Config {
 	return Config{
-		GrpcAddr:             ":5555",
-		EnableTLS:            true,
-		AdvertHashesInterval: 2000,
+		GrpcAddr:                     ":5555",
```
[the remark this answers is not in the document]
reinkrul: I'd say this function is the const itself. They won't be repeated, they're immutable (since the function returns a new instance every time) and they're named.

network/proto/impl.go
reinkrul: When the program dies, the goroutines die with it (not like Java's non-daemon threads). The network engine's `Stop()` is called when the program halts, which in turn shuts down the P2P layer, which closes all active gRPC connections.  Since all access to these connections is protected using mutexes, race conditions should not occur. It could happen that diagnostics are collected but there are no more peer connections to broadcast it to, because the program is halting, but no harm would be done.

network/network.go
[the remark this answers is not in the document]
reinkrul: Discussed with @woutslakhorst, made an issue for it so it can be refactored later: https://github.com/nuts-foundation/nuts-node/issues/330

network/proto/impl.go
```diff
@@ -39,14 +39,21 @@ type protocol struct {
 	receivedPeerHashes        *chanPeerHashQueue
 	receivedTransactionHashes *chanPeerHashQueue
 
-	peerOmnihashChannel chan PeerOmnihash
+	// peerDiagnostics contains diagnostic information of the node's peers. The key contains the remote peer's ID. Access must be protected using peerDiagnosticsMutex
+	peerDiagnostics      map[p2p.PeerID]Diagnostics
+	peerDiagnosticsMutex *sync.Mutex
```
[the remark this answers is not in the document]
reinkrul: It's written to in `updateDiagnostics()`

reinkrul:
@woutslakhorst ready for re-review

docs/_static/network/v1.yaml
```diff
@@ -101,7 +101,6 @@ paths:
           content:
             application/json:
               schema:
-                type: object
```
woutslakhorst: must not be removed (actual a conflict with master)

woutslakhorst (changes requested):
only a breaking change

docs/_static/network/v1.yaml
reinkrul: Reverted it. Didn't actually change code since we override that type with our own, since the code generator doesn't properly handle additionalProperties (making it impossible to consume/produce a map)

woutslakhorst (approved):

reinkrul:
Discussed with @stevenvegt: operators should be able to disable the broadcasting and it should be noted in the production considerations, since some parties might be allergic to (automatic) reporting processes.

PR #318, opened 2021-06-01
VDR: Use TX's signing time when resolving a DID document's controllers when updating

### LINKED ISSUE #311: VDR: Resolve controllers using transaction's signing time
`ResolveControllers` doesn't take the signing time of the transaction containing the document's controller being resolved into consideration.

When `ResolveControllers` should optionally take resolver input metadata so VDR's ambassador can resolve controllers in relation to the TXs signing time.

reinkrul:
It currently resolves to the latest versions, which is incorrect.

Fixes #311


woutslakhorst (changes requested):
Missing test on resolving controllers in the past.

woutslakhorst (approved):

PR #322, opened 2021-06-02
fix correct types for service

woutslakhorst:
inside a did.Document.service, an `endpoint` needs to be a `map[string]interface{}`, `[]interface{}` or `string`. In didman a `ContactInformation` struct is set as `endpoint`, this trips the validation. This only recently became a problem when validation was added before publishing.

an alternative would be to change the adding of a service.
It should then marshal/unmarshal which would always fool the validator....

### DIFF
--- a/didman/didman.go
+++ b/didman/didman.go
@@ -82,7 +82,7 @@ func (d *didman) AddCompoundService(id did.DID, serviceType string, references m
 		return err
 	}
 
-	// transform ContactInformation to map[string]interface{}
+	// transform service references to map[string]interface{}
 	serviceEndpoint := map[string]interface{}{}
 	for k, v := range references {
 		serviceEndpoint[k] = v.String()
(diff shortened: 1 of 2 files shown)


reinkrul:
Can you please state the problem in the title or description? Otherwise I'd have to dive into the code to see what problem it actually tries to fix

reinkrul (changes requested):
Can you please state the problem in the title or description? Otherwise I'd have to dive into the code to see what problem it actually tries to fix

didman/didman.go
```diff
@@ -81,7 +81,14 @@ func (d *didman) AddCompoundService(id did.DID, serviceType string, references m
 	if err := d.validateCompoundServiceEndpoint(references); err != nil {
 		return err
 	}
-	err := d.addService(id, serviceType, references, nil)
+
+	// transform ContactInformation to map[string]interface{}
```
reinkrul, line 85: incorrect comment

reinkrul (approved):

reinkrul:
Trivial thing left to fix

PR #333, opened 2021-06-07
removed one-of from vdr spec

### LINKED ISSUE #328: Remove one-of from VDR api spec
one-of, any-of and any other X-of are evil when using code generation.

woutslakhorst:
fixes #328 

`one-of` can cause problems for some open api spec code generators. Better to stick to a single type, eg `object`.

### DIFF
--- a/docs/_static/vdr/v1.yaml
+++ b/docs/_static/vdr/v1.yaml
@@ -345,9 +345,7 @@ components:
           type: string
         serviceEndpoint:
           description: Either a URI or a complex object.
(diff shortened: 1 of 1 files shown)


reinkrul (commented):
no changes in code generation?

woutslakhorst:
> no changes in code generation?

the change is in the `service` type which is part of `DIDDocument` which we aliased to `did.Document`.
Talking about not eating your own dogfood

reinkrul (approved):

PR #343, opened 2021-06-22
Rework connection handling to solve concurrency issues

### LINKED ISSUE #342: Network: peer diagnostics/state incorrect after reconnect
After a peer (Raspberry PI in electrical cabinet at home) disconnected and directly reconnected again to a node hosted in the cloud, the diagnostics on the cloud node are incorrect. It shows up in the "omnihashes" statistic, but not in the "connected peers":

```
[Protocol] Peer omnihashes: a776b8c9ac6f5349065612521de7a7ec9eb608af1a00f45f2c98dd58bb1cc41e={f7e7ef59-dbf4-46dd-b2f6-90c3dec4b3a0, ce26f36a-7dfb-4733-838b-c35e633a5cec, c1323417-ea15-40f1-a190-546b5b46bd5b, 6488eaf3-cec8-4986-a2c3-fd2f6a768a5e, 3a6d9005-7b12-464f-894d-c9057ef9bd96}
```

Log of Raspberry PI node:

```
time="2021-06-16T23:34:59Z" level=warning msg="Peer connection error (peer-id=edc2f210-6c39-4823-ab49-b47953ec57fc): rpc error: code = Unavailable desc = transport is closing" module=Network
time="2021-06-16T23:34:59Z" level=warning msg="Peer connection error (peer-id=c1323417-ea15-40f1-a190-546b5b46bd5b): rpc error: code = Unavailable desc = transport is closing" module=Network
time="2021-06-16T23:34:59Z" level=info msg="Connected to peer (id=edc2f210-6c39-4823-ab49-b47953ec57fc,addr=nuts.reinkrul.nl:5555)" module=Network
time="2021-06-16T23:35:00Z" level=info msg="Connected to peer (id=c1323417-ea15-40f1-a190-546b5b46bd5b,addr=********:5555)" module=Network
```

Log of cloud node:

```
time="2021-06-16T23:34:59Z" level=info msg="New peer connected (peer=f7e7ef59-dbf4-46dd-b2f6-90c3dec4b3a0@213.127.27.252:29120)" module=Network
time="2021-06-16T23:35:22Z" level=warning msg="Peer connection error (peer-id=f7e7ef59-dbf4-46dd-b2f6-90c3dec4b3a0): rpc error: code = Canceled desc = context canceled" module=Network
```

So the peers is actually connected, but the diagnostics are screwed up.

reinkrul:
Fixes #342

When a second connection for the same peer arrived, the end state of the diagnostics was undetermined because the handling was racy.

Ultimately fixed it by closing the first connection (making sure it's proper synchronized), but the code became even more spaghetti-ish (than it already was) so I introduced `connectionManager` to create/close connections.


network/p2p/connection.go
```diff
@@ -82,32 +87,166 @@ func (conn *connection) send(message *transport.NetworkMessage) error {
...
-func (conn connection) sendMessages() {
-	for message := range conn.outMessages {
-		if conn.messenger.Send(message) != nil {
-			log.Logger().Warnf("Unable to send message to peer (peer=%s)", conn.Peer)
+func (conn *managedConnection) exchange(receivedMessages messageQueue) {
+	conn.mux.Lock()
+	out := conn.outMessages
+	in := conn.receiveMessages()
+	conn.mux.Unlock()
+	for {
```
woutslakhorst, line 95: sending is preferred over reading. If you send a lot, would this overflow the in channel?

network/p2p/connection.go
```diff
@@ -82,32 +87,166 @@ func (conn *connection) send(message *transport.NetworkMessage) error {
...
+	conn.outMessages = nil
+}
+
+func (conn *managedConnection) receiveMessages() chan *PeerMessage {
+	peerID := conn.ID
+	messenger := conn.messenger
+	result := make(chan *PeerMessage, 10)
+	go func() {
+		for {
+			msg, recvErr := messenger.Recv()
```
woutslakhorst, line 149: What happens to this call when we close the connection?

network/p2p/connection.go
```diff
@@ -82,32 +87,166 @@ func (conn *connection) send(message *transport.NetworkMessage) error {
...
-	for {
-		msg, recvErr := conn.messenger.Recv()
-		if recvErr != nil {
-			if recvErr == io.EOF {
-				log.Logger().Infof("Peer closed connection (peer-id=%s)", conn.ID)
-			} else {
-				log.Logger().Warnf("Peer connection error (peer-id=%s): %v", conn.ID, recvErr)
-			}
-			break
+func (conn *managedConnection) close() {
```
woutslakhorst, line 118: Does close need to do anything with `conn.messenger`?

network/p2p/impl.go
```diff
@@ -300,18 +282,10 @@ func (n adapter) ConnectToPeer(address string) bool {
...
-func (n *adapter) startSendingAndReceiving(conn *connection) {
-	go conn.sendMessages()
-	n.registerConnection(conn)
-	conn.receiveMessages(n.receivedMessages)
-	// When we reach this line, receiveMessages has exited which means the connection has been closed.
-	n.unregisterConnection(conn)
-}
-
 // connectToNewPeers reads from connectorAddChannel to start connecting to new peers
 func (n *adapter) connectToNewPeers() {
```
woutslakhorst, line 286: You probably want to migrate this to the connectionManager in the future.

woutslakhorst (changes requested):

network/network_integration_test.go
```diff
@@ -71,7 +71,10 @@ func TestNetworkIntegration_HappyFlow(t *testing.T) {
 	defer func() {
 		node2.Shutdown()
 		node1.Shutdown()
-		bootstrap.Shutdown()
+		err := bootstrap.Shutdown()
+		if !assert.NoError(t, err) {
```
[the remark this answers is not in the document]
reinkrul: changed to `FailNow`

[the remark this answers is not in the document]
reinkrul: Actually, removed it altogether because it was for debbuging the test

network/p2p/connection.go
```diff
@@ -82,32 +87,166 @@ func (conn *connection) send(message *transport.NetworkMessage) error {
 	return nil
 }
 
-// sendMessages (blocking) reads messages from its outMessages channel and sends them to the peer until the channel is closed.
-func (conn connection) sendMessages() {
-	for message := range conn.outMessages {
-		if conn.messenger.Send(message) != nil {
-			log.Logger().Warnf("Unable to send message to peer (peer=%s)", conn.Peer)
+func (conn *managedConnection) exchange(receivedMessages messageQueue) {
+	conn.mux.Lock()
```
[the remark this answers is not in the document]
reinkrul: otherwise access to `conn.outMessages` (and fields that `receiveMessages` accesses) is racy (according to `-race`).

E.g. calls to `close()` (called async from this func) nill those fields, thus theoretically causing racy behaviour (in practce not, but OK).

network/p2p/connection.go
reinkrul: Good point, I actually don't know how this functions.

Tested it, turns out it's some it's sort of random although the first select `case` will always have precidence:

```
func TestSelectChanPrio(t *testing.T) {
	const count = 5
	chanA := make(chan string, count)
	chanB := make(chan string, count)
	for i := 0; i < count; i++ {
		chanA <- "A"
	}
	for i := 0; i < count; i++ {
		chanB <- "B"
	}

	for {
		select {
		case item := <-chanA:
			println(item)
		case item := <-chanB:
			println(item)
		}
	}
}
```

Outputs:

```
AABABABBBA
AABABBABAB
AABAABABBB
ABABABBBAA
```

It's not 100% fair, but good enough I think?

network/p2p/connection.go
reinkrul: This is a bit complex and depends on the side of the connection:

- client side `Recv()` is cancelled by closing the `ClientStream` (status = cancelled)
- server side `Recv()` is cancelled by letting the gRPC service handler (implementation of `Connect()`) exit, which closes the underlying gRPC channel, which in turn cancels `Recv()` (status = EOF)

I documented this behaviour a bit better.

network/p2p/connection.go
reinkrul: Not necessarily (because the connection will get GC'd anyways), but it might be more clean (`nil` all fields when `close()` is called) and makes it impossible to accidentally receive/send messages after it has been closed. Will add.

network/p2p/impl.go
reinkrul: It might. I'm not entirely sure how the design will end up at this point, because it risks connectionManager becomes the God object that `adapter` was. I think OK-ish for now, will have to draw diagram to work out the exact components and reponsibilities. (adapter, connector, connection, connectionManger)

network/p2p/connection.go
```diff
@@ -33,17 +34,29 @@ type grpcMessenger interface {
 	Recv() (*transport.NetworkMessage, error)
 }
 
-func createConnection(peer Peer, messenger grpcMessenger) *connection {
-	return &connection{
+type connection interface {
+	// exchange must be called on the connection for it to start sending (using send()) and receiving messages.
+	// Received messages are passed to the receivedMessages message queue. It blocks until the connection is closed,
+	// by either the peer or the local node.
+	exchange(receivedMessages messageQueue)
```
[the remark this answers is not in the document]
reinkrul: good idea. I also think `messageQueue` doesn't add value over just `chan Message`, so it should probably be removed. Out of scope for this PR.

I'll rename the argument.

network/p2p/connection.go
woutslakhorst: It behaves a bit like go routines, so no need to use go routines 👍

network/p2p/connection.go
woutslakhorst: 👍

woutslakhorst (approved):

PR #364, opened 2021-07-07
first trial run for codeclimate coverage

### LINKED ISSUE #357: Migrate away from codecov
The coverage numbers from codecov are very unpredictable.

Codeclimate would be an options since we already use it for code quality.

woutslakhorst:
closes #357 

check the beautiful check marks in the build overview

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -20,8 +20,11 @@ jobs:
             - "/go/pkg/mod"
 
       - run: go test ./... -race
-      - run: go test ./... -coverprofile=c.out
-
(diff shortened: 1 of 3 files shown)


reinkrul:
Total coverage would be 64%? I wonder if that's correct. And it should not be passable imo?

woutslakhorst:
> Total coverage would be 64%? I wonder if that's correct. And it should not be passable imo?

it's all about the settings, I'll check what can be ignored (mock, generated)

woutslakhorst:
It's not ignoring the excluded files for coverage

reinkrul (approved):

PR #368, opened 2021-07-12
Iss351 vc resolve period validation

### LINKED ISSUE #351: VCR: Take end date into consideration when searching for credentials
Part of validation

woutslakhorst:
Closes #351 
Also part of nuts-foundation/nuts-specification#108, it solves the validity problem of VCs where the issuer has been deactivated.

This PR adds a `validate` and `validateForUse` private function to the vcr which does runtime checks.
The Resolve API now also has an optional `at` query param for resolving at a specific time.

### DIFF
--- a/docs/_static/vcr/v1.yaml
+++ b/docs/_static/vcr/v1.yaml
@@ -54,7 +54,7 @@ paths:
         * 500 - An error occurred while processing the request
       operationId: "resolve"
       parameters:
-        - name: at
+        - name: resolveTime
           in: query
           description: a rfc3339 time string for resolving a VC at a specific moment in time
           example: "2012-01-02T12:00:00Z"
--- a/docs/_static/vcr/v1.yaml
+++ b/docs/_static/vcr/v1.yaml
@@ -57,6 +57,7 @@ paths:
         - name: at
           in: query
           description: a rfc3339 time string for resolving a VC at a specific moment in time
+          example: "2012-01-02T12:00:00Z"
           schema:
             type: string
       tags:
--- a/docs/_static/vcr/v1.yaml
+++ b/docs/_static/vcr/v1.yaml
@@ -53,6 +53,12 @@ paths:
         * 404 - Corresponding credential could not be found
         * 500 - An error occurred while processing the request
       operationId: "resolve"
+      parameters:
+        - name: at
+          in: query
+          description: a rfc3339 time string for resolving a VC at a specific moment in time
+          schema:
+            type: string
       tags:
         - credential
       responses:
--- a/vcr/api/v1/api.go
+++ b/vcr/api/v1/api.go
@@ -21,6 +21,7 @@ package v1
 
 import (
 	"net/http"
+	"time"
 
 	ssi "github.com/nuts-foundation/go-did"
 
@@ -83,7 +84,7 @@ func (w *Wrapper) Search(ctx echo.Context, conceptTemplate string) error {
 		query.AddClause(concept.Prefix(kvp.Key, kvp.Value))
 	}
 
-	VCs, err := w.R.Search(query)
+	VCs, err := w.R.Search(query, nil)
 	if err != nil {
 		return err
 	}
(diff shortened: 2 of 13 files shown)


docs/_static/vcr/v1.yaml
```diff
@@ -53,6 +53,13 @@ paths:
         * 404 - Corresponding credential could not be found
         * 500 - An error occurred while processing the request
       operationId: "resolve"
+      parameters:
+        - name: at
```
dmeijboom, line 57: Shouldn't this be something similar to `resolveTime` or anything like that?

vcr/vcr.go
```diff
@@ -565,7 +584,8 @@ func (c *vcr) Get(conceptName string, subject string) (concept.Concept, error) {
 
 	q.AddClause(concept.Eq(concept.SubjectField, subject))
 
-	vcs, err := c.Search(q)
+	// finding a VC that backs a concept is a runtime call, eg: usage is now
+	vcs, err := c.Search(q, nil)
```
reinkrul, line 588: shouldn't `at` be `time.Now()` in that case?

reinkrul (changes requested):

vcr/api/v1/api.go
```diff
@@ -135,16 +136,24 @@ func (w *Wrapper) Create(ctx echo.Context) error {
...
 	idURI, err := ssi.ParseURI(id)
 	// return 400 for malformed input
 	if err != nil {
 		return core.InvalidInputError("failed to parse credential ID: %w", err)
 	}
 
+	// resolve time
+	at := time.Now()
+	if params.At != nil {
+		at, err = time.Parse(time.RFC3339, *params.At)
```
dmeijboom, line 149: I'm wondering, what's the use-case for using another resolve time (in the past for example)?

vcr/store_test.go
```diff
@@ -52,9 +52,9 @@ func TestVcr_StoreCredential(t *testing.T) {
 	t.Run("ok", func(t *testing.T) {
 		ctx := newMockContext(t)
 
-		ctx.tx.EXPECT().Subscribe(gomock.Any(), gomock.Any()).Times(2)
 		ctx.vcr.Configure(core.ServerConfig{Datadir: io.TestDirectory(t)})
 		ctx.keyResolver.EXPECT().ResolveSigningKey(gomock.Any(), nil).Return(pk, nil)
+		ctx.docResolver.EXPECT().Resolve(gomock.Any(), gomock.Any()).Return(nil, nil, nil)
```
dmeijboom, line 57: Why do we use `gomock.Any()` here? We do want to verify the mocked methods are called with the actual paremeters right? Unless it's something like a `Context` I wouldn't use `gomock.Any()`

vcr/test.go
```diff
@@ -60,16 +61,22 @@ type mockContext struct {
...
+	keyResolver := types.NewMockKeyResolver(ctrl)
+	docResolver := types.NewMockDocResolver(ctrl)
+	vcr := NewVCRInstance(crypto, docResolver, keyResolver, tx).(*vcr)
 	vcr.trustConfig = trust.NewConfig(path.Join(testDir, "trust.yaml"))
+	if err := vcr.Configure(core.ServerConfig{Datadir: testDir}); err != nil {
+		t.Fatal(err)
+	}
 	t.Cleanup(func() {
 		ctrl.Finish()
 	})
```
dmeijboom, line 82: Nitpick: this can be removed the test controller will automatically do this when running `gomock.NewController` in a unittest.

vcr/vcr.go
```diff
@@ -297,26 +293,61 @@ func (c *vcr) Issue(template vc.VerifiableCredential) (*vc.VerifiableCredential,
...
+	at := time.Now()
+	if validAt != nil {
+		at = *validAt
+	}
+	issuer, err := did.ParseDIDURL(credential.Issuer.String())
+	if err != nil {
+		return err
+	}
+
+	if credential.IssuanceDate.After(at) {
```
dmeijboom, line 342: Nitpick: should this be after or equal or only after?

dmeijboom (changes requested):
Some minor feedback

vcr/vcr.go
woutslakhorst: I'll adjust the description. At the lowest level, time.Now is used.

vcr/api/v1/api.go
woutslakhorst: mostly for when things go wrong. If an access token requests fails for invalid credentials, you would be able to redo the request with the same timestamp.

vcr/vcr.go
woutslakhorst: Given that `at` would mostly be `now` and `IssuanceDate` only has an accuracy on the seconds, equal will do nothing.

vcr/api/v1/api.go
dmeijboom: Check!

vcr/store_test.go
woutslakhorst: removed lots of gomock.Any

dmeijboom (approved):
LGTM

reinkrul (approved):

PR #370, opened 2021-07-12
Return DID.controller as object

### LINKED ISSUE #365: VDR API for creating a new Nuts DID returns result conflicting with the spec
When I run a POST call to generate a customer DID (https://nuts-node.readthedocs.io/en/latest/pages/getting-started/4-connecting-crm.html#create-and-store-a-customer-did) the returned data provides a `string` in the `controller` field whereas the specification tells me it is of type `[string]`. This makes generated Open API clients fail on these requests.

A bit of history; client generation didn't work when `OneOf` was being used earlier (for some languages at least).

woutslakhorst:
Fixes #365 

Returning document.controller as object is the quickest and simplest way to fix this. 
The API returns a DIDDocument according to the go-did library not according to spec.

### DIFF
--- a/docs/_static/vdr/v1.yaml
+++ b/docs/_static/vdr/v1.yaml
@@ -204,10 +204,8 @@ components:
           items:
             type: string
         controller:
-          description: List of DIDs that have control over the DID document
-          type: array
-          items:
-            type: string
+          description: Single DID (as string) or List of DIDs that have control over the DID document
+          type: object
         id:
           description: DID according to Nuts specification
           example: "did:nuts:1"
--- a/docs/_static/vdr/v1.yaml
+++ b/docs/_static/vdr/v1.yaml
@@ -205,7 +205,6 @@ components:
             type: string
         controller:
           description: Single DID (as string) or List of DIDs that have control over the DID document
-          type: object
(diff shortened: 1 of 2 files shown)


dmeijboom:
How does this work? A controller can be a string or a set of strings right? But it it's encoded as a JSON object this won't do right or am I missing something?

woutslakhorst:
> How does this work? A controller can be a string or a set of strings right? But it it's encoded as a JSON object this won't do right or am I missing something?

would `string` be a valid `object`?

woutslakhorst:
I removed the `type`, apparently typeless is the way to go

dmeijboom:
> > How does this work? A controller can be a string or a set of strings right? But it it's encoded as a JSON object this won't do right or am I missing something?
> 
> would `string` be a valid `object`?

The `type` maps directly to an JSON type in OpenAPI 3 right?

Reference in the [docs](https://swagger.io/docs/specification/data-models/data-types/#object):
```
 An object is a collection of property/value pairs. The properties keyword is used to define the object properties – you need to list the property names and specify a schema for each property. 
```

reinkrul (approved):

PR #371, opened 2021-07-13
E2E tests

dmeijboom:


woutslakhorst (approved):
LGTM

And it doesn't add any time to the total build!
Nitpicky: you could use the SHA1 as docker tag instead of image name

PR #373, opened 2021-07-13
added check to vdr.Resolve for deactivated controller

### LINKED ISSUE #372: Add active controller check to DID Document resolve
added line to RFC006:

> A DID Document MUST at least have one non-deactivated controller, otherwise it's to be regarded as deactivated.

requires an additional check, this would become a recursive check, so a max depth precaution is required.

woutslakhorst:
closes #372 

It now checks if the controller is not deactivated. If it is, it'll return an ErrDeactivated.
It still honours metadata.AllowDeactivated

### DIFF
--- a/vdr/doc/resolvers.go
+++ b/vdr/doc/resolvers.go
@@ -23,6 +23,7 @@ package doc
 
 import (
 	"crypto"
+	"errors"
 	"fmt"
 	"time"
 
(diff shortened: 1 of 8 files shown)


vdr/doc/resolvers.go
```diff
@@ -64,7 +101,13 @@ func (d Resolver) ResolveControllers(doc did.Document, metadata *types.ResolveMe
...
 	// resolve all unresolved docs
 	for _, ref := range refsToResolve {
-		node, _, err := d.Store.Resolve(ref, metadata)
+		node, _, err := d.resolve(ref, metadata, depth)
+		if errors.Is(err, types.ErrDeactivated) {
+			continue
+		}
+		if errors.Is(err, ErrNestedDocumentsTooDeep) {
+			return nil, err
+		}
```
dmeijboom, line 110: Maybe it would be nice if we can return something like `ErrNoActiveController` or something similar when all of the controllers are deactivated.

vdr/doc/resolvers.go
```diff
@@ -31,17 +32,53 @@ import (
 	"github.com/nuts-foundation/nuts-node/vdr/types"
 )
 
+// ErrNestedDocumentsTooDeep is returned when a DID Document contains a multiple services with the same type
+var ErrNestedDocumentsTooDeep = errors.New("DID Document controller structure has too many indirections")
+
+const maxControllerDepth = 5
```
dmeijboom, line 38: I couldn't find anything in the spec about this. Maybe we can open an issue on https://github.com/w3c/did-core to clarify how we should handle this? I mean `5` is probably fine but it's feels quite arbitrary

dmeijboom (changes requested):
Looks good! Two questions

vdr/doc/resolvers.go
woutslakhorst: Yes, 5 is arbitrary chosen based on the expected hierarchy of DID Documents. 3 is to be expected: offlineRoot->activeSaasKeys->customers. 2 has been added as a safe margin. This could also easily be changed to 10 or 100 since that would protect from recursion.

I don't think it's something that should go in the w3c specs since adding a hierarchy using controllers is something we specified. I did sent an email to the public-did-wg mailing list.

vdr/doc/resolvers.go
woutslakhorst: It could help in debugging since an `ErrNoActiveController` would generate some other logging.

dmeijboom (approved):
LGTM 👍

Issue #358, opened 2021-07-06
Update OAuth flow according to specs


woutslakhorst:
This is a tracking issue

- [x] Updated API for token request
- [x] Config for TLS certs (combine network and oauth TLS certs into 1 config)
- [ ] CRL check for certs
- [x] Client API for access token request
- [x] Updated e2e test
- [x] Updated specification (which also mentions JWT grant instead of JWT bearer token)
- [ ] Using AuthorizationCredentials (#94)
- [ ] etc

dmeijboom:
Nut sure why it was closed.. It's not finished so should be open.

dmeijboom:
@woutslakhorst 

> Config for TLS certs (combine network and oauth TLS certs into 1 config)
Isn't this already the case? The auth API is part of the API routes and will use the same TLS config right? Or do you mean the Auth API HTTP Client?

woutslakhorst:
For using the credentials in the oauth flow. A minimum would be to add the IDs of the authorization credentials to the access token request API call. They must also be returned by the introspection call.

woutslakhorst:
TLS part continued in #479

PR #375, opened 2021-07-15
Removed `Injectable.ConfigKey()` and parse flat config

dmeijboom:
With this PR configuration can be shared between engines. This is a requirement for #358 but it also makes sense as the configuration doesn't always directly map to an engine.


core/config.go
```diff
@@ -31,7 +31,7 @@ import (
 const defaultPrefix = "NUTS_"
 const defaultDelimiter = "."
 
-func loadConfigIntoStruct(flags *pflag.FlagSet, target interface{}, configMap *koanf.Koanf) error {
+func loadConfigIntoStruct(flags *pflag.FlagSet, target interface{}, flat bool, configMap *koanf.Koanf) error {
```
woutslakhorst, line 34: In this PR, only `false` is passed for `flat`.

woutslakhorst (changes requested):
only the `flat` argument requires some explaining.

core/config.go
dmeijboom: Yeah indeed, will fix

woutslakhorst (approved):

PR #376, opened 2021-07-16
Add support for TLS for Auth API HTTP client

dmeijboom:
Adds support for TLS in the new OAuth flow. It uses the same configuration as the network module (the `truststorefile` to be specific).


woutslakhorst:
- [ ] missing client certificate config for auth (can be stolen from network)
- [ ] missing check in configure when `strictmode` is enabled (TLS must be enabled and a client cert must be available)

When merged, update e2e test and enable TLS config.

woutslakhorst:
The missing coverage is covered by nuts-foundation/nuts-go-e2e-test#9

woutslakhorst (approved):

PR #378, opened 2021-07-22
Refactored concepts

woutslakhorst:
* Removed concept config by template and replaced it with a more verbose config
* Using latest leia lib for more search capabilities
* NutsAuthorizationCredential indices tweaked


auth/services/oauth/oauth.go
```diff
@@ -196,11 +196,11 @@ func (s *service) validateIssuer(context *validationContext) error {
 	if err != nil {
 		return fmt.Errorf(errInvalidIssuerFmt, err)
 	}
-	ok := false
-	if context.actorName, ok = orgConcept.GetValue(concept.OrganizationName).(string); !ok {
+
+	if context.actorName, err = orgConcept.GetString(concept.OrganizationName); err != nil {
 		return fmt.Errorf(errInvalidIssuerFmt, errors.New("actor has invalid organization VC"))
```
reinkrul, line 201: `err` is ignored?

auth/services/oauth/oauth.go
```diff
@@ -196,11 +196,11 @@ func (s *service) validateIssuer(context *validationContext) error {
...
 	}
-	if context.actorCity, ok = orgConcept.GetValue(concept.OrganizationCity).(string); !ok {
+	if context.actorCity, err = orgConcept.GetString(concept.OrganizationCity); err != nil {
 		return fmt.Errorf(errInvalidIssuerFmt, errors.New("actor has invalid organization VC"))
```
reinkrul, line 204: here as well

docs/pages/technology/vc-concepts.rst
```diff
@@ -13,84 +13,87 @@ Multiple credentials can give a DID its name, the concept mappings makes sure on
...
+Configuration
+*************
+
+Credentials are configured to have indices and a template for transformation.
+
+The organization credential configuration looks like this:
+
+.. code-block:: yaml
+
+    concept: organization
```
reinkrul, line 25: I like this, it's much clearer than the obscure DSL we had before 👍

docs/pages/technology/vc-concepts.rst
```diff
@@ -13,84 +13,87 @@ Multiple credentials can give a DID its name, the concept mappings makes sure on
...
+An index part can also contain a transformer and/or a tokenizer. A tokenizer will split a value into multiple values to be indexed.
+For example: the whitespace tokenizer will split a sentence into a list of words to be indexed.
+The name *healthcare organization the Nutty professor* will be split and can be found by just searching for *healthcare*, *organization*, *the*, *Nutty* or *professor*.
+Possible values for tokenizer: `whitespace`.
 
-.. code-block:: text
+A transformer transforms a value before the value is used as an index key. A transformer will also transform search parameters that use that index.
+For example: the lowercase transformer will transform all indexed values and search params to lowercase.
+The word *Nutty* in the previous example could then be found with *nutty*, *NUTTY* or any other combination of upper- and lowercase letters.
+Possible values for transformer: `lowerCase`, `cologne`.
```
reinkrul, line 87: is `lowerCase` case sensitive in itself? Maybe `lowercase` is less error prone in that case.

vcr/assets/NutsAuthorizationCredential.config.yaml
```diff
@@ -0,0 +1,18 @@
+concept: authorization
```
reinkrul, line 1: no template?

vcr/assets/NutsAuthorizationCredential.config.yaml
```diff
@@ -0,0 +1,18 @@
...
+credentialType: NutsAuthorizationCredential
+indices:
+  - name: index_id
```
reinkrul, line 4: do they need to be prefixed with `index_`, what if you don't?

Maybe a bit odd when you have to prefix every index name with `index_` (since in that case, the software should probably do it?)

vcr/assets/NutsAuthorizationCredential.config.yaml
```diff
@@ -0,0 +1,18 @@
...
+    parts:
+      - path: id
+  - name: index_issuer
+    parts:
+      - path: issuer
+  - name: index_holder
+    parts:
+      - path: credentialSubject.id
+      - path: credentialSubject.purposeOfUse
+      - path: issuanceDate
```
reinkrul, line 14: why is `issuanceDate` part of this index?

Second though: unless we'll be searching for an exact value, using the normal index type for dates isn't really the best solution? (probably better to convert it to a number and index that, for range queries and such?)

vcr/assets/NutsOrganizationCredential.config.yaml
```diff
@@ -0,0 +1,34 @@
+concept: organization
```
reinkrul, line 1: should we add the license comment block here as well? (same for `NutsAuthorizationCredential.config.yaml`)

vcr/concept/config.go
```diff
@@ -0,0 +1,91 @@
...
+	// CredentialType defines the type of the credential. 'VerifiableCredential' is omitted.
+	CredentialType string `yaml:"credentialType"`
+	// Indices contains a set of Index values
+	Indices []Index `yaml:"indices"`
+	// Template is the string template for outputting a credential to a common format
+	// Each <<JSONPath>> value is substituted with the outcome of the JSONPath query
+	Template *string `yaml:"template"`
+}
+
+var templateStringMatcher, _ = regexp.Compile(`<<([a-zA-Z\\.]+)>>`)
```
reinkrul, line 44: `_` and `-` not allowed? `_` is allowed in JSON field names (think `-` isn't)

vcr/concept/config.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+
+	// if no template is present, we'll return the credential
+	if c.Template == nil {
+		template = vcString
+	} else {
+		template = *c.Template
+
+		// find all << refs >>
+		for _, match := range templateStringMatcher.FindAllStringSubmatch(template, -1) {
```
reinkrul, line 62: isn't this the same for all credentials of this type? Shouldn't this regexing be done once, when the config is created instead of for every VC being transformed (for performance reasons)?

vcr/concept/config.go
reinkrul, line 44: use `MustCompile`, which panics when it fails (which is good, in this case)

vcr/concept/query.go
```diff
@@ -119,7 +114,7 @@ func (e eq) Match() string {
 	return e.value
 }
 
-// EqType is the identifier for an equals clause
+// PrefixType is the identifier for an prefix clause
```
reinkrul, line 117: a prefix clause

vcr/concept/registry.go
```diff
@@ -57,59 +57,34 @@ const (
...
-		current = append(current, conceptTemplate)
-		r.conceptTemplates[c] = current
-	}
-
-	// add to map of specific VC type to template
-	v, ok := conceptTemplate.fixedValues[TypeField]
-	if !ok || v == "" {
+// Add adds a new template to a concept and parses it.
+func (r *registry) Add(config Config) error {
+	// check for type
```
reinkrul, line 82: good thing this comment is here, otherwise I wouldn't have understood the code ;-)

vcr/concept/types.go
```diff
@@ -40,60 +42,29 @@ var ErrNoValue = errors.New("no value for given path")
...
-		}
-		if sub, ok := current[p]; ok {
-			ok2 := false
-			if current, ok2 = sub.(Concept); ok2 {
-				continue
-			}
-		}
-		break
-	}
+// SubjectField defines the concept JSONPath to a VC humanSubject
```
reinkrul, line 54: humanSubject?

vcr/interface.go
```diff
@@ -60,7 +60,7 @@ var revocationDocumentType = "application/vc+json;type=revocation"
 
 // ConceptFinder can resolve VC backed concepts for a DID.
 type ConceptFinder interface {
-	// Get returns the requested concept for the subject or ErrNotFound
+	// Get returns the requested concept as JSON string for the subject or ErrNotFound
```
reinkrul, line 63: it returns it as concept, which is a `map[string]interface{}`? (not as JSON string)

vcr/vcr.go
```diff
@@ -364,22 +360,23 @@ func (c *vcr) find(ID ssi.URI) (vc.VerifiableCredential, error) {
...
-	docs, err := gIndex.Find(q)
-	if err != nil {
-		return credential, err
-	}
-
-	if len(docs) != 1 {
-		return credential, ErrNotFound
-	}
+	for _, t := range c.registry.Concepts() {
+		docs, err := c.store.Collection(t.CredentialType).Find(q)
```
reinkrul, line 364: this concept might not have a collection of the given credential type?

reinkrul (changes requested):

auth/services/oauth/oauth.go
woutslakhorst: Added error output, double ErrorF ftw

docs/pages/technology/vc-concepts.rst
woutslakhorst: did both, suggested to use `lowercase` and comparing after applying `strings.ToLower`

vcr/assets/NutsAuthorizationCredential.config.yaml
woutslakhorst: Nope, that means the raw credential is returned.

vcr/assets/NutsAuthorizationCredential.config.yaml
woutslakhorst: It translates to a sub-bucket on the collection. Only using the field as name would be weird as well. There are no requirements on the name.

vcr/assets/NutsAuthorizationCredential.config.yaml
woutslakhorst: It's currently not possible to use optional fields for an index (with nil and such, it would create large index entries). An index on only holder and purpose would not be specific enough. The extra index on date would help in finding eTransfer authorization credentials from the last 4 weeks or so. the issuanceDate field is specced as RFC3339 date which when translated from string to bytes is comparable.

vcr/assets/NutsOrganizationCredential.config.yaml
woutslakhorst: no intellectual property to protect is there?

vcr/concept/config.go
woutslakhorst: indeed, `_` should be there.

vcr/concept/config.go
woutslakhorst: The output is specific per credential. The values between << and >> have to be replaced with the JSONPath result. It's not something you can do genericly

vcr/concept/types.go
woutslakhorst: find replace ftw

vcr/interface.go
woutslakhorst: ahhh yes, I wanted it to be a string but it didn't work out.

vcr/vcr.go
woutslakhorst: Then it would be created and it would return no results.

vcr/assets/NutsAuthorizationCredential.config.yaml
reinkrul: Doing a range query over an RFC3339 string date is a pain though, but I guess good enough for now. Future feature.

reinkrul (approved):

PR #379, opened 2021-07-22
Network: validate DAG TXs on boot time the same order as run-time

### LINKED ISSUE #366: Network: validate/process order at boot-time differs from runtime order
Because runtime transactions are validated before they're processed, at boot-time the DAG is loaded and then validated.

This causes rules to be executed differently, causing issues (TX is valid at run-time, but not at boot-time).

Boot-time validation order needs to be the same as run-time.

reinkrul:
FIxes #366


reinkrul:
Review round 1

reinkrul:
Already fixed because `dag.FindBetween()` was updated to return the TXs in DAG walking order.

PR #385, opened 2021-07-27
set env.SHA so e2e tests work for both branches and PRs

### LINKED ISSUE #382: Broken e2e tests for master
To serve the PRs, the e2e test config uses `github.event.pull_request.head.sha` which is empty for master. For master it must  be `github.sha`.

woutslakhorst:
Fixes #382

we'll now if it works for master after merging ;)

### DIFF
--- a/.github/workflows/e2e-tests.yaml
+++ b/.github/workflows/e2e-tests.yaml
@@ -34,13 +34,11 @@ jobs:
 
       - name: Setting git SHA to PR head
         if: ${{ github.event_name == 'pull_request' }}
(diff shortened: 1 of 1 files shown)


reinkrul (approved):
👏

.github/workflows/e2e-tests.yaml
```diff
@@ -5,6 +5,7 @@ on:
   push:
     branches:
       - master
+      - iss382_fix_master_e2e_tests
```
dmeijboom: This has to be removed right?

PR #389, opened 2021-07-31
Add codeQL analyser

stevenvegt:


.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,71 @@
...
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ master ]
+  schedule:
+    - cron: '21 10 * * 2'
```
reinkrul, line 21: > "At 10:21 on Tuesday."

Chosen on purpose or did we just need whatever moment each week to automatically run it?

reinkrul (approved):

.github/workflows/codeql-analysis.yml
stevenvegt: Whatever moment indeed. This was auto generated by github, I suspect the try to load balance these cron job by randomizing trigger times.

PR #391, opened 2021-08-03
Allow VerifiableCredentials in the oauth2 flow

dmeijboom:
Allow VerifiableCredentials in the oauth2 flow. If present they will be validated.


docs/_static/auth/v1.yaml
```diff
@@ -279,6 +279,67 @@ paths:
 
 components:
   schemas:
+    VerifiableCredential:
+      type: object
+      description: A credential according to the W3C and Nuts specs.
+      required:
+        - "@context"
```
stevenvegt, line 286: Many of these fields can also contain a string when there is only 1 value.

stevenvegt (commented):
If a token does not have credentials, it is valid? Feels like we might need a higher level DIDMan method which actually checks if the correct credentials are present. 🤔

docs/_static/auth/v1.yaml
dmeijboom: Note that the `VerifiableCredential` type is a 1-on-1 copy from the `vcr` OpenAPI spec. Currently, this type is unused in Go as it's aliased to `vc.VerifiableCredential`. For now this feels like an "OK" solution.

dmeijboom:
> If a token does not have credentials, it is valid? Feels like we might need a higher level DIDMan method which actually checks if the correct credentials are present. 🤔

It's valid without credentials indeed (they are optional in the spec).

auth/services/oauth/oauth.go
```diff
@@ -273,6 +274,12 @@ func (s *service) CreateJwtGrant(request services.CreateJwtGrantRequest) (*servi
 		return nil, err
 	}
 
+	for _, verifiableCredential := range request.Credentials {
```
reinkrul, line 277: I think these validation steps are missing (chapter 5.2.1.7 of RFC003):

> The credential issuer equals the sub field of the JWT.
> The credential credentialSubject.id equals the iss field of the JWT.

auth/services/oauth/oauth.go
```diff
@@ -273,6 +274,12 @@ func (s *service) CreateJwtGrant(request services.CreateJwtGrantRequest) (*servi
...
+		if err := credential.Validate(verifiableCredential); err != nil {
+			return nil, fmt.Errorf("invalid VerifiableCredential: %w", err)
+		}
+	}
+
 	endpointID, _, err := services.ResolveCompoundServiceURL(s.docResolver, *custodian, request.Service, services.OAuthEndpointType, nil)
```
reinkrul, line 283: `NutsAuthorizationCredential` specific part is not implemented yet?

> The name and city MUST exactly (case sensitive) match with a valid Verifiable Credential issued to the actor. This must be a NutsOrganizationCredential.

(RFC003 5.2.1.5 )

reinkrul (changes requested):

auth/services/oauth/oauth.go
dmeijboom: Note that this PR is based on: https://github.com/nuts-foundation/nuts-specification/pull/119 where any verifiable credential can be used. However, we could validate known types (such as `NutsAuthorizationCredential`)?

auth/services/oauth/oauth.go
dmeijboom: See: https://github.com/nuts-foundation/nuts-specification/pull/119

reinkrul: Ah right I missed that.

auth/services/oauth/oauth.go
reinkrul: Well it must be validated _somewhere_, maybe we can have specialised validators per credential type?

dmeijboom: I think we already have something like that. I'll take a look!

reinkrul (approved):

PR #405, opened 2021-08-16
Fix oauth flow without identity

dmeijboom:
Currently, the oath flow doesn't work without user identity because of two bugs:

1. In the OpenAPI spec the identity is a string but in our code it's a string pointer. Before this fix the empty string would be converted into an empty string pointer instead of `nil`
2. No identity means no contract validation (if I understand it correctly) this wasn't taken into account when creating the access token

### DIFF
--- a/auth/services/oauth/oauth.go
+++ b/auth/services/oauth/oauth.go
@@ -147,9 +147,6 @@ func (s *service) CreateAccessToken(request services.CreateAccessTokenRequest) (
 	return &services.AccessTokenResult{AccessToken: accessToken}, nil
 }
 
-// ErrLegalEntityNotProvided indicates that the legalEntity is missing
-var ErrLegalEntityNotProvided = errors.New("legalEntity not provided")
-
 // checks if the name from the login contract matches with the registered name of the issuer.
 func (s *service) validateActor(context *validationContext) error {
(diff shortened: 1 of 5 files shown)


auth/services/oauth/oauth.go
```diff
@@ -112,21 +112,26 @@ func (s *service) CreateAccessToken(request services.CreateAccessTokenRequest) (
...
 		}
-	}
-	if context.contractVerificationResult.Validity == contract.Invalid {
-		return nil, errors.New("identity validation failed")
-	}
-	// checks if the name from the login contract matches with the registered name of the issuer.
-	if err := s.validateActor(&context); err != nil {
-		return nil, err
+
+		if context.contractVerificationResult.Validity == contract.Invalid {
```
reinkrul, line 127: I know this is existing code, but wouldn't it be safer to reverse this check, to fail if `Validity != contract.Valid`? Then it's also deemed invalid when a bug (or attack) occurs (e.g. `""`)

reinkrul (changes requested):
The code change itself looks fine. Reading the Bolt, I'm not sure it's stated explicitly enough what the consequences are of having/not having `usi` (user identity) in the token (https://nuts-foundation.gitbook.io/drafts/rfc/rfc003-oauth2-authorization#5-2-validation). Or do we have it explained more thoroughly somewhere else?

auth/services/oauth/oauth.go
dmeijboom: I was actually thinking about something like this as well. You're right, will fix

reinkrul (approved):

PR #414, opened 2021-08-19
DIDMan: make service reference checks compliant with RFC006

reinkrul:
In preparation of adding a method to resolve a (nested) refenced service from a compound endpoint.

Specified by https://github.com/nuts-foundation/nuts-specification/pull/122

Fixes:

- Compound service can either contain URLs, references or both
- Reference must have `/serviceEndpoint` as path
- Made `type` query param checking stricter
- Max 5 levels of ref-ing.


dmeijboom (changes requested):
Looks good but why the `wrapper` package? You can simply use `fmt.Errorf` with the `%w` modifier. Then `errors.Is` also works on the wrapped error. Or am I missing something?

reinkrul:
> Looks good but why the `wrapper` package? You can simply use `fmt.Errorf` with the `%w` modifier. Then `errors.Is` also works on the wrapped error. Or am I missing something?

You can only wrap 1 error with `fmt.Errorf`'s `%w` directive :-( and since we generally like we define and export our errors (so you can check 'm), you'd have to choose which error can be checked (the 'outer' or 'inner' one), while `errors.Is` should work on both just fine (imo).

We have some libraries (transitive) that already do sort of the same, but not exactly what we need. I figured it's an extremely simple type to write.

dmeijboom:
> > Looks good but why the `wrapper` package? You can simply use `fmt.Errorf` with the `%w` modifier. Then `errors.Is` also works on the wrapped error. Or am I missing something?
> 
> You can only wrap 1 error with `fmt.Errorf`'s `%w` directive :-( and since we generally like we define and export our errors (so you can check 'm), you'd have to choose which error can be checked (the 'outer' or 'inner' one), while `errors.Is` should work on both just fine (imo).
> 
> We have some libraries (transitive) that already do sort of the same, but not exactly what we need. I figured it's an extremely simple type to write.

Seems to work just fine, see this example I made: https://play.golang.org/p/2pt1ybE4qK2
Or is this not what you mean?

reinkrul:
I meant this: https://play.golang.org/p/DLlqC_-jXlp

dmeijboom:
> I meant this: https://play.golang.org/p/hecNruBYL_Q

Ah right, well it depends. If you want to add context to the error or there is a something that can go wrong but can have many reasons why I would create a custom error type:

```go
type GeolocationError struct {
    Cause error
}

func (err *GeolocationError) Error() string {
    return fmt.Errorf("geolocation error: %w", err.Cause)
}
```

It can also contain context-specific information in this case. Otherwise I would just use `fmt.Errorf` if I don't care about the specific error.

reinkrul:
Well we usually test our returned errors (`outer`, not generally not the `inner` errors) and the only way to do that with errors that wrap other errors is to either:

1. assert the exact error message (which includes the inner error message). Which is an anti-pattern imo, since it forces you to test things which are not relevant for the unit test. And it might break when upgrading libraries (when error messages change)
2. assert the error using `Contains` to assert on the outer error message, which is even uglier to write
3. define a type for that specific error which implements `Is(err)`, which is incovenient

So although not ideal (defining our own wrapped error type) I think it's still the most elegant solution. Since an error that wraps a cause (error), `Is(..)` should really work on both, imo.

@woutslakhorst ?

woutslakhorst:
If you wrap an error and pass it along, you're creating an abstraction which usually means you won't check what the underlying error is. It looks like you're trying to add inheritance. The introduced `wrap` function is only used once here, so there's no use in making it generic.

Are you planning to use `errors.Is()` on the return error of `validateCompoundServiceEndpoint`? If not, use `fmt.Errorf`, if so, define a custom error and only check for that type. Checking for underlying types of errors will create some weird code. It's only used in `validateCompoundServiceEndpoint`. The function only returns `error` meaning that you shouldn't do any branch logic on the outcome, only pass the error along and add some info about the location it occured in. It looks like `fmt.Errorf` is a better fit here.

reinkrul:
Although we use it in some places, I agree with it being an anti-pattern to check for specific underlying causes. I guess we can leave out the `error.Is()`-ing of underlying causes.

At the same time we export (and document) the exact errors an API function returns ("it returns ErrSomeError when XYZ"), so we should return these actual (assertable) errors. Because when you just use the error message of the error you export to wrap an error message, it isn't actually the same error any more:

```
var EOF = errors.New("IO error")
var cause = /* some error, e.g. closed pipe */
return Sprintf("%s: %w", EOF.String(), cause)
```

In that case we shouldn't export our errors at all, because callers can't actually use it for anything (since only its message will/might be used, while creating a wrapped error).

I'll create a specific type for it.

reinkrul:
@woutslakhorst @dmeijboom  d'accord?

woutslakhorst (approved):

dmeijboom (approved):
LGTM 👍

Issue #340, opened 2021-06-15
DIDMan: Resolve endpoints or compound services.


stevenvegt:
Allow an application to resolve endpoints for a certain DID.
Use the spec at https://github.com/nuts-foundation/nuts-specification/issues/78.

Create api endpoint in DIDMan to resolve endpoints for a did by type. This should automatically follow refs.
It returns a `map[name]URI`.

`GET nodeaddress:port/internal/didman/v1/did/{did}/serviceEndpoint?type={type}`

Example:

`GET nodeaddress:port/internal/didman/v1/did/did:nuts:123/serviceEndpoint?type=eOverdracht`

```json
{
  "fhir": "ref:did:nuts:555/serviceEndpoint?type=fhir-production",
  "auth": "ref:did:nuts:555/serviceEndpoint?type=oauth-production",
}
```

Gets internally resolved to:
```json
{
  "fhir": "http://api-prod.example.com/fhir",
  "auth": "https://auth.example.com/token"
}
```

reinkrul:
@stevenvegt isn't this fixed now?

stevenvegt:
It was not broken. This is a proposal for endpoint resolving. Will be build along side the demo ehr needs.

reinkrul:
I meant "solved", of course :-)

woutslakhorst:
as discussed: `didman`

woutslakhorst:
Currently `didman` has only a single API for 'getting' endpoints: `getCompoundServices` which returns a list of compound services for a DID. It should specify if these use resolved URLs or not.

PR #415, opened 2021-08-19
DIDMan: introduce getCompoundServiceEndpoint to resolve an URL endpoint from a compound service

reinkrul:
Fixes https://github.com/nuts-foundation/nuts-node/issues/340

Based on https://github.com/nuts-foundation/nuts-node/pull/414 (needs to be merged first)

After this has been merged `auth` should use this new function to resolve the OAuth endpoints.


docs/_static/didman/v1.yaml
```diff
@@ -295,12 +348,12 @@ components:
...
         serviceEndpoint:
-          description: A map containing service references.
-          example: {'auth': 'did:nuts:1312321?type=auth'}
+          description: A map containing service references and/or endpoints.
+          example: { 'auth': 'did:nuts:1312321?type=auth' }
           type: object
     CompoundService:
       type: object
-      description: A creation request for a compound service that references endpoints.
+      description: A creation request for a compound service that references endpoints and/or endpoints.
```
woutslakhorst, line 356:
```suggestion
      description: A creation request for a compound service with endpoints and/or references to endpoints.
```

docs/_static/didman/v1.yaml
```diff
@@ -295,12 +348,12 @@ components:
           description: type of the endpoint. May be freely choosen.
           type: string
         serviceEndpoint:
-          description: A map containing service references.
-          example: {'auth': 'did:nuts:1312321?type=auth'}
+          description: A map containing service references and/or endpoints.
+          example: { 'auth': 'did:nuts:1312321?type=auth' }
```
woutslakhorst, line 352: example should contain `/serviceEndpoint`

didman/didman.go
```diff
@@ -109,26 +119,61 @@ func (d *didman) GetCompoundServices(id did.DID) ([]did.Service, error) {
...
+			}
+			if resolveReferences {
+				endpointURI, err := ssi.ParseURI(endpointStr)
+				if err != nil {
+					// Not sure when this could ever happen
+					return "", err
+				}
+				resolvedEndpoint, err := d.resolveAbsoluteURLEndpoint(*endpointURI, 0, maxServiceReferenceDepth, documentsCache)
+				return resolvedEndpoint.String(), err
+			} else {
```
woutslakhorst, line 169: else not needed

docs/_static/didman/v1.yaml
```diff
@@ -188,6 +188,59 @@ paths:
...
+    get:
+      operationId: getCompoundServiceEndpoint
+      summary: |
+        Retrieves the endpoint with the specified endpointType from the specified compound service.
+        It returns the serviceEndpoint of the specified service (which must be an absolute URL endpoint).
+
+        error responses:
+        * 400 - incorrect input (e.g. the given service type isn't a compound service)
+        * 404 - unknown DID, compound service or endpoint
+        * 406 - service references are nested too deep or reference is invalid in other ways
```
woutslakhorst, line 224: maybe a 400 is better. 406 is returned based on the accept headers. The underlying error is only returned when `resolve == true`, when it is `false` it'll succeed. Hence bad query input??

woutslakhorst (changes requested):
rebase after other PR merge.

docs/_static/didman/v1.yaml
reinkrul: It can also mean the service isn't correct (nested too deep), but which is done by the other party. 400 bad request implies (for me) that you need to alter your input for your request to succeed. But it's the data (entered by the remote party) that corrupt, so you won't be able to make the call succeed by altering your input.

Hence the status code != 400?

reinkrul: There's really nothing that naturally fits, and it also doesn't matter too much. As long as the returned code isn't confusing. I associate status 500 with something that's gone terribly bad in the internals of the server, which isn't actually the case either.

Maybe 422 Unprocessable Entity?

> The HyperText Transfer Protocol (HTTP) 422 Unprocessable Entity response status code indicates that the server understands the content type of the request entity, and the syntax of the request entity is correct, but it was unable to process the contained instructions.

woutslakhorst: one of the most vague codes and therefore a good match? ;)

PR #431, opened 2021-08-25
added DID sourceTX to VC issue TX.prevs

### LINKED ISSUE #430: Add DID tx reference to prevs for VC transaction
see nuts-foundation/nuts-specification#117

woutslakhorst:
depends on nuts-foundation/nuts-specification#117

closes #430 

this is the first half of the required functionality. First we add the prevs, in the next iteration we're going to validate the prevs and use those TXs to retrieve the correct DID doc.


dmeijboom:
If I understand correctly `meta.SourceTransactions` refers to the transactions that created the DID document while `prevs` refers to the previous transactions in the graph. In the `CreateTransaction` method the `meta.SourceTransactions` are added to the `prevs` but I don't understand the relationship between the two. Can you explain why they are added to the `prevs`?

PR #436, opened 2021-08-27
Validate auth creds for access token creation

### LINKED ISSUE #94: Auth: Check for consent during OAuth flow
Currently not implemented, because we have no consents in the new system yet.

### LINKED ISSUE #437: In the spec: purposeOfUse, in the JWT-grant, allows for a list. In the impl it's a single value.
Also in the APIs it's called `service`

woutslakhorst:
Validate authorization credentials within the JWT-grant according to RFC003 5.2.1.7

Closes #94 
Closes #437

### DIFF
--- a/auth/services/oauth/oauth.go
+++ b/auth/services/oauth/oauth.go
@@ -123,7 +123,7 @@ func (c validationContext) verifiableCredentials() ([]vc2.VerifiableCredential, 
 	for i, rawVC := range rawVCs {
 		vc := vc2.VerifiableCredential{}
 		if err := json.Unmarshal(rawVC, &vc); err != nil {
-			return vcs[:0], errors.New("cannot unmarshal authorization credential JSON")
+			return vcs[:0], fmt.Errorf("cannot unmarshal authorization credential: %w", err)
 		}
 		vcs[i] = vc
 	}
@@ -315,8 +315,7 @@ func (s *service) validateSubject(context *validationContext) error {
 	return nil
 }
 
-// validate the legal base, according to RFC003 §5.2.1.7 if sid is present
-// use consent store
+// validate the authorization credentials according to §5.2.1.7
 func (s *service) validateAuthorizationCredentials(context validationContext) error {
 	// filter on authorization credentials
 	vcs, err := context.verifiableCredentials()
@@ -350,7 +349,7 @@ func (s *service) validateAuthorizationCredentials(context validationContext) er
 
 		//The credential issuer equals the sub field of the JWT.
 		if authCred.Issuer.String() != sub {
-			return fmt.Errorf("issuer of authorization credential with ID: %s does not match jwt.sub: %s", authCred.Issuer.String(), sub)
+			return fmt.Errorf("issuer %s of authorization credential with ID: %s does not match jwt.sub: %s", authCred.Issuer.String(), authCred.ID.String(), sub)
 		}
 
 		//The credential credentialSubject.id equals the iss field of the JWT.
@@ -361,7 +360,7 @@ func (s *service) validateAuthorizationCredentials(context validationContext) er
 		// should be only 1 credentialSubject, but we do the range just to make sure and to avoid [0] specific code.
 		for _, authCredSubject := range authCredSubjects {
 			if authCredSubject.ID != iss {
-				return fmt.Errorf("credentialSubject.ID of authorization credential with ID: %s does not match jwt.iss: %s", authCred.Issuer.String(), iss)
+				return fmt.Errorf("credentialSubject.ID %s of authorization credential with ID: %s does not match jwt.iss: %s", authCredSubject.ID, authCred.ID.String(), iss)
 			}
 		}
 	}
--- a/auth/services/oauth/oauth_test.go
+++ b/auth/services/oauth/oauth_test.go
@@ -455,7 +455,7 @@ func TestService_validateAuthorizationCredentials(t *testing.T) {
 		if !assert.Error(t, err) {
 			return
 		}
-		assert.EqualError(t, err, "invalid jwt.vcs: cannot unmarshal authorization credential JSON")
+		assert.EqualError(t, err, "invalid jwt.vcs: cannot unmarshal authorization credential: json: cannot unmarshal string into Go value of type map[string]interface {}")
 	})
 
 	t.Run("error - jwt.iss <> credentialSubject.ID mismatch", func(t *testing.T) {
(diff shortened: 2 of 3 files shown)


auth/services/oauth/oauth.go
```diff
@@ -62,20 +69,93 @@ type service struct {
...
+		return ""
+	}
+	switch v := val.(type) {
+	case string:
+		return v
+	case []string:
+		if len(v) > 0 {
+			return v[0]
+		}
+	}
```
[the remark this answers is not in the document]
woutslakhorst: due to spec change, this has been removed.

auth/services/oauth/oauth.go
```diff
@@ -239,6 +318,54 @@ func (s *service) validateSubject(context *validationContext) error {
...
+		return nil
+	}
+
+	iat := context.jwtBearerToken.IssuedAt()
+	iss := context.jwtBearerToken.Issuer()
+	sub := context.jwtBearerToken.Subject()
+
+	for _, authCred := range vcs {
+		// first check if the VC is valid
+		if err := s.vcValidator.Validate(authCred, true, &iat); err != nil {
```
reinkrul, line 347: In the eOverdracht Leveranciersspecificatie "Retrieve eTransfer" and "Complete Transfer" sections of the sequence diagram, where the receiving care organization retrieves the FHIR compositions and updates the FHIR task it must send a Nuts Auth VC for authorisation, as per spec. It also provides the access token (containing the VC?) which appears to be issued by the sending care organization at step 19. In that case, the issuer of the JWT and VC are the same, so do these checks below apply to those steps as well? (`jwt.issuer == vc.subject.id && jwt.subject == vc.issuer`)

https://nuts-foundation.gitbook.io/bolts/eoverdracht/leveranciersspecificatie

auth/services/oauth/oauth.go
```diff
@@ -239,6 +318,54 @@ func (s *service) validateSubject(context *validationContext) error {
...
+			return fmt.Errorf(errInvalidVCClaim, err)
+		}
+
+		//The credential issuer equals the sub field of the JWT.
+		if authCred.Issuer.String() != sub {
+			return fmt.Errorf("issuer of authorization credential with ID: %s does not match jwt.sub: %s", authCred.Issuer.String(), sub)
```
reinkrul, line 353: "with ID" implies VC ID, but the issuer is printed

auth/services/oauth/oauth.go
```diff
@@ -239,6 +318,54 @@ func (s *service) validateSubject(context *validationContext) error {
...
+		}
+
+		//The credential credentialSubject.id equals the iss field of the JWT.
+		authCredSubjects := make([]credential.NutsAuthorizationCredentialSubject, 0)
+		if err := authCred.UnmarshalCredentialSubject(&authCredSubjects); err != nil {
+			return fmt.Errorf(errInvalidVCClaim, err)
+		}
+		// should be only 1 credentialSubject, but we do the range just to make sure and to avoid [0] specific code.
+		for _, authCredSubject := range authCredSubjects {
```
reinkrul, line 362: can there be 0 subjects (in some odd case/attacker/bug), making this check never run?

auth/services/oauth/oauth.go
```diff
@@ -239,6 +318,54 @@ func (s *service) validateSubject(context *validationContext) error {
...
+			if authCredSubject.ID != iss {
+				return fmt.Errorf("credentialSubject.ID of authorization credential with ID: %s does not match jwt.iss: %s", authCred.Issuer.String(), iss)
```
reinkrul, line 364: incorrect log (`credentialSubject.ID` v.s. `authCred.Issuer`)

reinkrul (changes requested):

auth/services/oauth/oauth.go
```diff
@@ -239,6 +318,54 @@ func (s *service) validateSubject(context *validationContext) error {
 // validate the legal base, according to RFC003 §5.2.1.7 if sid is present
 // use consent store
```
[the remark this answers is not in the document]
woutslakhorst: completely incorrect yes ;)

auth/services/oauth/oauth.go
woutslakhorst: No the issuer of the JWT grant token is the actor.

auth/services/oauth/oauth.go
woutslakhorst: should have been DID

auth/services/oauth/oauth.go
woutslakhorst: could be the case, then the actor will be denied access through the access policy at the resource server level.

auth/services/oauth/oauth.go
woutslakhorst: indeed.

auth/services/oauth/oauth.go
```diff
@@ -62,20 +68,77 @@ type service struct {
...
+
+	// If no credentials then OK
+	if !ok || claim == nil {
+		return vcs, nil
+	}
+
+	// vcs should contain a slice of map[string]interface{}
+	vcMaps, ok := claim.([]interface{})
+	if !ok {
+		return vcs, errors.New("field does not contain an array of credentials")
```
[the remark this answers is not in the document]
woutslakhorst: it prints: `invalid jwt.vcs: field does not contain an array of...` due to calling func

auth/services/oauth/oauth.go
```diff
@@ -62,20 +68,77 @@ type service struct {
...
+		rawVC, _ := json.Marshal(vcMap)
+		rawVCs[i] = rawVC
+	}
+
+	// filter on authorization credentials
+	vcs = make([]vc2.VerifiableCredential, len(rawVCs))
+	for i, rawVC := range rawVCs {
+		vc := vc2.VerifiableCredential{}
+		if err := json.Unmarshal(rawVC, &vc); err != nil {
+			return vcs[:0], errors.New("cannot unmarshal authorization credential JSON")
```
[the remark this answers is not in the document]
woutslakhorst: the error is ugly though: 

```
cannot unmarshal string into Go value of type map[string]interface {}
```

auth/services/oauth/oauth.go
reinkrul: because the access policy is derived from the Verifiable Credential, not from the token?

reinkrul (approved):

PR #440, opened 2021-08-27
DIDMan: GetCompoundServiceEndpoint should resolve reference to compound service as well

reinkrul:
Before it only resolved references from the endpoints specified by the compound service, not a reference to the compound service itself (the latter is how Nuts Registry Admin Demo currently registers endpoints).


woutslakhorst (commented):
e2e tests are broken, so we're missing a unit tests somewhere as well?

woutslakhorst (approved):

PR #444, opened 2021-08-31
DIDMan: fixed following ref to compound service endpoint with unmarshal instead of cast

reinkrul:
Missed this test before because the unit tests create a DID Document using the Go structs, instead of unmarshaling one.  I added a test to exactely mimic the DID Document structure as registry-admin-demo does it.

Before I cast the `serviceEndpoint` to a string, which is plainly wrong because many DID elements are always an array under the hood, so I had to use the unmarshal endpoint func instead.


woutslakhorst (approved):
the blobs of JSON could have had some formatting, easier for the eyes.

Issue #416, opened 2021-08-19
Network: split TransactionList in pages


reinkrul:
So we don't exceed the 512k max. message size, crippling the network (our dev network currently has 410 TXs, probably nearing the current limitation).

PR #448, opened 2021-09-03
Network: split TransactionList response message in pages 

reinkrul:
When number of transactions exceeds the maximum Protobuf message size.

Fixes https://github.com/nuts-foundation/nuts-node/issues/416

Also altered the behaviour of `dag.FindBetween`, it now returns the transactions in DAG walking order, to prevent out-of-order transactions (signing time) to be in the from page, preventing the receiving peer from processing the transactions (because it's previous transaction(s) are sent in a succeeding page, because its signing time lies before them). (edge case)

### DIFF
--- a/network/proto/senders.go
+++ b/network/proto/senders.go
@@ -6,8 +6,14 @@ import (
 	"github.com/nuts-foundation/nuts-node/network/log"
 	"github.com/nuts-foundation/nuts-node/network/p2p"
 	"github.com/nuts-foundation/nuts-node/network/transport"
+	"google.golang.org/protobuf/proto"
+	"math"
 	"time"
 )
+
+// estimatedMessageSizeMargin defines the factor by which the message size is multiplied, as a safety measure to avoid
+// accidentally exceeding the max Protobuf message size.
+const estimatedMessageSizeMargin = 0.75
 
 // messageSender provides a domain-specific abstraction for sending messages to the network. It aids testing and
 // implementation of non-functional requirements like throttling.
@@ -21,7 +27,8 @@ type messageSender interface {
 }
 
 type defaultMessageSender struct {
-	p2p p2p.Adapter
+	p2p            p2p.Adapter
+	maxMessageSize int
 }
 
 func (s defaultMessageSender) doSend(peer p2p.PeerID, envelope *transport.NetworkMessage) {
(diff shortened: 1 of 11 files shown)


network/proto/senders.go
```diff
@@ -63,10 +70,37 @@ func (s defaultMessageSender) sendTransactionListQuery(peer p2p.PeerID, blockDat
 }
 
 func (s defaultMessageSender) sendTransactionList(peer p2p.PeerID, transactions []dag.Transaction, blockDate time.Time) {
-	envelope := createEnvelope()
+	if len(transactions) == 0 {
+		envelope := createEnvelope()
```
woutslakhorst, line 74: I think we can just skip sending anything. The protocol doesn't demand a response to a transactionListQuery.

network/proto/senders.go
```diff
@@ -63,10 +70,37 @@ func (s defaultMessageSender) sendTransactionListQuery(peer p2p.PeerID, blockDat
...
+	// When the DAG grows it the transactions might not fit in 1 network message (defined by p2p.MaxMessageSizeInBytes),
+	// so we have to split it up in pages. Protobuf can calculate the size of a message, so we can check whether it will fit in the max. message size.
+	// However, this can be quite expensive since it will marshal the message to the wire format (thus, marshalling will happen twice).
+	// Furthermore, we'd have to do this for every transaction we add to the message to find the max. number of TXs that
+	// will fit in the message.
+	// Since this sounds all horribly inefficient, we just calculate the size of a network message with 1 transaction,
+	// and use that to guesstimate the number of transactions that will fit in a message, decreased by a safety margin.
+	// Since transactions just contain metadata of the transaction (and not the payload itself), all transactions should
+	// serialize to more or less the same number of bytes, making the calculation safe.
+	sizeMsg := createEnvelope()
```
woutslakhorst, line 88: this calculation can be done in the initializer.

Things sounds horribly inefficient, but given current CPU power, how inefficient is it compared to other operations (eg compared to cryptographic operations)?

network/proto/senders_test.go
```diff
@@ -55,17 +57,63 @@ func Test_defaultMessageSender_broadcastDiagnostics(t *testing.T) {
...
+		sender.sendTransactionList(peer, txs, blockDate)
+
+		assert.Len(t, sentMessages, numberOfTXs)
+		for i := 0; i < numberOfTXs; i++ {
+			if !sentMessages[byte(i)] {
+				t.Fatalf("Missing message (idx: %d)", i)
+			}
+		}
+	})
+	t.Run("ok - no transactions sends empty message", func(t *testing.T) {
```
woutslakhorst, line 109: questionable

woutslakhorst (commented):
peer review on my comments are welcome.

dmeijboom:
The only thing I don't understand is where the `512K` message limit comes from? In Go's gRPC library by default it sets the send message limit to `2GB` (or actually `math.MaxInt32` bytes) which could be a lot more on 64-bit platforms and the receive message limit to `4MB` which could be extended as well (on both 32-bit and 64-bit platforms).

Nevertheless this PR makes sense as someday we will run into the receive or send message limit.

woutslakhorst:
> The only thing I don't understand is where the `512K` message limit comes from? In Go's gRPC library by default it sets the send message limit to `2GB` (or actually `math.MaxInt32` bytes) which could be a lot more on 64-bit platforms and the receive message limit to `4MB` which could be extended as well (on both 32-bit and 64-bit platforms).
> 
> Nevertheless this PR makes sense as someday we will run into the receive or send message limit.

We specify a max message size in the specs. This is to prevent attackers to send huge messages which could require more resources by the recipient. Rate limiting is also on the radar to prevent attackers from sending a lot of small messages ;)

network/proto/senders.go
```diff
@@ -63,10 +72,32 @@ func (s defaultMessageSender) sendTransactionListQuery(peer p2p.PeerID, blockDat
...
+
+	for i := 0; i < numberOfMessages; i++ {
+		upper := (i + 1) * transactionsPerMessage
+		if upper > len(tl) {
+			upper = len(tl)
+		}
+		envelope := createEnvelope()
+		envelope.Message = &transport.NetworkMessage_TransactionList{TransactionList: &transport.TransactionList{Transactions: tl[i*transactionsPerMessage : upper], BlockDate: uint32(blockDate.Unix())}}
+		s.doSend(peer, &envelope)
+	}
```
dmeijboom, line 100: There are two issues with this:

1. We can calculate the size of a single message once by using an `init` function to set a global variable or something like that (instead of doing it each time)
2. Currently, we use the `NetworkMessage` protobuf type to calculate the size of a single transaction but this will be very inefficient as it will count the overhead of the `NetworkMessage` and `TransactionList` type multiple times instead of only once
3. We can't safely predict the size of the transaction data as the JWS payload has no maximum size. Even with an error margin this could go wrong. Maybe it would be better to either add a restriction on the JWS payload (so that we can calculate the maximum possible size of the transaction data).

dmeijboom (changes requested):

network/proto/senders.go
woutslakhorst: 1. is now being done, but it requires a transaction to do. So it won't be possible at initialize.
2. is only done once, so?
3. our JWS payload only differs by the number of hashes in the `prev` header. The payload is always a SHA256

dmeijboom: I misread part of the code so `2` is definitely not an issue. But for `1` we could create a fake transaction and base the size on that right? It shouldn't really matter whether it's one of the actual transactions or a fake one as we already take an error margin into account.

And `3` still stands right? As we don't know the number of hashes in the `prev` header or do you think that it shouldn't differ that much?

dmeijboom: 1. Currently it does but we could create a 'fake' transaction and base the size on that right? Or do you think that the size will differ too much from a real-world transaction?
2. I meant that it will over-estimate the size of a single transaction as `messageSizePerTX` (on line `127`) will include the overhead of a network message and transaction list structure. So if there are 10 transactions it will count this overhead ten times (which results in less transactions being sent per page) instead of only once. Or am I missing something?
3. Check!

woutslakhorst: 0. when we switch to didcomm, all will change anyway
1. could be, but will probably be more sensitive to bugs since we have to make some assumptions
2. so we're on the safe side twice! Currently the estimate is that we will send 500 TXs per page (instead of a 1000 or so). This should not be reached in the future when we select TXs based on lampart clocks. Also historical syncs for new nodes should only be done from trusted nodes in a streaming way (not through messages).

in all, it's temporary

reinkrul: > We can calculate the size of a single message once by using an init function to set a global variable or something like that (instead of doing it each time)

We could do that, with some fake transaction. Would need some refactoring an maybe some clever way to assure the "guesstimate" fake-TX is as close to a real-world TX as possible. But I don't think this calculation (once per `TransactionList` sent back) is going to be the bottleneck. However, I could cache it?

> Currently, we use the NetworkMessage protobuf type to calculate the size of a single transaction but this will be very inefficient as it will count the overhead of the NetworkMessage and TransactionList type multiple times instead of only once

I figured the overhead is minimal since the protobuf definition (as text) is compiled to a TLV-like (tag-length-value), very efficient optimized format. Meaning that an additional (almost empty) type just adds a few bytes, which is a very small overhead when compared to the size of the text-based JWS. More overhead means also we're more safe.
And since an extra page because our guestimating algorithm isn't optimized won't be much slower: the same amount of data is sent, just in separate messages.  

> We can't safely predict the size of the transaction data as the JWS payload has no maximum size. Even with an error margin this could go wrong. Maybe it would be better to either add a restriction on the JWS payload (so that we can calculate the maximum possible size of the transaction data).

The JWS payload size is fixed, since it's the SHA256 hash of the actual data. The rest of the JWS (headers) is strictly spec'd, so there will be some variation in size but not too much. That's why there's an additonal safety margin (fill to 75% capacity).

reinkrul:
> > The only thing I don't understand is where the `512K` message limit comes from? In Go's gRPC library by default it sets the send message limit to `2GB` (or actually `math.MaxInt32` bytes) which could be a lot more on 64-bit platforms and the receive message limit to `4MB` which could be extended as well (on both 32-bit and 64-bit platforms).
> > Nevertheless this PR makes sense as someday we will run into the receive or send message limit.
> 
> We specify a max message size in the specs. This is to prevent attackers from sending huge messages which could require more resources by the recipient. Rate limiting is also on the radar to prevent attackers from sending a lot of small messages ;)

The 512Kb limit is arbitrarily chosen, though. We can always tweak this in future versions if required.

woutslakhorst (approved):

PR #449, opened 2021-09-08
vcs in access token

woutslakhorst:
This adds VC IDs from the access token request to the access token. It will only add the IDs of the credentials (due to header size limitations).
The token introspection will also only return the IDs.


dmeijboom:
The code looks fine but there should also be a PR for the demo EHR as the FHIR Proxy uses `token.Vcs` to validate the `NutsAuthorizationCredential` against the current route.

auth/services/messages.go
```diff
@@ -53,19 +53,19 @@ type JwtBearerTokenResult struct {
...
-	Subject     string                    `json:"sub"`
-	Audience    string                    `json:"aud"`
-	Credentials []vc.VerifiableCredential `json:"vcs"`
+	KeyID       string    `json:"-"`
+	Expiration  int64     `json:"exp"`
+	IssuedAt    int64     `json:"iat"`
+	Issuer      string    `json:"iss"`
+	Subject     string    `json:"sub"`
+	Audience    string    `json:"aud"`
+	Credentials *[]string `json:"vcs"`
```
dmeijboom, line 68: Why is it a pointer to a slice? A slice can already be `nil` right?

dmeijboom (changes requested):
Note that there should also be a PR for the demo EHR as the FHIR Proxy uses token.Vcs to validate the NutsAuthorizationCredential against the current route.

dmeijboom:
> Note that there should also be a PR for the demo EHR as the FHIR Proxy uses token.Vcs to validate the NutsAuthorizationCredential against the current route.

Looks like that's already done in: https://github.com/nuts-foundation/nuts-demo-ehr/pull/93 so nvm!

auth/services/messages.go
woutslakhorst: was following the generated code. The generator is ofcourse stupid and makes every optional a pointer.

dmeijboom (approved):
LGTM

PR #475, opened 2021-09-27
Network: periodically broadcast-query peers for missing payloads

### LINKED ISSUE #474: Auto-healing for missing payloads
Our current payload-retrieval function states:

```
// TODO: Currently we send the query to the peer that sent us the hash, but this peer might not have the
//   transaction contents. We need a smarter way to get it from a peer who does.
```

The function (to retrieve payload) is only called when a transaction is received. The peer however, might not have the payload or something might wrong while receiving the payload (network partition, disk fault, OS crash, etc) leading to the payload not being received. Since the process isn't retried, the payload will never be received and thus will be forever missing.

We have plans to make receiving both transaction and payload atomic, then this situation can't occur any more. But we need a fix for the meantime.

We could introduce a goroutine that periodically finds transactions which' payload is missing, and query all peers for the payload. That way the situation will be resolved over time, unless nobody has it any more, because the node has gone away or because of a network partition. In that case the branch should be garbage collected after some time.

reinkrul:
This adds a scheduled check to the network's protocol layer which checks for missing payloads (transactions which we have but whose' payload is missing) and queries them by broadcasting a `TransactionPayloadQuery` to all of our peers.

In the circumstances the initial payload retrieval from the original peer (we got the TX from) failed, this causes the node to retrieve the payload from another node in the network.

The hashes of the missing payloads are also shown on the diagnostics page.

Fixes #474 
Implements https://github.com/nuts-foundation/nuts-specification/pull/128


network/dag/bbolt_payloadstore.go
```diff
@@ -53,3 +68,16 @@ func (store bboltPayloadStore) WritePayload(payloadHash hash.SHA256Hash, data []
...
 	return err
 }
+
+type bboltPayloadReader struct {
+	payloadsBucket *bbolt.Bucket
+}
+
+func (reader bboltPayloadReader) IsPresent(payloadHash hash.SHA256Hash) (bool, error) {
+	data := reader.payloadsBucket.Get(payloadHash.Slice())
+	return len(data) > 0, nil
```
dmeijboom, line 78: According to the docs:
> Returns a nil value if the key does not exist or if the key is a nested bucket.

We have to check for `nil` instead

network/proto/impl.go
```diff
@@ -111,6 +133,7 @@ func (p *protocol) Start() {
 	go p.startUpdatingDiagnostics(peerConnected, peerDisconnected)
 	go p.startAdvertingHashes()
 	go p.startAdvertingDiagnostics()
+	go p.startCollectingMissingPayloads()
```
dmeijboom, line 136: I'm wondering. What's the strategy for shutting down?

dmeijboom (changes requested):
Overall the solution looks fine. Probably some 'waste' here and there (e.g. where we query for transaction payloads even though we might already found all of them). But as it's a temporary fix this should be fine.

network/proto/impl.go
reinkrul: Glad you asked, I'm wondering as well. Not where it is (because I simply did't write it) but why it seems to work fine without it.

As goroutines are shut down together with the process, the program *will* exit some way or another. Since we don't write to (local) resources from these goroutines, I think nothing will be corrupted by the lacking shutdown mechanic. But it isn't very neat either, because it may cause the occasional weird error message when shutting down the Nuts node.

Is it OK to create an issue to fix it for all goroutines in this engine, rather than fixing it for this specific one?

dmeijboom: I guess it's fine-ish as there shouldn't be any corruption indeed but like you said it would be nice to shut it down 'properly'.

> Is it OK to create an issue to fix it for all goroutines in this engine, rather than fixing it for this specific one?

Yeah, makes sense 🙂

network/dag/bbolt_dag.go
```diff
@@ -154,6 +154,23 @@ func (dag bboltDAG) GetByPayloadHash(payloadHash hash.SHA256Hash) ([]Transaction
 	return result, err
 }
 
+func (dag *bboltDAG) PayloadHashes(visitor func(payloadHash hash.SHA256Hash) error) error {
```
woutslakhorst, line 157: untested

network/dag/bbolt_dag.go
```diff
@@ -154,6 +154,23 @@ func (dag bboltDAG) GetByPayloadHash(payloadHash hash.SHA256Hash) ([]Transaction
...
+	return dag.db.View(func(tx *bbolt.Tx) error {
+		payloadIndex := tx.Bucket([]byte(payloadIndexBucket))
```
woutslakhorst, line 159: How is this bucket structured. Is it payloadHash => payload or txRef => payload or?

woutslakhorst (changes requested):
I'm correct to assume that every 60s the entire DB is scanned for missing payloads? What would be the impact of keeping a separate DB for missing payloads?

reinkrul:
> I'm correct to assume that every 60s the entire DB is scanned for missing payloads? What would be the impact of keeping a separate DB for missing payloads?

More complex implementation with more state (= potentially more bugs). Can be optimization when the current approach proves too slow. Which I doubt (especially when there aren't 100.000's of TXs) because they use 1 TX, which should be quite quick using BBolt.

network/dag/bbolt_dag.go
woutslakhorst: so it's a reverse index ;)

woutslakhorst:
> > I'm correct to assume that every 60s the entire DB is scanned for missing payloads? What would be the impact of keeping a separate DB for missing payloads?
> 
> More complex implementation with more state (= potentially more bugs). Can be optimization when the current approach proves too slow. Which I doubt (especially when there aren't 100.000's of TXs) because they use 1 TX, which should be quite quick using BBolt.

Ok, so only a missing test

network/proto/impl.go
reinkrul: https://github.com/nuts-foundation/nuts-node/issues/480

reinkrul:
> Overall the solution looks fine. Probably some 'waste' here and there (e.g. where we query for transaction payloads even though we might already found all of them). But as it's a temporary fix this should be fine.

We only query payloads that we're missing. A trade-off would be querying a random peer (or peers) for it, but that could lead to being it unavailable longer. (less network traffic v.s. more time before the network is sync). But since the number of peers is limited, I think this will work fine for now.

dmeijboom (approved):

woutslakhorst (approved):

PR #483, opened 2021-09-28
Fix error message of drawup contract call

stevenvegt:
When the wrong date is used, the error is misleading.


woutslakhorst (changes requested):
Ok, but test fails because it's matching the error string (which has changed)

woutslakhorst (approved):

PR #485, opened 2021-09-29
Fix credential clock skew

dmeijboom:
Currently allows for 5 seconds max

### DIFF
--- a/vcr/vcr.go
+++ b/vcr/vcr.go
@@ -48,6 +48,8 @@ import (
 	"gopkg.in/yaml.v2"
 )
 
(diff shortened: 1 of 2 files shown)


stevenvegt (commented):
Not sure if I like the change in time during creation. I would expect it during validation.

dmeijboom:
> Not sure if I like the change in time during creation. I would expect it during validation.

I agree. Fixed

stevenvegt:
Add a little test?

woutslakhorst (approved):

PR #500, opened 2021-10-04
remove usi from introspection

### LINKED ISSUE #498: Introspection response lists `usi` as return field
But the access token doesn't contain this field anymore (too big), instead it lists specific attributes of the user.

woutslakhorst:
fixes #498 

was there incorrectly


dmeijboom (approved):
LGTM, should be removed from demo EHR as well

Issue #487, opened 2021-09-29
confusing terms in request-access-token


woutslakhorst:
it uses *custodian* and *actor*. For the eOverdracht notification these roles are reversed. Better terms would be *requestor* and *authorizer*.

PR #502, opened 2021-10-04
Rename custodian/actor to requester/authorizer

dmeijboom:
Closes: #487

TODO:
- Demo EHR
- Specs
- E2E tests
- Workshop manuals


dmeijboom:
Should this be backwards compatible? I guess not right?

woutslakhorst:
- workshop manuals

yes this would be a breaking change. We could warn for it on Slack after merge. Setup a #development channel for it

woutslakhorst (approved):

PR #506, opened 2021-10-04
check VC sig when creating access token

### LINKED ISSUE #494: AuthorizationCredential validation when requesting access token, does not check the proof
it validates fields, but seems to forget checking the signature....

woutslakhorst:
closes #494

The Validate method on the VCR now accepts an additional `checkSignature` param. VCs returned from local storage do not need this check, but VCs given via the API do.

### DIFF
--- a/vcr/vcr.go
+++ b/vcr/vcr.go
@@ -227,7 +227,7 @@ func (c *vcr) search(query concept.Query, allowUntrusted bool, resolveTime *time
 				return nil, errors.Wrap(err, "unable to parse credential from db")
 			}
 
-			if err = c.Validate(foundCredential, allowUntrusted, resolveTime); err == nil {
+			if err = c.Validate(foundCredential, allowUntrusted, false, resolveTime); err == nil {
 				VCs = append(VCs, foundCredential)
 			}
 		}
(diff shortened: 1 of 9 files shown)


vcr/vcr.go
```diff
@@ -227,7 +227,7 @@ func (c *vcr) search(query concept.Query, allowUntrusted bool, resolveTime *time
 				return nil, errors.Wrap(err, "unable to parse credential from db")
 			}
 
-			if err = c.Validate(foundCredential, allowUntrusted, resolveTime); err == nil {
+			if err = c.Validate(foundCredential, allowUntrusted, false, resolveTime); err == nil {
```
reinkrul, line 230: just curious, what would be the use case for searching for VCs with invalid signatures v.s. VCs with valid signatures?

reinkrul (commented):
question

vcr/vcr.go
woutslakhorst: VCs signatures are checked before a VC is written to the store. So the store only has valid VCs.
Checking them on return slows things down (and makes the unit tests a lot harder)

reinkrul: I see in VCR's `Ambassador`:
> // vcCallback gets called when new Verifiable Credentials are received by the network. All checks on the signature are already performed.

I assume the godoc means the network layer checks the signature. That's correct, for the TX signature. But there's also a `proof` in our Nuts VCs with a `jws` field, which isn't checked?

Update: just checked, VCR's `store.StoreCredential` verifies the VC signature. So the ambassador's comment is a bit misleading ("signatures" being plural, while just meaning "TX's signature"). Maybe we should change the comment to "the transaction's signature has already been verified".

reinkrul (approved):
Just a comment about a related piece of code.

PR #512, opened 2021-10-05
resolve DID Docs by TX hash after 1634446887

### LINKED ISSUE #432: Resolve DID Document via hash in VC validation
#430 introduced references to DID Document TXs when issuing/revoking VCs. Resolving the DID Document based on this information instead of timestamp has to be implemented.

The vcr probably needs access to the raw transactions. The prevs of the transaction point to transactions, we need the payload hash from those transactions to resolve the DID Document.

implement after 2021/10/01 (to reduce broken dev nodes)

spec: nuts-foundation/nuts-specification#117

woutslakhorst:
closes #432 

the epoch is at GMT: Sunday, 17 October 2021 05:01:27
arbitrary chosen, it ends on 6889 which is T9 for *nuts*

After that date DID Documents/Public Keys are resolved by TX hash @ DAG verification.

- [ ] use correct prevs

### DIFF
--- a/network/dag/verifier.go
+++ b/network/dag/verifier.go
@@ -4,10 +4,11 @@ import (
 	crypto2 "crypto"
 	"errors"
 	"fmt"
+	"time"
+
 	"github.com/lestrrat-go/jwx/jwa"
 	"github.com/lestrrat-go/jwx/jws"
 	"github.com/nuts-foundation/nuts-node/vdr/types"
-	"time"
 )
 
 // ErrPreviousTransactionMissing indicates one or more of the previous transactions (which the transaction refers to)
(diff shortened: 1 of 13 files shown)


network/dag/verifier.go
```diff
@@ -28,11 +32,20 @@ func NewTransactionSignatureVerifier(resolver types.KeyResolver) Verifier {
...
-				return fmt.Errorf("unable to verify transaction signature, can't resolve key (kid=%s): %w", tx.SigningKeyID(), err)
+			if signingTime.After(didDocumentResolveEPoch) {
+				pk, err := resolver.ResolvePublicKeyFromOriginatingTransaction(tx.SigningKeyID(), tx.Ref())
+				if err != nil {
+					return fmt.Errorf("unable to verify transaction signature, can't resolve key by TX ref (kid=%s, txhash=%s): %w", tx.SigningKeyID(), tx.Ref().String(), err)
+				}
+				signingKey = pk
+			} else {
+				// legacy resolving for older documents
+				pk, err := resolver.ResolvePublicKey(tx.SigningKeyID(), &signingTime)
```
dmeijboom, line 43: Nitpick: Wouldn't it be better to use `resolve.ResolvePublicKey` for the current implementation and something like: `resolver.ResolveTimeBasedPublicKey` for the legacy implementation?

dmeijboom (approved):

Issue #491, opened 2021-09-30
Parallel DID creates/updates makes DAG walker panic


reinkrul:
Tested using https://github.com/nuts-foundation/nuts-node/blob/vdr-load-test/generate-loads-of-dids_test.go:

```
node_1   | time="2021-09-30T08:45:46Z" level=info msg="DID document updated (tx=b4325e9b57b090c67030b9f56350d49dbd5bce241deda8acf142e840f17c16c6,did=did:nuts:5bL9Qw9mdwSs7puyDi5xpycDfhstMb15E78VbhLfGphP)" module=VDR
node_1   | fatal error: concurrent map read and map write
node_1   |
node_1   | goroutine 111 [running]:
node_1   | runtime.throw({0xf563ff, 0xc000ed9d20})
node_1   | 	/usr/local/go/src/runtime/panic.go:1198 +0x71 fp=0xc000ed9cf0 sp=0xc000ed9cc0 pc=0x4355b1
node_1   | runtime.mapaccess2(0xebc140, 0xc0015f1860, 0x99b631a1d9ad8401)
node_1   | 	/usr/local/go/src/runtime/map.go:469 +0x205 fp=0xc000ed9d30 sp=0xc000ed9cf0 pc=0x40e565
node_1   | github.com/nuts-foundation/nuts-node/network/dag.bfsWalkerAlgorithm.walk({0xc0002a5560, 0xc0002a5590}, 0xc000d05550, {0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...}, ...)
node_1   | 	/go/network/dag/bfswalker.go:56 +0x40e fp=0xc000ed9ee8 sp=0xc000ed9d30 pc=0xa66c6e
node_1   | github.com/nuts-foundation/nuts-node/network/dag.(*bfsWalkerAlgorithm).walk(0xc0004c4a00, 0xc0000b9440, {0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...}, ...)
node_1   | 	<autogenerated>:1 +0x55 fp=0xc000ed9f40 sp=0xc000ed9ee8 pc=0xa6c735
node_1   | github.com/nuts-foundation/nuts-node/network/dag.bboltDAG.Walk.func1(0x0)
node_1   | 	/go/network/dag/bbolt_dag.go:211 +0x182 fp=0xc000ed9fe8 sp=0xc000ed9f40 pc=0xa64302
[...]

PR #516, opened 2021-10-06
Network: provide PayloadReader to DAG visitor to avoid nested calls to BBoltDB.View()

reinkrul:
Fixes the deadlock in #491


reinkrul:
Data race seems to be an existing issue in VDR's `Ambassador` in conjunection with `MemoryStore`?

woutslakhorst (changes requested):
This fixes it since the `payloadStore` is no longer directly referenced from the DAG. I would like to see some more isolation, for example place the payload store in a different package to prevent the reader to leak to other layers.

woutslakhorst:
isolate storage from logic? bbolt_dag represents dag logic connected to the underlying store. What if all access to the store is abstracted? There's a payloadStore but not a txStore, that part is embedded in the dag.

reinkrul:
> Ah, check. I'd prefer the second option but we could run into issues when consumers need to write to bbolt while the transaction is read-only. Not sure if that's currently already the case.

I'm currently implementing the second option. The advantage of the new approach is that previously it would cause a deadlock, now we can detect it (and return an error/panic/whatever).

PR #528, opened 2021-10-08
made clockskew configurable with a default of 5s

### LINKED ISSUE #484: make clock-skew configurable
something like 5 seconds as default

This is only needed when requesting the access token

woutslakhorst:
closes #484 

had to push the config a bit further down the line that I hoped for.


woutslakhorst:
> This doesn't cover the clock skew configuration for vc validation right?

correct, that's a different case. This case caused persistent/replayable errors in the hackathon.

PR #532, opened 2021-10-08
Network: drop messages when peer's outbound queue has reached capacity

reinkrul:
Happens under heavy stress.

Turns out it deadlocks certain processes in the Nuts node, since writing to a full channel blocks until there's room for the entry.

Next to find out is why the channel isn't being emptied on the reader's side.

### DIFF
--- a/network/p2p/connection.go
+++ b/network/p2p/connection.go
@@ -20,6 +20,7 @@ package p2p
 
 import (
 	"errors"
+	"fmt"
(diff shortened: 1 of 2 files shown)


reinkrul:
> Looks good but shouldn't we drop messages if the queue has reached it's max capacity? Otherwise the node will exit because of the error right?

The error is catched and then logged as warning.

PR #534, opened 2021-10-11
Core: serve landing page when / is requested

### LINKED ISSUE #464: Serve landing page at HTTP API root
To avoid questions about "why doesn't it work" when people spin up the node and go to `http://localhost:1323` and see a 404 error.

Something like:

"It works! But there's nothing to see here. You can view the diagnostics page (link) to get some details about your node's status".

reinkrul:
Fixes #464


woutslakhorst:
some codeclimate issues

PR #535, opened 2021-10-11
Validate revoked certificates

dmeijboom:
The `RevocationDB` downloads or reloads CRLs based on an endpoint (signature is not validated yet). All CLRs are stored in memory where revoked certificates can be looked up. It's backed by a bloom filter (hash is generated by `math.abs(murmur3(issuer + subject.serialNumber))`) so checking revoked certificates should be really cheap in most cases.

### DIFF
--- a/auth/services/x509/uzi_validator_test.go
+++ b/auth/services/x509/uzi_validator_test.go
@@ -1,6 +1,9 @@
 package x509
 
 import (
+	"crypto/x509"
+	"crypto/x509/pkix"
+	"encoding/asn1"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


dmeijboom:
@woutslakhorst 

> Sync() is not a periodic process but only executed at startup, hence IsValid will start to fail if the node is not restarted


Oops, fixed

dmeijboom:
> The CRL for the network layer is now synced from a go procedure. The sync from NewJwtX509Validator however is still only run once.

@woutslakhorst That's not correct. In the `JwtX509Validator.Verify(...)` method the `Sync(...)` method is called so it should be synchronised as well. It's lame that it will block but that was already the case in the previous implementation.

PR #545, opened 2021-10-12
speedup tests

woutslakhorst:


woutslakhorst:
doesn't do a lot for the automated tests

Issue #550, opened 2021-10-15
Implement internal event engine


woutslakhorst:
Both #549 and parts of the newer protocols will require event based communication.
The retry queues for private credentials also require persistent storage.

In the past we used Nats for this in combination with Stan (Nats streaming). It always worked quite well and luckily for us Stan is now embedded in Nats (new Jetstream feature).

Proposal is to introduce an engine that handles app wide events (non-persistent) and the ability to introduce engine local persistent queues. It looks like Nats can produce metrics for prometheus using https://github.com/nats-io/prometheus-nats-exporter. It also supports clustering, secured connections and encryption at rest.

- [x] add events engine #552
- [ ] make smart config, no host == localhost, no port = 4222. If no host is given, start a server otherwise connect to an existing one. 
- [ ] Expose diagnostics/basic metrics from Nats.
- [ ] Add simple pub/sub to interface for system wide events, add rules for using events and add them to dev docs
- [ ] Add builder for retry queue. A retry queue has a simple interface: pub/sub/ack/nack where a nack or lack of ack will retrigger the event after a certain period of time (configurable)

woutslakhorst:
outdated

PR #552, opened 2021-10-15
add "events" engine

woutslakhorst:
This PR only adds the engine and starts a Nats server. No other interaction yet.

I made the different log/logger/logging packages consistent.

part of #550


reinkrul:
Normally additional improvements like the package refactors really deserve their own PR, because if we for some reason don't merge the Nats server feature, we don't get the other improvement. Or you get conflicts.

PR #573, opened 2021-11-01
Implemented a persisent version of the vdr store

### LINKED ISSUE #109: VDR: Make in-memory DID Document store persistent

dmeijboom:
This is only the 'bare' implementation + tests. Need to configure it by default and test it using my local nuts node (but maybe that could be in a different PR?).

The tests are copied from the `MemoryStore` tests so we can verify it works more or less the same.

Fixes #109

### DIFF
--- a/vdr/store/bbolt.go
+++ b/vdr/store/bbolt.go
@@ -34,6 +34,10 @@ type documentVersionList struct {
 }
 
 func (entry documentVersionList) Latest() hash.SHA256Hash {
+	if len(entry.Versions) == 0 {
+		return hash.EmptyHash()
+	}
+
 	return entry.Versions[len(entry.Versions)-1]
 }
 
--- a/vdr/store/bbolt.go
+++ b/vdr/store/bbolt.go
@@ -125,7 +125,7 @@ func (store *bboltStore) filterDocument(doc *documentVersion, metadata *vdr.Reso
 		return vdr.ErrNotFound
 	}
 
-	// Verify creation and update time
+	// Filter on creation and update time
 	if metadata.ResolveTime != nil {
 		resolveTime := *metadata.ResolveTime
 
@@ -140,7 +140,7 @@ func (store *bboltStore) filterDocument(doc *documentVersion, metadata *vdr.Reso
 		}
 	}
 
-	// Verify KeyTransaction
+	// Filter on SourceTransaction
 	if metadata.SourceTransaction != nil {
 		for i, keyTx := range doc.Metadata.SourceTransactions {
 			if keyTx.Equals(*metadata.SourceTransaction) {
--- a/vdr/store/bbolt.go
+++ b/vdr/store/bbolt.go
@@ -17,6 +17,7 @@ package store
 
 import (
 	"encoding/json"
+	"errors"
 	"github.com/nuts-foundation/go-did/did"
 	"github.com/nuts-foundation/nuts-node/crypto/hash"
 	vdr "github.com/nuts-foundation/nuts-node/vdr/types"
@@ -175,7 +176,10 @@ func (store *bboltStore) filterDocument(doc *documentVersion, metadata *vdr.Reso
 	return nil
 }
 
-// Resolve returns the DID Document for the provided DID
+// Resolve returns the DID Document for the provided DID.
+// If metadata is not provided the latest version is returned.
+// If metadata is provided then the result is filtered or scoped on that metadata.
+// It returns ErrNotFound if there are no corresponding DID documents or when the DID Documents are disjoint with the provided ResolveMetadata
 func (store *bboltStore) Resolve(id did.DID, metadata *vdr.ResolveMetadata) (document *did.Document, documentMeta *vdr.DocumentMetadata, txErr error) {
 	txErr = store.db.View(func(tx *bbolt.Tx) error {
 		documents := tx.Bucket(documentsBucket)
@@ -199,7 +203,13 @@ func (store *bboltStore) Resolve(id did.DID, metadata *vdr.ResolveMetadata) (doc
 			return err
 		}
 
-		data = documents.Get(versionList.Latest().Slice())
+		versionHash := versionList.Latest()
+
+		if metadata != nil && metadata.Hash != nil {
+			versionHash = *metadata.Hash
+		}
+
+		data = documents.Get(versionHash.Slice())
 		if data == nil {
 			return vdr.ErrNotFound
 		}
@@ -257,6 +267,10 @@ func (store *bboltStore) Write(document did.Document, metadata vdr.DocumentMetad
 
 // Update replaces the DID document identified by DID with the nextVersion
 func (store *bboltStore) Update(id did.DID, current hash.SHA256Hash, next did.Document, metadata *vdr.DocumentMetadata) error {
+	if metadata == nil {
+		return errors.New("unable to update document without metadata")
+	}
+
 	return store.db.Update(func(tx *bbolt.Tx) error {
 		versions, err := tx.CreateBucketIfNotExists(versionsBucket)
 		if err != nil {
--- a/vdr/store/bbolt.go
+++ b/vdr/store/bbolt.go
@@ -36,10 +36,6 @@ func NewBBoltStore(db *bbolt.DB) vdr.Store {
 	return &bboltStore{db: db}
 }
 
-func (store *bboltStore) getDocumentKey(id did.DID, hash hash.SHA256Hash) []byte {
-	return append([]byte(id.String()), hash.Slice()...)
-}
-
 func (store *bboltStore) storeDocument(tx *bbolt.Tx, document did.Document, metadata vdr.DocumentMetadata) error {
 	documents, err := tx.CreateBucketIfNotExists(documentsBucket)
 	if err != nil {
@@ -54,15 +50,15 @@ func (store *bboltStore) storeDocument(tx *bbolt.Tx, document did.Document, meta
 		return err
 	}
 
-	if err := documents.Put(store.getDocumentKey(document.ID, metadata.Hash), data); err != nil {
+	if err := documents.Put(metadata.Hash.Slice(), data); err != nil {
 		return err
 	}
 
 	return nil
 }
 
-func (store *bboltStore) getDocumentVersion(bucket *bbolt.Bucket, id did.DID, hash hash.SHA256Hash) (*documentVersion, error) {
-	data := bucket.Get(store.getDocumentKey(id, hash))
+func (store *bboltStore) getDocumentVersion(bucket *bbolt.Bucket, hash hash.SHA256Hash) (*documentVersion, error) {
+	data := bucket.Get(hash.Slice())
 	if data == nil {
 		return nil, nil
 	}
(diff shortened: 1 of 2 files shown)


reinkrul:
Does it have the same tests as the in-memory store? (that's part of the proof it's a 1:1 replacement)

dmeijboom:
> Does it have the same tests as the in-memory store? (that's part of the proof it's a 1:1 replacement)

Right, I should've explained that in the description 😅 I copied the unit tests from the memory-store so we can more or less verify that it works the same way. So, yes 🙂

vdr/store/bbolt.go
```diff
@@ -0,0 +1,305 @@
...
+
+func (store *bboltStore) getDocumentVersion(bucket *bbolt.Bucket, id did.DID, hash hash.SHA256Hash) (*documentVersion, error) {
+	data := bucket.Get(store.getDocumentKey(id, hash))
+	if data == nil {
+		return nil, nil
+	}
+
+	result := &documentVersion{}
+
+	if err := json.Unmarshal(data, result); err != nil {
```
reinkrul, line 72: You could encode this as a byte (for boolean `Deactivated`) + blocks of hashes, since they're always the same length. Then you don't need JSON marshalling. That's how `bboltDAG` does it:

```
// parseHashList splits a list of concatenated hashes into separate hashes.
func parseHashList(input []byte) []hash.SHA256Hash {
	if len(input) == 0 {
		return nil
	}
	num := (len(input) - (len(input) % hash.SHA256HashSize)) / hash.SHA256HashSize
	result := make([]hash.SHA256Hash, num)
	for i := 0; i < num; i++ {
		result[i] = hash.FromSlice(input[i*hash.SHA256HashSize : i*hash.SHA256HashSize+hash.SHA256HashSize])
	}
	return result
}
```

vdr/store/bbolt.go
```diff
@@ -0,0 +1,305 @@
...
+		}
+
+		if err := versions.ForEach(func(key, data []byte) error {
+			versionList := &documentVersionList{}
+
+			if err := json.Unmarshal(data, versionList); err != nil {
+				return err
+			}
+
+			id, _ := did.ParseDID(string(key))
```
reinkrul, line 99: chances of an error are very small here, but now leads to a nil deref. Maybe just skip this one if empty? (or then just return the error).

vdr/store/bbolt.go
```diff
@@ -0,0 +1,305 @@
...
+			return err
+		}
+
+		return nil
+	})
+}
+
+func (store *bboltStore) filterDocument(doc *documentVersion, metadata *vdr.ResolveMetadata) error {
+	// Verify deactivated
+	if IsDeactivated(doc.Document) {
```
reinkrul, line 120: what about `metadata.AllowDeactivated`?

vdr/store/bbolt.go
```diff
@@ -0,0 +1,305 @@
...
+	if err := json.Unmarshal(data, result); err != nil {
+		return nil, err
+	}
+
+	return result, nil
+}
+
+// Iterate loops over all the latest versions of the stored DID Documents and applies fn
+func (store *bboltStore) Iterate(fn vdr.DocIterator) error {
+	return store.db.View(func(tx *bbolt.Tx) error {
```
reinkrul, line 81: make sure not to call any BBolt functions from the `fn` callback. If probable, you can reuse the transaction-from-context concept from `Network` in `network/dag/bbolt_context.go`.

vdr/store/bbolt.go
```diff
@@ -0,0 +1,305 @@
...
+	if IsDeactivated(doc.Document) {
+		return vdr.ErrDeactivated
+	}
+
+	if metadata == nil {
+		return nil
+	}
+
+	// Verify creation and update time
+	if metadata.ResolveTime != nil {
```
reinkrul, line 129: there's some checks in the in-memory implementation that are missing here, e.g. `vdr/store/memory.go:171` (check `Updated` of `next` as well)

vdr/store/bbolt.go
```diff
@@ -0,0 +1,305 @@
...
+		}
+
+		if doc.Metadata.Updated != nil {
+			if doc.Metadata.Updated.After(resolveTime) {
+				return vdr.ErrNotFound
+			}
+		}
+	}
+
+	// Verify KeyTransaction
```
reinkrul, line 143: incorrect and non-descriptive comment.

vdr/store/bbolt.go
```diff
@@ -0,0 +1,305 @@
...
+		}
+
+		if err := json.Unmarshal(data, versionList); err != nil {
+			return err
+		}
+
+		for _, versionHash := range versionList.Versions {
+			data = documents.Get(store.getDocumentKey(id, versionHash))
+			if data == nil {
+				continue
```
reinkrul, line 203: this would be very odd?

vdr/store/bbolt.go
```diff
@@ -0,0 +1,305 @@
...
+			}
+
+			doc := &documentVersion{}
+
+			if err := json.Unmarshal(data, doc); err != nil {
+				return err
+			}
+
+			if err := store.filterDocument(doc, metadata); err == nil {
```
reinkrul, line 212: This leads `Resolve()` to return the first version, instead of the most recent (latest) one?

reinkrul (changes requested):

vdr/store/bbolt.go
```diff
@@ -0,0 +1,305 @@
...
+	db *bbolt.DB
+}
+
+// NewBBoltStore returns an instance of a BBolt based VDR store
+func NewBBoltStore(db *bbolt.DB) *bboltStore {
+	return &bboltStore{db: db}
+}
+
+func (store *bboltStore) getDocumentKey(id did.DID, hash hash.SHA256Hash) []byte {
+	return append([]byte(id.String()), hash.Slice()...)
```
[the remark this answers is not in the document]
dmeijboom: Indeed, and collisions are extremely rare in sha256 so should be fine to just use the hash indeed

vdr/store/bbolt.go
dmeijboom: Ah right, will fix

dmeijboom: And I'll add a test as well as this should've been covered by the unit tests.

vdr/store/bbolt.go
dmeijboom: I'll check if we do that somewhere

vdr/store/bbolt.go
dmeijboom: Note that these comments are from `memory_store.go:87` and on with `s/filter/verify/g`. Apart from that, why is it incorrect?

vdr/store/bbolt.go
dmeijboom: Indeed. But even then I'd argue that we shouldn't rely on `data` to be non-`nil` here.

vdr/store/bbolt.go
dmeijboom: Ah right, will fix (and add a test)

vdr/store/bbolt.go
```diff
@@ -0,0 +1,305 @@
...
+	return store.db.Update(func(tx *bbolt.Tx) error {
+		versions, err := tx.CreateBucketIfNotExists(versionsBucket)
+		if err != nil {
+			return err
+		}
+
+		key := []byte(document.ID.String())
+
+		if versions.Get(key) != nil {
+			return vdr.ErrDIDAlreadyExists
```
woutslakhorst, line 237: This will sting when rewalking the DAG

woutslakhorst (changes requested):
Apart from the comments that need changes:

I know it's a 1-on-1 copy of the in-memory variant. But you could extend it with indexed search capabilities and save some lines of code. I would try to extend the metadata with all fields to search on, create 2 buckets: metadata and document, both with the doc hash as key. Then index the metadata files with go-leia (biased, yes).

vdr/store/bbolt.go
dmeijboom: We can also return `nil` here instead? In the sense that nothing really went wrong.

vdr/store/bbolt.go
dmeijboom: I don't like premature optimisations so I ran a quick benchmark:

```
BenchmarkBBoltStore_VersionListBinary-8         32568783                36.84 ns/op
BenchmarkBBoltStore_VersionListJSON-8             383474              3036 ns/op
BenchmarkResolve-8                                105624              9695 ns/op
```

The resolve will drop from `9695 ns/op` to ~ `6696 ns/op` which IMO is significant enough to encode/decode  `documentVersionList` manually.

dmeijboom:
> Apart from the comments that need changes:
> 
> I know it's a 1-on-1 copy of the in-memory variant. But you could extend it with indexed search capabilities and save some lines of code. I would try to extend the metadata with all fields to search on, create 2 buckets: metadata and document, both with the doc hash as key. Then index the metadata files with go-leia (biased, yes).

I'm not sure if this is the right approach. Iterating will be a quite a lot faster but writing and updating documents is no longer atomic as go-leia creates a one-off transaction for each call to `.Add()`. The current implementation handles both adding the actual document and appending to the version-list in a single transaction.

woutslakhorst:
> I'm not sure if this is the right approach. Iterating will be a quite a lot faster but writing and updating documents is no longer atomic as go-leia creates a one-off transaction for each call to .Add(). The current implementation handles both adding the actual document and appending to the version-list in a single transaction.

We can ofcourse make that atomic since we control the library. I don't think there'll be a lot of difference in performance. The main reason I suggested it, is to reduce complexity from this implementation, especially when more search features will be added. I'll create an example based on this PR

dmeijboom:
> > I'm not sure if this is the right approach. Iterating will be a quite a lot faster but writing and updating documents is no longer atomic as go-leia creates a one-off transaction for each call to .Add(). The current implementation handles both adding the actual document and appending to the version-list in a single transaction.
> 
> We can ofcourse make that atomic since we control the library. I don't think there'll be a lot of difference in performance. The main reason I suggested it, is to reduce complexity from this implementation, especially when more search features will be added. I'll create an example based on this PR

> We can ofcourse make that atomic since we control the library

Sure but...

> I don't think there'll be a lot of difference in performance. The main reason I suggested it, is to reduce complexity from this implementation, especially when more search features will be added.

Check. Especially the latter makes sense as searching for documents based on their metadata requires a lookup for each document currently.


> I'll create an example based on this PR

All right. Note that even though I had some doubts I already started integrating Go-leia.

vdr/store/bbolt.go
```diff
@@ -0,0 +1,305 @@
...
+		if err != nil {
+			return err
+		}
+
+		if err := versions.Put(versionKey, data); err != nil {
+			return err
+		}
+
+		// Store the document
+		return store.storeDocument(tx, next, *metadata)
```
woutslakhorst, line 303: Possible nill dereference

woutslakhorst (changes requested):

vdr/store/bbolt.go
dmeijboom: Not really sure if this makes sense to add in the bbolt store implementation. The caller of the `Update` method is responsible for adding a version that comes after the latest version (it also passes a hash for the latest version which is validated).

woutslakhorst (approved):

reinkrul (approved):

reinkrul:
What is the status of this PR?

dmeijboom:
> What is the status of this PR?

In theory it can be merged but it does nothing (as the bbolt store isn't actually used). Implementing this requires refactoring the engines stuff iirc. That still has to be done.

PR #584, opened 2021-11-04
Network: enable TLS by default, do not derive from other properties

reinkrul:
Was previously derived from `CertFile` and `CertKeyFile` being set, but this didn't actually work. Changed it to be set explicitly, with default = `true`. Also logs a warning when one of the cert/key properties is set, but TLS is disabled.


dmeijboom (approved):
LGTM

woutslakhorst (approved):

PR #585, opened 2021-11-04
Network: fix gRPC package which breaks compatibility with older Nuts node versions

reinkrul:
When renaming the package the protobuf spec is in from `transport` to `protobuf`, only `go_package` should've been changed in the protobuf spec. The `package` directive is used in the fully qualified gRPC method name. When it's changed, other clients can't call the gRPC method.


woutslakhorst (approved):

dmeijboom (approved):
LGTM

PR #592, opened 2021-11-08
Quick (and dirty) copyright fixer

reinkrul:
Got tired of going through the files by hand. It adds the notice to Go files when it's missing.

Run it using:
```go run ./doc copyright```

README generation must now be invoked with:
```go run ./doc docs```

Will run and commit when this is merged.

### CODE REVIEW

File: docs/check_copyright.go [hunk 1]
```diff
@@ -0,0 +1,79 @@
...
+
+import (
+	"os"
+	"path"
+	"path/filepath"
+	"strings"
+)
+
+const copyrightText = `/*
+ * Copyright (C) 2021 Nuts community
```
woutslakhorst, line 11: Use current year and do find/replace?

dmeijboom:
A simple solution using command line utilities would be something like this (not tested):

```
export COPYRIGHT="Copyright contents (newlines should be replaced with \n)"
grep -Ri --include "*.go" -L "Copyright (C) 2021 Nuts community" . | xargs -I{} sed -i "1s;^;${COPYRIGHT}\n;" {}
```

docs/check_copyright.go
reinkrul: Fixed

reinkrul:
> A simple solution using command line utilities would be something like this (not tested):
> 
> ```
> export COPYRIGHT="Copyright contents (newlines should be replaced with \n)"
> grep -Ri --include "*.go" -L "Copyright (C) 2021 Nuts community" . | xargs -I{} sed -i "1s;^;${COPYRIGHT}\n;" {}
> ```

My Bash-fu has a terrible track record, so writing something quickly in Go is much faster for me.

docs/check_copyright.go
```diff
@@ -0,0 +1,92 @@
...
+			if info.IsDir() {
+				return nil
+			}
+			if !strings.HasSuffix(info.Name(), ".go") {
+				// only Go files
+				return nil
+			}
+			if strings.Contains(info.Name(), "mock") {
+				return nil
+			}
```
dmeijboom, line 59: Nitpick: could be an `or` instead

dmeijboom (approved):
LGTM if it works

PR #593, opened 2021-11-08
Fix copyrights

reinkrul:


dmeijboom (approved):
Looks like the first line of the copyright is changed from: "Copyright (C) 2021. Nuts community" to "Copyright (C) 2021 Nuts community".

I don't mind though as long as it's consistent so LGTM

PR #598, opened 2021-11-11
Network: clean up closed inbound connections

reinkrul:
This could prevent a client from reconnecting in certain edge cases (e.g. TCP connection got closed by a network device). Fix is backported from https://github.com/nuts-foundation/nuts-node/pull/571

### DIFF
--- a/network/transport/grpc/connection_manager_test.go
+++ b/network/transport/grpc/connection_manager_test.go
@@ -30,8 +30,10 @@ import (
 	"github.com/stretchr/testify/assert"
 	"google.golang.org/grpc/metadata"
 	"google.golang.org/grpc/peer"
+	"hash/crc32"
 	"net"
 	"testing"
+	"time"
 )
 
 func Test_grpcConnectionManager_Connect(t *testing.T) {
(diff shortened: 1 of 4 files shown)


network/transport/grpc/connection_manager_test.go
```diff
@@ -128,12 +130,12 @@ func Test_grpcConnectionManager_acceptGRPCStream(t *testing.T) {
...
 		cm := NewGRPCConnectionManager(Config{peerID: "server-peer-id"}).(*grpcConnectionManager)
 
-		serverStream := &stubServerStream{clientMetadata: constructMetadata("client-peer-id")}
+		serverStream := newServerStream("client-peer-id")
 		accepted, peerInfo, closer := cm.acceptGRPCStream(serverStream)
 
 		assert.True(t, accepted) // not already connected
 		assert.Equal(t, transport.PeerID("client-peer-id"), peerInfo.ID)
-		assert.Equal(t, "127.0.0.1", peerInfo.Address)
+		assert.Equal(t, "127.0.0.1:9522", peerInfo.Address)
```
woutslakhorst, line 138: better to take a random available port?

woutslakhorst (changes requested):
you may want to fix that port assignment in the tests. It might clash in the future.

Is it ok if one stream dies and the other is still alive? Don't you want to close the entire connection or is it extremely likely the other will be closed short after?

network/transport/grpc/connection_manager_test.go
reinkrul: It doesn't actually open a port, because it's a stub gRPC stream. For the unit test the port is derived from the peer ID (some crc32 + `mod`-ing). Our existing integration tests, however, do take hardcoded ports. We might want to fix that at some point.

woutslakhorst:
> Can't properly test that until we have multiple protocols though, so I'll add it as a TODO for v2, ok?

Ok, create an issue to keep track of things

reinkrul:
Added it as TODO of https://github.com/nuts-foundation/nuts-node/issues/600

woutslakhorst (approved):

PR #599, opened 2021-11-11
Add lamport clock values to V1 transactions

### LINKED ISSUE #451: extremely large `Tx.next`
The `Tx.next` is populated with all TXs that have the Tx listed in the prev header.
Since more TX will now refer to a TX (DID Document updates, VC issuance), the next field can become quite large.
This is only internally a problem. The TX itself only refers to a TX via prev.

### LINKED ISSUE #596: Introduce calculated lamport clock for v1 TXs
V1 transactions will not have a LC value. This will only be introduced with the new V2 protocol on the V2 Tx.
To be backwards compatible with the older V1 transactions, we can already calculate what the values should have been and change storage to accommodate this.

woutslakhorst:
closes #596 
fixes #451

- implemented the logical clock as an index from clock to hashList, but also a reverse index since current transactions do not have a clock value (`lc`) yet
- Simplified the DAG walker by removing the `algo`. Now the clock value is used to walk the DAG.
- Added a migration that will only run once and migrate from `nexts` transactions to clock based ordering.
- This means that the DAG will no longer accept out of order transactions. (network already prevented this)
- Added check in `Walk` that it starts at root when an empty hash is given (according to its description)
- Removed `Root` method from the DAG as it was no longer used.

Also fixed a small bug in the VDR that popped up during testing. It was possible to add duplicate hashes in the sourceTransactions of a DID Document.

- [x] migration of nexts to LC bucket
- [x] tests


network/dag/bbolt_dag.go
```diff
@@ -127,7 +210,7 @@ func (dag bboltDAG) Get(ctx context.Context, ref hash.SHA256Hash) (Transaction,
 	var err error
 	err = bboltTXView(ctx, dag.db, func(ctx context.Context, tx *bbolt.Tx) error {
 		if transactions := tx.Bucket([]byte(transactionsBucket)); transactions != nil {
```
reinkrul, line 212: this `if` has become useless since `getTransaction` performs the same check

network/dag/bbolt_dag.go
```diff
@@ -227,42 +306,53 @@ func (dag *bboltDAG) Add(ctx context.Context, transactions ...Transaction) error
...
-}
 
-func (dag bboltDAG) Root(ctx context.Context) (hash hash.SHA256Hash, err error) {
-	err = bboltTXView(ctx, dag.db, func(ctx context.Context, tx *bbolt.Tx) error {
-		if transactions := tx.Bucket([]byte(transactionsBucket)); transactions != nil {
-			if roots := getRoots(transactions); len(roots) >= 1 {
-				hash = roots[0]
+		// we find the clock value of the TX ref
+		// an empty hash means start at root
+		clockValue, err := getClock(tx, startAt)
```
reinkrul, line 320: What if there are multiple TXs with the same clock value (due to a branch)? I'd expect it to only walk TXs from the given TX, not TXs (and subsequent TXs) from parallel branches with the same LC.

network/dag/bbolt_dag.go
```diff
@@ -227,42 +306,53 @@ func (dag *bboltDAG) Add(ctx context.Context, transactions ...Transaction) error
...
 	return bboltTXView(ctx, dag.db, func(ctx context.Context, tx *bbolt.Tx) error {
 		transactions := tx.Bucket([]byte(transactionsBucket))
-		nexts := tx.Bucket([]byte(nextsBucket))
-		if transactions == nil || nexts == nil {
+		clocksBucket := tx.Bucket([]byte(clockBucket))
+		if transactions == nil {
 			// DAG is empty
 			return nil
 		}
-		return algo.walk(ctx, visitor, startAt, func(hash hash.SHA256Hash) (Transaction, error) {
```
reinkrul, line 238: I believe the `BFSAlgorithmWalker` stopped at TXs that missed payload (correct me if I'm wrong), is that still the case?

network/dag/bbolt_dag.go
```diff
@@ -351,38 +509,32 @@ func getBuckets(tx *bbolt.Tx) (transactions, nexts, payloadIndex, heads *bbolt.B
...
 	if transactionBytes == nil {
 		return nil, nil
 	}
-	transaction, err := ParseTransaction(transactionBytes)
+	parsedTx, err := ParseTransaction(transactionBytes)
 	if err != nil {
 		return nil, fmt.Errorf("unable to parse transaction %s: %w", hash, err)
 	}
-	return transaction, nil
+	castTx := parsedTx.(*transaction)
```
reinkrul, line 534: hmm...

network/dag/bbolt_dag.go
```diff
@@ -227,42 +306,53 @@ func (dag *bboltDAG) Add(ctx context.Context, transactions ...Transaction) error
...
+		// initiate a cursor and start from the given lcValue
+		cursor := clocksBucket.Cursor()
+
+	outer:
+		for _, list := cursor.Seek(clockToBytes(clockValue)); list != nil; _, list = cursor.Next() {
+
+			parsed := parseHashList(list)
+			// according to RFC004, lower byte value refs go first
+			sort.Slice(parsed, func(i, j int) bool {
+				return parsed[i].Compare(parsed[j]) <= 0
```
reinkrul, line 334: untested

reinkrul (changes requested):

network/dag/bbolt_dag.go
woutslakhorst: Yes, the pushback of the resume point is now at the 2 different callback funcs (payload and tx)

network/dag/bbolt_dag.go
woutslakhorst: it's true that a TX can be hit twice, although the `visitedTransactions` hash will block it. Also the callbacks should be resilient (and currently are) to duplicates. I added logic for checking if a TX is processed at this level, but it adds way too much lines of ugliness which is already caught at different points.
I also suspect the `startAt` will be replaced with clock value somewhere down the line anyway since it would simplify a lot.

network/dag/bbolt_dag.go
woutslakhorst: added test

network/dag/bbolt_dag.go
reinkrul: Given that:

- One of the important rules of thumb of the DAG is that we minimize branches (and have them as short as possible), and thus minimize the number of TXs that would be revisited.
- We specify in the contract of `Walk()` that more TXs than just the requested branch can be visited (e.g. if you want to build a visualization from TX `XYZ` you might also get parallel branch `ABC`).
- And thus make sure all consumers can handle consuming TXs multiple times (there was a good term for this, but I forgot).

I think it's probably safe for now.

reinkrul (approved):

reinkrul:
Maybe add some comments on the DAG walker func regarding more TXs than the requested one might be visited in case of branches.

PR #601, opened 2021-11-11
Network: implement v2 skeleton protocol

reinkrul:
This PR adds the v2 gRPC protocol. It contains a single, "Hello World"-style message to prove it works (and to test whether the other side handles v2 messages).


network/transport/v2/protocol.proto
```diff
@@ -0,0 +1,36 @@
...
+ *
+ */
+syntax = "proto3";
+
+option go_package = "github.com/nuts-foundation/nuts-node/network/transport/v2";
+
+package v2;
+
+service Protocol {
+  rpc Stream (stream Message) returns (stream Message) {
```
dmeijboom, line 25: Currently, we're using the method name (in this case `Stream`) in the `grpcOutboundStream`/`grpcInboundStream` maps right? Wouldn't it make more sense to call this something like: `ConnectV2` then?

dmeijboom (changes requested):

network/transport/v2/protocol.proto
reinkrul: I want to stay clear of `Connect` since we have connections and connection managers all over the place. Maybe `Invoke`, `Exchange`?

dmeijboom: Both would be fine but it should have the `V2` suffix right? Otherwise the v3 protocol has to choose a different method name. That would be a mess. Or we should choose a different key for the `grpcOutboundStream` and inbound stream maps.

network/transport/grpc/connection.go
```diff
@@ -119,36 +147,53 @@ func (mc *conn) verifyOrSetPeerID(id transport.PeerID) bool {
...
 
-func (mc *conn) registerClientStream(clientStream grpc.ClientStream) {
+func (mc *conn) registerClientStream(clientStream grpc.ClientStream, method string) error {
 	mc.mux.Lock()
 	defer mc.mux.Unlock()
-	mc.grpcOutboundStreams = append(mc.grpcOutboundStreams, clientStream)
+
+	if mc.ctx == nil {
+		mc.ctx, mc.cancelCtx = context.WithCancel(clientStream.Context())
+	}
```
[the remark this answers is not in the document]
reinkrul: I tried to merge contexts but that became a bit complex, so I changed it a bit so it creates a new background context, which is cancelled when the stream context is done (in a separate goroutine, which was already present in the inbound-stream branch of things).

network/transport/v2/protocol.proto
woutslakhorst: > Both would be fine but it should have the V2 suffix right? Otherwise the v3 protocol has to choose a different method name. That would be a mess. Or we should choose a different key for the grpcOutboundStream and inbound stream maps.

The package name is already `v2`

dmeijboom: > > Both would be fine but it should have the V2 suffix right? Otherwise the v3 protocol has to choose a different method name. That would be a mess. Or we should choose a different key for the grpcOutboundStream and inbound stream maps.
> 
> The package name is already `v2`

I'm aware. That's not what I mean. If you look at this for example: https://github.com/nuts-foundation/nuts-node/pull/601/files#diff-56b8a578980a5f6863c2b3e7dc55ca89c600a67014b54b4287fc7e4dfded819cR154

You'll see that the gRPC method name is used to differentiate different gRPC streams. This means that if a v2 protocol gRPC stream will be opened it will be stored in `grpcOutboundStreams["Stream"]`. For v1 this will be: `grpcOutboundStreams["Connect"]`. This doesn't really make sense to me as it says nothing about the protocol version and assumes new protocol versions always have a different method (which is not in any way a requirement).

network/transport/grpc/connection.go
```diff
@@ -160,24 +211,34 @@ func (mc *conn) open(tlsConfig *tls.Config, connectedCallback func(grpcConn *grp
...
+	defer mc.mux.RUnlock()
+
+	if protocol != AnyProtocol {
+		// Check for specific protocol
+		if mc.grpcOutboundStreams[protocol] != nil {
+			return true
+		}
+		if mc.grpcInboundStreams[protocol] != nil {
+			return true
+		}
```
dmeijboom, line 237: This could panic right? Shouldn't we use `if _, ok := mc.grpcOutboundStreams[protocol]; ok` here?

network/transport/v2/protocol.go
```diff
@@ -0,0 +1,111 @@
...
+	go func() {
+		p.receiveMessages(peer, stream)
+		cancelFn()
+	}()
+
+	return ctx, nil
+}
+
+// Stream is called when a peer opens a stream to the local node (inbound connections).
+func (p protocol) Stream(stream Protocol_StreamServer) error {
```
woutslakhorst, line 71: in v1, this is called `RegisterService`? Why the difference in approaches?

network/transport/v2/protocol.go
```diff
@@ -0,0 +1,111 @@
...
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+// New creates an instance of the v2 protocol.
+func New() transport.Protocol {
+	return &protocol{}
+}
+
+type protocol struct {
```
woutslakhorst, line 20: I agree with the `v2/protocol` naming, but for v1 it's still `v1/ protocolV1`.....

dmeijboom (changes requested):
I tried following all locks/channels/goroutines and contexts but I can't really verify if there won't be any deadlocks or race conditions. Would be nice if we could add some integrations tests were we open/close a lot of gRPC streams.

network/transport/v2/protocol.go
```diff
@@ -0,0 +1,111 @@
+package v2
```
woutslakhorst, line 1: missing copyright.

network/transport/v2/protocol.proto
woutslakhorst: Less generic could be very helpful when doing a global find on the exact term. `stream` would get to many hits.
`InitiateV2ProtocolChannel` (but then in German because they are master in combining words.)

woutslakhorst (changes requested):

network/transport/v2/protocol.proto
reinkrul: The `method` format is `/<protobuf package>.<service>/<call>`, so for v1:

`transport.Network/Connect`

and for v2:

`/v2.Protocol/Stream`

(I've got a feeling that it's also the HTTP/2 URL which is used...)

They need to be unique, so I think it's good enough?

reinkrul:
> I tried following all locks/channels/goroutines and contexts but I can't really verify if there won't be any deadlocks or race conditions. Would be nice if we could add some integrations tests were we open/close a lot of gRPC streams.

We have integration tests in `network_integrations_tests.go` with edge cases (node shuts down/restarts and then needs to reconnect, nodes connect to each other) which I used to solve deadlocks.

The previous implementation with the `closers` (which were complex and all over the place) I couldn't get deadlock-free. The updated implementation with the cancellable context fixed this all and is (imo) much simpler and easier to read. It's basically: 1 context per connection, which is cancelled when one of the gRPC stream contexts are done.

network/transport/v2/protocol.go
reinkrul: For the sake of small PRs, can I (or someone else) fix that in a subsequent PR?

(made an issue: https://github.com/nuts-foundation/nuts-node/issues/628)

network/transport/grpc/connection.go
reinkrul: Reading a non-existing item from a map returns either `nil` (in case of an interface type) or the default value (in case of a value type). `ok` allows you to differentiate with the entry being `nil`/default or non-existent. The value here is an interface, so it just returns nil. Proof:

```go
func Test_t(t *testing.T) {
	println(fmt.Sprintf("%v", map[string]interface{}{"foo": "bar"}["fuu"]))
}
```

```
=== RUN   Test_t
<nil>
--- PASS: Test_t (0.00s)
```

network/transport/v2/protocol.proto
dmeijboom: Check. Wasn't aware that the service name was already present in the method name

network/transport/grpc/connection.go
dmeijboom: Wow. I never knew... Thanks 😅

dmeijboom (approved):

network/transport/v2/protocol.go
reinkrul: Not exactly; `RegisterService` is defined by the gRPC Protocol interface and called to register the gRPC stream implementation (present in both v1 and v2). This `Stream` function is dictated by the RPC call function in the protobuf spec, which is `Stream` for v2. In v1 it's `Connect` which was incredible ambiguous.

woutslakhorst (approved):

PR #611, opened 2021-11-16
VDR: List DIDs managed by the local node

reinkrul:
This allows node operators to retrieve the list of DIDs that are managed by the local node (which can be updated). It works by retrieving the list of private keys in the key store and resolving the DID document (based on KID).

It answers the question, "how do I retrieve the DIDs I previously created?"

It does not return the documents controlled by the node via a `controller`. E.g., a vendor's care organizations, where the care organization DID documents don't contain any key material (available to the vendor). That would be a future improvement.

### DIFF
--- a/crypto/storage/fs.go
+++ b/crypto/storage/fs.go
@@ -23,8 +23,10 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io/fs"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/nuts-foundation/nuts-node/crypto/util"
 )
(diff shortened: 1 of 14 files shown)


crypto/storage/fs.go
```diff
@@ -75,12 +77,26 @@ func NewFileSystemBackend(fspath string) (Storage, error) {
 	return fsc, nil
 }
 
+func (fsc *fileSystemBackend) ListPrivateKeys() []string {
+	var result []string
+	_ = filepath.Walk(fsc.fspath, func(path string, info fs.FileInfo, err error) error {
+		if !info.IsDir() && strings.HasSuffix(info.Name(), string(privateKeyEntry)) {
+			upper := len(info.Name()) - len(privateKeyEntry) - 1
+			if upper < len(info.Name()) {
```
dmeijboom, line 85: If we do this:

`upper := len(info.Name()) - len(privateKeyEntry) - 1`

Shouldn't `upper` always be less than `len(info.Name())` (because of the `- 1` part)?

dmeijboom (approved):

woutslakhorst:
There's a purpose behind this change? Because it only returns the union of `controller` and `capabilityInvocation` . If you vendor controller key is locked in a physical vault, the actual returned set will always be empty.

reinkrul:
> There's a purpose behind this change? Because it only returns the union of `controller` and `capabilityInvocation` . If you vendor controller key is locked in a physical vault, the actual returned set will always be empty.

When the vendor controller key is locked in a physical vault, the node can't use it to manage the vendor DID document anyways so that would be a logical outcome?

woutslakhorst:
> The reason behind this PR is that we need a more user-friendly answer to the question "what are my DIDs?" than replying with "look on disk in data/crypto, find the files ending with _private.json, remove everything after '#', those are your DIDs"?

Then you must support the controller route. Physical vault manages root DID, root DID manages vendor DID, vendor DID manages customer DID. Each step with an decrease in private key rotation time.

reinkrul:
> > The reason behind this PR is that we need a more user-friendly answer to the question "what are my DIDs?" than replying with "look on disk in data/crypto, find the files ending with _private.json, remove everything after '#', those are your DIDs"?
> 
> Then you must support the controller route. Physical vault manages root DID, root DID manages vendor DID, vendor DID manages customer DID. Each step with an decrease in private key rotation time.

I agree we should support this case right from the start. To avoid doing work that has to be redone when we have persisted VDR storage, I'll put this on hold until https://github.com/nuts-foundation/nuts-node/pull/573 has been merged. Then we can more easily create an extra index that can quickly resolve `X is controller of Y` relations.

PR #619, opened 2021-11-22
Ignore transactions with a 'to' header in v1

dmeijboom:


reinkrul:
Missing the following things:

- The check to not query the payload its initially received, see `checkTransactionOnLocalNode` in `network/transport/v1/logic/handlers.go:217`. The `missingPayloadCollector` is only for those edge cases where we received the transaction but couldn't immediately resolve the payload (which can happen in a number of situations).
- The DAG publisher, which is subscribed on by VDR/VCR, stops publishing a branch when it encounters a TX with missing payload. This will be the case when encountering a private VC TX for another organization. So the publisher will need to continue when it encounters a TX with a `to` header. Not sure whether it should publish the TX (without payload) or just skip it, since it will be published (through an event) when the payload is ultimately resolved by `v2`. See `network/dag/publisher.go:132`

dmeijboom:
> Missing the following things:
> 
> * The check to not query the payload its initially received, see `checkTransactionOnLocalNode` in `network/transport/v1/logic/handlers.go:217`. The `missingPayloadCollector` is only for those edge cases where we received the transaction but couldn't immediately resolve the payload (which can happen in a number of situations).

See line 243 ;)

> * The DAG publisher, which is subscribed on by VDR/VCR, stops publishing a branch when it encounters a TX with missing payload. This will be the case when encountering a private VC TX for another organization. So the publisher will need to continue when it encounters a TX with a `to` header. Not sure whether it should publish the TX (without payload) or just skip it, since it will be published (through an event) when the payload is ultimately resolved by `v2`. See `network/dag/publisher.go:132`

Done

> * Maybe an integration test in `network/network_integration_test.go` to assert the behavior stated above?

I'll add unit tests in the right places for each component. Not sure if we should have a separate integration test as well

network/dag/parser_test.go
```diff
@@ -73,6 +73,17 @@ func TestParseTransaction(t *testing.T) {
 		assert.Nil(t, tx)
 		assert.EqualError(t, err, "unable to parse transaction: failed to unmarshal jws message: required field \"signatures\" not present")
 	})
+	t.Run("error - to header has invalid type", func(t *testing.T) {
```
woutslakhorst, line 76: You could also add the `to` header to the `OK` test. Then a 100% coverage score 🚀

network/dag/publisher.go
```diff
@@ -124,11 +124,17 @@ func (s *replayingDAGPublisher) publish(ctx context.Context) {
 }
 
 func (s *replayingDAGPublisher) publishTransaction(ctx context.Context, transaction Transaction) bool {
+	// We need to skip transactions with a to addr header as it should be handled by the v2 protocol
```
woutslakhorst, line 127: Shouldn't this be the place where a future event should be generated?
Is this the place where the `to` header decrypt should be performed? (If for me => check if payload is present)

Just some comments for future committers.

woutslakhorst (changes requested):
Just some questions to answer, some intelligent comments might be included in source.

network/dag/publisher.go
reinkrul: > Shouldn't this be the place where a future event should be generated? Is this the place where the `to` header decrypt should be performed? (If for me => check if payload is present)
> 
> Just some comments for future committers.

I think the `to` header should be decrypted by a listener (Nats/observer/whatever) in protocol v2.

network/dag/parser_test.go
dmeijboom: Done

network/dag/publisher.go
dmeijboom: I guess it *should* be decrypted here right? If that succeeds it can be pushed to an event listener in the v2 protocol.

woutslakhorst: so that will be a broader discussion ;)

woutslakhorst (approved):

reinkrul (approved):
LGTM

Issue #623, opened 2021-11-23
Network: allow node DID to be configured


reinkrul:
Must be configurable. If set, try to resolve it (boot time) and fail if it doesn't exist or is deactivated.

woutslakhorst:
an auto-detect would be really nice

reinkrul:
It is, that's why I made an issue for it so we don't forget about it: https://github.com/nuts-foundation/nuts-node/issues/621

PR #624, opened 2021-11-23
Network: configure NodeDID and send it as header

reinkrul:
Fixes https://github.com/nuts-foundation/nuts-node/issues/623


dmeijboom:
I'm not sure if we should even add a setting for this as it should be calculated based on your DID documents.

network/network.go
```diff
@@ -180,6 +195,14 @@ func (n *Network) Start() error {
 		return err
 	}
 
+	// Sanity check for configured node DID: can we resolve it?
+	if n.configuredNodeDID != nil {
```
woutslakhorst, line 199: You should also check for the existence of the private key!

woutslakhorst (changes requested):

woutslakhorst:
> I'm not sure if we should even add a setting for this as it should be calculated based on your DID documents.

There might be multiple available because you made an error somewhere along the line (100% change in workshops), being able to select the right one could fix that. Agree that in 99% of the cases an autoselect should work.

reinkrul:
> 

I think in a controlled production environment you want to set it explicitly. Since it's something you do once and it'll probably never change (if you manage your keys correctly). And you don't want it to accidentally change when someone creates a new DID on (one of) your node.

network/network.go
reinkrul: The question is... which one? We haven't spec'd that yet.

I think we should introduce a DID `keyAgreement` key here (and in the spec), because it will be used to decrypt the `to` header. We should probably check here whether the first `keyAgreement` key of the DID document is present?

dmeijboom (approved):

network/network.go
reinkrul: @woutslakhorst I added these checks

network/network.go
```diff
@@ -172,20 +188,38 @@ func (n *Network) Config() interface{} {
...
+		doc, _, err := n.didDocumentResolver.Resolve(*n.configuredNodeDID, nil)
+		if err != nil {
+			return fmt.Errorf("invalid NodeDID configuration: DID document can't be resolved (did=%s): %w", n.configuredNodeDID, err)
+		}
+		if len(doc.KeyAgreement) == 0 {
+			return fmt.Errorf("invalid NodeDID configuration: DID document does not contain a keyAgreement key (did=%s)", n.configuredNodeDID)
+		}
+		kid := doc.KeyAgreement[0].ID.String()
```
dmeijboom, line 208: I'm wondering. Does it *have* to be the first `KeyAgreement` or can it also be the second one for example?

dmeijboom (changes requested):
One question

network/network.go
reinkrul: It now checks all.

dmeijboom (approved):
LGTM 💯

Issue #627, opened 2021-11-25
VDR: Create keyAgreement key by default


reinkrul:
For decrypting transaction `to` headers (introduced in protocol v2 for exchanging private credentials) there needs to be an encryption key in vendor DID documents. This will be a `keyAgreement` verification relationship (see https://www.w3.org/TR/did-core/#key-agreement).

To make the setup easier, this key should be created by default when creating new DID documents.

Since care organizations probably don't need this key, I think we should also alter the "getting started" manual (https://nuts-node.readthedocs.io/en/latest/pages/getting-started/4-connecting-crm.html), such that no `keyAgreement` keys are created by default for care organizations, when following the manual (open for debate, though).

A really nice to have would be a VDR CLI command to generate and add a `keyAgreement` key to existing DID documents, to help parties who already have set up their vendor DID documents on the dev network.

PR #630, opened 2021-11-25
Enable KeyAgreement by default on DID document creation

dmeijboom:
* Only when the DID document is owned by the node

Fixes https://github.com/nuts-foundation/nuts-node/issues/627

### DIFF
--- a/vdr/api/v1/api.go
+++ b/vdr/api/v1/api.go
@@ -144,7 +144,7 @@ func (a Wrapper) CreateDID(ctx echo.Context) error {
 		return err
 	}
 
-	// this API returns a DIDDocument according to spec so it may return the business object
+	// this API returns a DIDDocument according to spec, so it may return the business object
 	return ctx.JSON(http.StatusOK, *doc)
 }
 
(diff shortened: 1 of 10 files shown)


reinkrul (approved):

reinkrul:
One minor comment is that the default isn't really a default now, but a derivative, which might not be the expected behaviour of someone creating a DID document.

vdr/api/v1/api.go
```diff
@@ -119,6 +119,11 @@ func (a Wrapper) CreateDID(ctx echo.Context) error {
 		}
 	}
 
+	// Enable `KeyAgreement` by default only when it's a DID document owned by the node.
+	if req.KeyAgreement == nil && (req.Controllers == nil || len(*req.Controllers) == 0) {
```
woutslakhorst, line 123: not true, the DID Document can be owned by an offline key. I don't see any objections in being it enabled by default for all DID Documents created

woutslakhorst (changes requested):

vdr/api/v1/api.go
dmeijboom: All right. Then let's do that

woutslakhorst (changes requested):
Ok for the code, but:

- docs/_static/vdr/v1.yaml#DIDCreateRequest doesn't reflect the updted default value
- docs/getting-started/4-connecting-crm.rst has outdated examples

woutslakhorst (approved):

woutslakhorst:
@dmeijboom due to default changes a test now fails.

PR #632, opened 2021-11-25
Network: connect to bootstrap nodes during start instead of configuration phase

reinkrul:
Currently, the node starts connecting to bootstrap nodes while still configuring and starting, which causes DAG sync'ing because the node looks out of date, while the DAG is still loading.

It needs to wait to start connecting to nodes until all other services are started.


dmeijboom (approved):
It's fine to change the order anyway but I'm wondering if it will actually make a difference. After running `.Start()` we're still not sure if it's actually ready right?

woutslakhorst (approved):

reinkrul:
> It's fine to change the order anyway but I'm wondering if it will actually make a difference. After running `.Start()` we're still not sure if it's actually ready right?

Loading the DAG is blocking/sync, so it's actually ready.

PR #636, opened 2021-11-26
Network: reset peer ID when connection closes

reinkrul:
Based on https://github.com/nuts-foundation/nuts-node/pull/635 (rebase first).

Currently, the peer ID is never reset, while it is randomly generated when a node starts. That way, when a bootstrap node restarts, its peers aren't able to reconnect (because it sends a new peer ID after the restart). Peer ID should be reset/emptied when a peer disconnects, then it can be set again when it reconnects.


woutslakhorst (approved):
seems logical and simple. Let's see what it does in practise ;)

PR #643, opened 2021-11-30
update to go-leia/v2

### LINKED ISSUE #612: Searching for auth creds with only credentialSubject.id doesn't give any results for creds with missing credentialSubject.subject
`credentialSubject.id` is part of one of the indices for `NutsAuthorizationCredential`. We have a case where it does not give a result. We're sure it is present and can be resolved by other queries.

Faulty data is available. Will have to investigate if this is the case with all NutsAuthorizatoinCredentials

woutslakhorst:
fixes #612 

a migration is added that removes the current VCR DB. This is allowed as long as the DAG is walked at startup. When that stops the migration must be removed.

- [ ] create an issue after merge for removing migration

### DIFF
--- a/network/network.go
+++ b/network/network.go
@@ -186,7 +186,6 @@ func (n *Network) Config() interface{} {
 
 // Start initiates the Network subsystem
 func (n *Network) Start() error {
-	log.Logger().Debugf("starting %s", ModuleName)
 	n.startTime.Store(time.Now())
 
 	// Load DAG and start publishing
@@ -229,8 +228,6 @@ func (n *Network) Start() error {
 		}
 		n.connectionManager.Connect(bootstrapNode)
 	}
-
-	log.Logger().Infof("started %s", ModuleName)
 
 	return nil
 }
--- a/network/network.go
+++ b/network/network.go
@@ -22,12 +22,6 @@ import (
 	"context"
 	"crypto/tls"
 	"fmt"
-	"github.com/nuts-foundation/go-did/did"
-	"github.com/nuts-foundation/nuts-node/network/transport"
-	"github.com/nuts-foundation/nuts-node/network/transport/grpc"
-	"github.com/nuts-foundation/nuts-node/network/transport/v1"
-	v2 "github.com/nuts-foundation/nuts-node/network/transport/v2"
-	"github.com/pkg/errors"
 	"os"
 	"path"
 	"path/filepath"
@@ -35,6 +29,13 @@ import (
 	"sync"
 	"sync/atomic"
 	"time"
+
+	"github.com/nuts-foundation/go-did/did"
+	"github.com/nuts-foundation/nuts-node/network/transport"
+	"github.com/nuts-foundation/nuts-node/network/transport/grpc"
+	"github.com/nuts-foundation/nuts-node/network/transport/v1"
+	v2 "github.com/nuts-foundation/nuts-node/network/transport/v2"
+	"github.com/pkg/errors"
 
 	"github.com/google/uuid"
 	"github.com/nuts-foundation/nuts-node/core"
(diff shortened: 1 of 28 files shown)


network/network.go
```diff
@@ -230,6 +230,8 @@ func (n *Network) Start() error {
 		n.connectionManager.Connect(bootstrapNode)
 	}
 
+	log.Logger().Infof("started %s", ModuleName)
```
reinkrul, line 233: intentional, of leftover debugging?

If we want this logged for all engines, we'd better do it in `System`

vcr/concept/phonetics_test.go
```diff
@@ -22,7 +22,7 @@ package concept
 import (
 	"testing"
 
-	"github.com/nuts-foundation/go-leia"
+	"github.com/nuts-foundation/go-leia/v2"
```
reinkrul, line 25: why the v2 package?

vcr/vcr.go
```diff
@@ -406,7 +436,7 @@ func (c *vcr) find(ID ssi.URI) (vc.VerifiableCredential, error) {
 	q := leia.New(qp)
 
 	for _, t := range c.registry.Concepts() {
-		docs, err := c.store.Collection(t.CredentialType).Find(q)
+		docs, err := c.store.Collection(t.CredentialType).Find(context.TODO(), q)
```
reinkrul, line 439: TODO context?

vcr/vcr.go
```diff
@@ -91,9 +97,8 @@ func (c *vcr) Configure(config core.ServerConfig) error {
 	var err error
 
 	// store strictMode
```
reinkrul, line 99: now useless and incorrect comment

vcr/vcr.go
```diff
@@ -720,7 +750,7 @@ func (c *vcr) isRevoked(ID ssi.URI) (bool, error) {
 	q := leia.New(qp)
 
 	gIndex := c.revocationIndex()
-	docs, err := gIndex.Find(q)
+	docs, err := gIndex.Find(context.TODO(), q)
```
reinkrul, line 753: TODO

reinkrul (changes requested):

vcr/concept/phonetics_test.go
woutslakhorst: Not backwards compatible with v1, then go dictates to add the version number after the package

vcr/vcr.go
woutslakhorst: Yes, to start a discussion, do we want unlimited time for a find by ID, or do we want to limit it tox seconds?

vcr/vcr.go
woutslakhorst: same discussion

vcr/vcr.go
reinkrul: if we expect calls to be returned within X seconds (e.g. 1s), I'd limit it at that.

Would be odd though, if it were to take a very long time? Would indicate a bug in storage, I'd say.

woutslakhorst: Could be, then it would be best to fail fast.

woutslakhorst:
@reinkrul replaced `context.TODO()` with a timeout context.

vcr/vcr.go
reinkrul: The limit is arbitrary, and 1s feels awfully long waiting for a fast memory-mapped store to find something. On the other side, if we hardcode it here and it occurs in a production scenario (and doesn't auto-recover), it's really blocking until we release a patched version, which is also isn't very nice..

woutslakhorst: APIs also have a timeout, so the behaviour would be sort of the same.

reinkrul (approved):

Issue #564, opened 2021-10-25
VCR: Issue credentials privately


woutslakhorst:
We need a way to detect when a credential should be private. Since the basics of credentials is that they are issued to the holder, all credentials are private unless stated otherwise.

We'll add a `public` param to the credential config. If a credential is public, it'll be added to the DAG without `to` header. Otherwise the `to` header will contain the (encrypted) node DID of the receiver, making it private.

woutslakhorst:
already fixed

PR #647, opened 2021-12-02
Network: create private TXs on the DAG with encrypted `pal` header

reinkrul:
Required by https://github.com/nuts-foundation/nuts-node/issues/564


network/dag/pal.go
```diff
@@ -0,0 +1,88 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("unable to resolve keyAgreement key (recipient=%s): %w", recipient, err)
+		}
+		kak, ok := rawKak.(*ecdsa.PublicKey)
+		if !ok {
+			return nil, fmt.Errorf("resolved keyAgreement key is not an elliptic curve key (recipient=%s)", recipient)
+		}
+		encryptionKeys = append(encryptionKeys, kak)
+	}
+	plaintext := []byte(strings.Join(recipientStrs, "\n"))
```
dmeijboom, line 32: Nitpick: if we want a byte slice anyway we could use `recipientStrs` as a `[]byte` and use `bytes.Join` here

network/tx.go
```diff
@@ -0,0 +1,53 @@
...
+func NewTransaction(payloadType string, payload []byte, key crypto.Key) TransactionBuilder {
+	return TransactionBuilder{
+		PayloadType: payloadType,
+		Payload:     payload,
+		Key:         key,
+	}
+}
+
+// TransactionBuilder is used to build a spec for new transactions.
+type TransactionBuilder struct {
```
dmeijboom, line 20: +1

vdr/doc/resolvers.go
```diff
@@ -190,6 +190,19 @@ func (r KeyResolver) ResolveAssertionKeyID(id did.DID) (ssi.URI, error) {
...
+// If the document has no KeyAgreements, types.ErrKeyNotFound is returned.
+func (r KeyResolver) ResolveKeyAgreementKey(id did.DID) (crypto.PublicKey, error) {
+	doc, _, err := r.Store.Resolve(id, nil)
+	if err != nil {
+		return ssi.URI{}, err
+	}
+	if len(doc.KeyAgreement) == 0 {
+		return nil, types.ErrKeyNotFound
+	}
+	return doc.KeyAgreement[0].PublicKey()
```
dmeijboom, line 203: I think I've mentioned this before. In theory a DID could contain multiple public keys that can be used for key agreements. I'm fine with picking the first one but afaic it should be in the RFC

dmeijboom (approved):

vdr/doc/resolvers.go
reinkrul: Will do

network/dag/pal.go
```diff
@@ -0,0 +1,89 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("unable to resolve keyAgreement key (recipient=%s): %w", recipient, err)
+		}
+		kak, ok := rawKak.(*ecdsa.PublicKey)
+		if !ok {
+			return nil, fmt.Errorf("resolved keyAgreement key is not an elliptic curve key (recipient=%s)", recipient)
+		}
+		encryptionKeys = append(encryptionKeys, kak)
+	}
+	plaintext := bytes.Join(recipients, []byte("\n"))
```
woutslakhorst, line 33: extract to const

network/dag/pal.go
```diff
@@ -0,0 +1,89 @@
...
+		return nil, nil
+	}
+
+	var participants []did.DID
+	for _, curr := range strings.Split(string(decrypted), "\n") {
+		participant, err := did.ParseDID(curr)
+		if err != nil {
+			return nil, fmt.Errorf("invalid participant (did=%s): %w", curr, err)
+		}
+		participants = append(participants, *participant)
```
woutslakhorst, line 85: you may want to deduplicate

network/dag/transaction.go
```diff
@@ -114,7 +117,8 @@ type Transaction interface {
 
 // NewTransaction creates a new unsigned transaction. Parameters payload and payloadType can't be empty, but prevs is optional.
 // Prevs must not contain empty or invalid hashes. Duplicate prevs will be removed when given.
-func NewTransaction(payload hash.SHA256Hash, payloadType string, prevs []hash.SHA256Hash) (UnsignedTransaction, error) {
+// The pal byte slice (may be nil) holds the encrypted recipient address, if it is a private transaction.
+func NewTransaction(payload hash.SHA256Hash, payloadType string, prevs []hash.SHA256Hash, pal [][]byte) (UnsignedTransaction, error) {
```
woutslakhorst, line 121: custom type for `[][]byte` and some convenience methods to transform []did to that?

network/tx.go
```diff
@@ -0,0 +1,53 @@
...
+
+import (
+	"github.com/nuts-foundation/go-did/did"
+	"github.com/nuts-foundation/nuts-node/crypto"
+	"github.com/nuts-foundation/nuts-node/crypto/hash"
+	"time"
+)
+
+// NewTransaction creates a new TransactionBuilder with the given required properties.
+func NewTransaction(payloadType string, payload []byte, key crypto.Key) TransactionBuilder {
```
woutslakhorst, line 11: No need for this method, struct can be used directly.  This name is already in use by the DAG anyway.

woutslakhorst: unless you want to enforce a minimum use. But then it would also require an interface and private struct

vcr/credential/builder.go
```diff
@@ -42,7 +42,8 @@ type defaultBuilder struct {
 	vcType string
 }
 
-var nowFunc = time.Now
+// NowFunc is used to store a function that returns the current time. This can be changed when you want to mock the current time.
+var NowFunc = time.Now
```
woutslakhorst, line 46: I'm guessing this is because some test from another package needs it. Still not very nice to expose something only changed for test...

woutslakhorst (changes requested):

vcr/credential/builder.go
reinkrul: Hmm I thought I reverted this change, was not meant to be changed after all.

network/tx.go
reinkrul: It started with Codeclimate (rightly) complaining about the number of arguments of `CreateTransaction`. I considered the Options pattern, but that became ugly because it's not mapped to a single struct (but rather plucked apart and used as arguments for several underlying function calls). So I chose a sort of Builder pattern, but without the actual `Build()` step (and as you noticed, protected inners).

Reason is it can't build a `Transaction`, since that required multiple calls (new TX, sign it) and checks performed using dependencies (DAG to check the presence of all specified `prev`s, fetching DAG heads).  So I can't have a `TransactionBuilder` building a `Transaction` (or we could, but would require refactoring the complete transaction creation layer).

Most pure application of the pattern would be introducing a `TransactionSpec` (like a "blueprint") and having a `TransactionSpecBuilder`, which's `Build()` function would yield an immutable `TransactionSpec`. But that feels severly overengineered to me.

Would it be better if I unexported the fields of `TransactionBuilder` and renamed it to `TransactionTemplate`? Because it's not a true application of the Builder pattern, and more like a template with convenience methods.

woutslakhorst: I was more triggered by the `New` in `NewTransaction` since that pattern usually returns an interface to an underlying (pointer to a) struct. You could just do: `TransactionTemplate{A, B, C}`? We have so many `New` and `Create` methods all over the place, that those start to lose meaning.

reinkrul: I agree. Will remove the "New" method.

woutslakhorst (approved):

PR #655, opened 2021-12-03
Add basic vault storage backend

stevenvegt:
Uses the Vault KV version 1 secret store and token authentication.

Note: The public key methods are removed from the storage interface since they were no longer in use.

Note: The keys are stored in the KV store and not in the `transit` store since encryption for the curve algorithms is currently not supported.


crypto/crypto_test.go
```diff
@@ -121,11 +121,18 @@ func TestCrypto_Configure(t *testing.T) {
 		storageType := reflect.TypeOf(client.Storage).String()
 		assert.Equal(t, "*storage.fileSystemBackend", storageType)
 	})
+	t.Run("error - no backend in strict mode is now allowd", func(t *testing.T) {
```
dmeijboom, line 124: Nitpick: maybe 'error - backend must be explicitly set in strict mode' as it's really confusing to read like this 😅

crypto/storage/vault.go
```diff
@@ -0,0 +1,118 @@
...
+// If token is empty, the VAULT_TOKEN environment should be is set.
+func NewVaultKVStorage(token string, vaultAddr string) (Storage, error) {
+	client, err := configureVaultClient(token, vaultAddr)
+	if err != nil {
+		return nil, err
+	}
+
+	vaultStorage := vaultKVStorage{client: client.Logical()}
+	err = vaultStorage.checkConnection()
+	return vaultStorage, err
```
dmeijboom, line 35: I'm not a fan of returning errors like this. I'd rather see it like this for example:


```
if err := vaultStorage.checkConnection(); err != nil {
    return nil, err
}

return vaultStorage, nil
```

Otherwise it's hard to see wether the error was actually checked

crypto/storage/vault.go
```diff
@@ -0,0 +1,118 @@
...
+		return fmt.Errorf("unable to connect to Vault: unable to retrieve token status: %w", err)
+	}
+	if len(secret.Data) == 0 {
+		return fmt.Errorf("could not read token information on auth/token/lookup-self")
+	}
+	return nil
+}
+
+func (v vaultKVStorage) GetPrivateKey(kid string) (crypto.Signer, error) {
+	path := fmt.Sprintf("%s/%s/%s", kvEnginePath, privateKeyPath, kid)
```
dmeijboom, line 65: We need to make sure `kid` doesn't contain slashes otherwise someone would be able to read arbitrary data from the Vault key value store.

dmeijboom: (if it contains `../` for example)

dmeijboom: Same for `PrivateKeyExists` and `SavePrivateKey`

dmeijboom (changes requested):
Nice

crypto/storage/fs.go
```diff
@@ -93,21 +91,7 @@ func (fsc *fileSystemBackend) GetPrivateKey(kid string) (crypto.Signer, error) {
...
-		return publicKeyEntry, err
-	}
-
-	err = json.Unmarshal(data, &publicKeyEntry)
-
-	return publicKeyEntry, err
-}
-
-// Save the private key for the given key to disk. Files are  postfixed with '_private.pem'. Keys are stored in pem format. It also store the public key
+// SavePrivateKey saves the private key for the given key to disk. Files are  postfixed with '_private.pem'. Keys are stored in pem format. It also store the public key
```
woutslakhorst, line 94: It doesn't save the public key anymore...

crypto/crypto_test.go
```diff
@@ -121,11 +121,18 @@ func TestCrypto_Configure(t *testing.T) {
...
+		cfg.Strictmode = true
+		err := client.Configure(cfg)
+		assert.EqualErrorf(t, err, "you must explicitly provide a crypto storage backend in strict-mode", "expected error")
+	})
 	t.Run("error - unknown backend", func(t *testing.T) {
 		client := createCrypto(t)
 		client.config.Storage = "unknown"
 		err := client.Configure(cfg)
-		assert.EqualErrorf(t, err, "only fs backend available for now", "expected error")
+		assert.EqualErrorf(t, err, "invalid config for crypto.storage. Available options are: vaultkv, fs", "expected error")
```
woutslakhorst, line 135: Not your fault, but it should actually be `EqualError` without the formatting.

crypto/storage/vault.go
```diff
@@ -0,0 +1,118 @@
...
+
+type vaultKVStorage struct {
+	client logicaler
+}
+
+// NewVaultKVStorage creates a new Vault backend using the kv version 1 secret engine: https://www.vaultproject.io/docs/secrets/kv
+// It currently only supports token authentication which should be provided by the token param.
+// If vaultAddr is empty, the VAULT_ADDR environment should be set.
+// If token is empty, the VAULT_TOKEN environment should be is set.
+func NewVaultKVStorage(token string, vaultAddr string) (Storage, error) {
```
woutslakhorst, line 27: Untested.

crypto/storage/vault.go
```diff
@@ -0,0 +1,118 @@
...
+	Write(path string, data map[string]interface{}) (*vault.Secret, error)
+}
+
+type vaultKVStorage struct {
+	client logicaler
+}
+
+// NewVaultKVStorage creates a new Vault backend using the kv version 1 secret engine: https://www.vaultproject.io/docs/secrets/kv
+// It currently only supports token authentication which should be provided by the token param.
+// If vaultAddr is empty, the VAULT_ADDR environment should be set.
```
woutslakhorst, line 25: This dual possibility of configuring is confusing. VAULT_ADDR and NUTS_CRYPTO_ADDR are the same? Would be clearer if only NUTS_ type config is used?

Or is this the result of default vault params injection in cloud environments? if so this should deserve a mention in the docs.

crypto/storage/vault.go
```diff
@@ -0,0 +1,118 @@
...
+	if vaultAddr != "" {
+		if err = client.SetAddress(vaultAddr); err != nil {
+			return nil, fmt.Errorf("vault address invalid: %w", err)
+		}
+	}
+	return client, nil
+}
+
+func (v vaultKVStorage) checkConnection() error {
+	secret, err := v.client.Read("auth/token/lookup-self")
```
woutslakhorst, line 54: some doc/link on why this particular call

crypto/storage/vault.go
woutslakhorst: Also: how does Vault handle the key fingerprint after the `#`? Is it ignored or part of the identifier?

woutslakhorst (changes requested):

crypto/storage/vault.go
stevenvegt: Yes, this desirable behavior since Nomad and other orchestrator set this env var. It is mentioned in the production configuration docs.

crypto/storage/vault.go
stevenvegt: Tested, the part after the `#` or _fragment_ is not ignored and returns different results. Also, I have added some path sanitation so the “dot-dot-slash” aka “directory traversal” attack is mitigated.

dmeijboom (approved):
LGTM

woutslakhorst (approved):

PR #656, opened 2021-12-06
Publish transaction with 'to' header to the v2 private transaction stream

dmeijboom:
This is the first part of the code of processing private transactions (or actually, transactions with private payloads). It detects transactions with an encrypted 'to' header and sends them to the event stream.


dmeijboom:
@woutslakhorst This PR contains code from two branches. It also seems like it contains some stuff that shouldn't be there anymore. I'll make sure those parts are removed and then you can check again

crypto/interface.go
```diff
@@ -67,4 +74,6 @@ type Key interface {
 	KID() string
 	// Public returns the public key. This is a short-hand for Signer().Public()
 	Public() crypto.PublicKey
+	// Private returns a crypto.PrivateKey.
+	Private() crypto.PrivateKey
```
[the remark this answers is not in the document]
dmeijboom: We need access to the private key for Ecies: https://github.com/ecies/go/blob/master/ecies.go#L61

crypto/interface.go
```diff
@@ -47,11 +47,18 @@ type KeyResolver interface {
...
 // KeyStore defines the functions for working with private keys.
 type KeyStore interface {
+	ECIES
 	KeyCreator
 	KeyResolver
 	JWTSigner
 }
 
+// ECIES is the interface to support Elliptic Curve Integrated Encryption Scheme
+type ECIES interface {
```
stevenvegt, line 57: Feels like this is a more generic Encryption interface?

crypto/storage/storage.go
```diff
@@ -32,11 +32,16 @@ import (
 // ErrNotFound indicates that the specified crypto storage entry couldn't be found.
 var ErrNotFound = errors.New("entry not found")
 
+type PrivateKey interface {
```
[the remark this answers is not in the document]
dmeijboom: Because we have to be able to cast it to `*ecdsa.PrivateKey`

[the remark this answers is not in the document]
stevenvegt: We made the assumption that a private key would only used for signing, so we created the `crypto.Signer` interface. But now we can also use it for decrypting. Perhaps we must rename it to something like "crypto.PrivateCryptoOperations" and add the `Decrypt` method to this interface?

reinkrul:
We need an issue to remind ourselves to support encryption/decryption for all supported (per-spec) key types

crypto/storage/storage.go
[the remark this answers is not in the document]
dmeijboom: I removed it as you can simply cast a `crypto.Signer` to `*ecdsa.PrivateKey` for instance. We can't create a proper interface for this as a private key could theoretically be any type (depends on the encryption method used)

crypto/ecies.go
```diff
@@ -0,0 +1,32 @@
+package crypto
+
+import (
+	"crypto/ecdsa"
+
+	ecies "github.com/ecies/go"
+)
+
+// EciesDecrypt decrypts the `cipherText` using the Elliptic Curve Integrated Encryption Scheme
+func EciesDecrypt(privateKey *ecdsa.PrivateKey, cipherText []byte) ([]byte, error) {
```
stevenvegt, line 10: Can you make this implement the `crypto.Decrypter` interface?

network/dag/publisher.go
```diff
@@ -123,9 +131,24 @@ func (s *replayingDAGPublisher) publish(ctx context.Context) {
 	}
 }
 
+func (s *replayingDAGPublisher) handlePrivateTransaction(tx Transaction) error {
+	body, err := json.Marshal(tx)
```
reinkrul, line 135: you could use `tx.Data()`, which contains the original TX data as byte slice (and does not return an error)

network/network.go
```diff
@@ -115,8 +117,21 @@ func (n *Network) Configure(config core.ServerConfig) error {
 		return fmt.Errorf("unable to migrate DAG: %w", err)
 	}
 
+	// NATS
+	natsAddr := fmt.Sprintf("%s:%d", n.config.NatsHostname, n.config.NatsPort)
```
reinkrul, line 121: maybe make this a function on a `NatsConfig` struct

network/network.go
```diff
@@ -115,8 +117,21 @@ func (n *Network) Configure(config core.ServerConfig) error {
...
+
+	conn, err := nats.Connect(natsAddr, nats.RetryOnFailedConnect(true), nats.Timeout(time.Duration(n.config.NatsTimeout)*time.Second))
```
reinkrul, line 123: should this be done here, or in some other engine (I believe we have the 'event bus' engine?)

crypto/ecies.go
dmeijboom: I did this on purpose as `EciesEncrypt`/`EciesDecrypt` are generic functions to encrypt/decrypt using ECIES with the standard Go structures for `crypto/ecdsa` private and public keys. The `store.Decrypt` method is the actual implementation of `crypto.Decrypter`

README.rst
```diff
@@ -175,6 +175,7 @@ network.enabletls                          true              Whether to enable T
 network.grpcaddr                           \:5555             Local address for gRPC to listen on. If empty the gRPC server won't be started and other nodes will not be able to connect to this node (outbound connections can still be made).
 network.nodedid                                              Specifies the DID of the organization that operates this node, typically a vendor for EPD software. It is used to identify the node on the network. If the DID document does not exist of is deactivated, the node will not start.
 network.truststorefile                                       PEM file containing the trusted CA certificates for authenticating remote gRPC servers.
+network.natstimeout                        30                Timeout for NATS connections (in seconds)
```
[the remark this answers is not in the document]
dmeijboom: See the events engine (the other configuration is re-used from the events engine)

network/network.go
dmeijboom: You can actually use an external NATS instance if you want. This is only the client NATS connection so in my opinion it should definitely be here

woutslakhorst: there is a `events.Connect` convenience method (with different args though) that could be put back in action. Could also be a nice coupling the the NatsConfig struct mentioned earlier?

crypto/ecies.go
```diff
@@ -0,0 +1,32 @@
...
+			Y:     privateKey.Y,
+		},
+		D: privateKey.D,
+	}
+
+	return ecies.Decrypt(eciesKey, cipherText)
+}
+
+// EciesEncrypt encrypts the `plainText` using the Elliptic Curve Integrated Encryption Scheme
+func (client *Crypto) EciesEncrypt(publicKey *ecdsa.PublicKey, plainText []byte) ([]byte, error) {
```
woutslakhorst, line 24: client receiver not used.

crypto/decryptor.go
```diff
@@ -0,0 +1,21 @@
+package crypto
+
+import (
+	"crypto/ecdsa"
+	"errors"
+)
+
+// Decrypt decrypts the `cipherText` with key `kid`
+func (client *Crypto) Decrypt(kid string, cipherText []byte) ([]byte, error) {
+	key, err := client.Storage.GetPrivateKey(kid)
```
woutslakhorst, line 10: missing test coverage

crypto/storage/fs_test.go
```diff
@@ -126,3 +128,31 @@ func Test_fs_KeyExistsFor(t *testing.T) {
 		assert.True(t, storage.PrivateKeyExists(kid))
 	})
 }
+
+func Test_fs_ListPrivateKeys(t *testing.T) {
```
woutslakhorst, line 132: Not testing ListPrivateKeys (which is gone)

crypto/ecies.go
```diff
@@ -0,0 +1,32 @@
+package crypto
```
woutslakhorst, line 1: missing copyright
missing test

woutslakhorst (changes requested):

crypto/ecies.go
stevenvegt: I mean the golang `crypto.Decrypter` which has the following signature `Decrypt(rand io.Reader, msg []byte, opts crypto.DecrypterOpts) (plaintext []byte, err error)`

>// Decrypter is an interface for an opaque private key that can be used for
// asymmetric decryption operations. An example would be an RSA key
// kept in a hardware module.

network/network.go
dmeijboom: That's true, will do 👍

README.rst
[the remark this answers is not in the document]
dmeijboom: nvm, fixed

crypto/ecies.go
dmeijboom: Ah okay, none of the ECIES libraries are compatible with `crypto.Decrypter` unfortunately (they require access to the actual private key instead)

dmeijboom: We could implement `crypto.Decrypter` for our crypto store but it doesn't really matter for the `EciesDecrypt` function as that's only an helper for ECIES encrypt/decrypt.

reinkrul (approved):

go.mod
```diff
@@ -41,9 +41,12 @@ require (
 	schneider.vip/problem v1.6.0
 )
 
+require github.com/ethereum/go-ethereum v1.10.13
```
reinkrul, line 44: wait, what?

reinkrul (changes requested):

go.mod
dmeijboom: Surprise! In all seriousness, I'll try to replace it with another library. If that fails we're going to create a repo with just the `crypto/ecies` package of Ethereum.

woutslakhorst (approved):

PR #660, opened 2021-12-06
Network: improve logging and backoff in case Node DID authentication fails

reinkrul:
Changes:

- When a gRPC stream fails to open (e.g. protocol v1 or v2), log this as a protocol-level error, rather than connection level failure.
- Outbound connections: don't try other protocols when node DID of the peer fails (optimization, declutters log)
- Outbound connections: backoff when node DID authentication fails


woutslakhorst:
This might just fix #664 ?

woutslakhorst (approved):

reinkrul:
> This might just fix #664 ?

It should not  :-P

PR #661, opened 2021-12-07
Set min TLS version to 1.3

dmeijboom:
When using `tls.Config` be sure to set the `MinVersion` because:

>	// If zero, TLS 1.0 is currently taken as the minimum.

Which is obviously not what we want.


woutslakhorst (approved):

woutslakhorst:
@dmeijboom e2e tests are now broken, apparently our nginx conf doesn't do 1.3 out of the box?

dmeijboom:
> @dmeijboom e2e tests are now broken, apparently our nginx conf doesn't do 1.3 out of the box?

I'll fix the E2E tests after merging

Issue #666, opened 2021-12-10
Network: v2 protocol


woutslakhorst:
This is a tracking issue

Implement v2 protocol according to RFC017

- [x] protobuf spec #669
- [x] TX Ref XOR binary tree #863
- [x] TX IBLT binary tree #863
- [x] conversation ID, issue + check #819 
- [x] State message #953
- [x] TransactionListQuery message #832 
- [x] TransactionRangeQuery message #950
- [x] TransactionPayloadQuery message
- [x] TransactionSet message #953
- [x] TransactionList message #832 
- [x] TransactionPayload message
- [x] attached payload validation for DID Documents
- [x] Gossip message (Hashes) #816 
- [x] Gossip message (XOR, LC) #951
- [x] Track created TXs for Gossip message

woutslakhorst:
All done

PR #669, opened 2021-12-10
added set reconciliation protocol to V2 protobuf spec

woutslakhorst:
part of #666 

implements protobuf spec of nuts-foundation/nuts-specification#159

- [ ] add latest changes from spec


reinkrul (changes requested):
2 things:

- I think conversationID should be in `Message`, so it can be used for any message type
- we're now using a mix of camelCase and UNDERSCORE_CASE naming, which looks sloppy

Also need to rebase with master.

network/transport/v2/protocol.proto
```diff
@@ -22,31 +22,123 @@ option go_package = "github.com/nuts-foundation/nuts-node/network/transport/v2";
 package v2;
 
 service Protocol {
-  rpc Stream (stream Envelope) returns (stream Envelope) {
+  rpc Stream (stream Message) returns (stream Message) {
```
reinkrul, line 25: Why change it back to `Message`?

reinkrul (changes requested):

network/transport/v2/protocol.proto
gerardsn: rebase mistake. It's `Envelope` again.

network/transport/v2/protocol.proto
```diff
@@ -27,13 +27,89 @@ service Protocol {
...
+  // hash contains the reference of the transaction, as specified by RFC004.
+  bytes hash = 1;
+  // data contains the data of the transaction, which is a JWS as specified by RFC004.
+  bytes data = 2;
+  // payload contains the payload when it may be attached but wasn't so in the transaction already
+  optional bytes payload = 3;
+}
+
+// Gossip is a message broadcast to inform peers of the node's DAG state and recent additions to it.
+message Gossip {
```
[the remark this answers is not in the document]
gerardsn: `State` does not contain a `conversationID`. I've moved this back in to the message to avoid confusion.
The messages should handles differently, so reusing is probably not a good idea.

woutslakhorst:
LGTM 👍

reinkrul (approved):

PR #673, opened 2021-12-13
Feature/handle private transactions

dmeijboom:
Picks up private transactions from the NATS event stream, decrypts the PAL header and sents the transaction payload query.


network/transport/v2/protocol.go
```diff
@@ -71,22 +116,135 @@ func (p protocol) UnwrapMessage(envelope interface{}) interface{} {
...
+	if err != nil {
+		return err
+	}
+
+	// We weren't able to decrypt the PAL, so it wasn't meant for us
+	if pal == nil {
+		return nil
+	}
+
+	conn := p.connectionList.Get(grpc.ByConnected(), grpc.ByNodeDID(pal[0]))
```
reinkrul, line 219: Don't think this is the right way to select the node that published the TX. I would either;

- derive it from the TX signer (then find out vendor node through NutsComm).
- query it from the node we got this TX from

network/transport/v2/protocol.go
```diff
@@ -71,22 +116,135 @@ func (p protocol) UnwrapMessage(envelope interface{}) interface{} {
...
+	js, err := conn.JetStream()
+	if err != nil {
+		return err
+	}
+
+	if _, err = js.Subscribe(events.PrivateTransactionsSubject, func(msg *nats.Msg) {
+		if err := p.handlePrivateTx(msg); err != nil {
+			logrus.Errorf("failed to handle private transaction: %v", err)
+
+			if err := msg.Nak(); err != nil {
```
reinkrul, line 138: Can I brick a peer's node by sending a TX that causes an error in `handlePrivateTx`, so that it can process no subsequent private TXs?

network/transport/v2/protocol.go
```diff
@@ -71,22 +116,135 @@ func (p protocol) UnwrapMessage(envelope interface{}) interface{} {
...
+
+func (p *protocol) Start() {
+	p.ctx, p.cancel = context.WithCancel(context.Background())
+
+	go func() {
+		for {
+			if err := p.setupNatsHandler(); err != nil {
+				logrus.Errorf("failed to setup NATS handler: %v", err)
+
+				select {
```
reinkrul, line 169: why is this in a loop?

network/transport/v2/protocol.go
```diff
@@ -71,22 +116,135 @@ func (p protocol) UnwrapMessage(envelope interface{}) interface{} {
...
+	}
+
+	js, err := conn.JetStream()
+	if err != nil {
+		return err
+	}
+
+	if _, err = js.Subscribe(events.PrivateTransactionsSubject, func(msg *nats.Msg) {
+		if err := p.handlePrivateTx(msg); err != nil {
+			logrus.Errorf("failed to handle private transaction: %v", err)
```
reinkrul, line 136: make sure we log something we can recognize the TX from (i'd say `ref`)

network/transport/v2/protocol.go
```diff
@@ -71,22 +116,135 @@ func (p protocol) UnwrapMessage(envelope interface{}) interface{} {
...
+
+	// We weren't able to decrypt the PAL, so it wasn't meant for us
+	if pal == nil {
+		return nil
+	}
+
+	conn := p.connectionList.Get(grpc.ByConnected(), grpc.ByNodeDID(pal[0]))
+
+	if conn == nil {
+		return fmt.Errorf("no connection found (DID=%s)", pal[0])
```
reinkrul, line 222: error is probably not descriptive enough (imagine random vendor getting this error, and having to solve it)

network/transport/v2/protocol.go
```diff
@@ -71,22 +116,135 @@ func (p protocol) UnwrapMessage(envelope interface{}) interface{} {
...
+	}
+
+	return conn.Send(p, &Envelope{Message: &Envelope_TransactionPayloadQuery{
```
reinkrul, line 225: this doesn't actually guarantee the payload will be retrieved (just that the message is put in the outbox). Do we have a system in place to retrieve it in case `Send()` succeeds, but we never receive the payload?

reinkrul (changes requested):

network/cmd/cmd.go
```diff
@@ -49,9 +49,9 @@ func FlagSet() *pflag.FlagSet {
 		"Required when `network.enabletls` is `true`.")
 	flagSet.String("network.truststorefile", defs.TrustStoreFile, "PEM file containing the trusted CA certificates for authenticating remote gRPC servers.")
 	flagSet.String("network.nodedid", defs.NodeDID, "Specifies the DID of the organization that operates this node, typically a vendor for EPD software. It is used to identify the node on the network. If the DID document does not exist of is deactivated, the node will not start.")
-	flagSet.String("network.nats.hostname", defs.Nats.Hostname, "Hostname for the NATS server")
-	flagSet.Int("network.nats.port", defs.Nats.Port, "Port where the NATS server listens on")
-	flagSet.Int("network.nats.timeout", defs.Nats.Timeout, "Timeout (in seconds) for the NATS connection")
+	flagSet.String("network.nats.hostname", defs.ProtocolV2.Nats.Hostname, "Hostname for the NATS server")
```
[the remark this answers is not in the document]
dmeijboom: Not sure what you mean exactly but note that you can also configure an external NATS instance

network/transport/grpc/connection_list.go
```diff
@@ -38,18 +38,24 @@ type connectionList struct {
 	list []Connection
 }
 
-func (c *connectionList) Get(peer transport.PeerID) Connection {
-	if len(peer.String()) == 0 {
+func (c *connectionList) Get(query ...Predicate) Connection {
```
woutslakhorst, line 41: leftover from @reinkrul PR, untested method

network/transport/grpc/predicate.go
```diff
@@ -0,0 +1,55 @@
+package grpc
```
woutslakhorst, line 1: copyright and untested file

network/transport/v2/protocol.go
dmeijboom: The NATS docs state that:

> Nak Signals that the message will not be processed now and processing can move onto the next message, NAK'd message will be retried

So I would hope that the next message won't be the NAK'd message but rather the actual next message in the stream.

network/transport/v2/protocol.go
dmeijboom: Because it should try to re-connect when it errors

network/transport/v2/protocol.go
```diff
@@ -20,30 +20,75 @@ package v2
...
 	"github.com/nuts-foundation/nuts-node/network/transport"
 	"github.com/nuts-foundation/nuts-node/network/transport/grpc"
-	grpcLib "google.golang.org/grpc"
+	vdr "github.com/nuts-foundation/nuts-node/vdr/types"
 )
 
 var _ grpc.Protocol = (*protocol)(nil)
 
+// Config specifies config for protocol v2
+type Config struct {
```
woutslakhorst, line 44: As earlier, this config feels out of place here.

Would different engines/protocols have a need for a different Nats config?

network/transport/v2/protocol.go
dmeijboom: Right. I'll think of something. We probably need this in more places anyways.

network/transport/v2/protocol.go
woutslakhorst: as trace logging

network/transport/v2/protocol.go
woutslakhorst: This requires a todo + issue for persistent retries

network/transport/v2/protocol.go
```diff
@@ -71,22 +116,135 @@ func (p protocol) UnwrapMessage(envelope interface{}) interface{} {
...
+			logrus.Errorf("failed to handle private transaction: %v", err)
+
+			if err := msg.Nak(); err != nil {
+				logrus.Errorf("failed to NACK private transaction event: %v", err)
+			}
+
+			return
+		}
+
+		if err := msg.Ack(); err != nil {
```
woutslakhorst, line 145: Isn't the default for Jetstream an automatic Ack?

network/transport/v2/protocol.go
dmeijboom: We can't do that here as we haven't parsed the NATS message yet. I added `ref` in the custom errors in `handlePrivateTx(...)` instead

network/transport/v2/protocol.go
woutslakhorst: It tries to connect to Nats until suc6 or cancelled?

network/transport/v2/protocol.go
```diff
@@ -71,22 +116,135 @@ func (p protocol) UnwrapMessage(envelope interface{}) interface{} {
...
+	}()
+
+	return nil
+}
+
+func (p *protocol) Start() {
+	p.ctx, p.cancel = context.WithCancel(context.Background())
+
+	go func() {
+		for {
```
woutslakhorst, line 165: Once disconnected, the subscription will not resume?

network/transport/v2/protocol.go
```diff
@@ -71,22 +116,135 @@ func (p protocol) UnwrapMessage(envelope interface{}) interface{} {
...
+
+			return
+		}
+	}()
 }
 
-func (p protocol) Stop() {
+func (p *protocol) handlePrivateTx(msg *nats.Msg) error {
+	if p.nodeDID == nil {
+		return errors.New("node DID is not set")
```
woutslakhorst, line 184: Although the business layer will prevent private TXs without a nodeDID, the logic in the network here will create an infinite loop. This error will NACK the event.

network/transport/v2/protocol.go
woutslakhorst: I would prefer the second since that is the most viable option. That connection will probably be there.

network/transport/v2/protocol.go
woutslakhorst: That should also be part of the retry mechanism

woutslakhorst (changes requested):
Please add development documentation on the new event(type, payload, sending engine, receiving engine, description)
An integration test for a happy flow is welcome as well. After the retry mechanism is added, that test can be extended with error condition tests.

network/cmd/cmd.go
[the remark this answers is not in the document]
dmeijboom: They are. If not, how would you be able to connect to an external NATS instance?

network/transport/v2/protocol.go
dmeijboom: See: https://github.com/nuts-foundation/nuts-node/issues/693

network/cmd/cmd.go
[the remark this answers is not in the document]
reinkrul: I think I would prefer to have all NATS config in the `Event Bus` module. Then inject the queue/stream/whatever into the specific modules. Maybe based on an interface the module can implement, which is then queried by the `System` (or maybe the `Event` module) when bootstrapping the Nuts node.

Then the `Event Bus` can be configured to redirect certain topics to an external NATS instance (future work).

E.g.:
```go
type EventPublisher interface {
  GetPublishTopics() []string
  InjectPublishers(topicPublishers map[string]Publisher)
}
```
(and having another one being `EventConsumer`, or make a composite one).

Or maybe better:

```go
type Eventable interface {
   StartPubSub(pubSubber) // called by System/Event Bus, allows engines to start listening on/publishing to topics.
}
```
(requires some better naming)

network/transport/v2/protocol.go
dmeijboom: Good one. I enabled explicit asking so this should work now

network/transport/v2/protocol.go
dmeijboom: Not sure what you mean. Note that the NATS client will automatically reconnect if the initial connection worked

network/transport/v2/protocol.go
reinkrul: Not just TRACE logging, we'll have to make sure it gets logged with default production settings imo. Enabling DEBUG/TRACE on production systems was often not allowed in previous environments, because it often contained sensitive information. And it yielded so much, that I hated having to spit through it trying to find out what went wrong.

I generally find Go errors OK (or maybe just in our Nuts node) for finding out _what_ went wrong (e.g. parsing error), but generally poor for finding out _where_ (record XYZ). Especially important when there's a lot of options (e.g. thousands of TXs)

network/transport/v2/protocol.go
dmeijboom: Good one. This is awful because it will just exhaust CPU by constantly NAKking messages if the Node DID isn't configured. I'll check for the node DID while setting up the NATS subscriber and error there.

network/cmd/cmd.go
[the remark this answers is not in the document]
dmeijboom: In my opinion the events engine should be optional. If it's disabled you have to use an external NATS instance. Adding yet another tight coupled integration to the engines is not a good idea in my opinion. It quickly becomes a mess because it's really hard to find the right abstraction.

network/transport/grpc/connection_list.go
dmeijboom: I'll add some tests

network/transport/v2/protocol.go
dmeijboom: We only have access to the transaction data. I don't think there is a way to retrieve the node that send the transaction right?

network/transport/v2/protocol.go
dmeijboom: This is not really an option as it's no longer fixed but uses an interface so it could change at runtime (in theory)

network/transport/v2/protocol.go
woutslakhorst: Ah yes, could as well come from a totally different node. Then filter the PAL list (exclude our own DID) and take first.

events/conn.go
```diff
@@ -39,11 +45,66 @@ type JetStreamContext interface {
...
+			if err == nil {
+				pool.conn.Store(connectionAndContext{conn, js})
+
+				return conn, js, nil
+			}
+		}
+
+		log.Logger().Errorf("failed to connect to %s: %s", addr, err.Error())
+
+		select {
```
woutslakhorst, line 103: untested cases

events/conn_test.go
```diff
@@ -0,0 +1,40 @@
...
+	err := manager.Start()
+	assert.NoError(t, err)
+
+	defer manager.Shutdown()
+
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	defer cancel()
+
+	wc := sync.WaitGroup{}
+	wc.Add(500)
```
woutslakhorst, line 25: when 10 just doesn't cut it?

network/network.go
```diff
@@ -89,16 +91,24 @@ func (n *Network) Walk(visitor dag.Visitor) error {
...
 
 // NewNetworkInstance creates a new Network engine instance.
-func NewNetworkInstance(config Config, keyResolver types.KeyResolver, privateKeyResolver crypto.KeyResolver, didDocumentResolver types.DocResolver) *Network {
-	result := &Network{
+func NewNetworkInstance(
+	config Config,
+	eventsConnectionPool events.ConnectionPool,
+	keyResolver types.KeyResolver,
+	privateKeyResolver crypto.KeyResolver,
+	decrypter crypto.Decrypter,
```
woutslakhorst, line 99: passing the crypto module twice is a bit weird. But lets tackle that in a refactor PR laer

events/conn.go
```diff
@@ -20,13 +20,19 @@
...
 	"time"
+
+	"github.com/nuts-foundation/nuts-node/vcr/log"
+
+	"github.com/nats-io/nats.go"
 )
 
 // Conn defines the methods required in the NATS connection structure
 type Conn interface {
+	Close()
```
woutslakhorst, line 35: Close not called by any one?
Which adds the question, who is going to call close? The events engine?

network/transport/v2/protocol.go
```diff
@@ -20,30 +20,67 @@ package v2
...
 	"github.com/nuts-foundation/nuts-node/network/transport"
 	"github.com/nuts-foundation/nuts-node/network/transport/grpc"
-	grpcLib "google.golang.org/grpc"
+	vdr "github.com/nuts-foundation/nuts-node/vdr/types"
 )
 
 var _ grpc.Protocol = (*protocol)(nil)
 
+// Config specifies config for protocol v2
+type Config struct {
```
woutslakhorst, line 43: Config is empty, omit?

network/transport/v2/protocol.go
```diff
@@ -71,22 +108,127 @@ func (p protocol) UnwrapMessage(envelope interface{}) interface{} {
...
-func (p protocol) Start() {
+func (p *protocol) setupNatsHandler() error {
+	_, js, err := p.eventsConnectionPool.Acquire(p.ctx)
+	if err != nil {
+		return err
+	}
+
+	if _, err = js.Subscribe(events.PrivateTransactionsSubject, func(msg *nats.Msg) {
+		if err := p.handlePrivateTx(msg); err != nil {
+			logrus.Errorf("failed to handle private transaction: %v", err)
```
woutslakhorst, line 119: not local logger

network/transport/v2/protocol.go
woutslakhorst: Let's fix it when implementing retry

woutslakhorst (changes requested):
summary:

- regenerate docs
- who closes Nats connection
- small stuff

events/conn_test.go
dmeijboom: I wanted to catch possible race conditions quickly but 500 is a bit too extreme 😅 Changed it to 50 which should be fine

events/conn.go
dmeijboom: > who is going to call close? The events engine?

Yes, I think that makes sense. Will fix

network/transport/v2/protocol.go
dmeijboom: I could've removed it but it's a nice boilerplate to have as I'm sure the v2 protocol will get it's own configuration at some point

woutslakhorst (approved):

Issue #674, opened 2021-12-13
Failing CRL updates should mention the issuer of the certificate


reinkrul:
```
CRL synchronization failed: synchronization failed: CRL signature could not be validated against known certificates
```

Otherwise, it's very hard to find out which CRL couldn't be updated.

PR #675, opened 2021-12-13
Improve CRL synchronization errors

reinkrul:
Log issuer name where available, otherwise log CRL endpoint (in case of network errors)

Fixes https://github.com/nuts-foundation/nuts-node/issues/674


dmeijboom (approved):
Nice 👍

woutslakhorst (approved):

PR #676, opened 2021-12-13
Auth: fix clockskew test by making the skew larger (slow CirlceCI)

reinkrul:
Randomly fails because the clock skew is too small now and then.

Also made the use of `time.Duration` appropriate.

### DIFF
--- a/auth/services/oauth/oauth_test.go
+++ b/auth/services/oauth/oauth_test.go
@@ -573,6 +573,7 @@ func TestService_parseAndValidateJwtBearerToken(t *testing.T) {
 	})
 
 	t.Run("valid token with clock diff", func(t *testing.T) {
+		// a token created 10 minutes ago, valid until 4 minutes ago. But due to clock skew of 5 minutes, it should still be valid.
 		ctx := createContext(t)
 		ctx.oauthService.clockSkew = 5 * time.Minute
 		tokenCtx := validContext()
--- a/auth/services/oauth/oauth_test.go
+++ b/auth/services/oauth/oauth_test.go
@@ -586,6 +586,21 @@ func TestService_parseAndValidateJwtBearerToken(t *testing.T) {
 		err := ctx.oauthService.parseAndValidateJwtBearerToken(tokenCtx)
 		assert.NoError(t, err)
 	})
+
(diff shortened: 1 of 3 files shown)


dmeijboom (changes requested):
`s.clockSkew` is already multiplied by `time.Millisecond`

stevenvegt (approved):

woutslakhorst:
> `s.clockSkew` is already multiplied by `time.Millisecond`

Then a test just outside the clock skew should be added to validate

reinkrul:
> `s.clockSkew` is already multiplied by `time.Millisecond`

It was, but in the wrong location. The clockSkew was being passed around as 5000 milliseconds as `time.Duration`, which then actually is 5000 nanoseconds. Before being used it was converted to milliseconds, (`* time.Milliseconds`), but that should've done at the moment the clockSkew is converted from `int` to `time.Duration`. Otherwise, one could falsely assume that the `5000` as `clockSkew int` == `5000 milliseconds clockSkew time.Duration` (while it was in nanoseconds).

Therefore I renamed `clockSkew int` to `clockSkewInMilliseconds int` to avoid confusing about the units, and then immediately convert it to actual milliseconds the moment we derive a `time.Duration` from it.

auth/services/oauth/oauth_test.go
```diff
@@ -586,6 +586,21 @@ func TestService_parseAndValidateJwtBearerToken(t *testing.T) {
 		err := ctx.oauthService.parseAndValidateJwtBearerToken(tokenCtx)
 		assert.NoError(t, err)
 	})
+
+	t.Run("expired token", func(t *testing.T) {
+		// a token created 10 minutes ago, valid until 4 minutes ago. Just a very small clock skew allowed, so it should be expired.
+		ctx := createContext(t)
+		ctx.oauthService.clockSkew = 1 * time.Millisecond // because 0 multiplied by 0 equals 0, rather use 1 millisecond (small clock skew), better test.
```
dmeijboom, line 593: Nitpick: `1 * time.Millisecond` is the same as `time.Millisecond`

dmeijboom (approved):
LGTM

PR #678, opened 2021-12-15
Network: fix call on nil context due to racy connection close

reinkrul:
Happens now and then on CircleCI, can theoretically happen runtime as well.

`close()` causes `context()` to return `nil` (which is valid), callers should handle it. Introduced `waitTilClosed()` to reduce chance of errornous use.

### DIFF
--- a/network/transport/grpc/connection.go
+++ b/network/transport/grpc/connection.go
@@ -40,8 +40,8 @@ type managedConnection interface {
 	// close shuts down active inbound or outbound streams and stops active outbound connectors.
 	// After calling close() the managedConnection cannot be reused.
 	close()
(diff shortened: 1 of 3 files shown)


network/transport/grpc/connection.go
```diff
@@ -124,6 +126,18 @@ func (mc *conn) context() context.Context {
 	return mc.ctx
 }
 
+func (mc *conn) waitTilClosed() {
```
dmeijboom, line 129: \* `waitUntilClosed`

dmeijboom (changes requested):
One tiny comment, after that: LGTM

dmeijboom (changes requested):
LGTM!

Issue #686, opened 2021-12-20
Human-friendly error should be returned when a private TX is created, but no node DID is configured


reinkrul:
To help parties already running a Nuts Node migrate to public private VCs as private TXs, the Nuts Node must return a readable, descriptive error when node DID is not configured.

Must be checked.

reinkrul:
There are a few perspectives on this:

- Technically you don't need to configure a node DID to publish a private TX: the TX is signed with a provided key (ID) and the PAL header is encrypted with key agreement keys of the specified PAL DIDs. Missing (or misconfigured) node DID isn't a problem until the parties in the PAL header try to retrieve the transaction payload.

woutslakhorst:
an assertion would be the best for now

dmeijboom:
> Maybe the best we could do for now, without adding lots of complexity, is asserting a node DID is configured when issuing a private transaction (although it might not be the correct one).

^ Looks like a plan. The Node DID is already validated right on startup? (as in: resolving the document and error if it doesn't exist)

PR #692, opened 2021-12-20
Network: return operator-friendly error when creating private TXs when node DID isn't configured

reinkrul:
Fixes https://github.com/nuts-foundation/nuts-node/issues/686

### DIFF
--- a/network/network.go
+++ b/network/network.go
@@ -222,18 +222,9 @@ func (n *Network) Start() error {
 		return err
 	}
 	if !nodeDID.Empty() {
-		doc, _, err := n.didDocumentResolver.Resolve(nodeDID, nil)
+		err := n.validateNodeDID(nodeDID)
 		if err != nil {
-			return fmt.Errorf("invalid NodeDID configuration: DID document can't be resolved (did=%s): %w", nodeDID, err)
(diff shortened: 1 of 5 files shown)


network/network.go
```diff
@@ -218,24 +217,28 @@ func (n *Network) Start() error {
 	}
 
 	// Sanity check for configured node DID: can we resolve it?
-	if n.configuredNodeDID != nil {
-		doc, _, err := n.didDocumentResolver.Resolve(*n.configuredNodeDID, nil)
+	nodeDID, err := n.nodeDIDResolver.Resolve()
+	if err != nil {
+		return err
+	}
+	if !nodeDID.Empty() {
```
woutslakhorst, line 224: It should also have a `NutsComm` endpoint?

woutslakhorst (changes requested):

network/network.go
reinkrul: OriginallynNot really part of this PR (sanity check on startup), these lines changed due to nodeDID not being cached any more from startup. But will add it.

woutslakhorst:
Tests seem to work ;)

PR #696, opened 2021-12-21
add Finder to VDR

woutslakhorst:
This adds a DocFinder interface to the VDR.
It supports finding DID Documents with specific predicates.

Stole Predicate setup from @dmeijboom ;)

This feature is required for service discovery. At startup, the network needs to find all did Documents with a concrete NutsComm address.

### DIFF
--- a/vdr/doc/finder.go
+++ b/vdr/doc/finder.go
@@ -1,0 +1,86 @@
+/*
+ * Nuts node
+ * Copyright (C) 2021 Nuts community
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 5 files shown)


vdr/doc/finder.go
```diff
@@ -0,0 +1,111 @@
...
+
+	"github.com/nuts-foundation/go-did/did"
+	"github.com/nuts-foundation/nuts-node/vdr/types"
+)
+
+// ByServiceType returns a predicate that matches on service type
+// it only matches on DID Documents with a concrete endpoint (not starting with "did")
+func ByServiceType(serviceType string) types.Predicate {
+	return servicePredicate{serviceType: serviceType}
+}
```
dmeijboom, line 34: I'm wondering does it really make sense to only return DID documents with a concrete endpoint? Or should that be in a separate predicate?

dmeijboom (commented):
One question

vdr/doc/finder.go
woutslakhorst: I had the same thoughts. I decided to make it explicit in the godoc, so whenever something else is needed, a simple refactor would suffice

dmeijboom: If we only care about DID documents with a concrete endpoint I think it would be okay like this. As long as we refactor it whenever we do want to return DID documents without a concrete endpoint

dmeijboom (approved):

PR #683, opened 2021-12-16
Network: move all connection management and message sending/receiving to gRPC layer

reinkrul:
- Introduced `grpc.Protocol` for gRPC protocol implementations (`v1` and `v2`)
- Removed connection management from protocols: now get a reference to `grpc.ConnectionList` to query peers
- removed message sending/receiving from protocol
  - Can now send using `grpc.Connection.Send()` (lookup using `grpc.ConnectionList`)
  - Receive loop is now managed by `grpc.ConnectionManager` which calls `grpc.Protocol.Handle()` when a message is received
  - Now completely non-blocking, all blocking (to avoid immediately closing streams by letting the gRPC functions end) is now done by `grpc.ConnectionManager`
- `v1.p2p` package is now removed (possible due to changes above)
- `v1.logic` can be merged into parent package in a future PR

PR #700, opened 2021-12-21
After-merge PR feedback on gRPC refactor

reinkrul:
See https://github.com/nuts-foundation/nuts-node/pull/683

### DIFF
--- a/events/event_test.go
+++ b/events/event_test.go
@@ -20,6 +20,8 @@
 package events
 
 import (
+	"fmt"
+	"github.com/nuts-foundation/nuts-node/test"
 	"testing"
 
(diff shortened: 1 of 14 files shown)


dmeijboom:
To be honest. I think `IsConnected` is way more clear than `isOpen`. It also makes sense as the `disconnect` method was called when `connected` is `false` and all gRPC related stuff is closed.

dmeijboom (approved):
LGTM

reinkrul:
> To be honest. I think `IsConnected` is way more clear than `isOpen`. It also makes sense as the `disconnect` method was called when `connected` is `false` and all gRPC related stuff is closed.

@woutslakhorst ?

PR #701, opened 2021-12-21
Feature/service discovery

woutslakhorst:
Added `enablediscovery` option (true by default) to disable connecting to other nodes. Useful in test and for clustered nodes that only want to connect to own nodes.

This does not yet listen to changed DID Documents at runtime


reinkrul:
Discovery now discovers (and connects to) the local node, which it shouldn't. Easiest way to counter this is by comparing found DID documents with the local node DID: if it's the same DID, don't discover `NutsComm` from that DID document.

This still allows other DIDs to define the local node's public address as their `NutsComm` address, potentially irritating/disturbing the node's operation.

To completely counter this, we should probably disconnect when a party presents a peer ID that matches the one of the local node.

reinkrul:
Rebased.

network/transport/types.go
```diff
@@ -33,12 +35,61 @@ func (p PeerID) String() string {
...
+func ParseAddress(input string) (Addr, error) {
+	parsed, err := url.Parse(input)
+	if err != nil {
+		return Addr{}, err
+	}
+	if parsed.Scheme != "grpc" {
+		return Addr{}, errors.New("invalid URL scheme")
+	}
+	return Addr{scheme: parsed.Scheme, target: parsed.Host}, nil
+}
```
dmeijboom, line 85: Why not use `url.URL` type instead of creating an `Addr` type?

network/network.go
```diff
@@ -253,14 +255,50 @@ func (n *Network) Start() error {
...
+				address, err := transport.ParseAddress(nutsCommStr)
+				if err != nil {
+					log.Logger().Warnf("invalid NutsComm address from service (did=%s, str=%s): %v", node.ID.String(), nutsCommStr, err)
+					continue inner
+				}
+				log.Logger().Infof("Discovered Nuts node (address=%s), published by %s", address, node.ID)
+				n.connectionManager.Connect(address)
+			}
+		}
+	}
```
dmeijboom, line 301: Is this only called once at startup?

dmeijboom (changes requested):

network/network.go
woutslakhorst: That was the initial thought. The best way to adjust things runtime is to receive an event in the network layer from the VDR.

network/transport/types.go
reinkrul: I initially had it as `url.URL` but then you can supply many variants which don't qualify as a valid gRPC target address. We currently only support the DNS-style addresses, but there are a few other styles I can imagine we want to support (e.g. for clustering and/or in IPv6 environments), see https://github.com/grpc/grpc/blob/master/doc/naming.md

Therefore I figured the safest and most future-proof way to implement this was to introduce a type which implementation can be extended to support new/other formats.

dmeijboom: This is the actual target parsing code in the `grpc-go` library: https://github.com/grpc/grpc-go/blob/14c11384b76b67f7b1b32a5d18f865762634c0ae/clientconn.go#L1663 Looking at it this seems wrong. To be honest I'm not even sure why we need an `Addr` type at all. What's wrong with using a `string` here?

woutslakhorst: If all that is needed is supported by the grpc library itself, then use that?

reinkrul: Reverted it back to `string`. We can't use the internal type since it's private, unfortunately.

dmeijboom (approved):

PR #702, opened 2021-12-22
allow connection to bootstrapnode

### LINKED ISSUE #699: Authentication issues when node is out of sync
When starting a new Nuts node, you need to configure a bootstrap node to sync the existing network DAG from. However, if the bootstrap node sends a node DID the new node will attempt to authenticate it upon connecting. It does this by resolving the node DID on the VDR. The VDR, however, is empty since it's a new node. So it can't authenticate the bootstrap node's connection, and the node will immediately disconnect. This prevents new nodes from getting in sync with bootstrap nodes that provide their node DID (which all nodes should do, eventually).

This not just happens with new nodes: it can also happens when the node is so out-of-sync (after some downtime) that it misses transactions crucial for authenticating the bootstrap node.

Thus, nodes need to be able to sync (at least with bootstrap nodes), when they can't authenticate the node DID (yet). Solutions:

- When a new connection can't be authenticated, but it is a bootstrap node, do not disconnect. But also don't mark the connection as trusted. That way it can be used for syncing, but not for exchanging private VCs. After syncing, the next (re)connect with the bootstrap node will be able to authenticate.
- Auto-trust node DIDs sent by bootstrap nodes (not a very good idea)
- Delay node DID authentication until required (exchanging private VCs). Opens up new issues to solve (what if multiple peers, some are attackers/spoofers, send the same node DID).

@woutslakhorst @dmeijboom

woutslakhorst:
fixes #699 

Adds a bool to `Connect` allowing unauthenticated connections to bootstrap nodes.

### DIFF
--- a/network/transport/grpc/authenticator.go
+++ b/network/transport/grpc/authenticator.go
@@ -20,13 +20,14 @@ package grpc
 
 import (
 	"fmt"
+	"net/url"
+
 	"github.com/nuts-foundation/go-did/did"
 	"github.com/nuts-foundation/nuts-node/network/log"
 	"github.com/nuts-foundation/nuts-node/network/transport"
 	"github.com/nuts-foundation/nuts-node/vdr/doc"
 	"google.golang.org/grpc/credentials"
 	grpcPeer "google.golang.org/grpc/peer"
-	"net/url"
 )
 
 // Authenticator verifies node identities.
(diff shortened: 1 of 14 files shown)


dmeijboom (changes requested):
It's a tiny bit more work but I'd be in favour of adding connection options instead as it's not really clear what the booleans stands for and there might be other options in the future. Something like this:

```
func (manager *ConnectionManager) Connect(peerAddress, opts ...ConnectionOpt) {
    (...)
}
```

Where you can call it like this (just an example):


```
manager.Connect("example123", WithUnauthenticated())
```

network/transport/grpc/authenticator.go
```diff
@@ -47,10 +48,18 @@ type tlsAuthenticator struct {
...
 
 func (t tlsAuthenticator) Authenticate(nodeDID did.DID, grpcPeer grpcPeer.Peer, peer transport.Peer) (transport.Peer, error) {
+	withOverride := func(peer transport.Peer, err error) (transport.Peer, error) {
+		if peer.AcceptUnauthenticated {
+			log.Logger().Warnf("Connection manually authenticated, authentication error: %v", err)
+			return peer, nil
+		}
+		return peer, err
+	}
+
```
dmeijboom, line 58: If `peer.AcceptUnauthenticated` is `true` can't we just skip the entire `Authenticate` method altogether? Doesn't really make a difference right?

dmeijboom (changes requested):

network/transport/grpc/authenticator.go
woutslakhorst: That would also remove the logging. So you wouldn't be aware of a faulty bootstrap node.

dmeijboom: Right, makes sense!

dmeijboom (approved):
LGTM!

PR #703, opened 2021-12-22
find NutsComm service owners for private credentials

### LINKED ISSUE #698: Incorrect handling of participants
In VCR the *participants* are selected from the issuer and credentialSubject. These are not the vendor IDs (probably not).
These DIDs need to be translated to DIDs that have a NutsComm endpoint and KeyAgreement key.

woutslakhorst:
fixes #698 

the vcr translates the DIDs in the VC to their `NutsComm` service owners.

After merge, notify Slack channel on changes since issuing a private VC will no longer work without a node identity.


dmeijboom (approved):
LGTM

PR #706, opened 2021-12-27
Feature/705 bypass node auth

### LINKED ISSUE #705: Bypass node identity authentication in development
add option `AllowUnauthenticatedNodeIdentity` in non-strict mode. In workshops, TLS might be disabled. The node identity can then not be validated (no SANs from certificates), then we can just trust the DID sent by the other party!

woutslakhorst:
Closes #705 

Added option to disable peer Node DID authentication when in development with `disableNodeAuthentication`.

Also refactored the events connection pool passing. A connection for Nats was started from Configure instead of Start


reinkrul:
> Also refactored the events connection pool passing. A connection for Nats was started from Configure instead of Start

This really, really is a separate PR.

reinkrul (approved):

PR #708, opened 2021-12-29
added retry mechanism to payload queries

### LINKED ISSUE #693: Retry logic in private transaction handler
Private transactions are handled by protocol v2. Currently, when it fails the NATS message is NAK'd and it will be retried based on the logic of the NATS server.

In the future there need to be some logic to retry it based on an exp backoff (or something similar).

_Originally posted by @woutslakhorst in https://github.com/nuts-foundation/nuts-node/pull/673#discussion_r772965719_

### LINKED ISSUE #707: Connection selection from PAL not correct
in protocol V2 handlePrivateTx, the first DID from the PAL is used to find the connection. This should be filtered to remove the node DID first.

woutslakhorst:
fixes #707 
closes #693

Adds retry mechanism to tx payload fetching.
It uses an incremental backoff with a maximum of 24 hours per retry and a count of 100.
The retry mechanism uses a bbolt DB for storing jobs. 

questions/todos:
- [x] future issues: gather some statistics/diagnostics so action can be undertaken before the count of 100 is reached?
- [x] do we have an e2e test with private TXs?
- [x] passing of config/default config for v2 protocol


reinkrul:
Made an issue for e2e end tests for private TXs: https://github.com/nuts-foundation/nuts-node/issues/713

reinkrul:
`payload_jobs.db` has been checked in, probably by unit tests? These unit tests should use a temp test directory (see integration tests, for instance), so files generated by tests don't end up in the repo itself.

network/transport/v2/retry.go
```diff
@@ -0,0 +1,195 @@
...
+			count, existing, err := p.readCount(hash)
+			if err != nil {
+				return retry.Unrecoverable(err)
+			}
+			if existing {
+				if err := p.writeCount(hash, count+1); err != nil {
+					return retry.Unrecoverable(err)
+				}
+				p.callback(hash)
+				return errContinue
```
reinkrul, line 139: what if the callback causes the payload to be retrieved successfully, is an error logged by the retry lib? (because we're returning an error here)

network/transport/v2/retry.go
```diff
@@ -0,0 +1,195 @@
...
+			return nil
+		},
+			retry.Attempts(100),          // should be enough
+			retry.MaxDelay(24*time.Hour), // maximum delay of an hour
+			retry.Delay(startDelay),      // first retry after 5 seconds, second after 10, 20, 40, etc
+			retry.DelayType(retry.BackOffDelay),
+			retry.Context(p.ctx),
+			retry.LastErrorOnly(true), // only log last error
+			retry.OnRetry(func(n uint, err error) {
+				log.Logger().Debugf("retrying payload (count=%d) with ref: %s", n, hash.String())
```
reinkrul, line 152: we're ignoring `err`?

network/transport/v2/retry.go
```diff
@@ -0,0 +1,195 @@
...
+		if err != nil {
+			log.Logger().Errorf("Failed to pass payload query to network: %v", err)
+		}
+	}()
+}
+
+func (p *payloadRetrier) writeCount(hash hash.SHA256Hash, count uint16) error {
+	return p.db.Update(func(tx *bbolt.Tx) error {
+		bucket := tx.Bucket([]byte("payload_jobs"))
+		data := make([]byte, 2)
```
reinkrul, line 164: magic number (2), can't we get/calculate the size of a uint16 (or whatever type we're using)?

network/transport/v2/retry.go
```diff
@@ -0,0 +1,195 @@
...
+func (p *payloadRetrier) Configure() error {
+	p.ctx, p.cancel = context.WithCancel(context.Background())
+
+	dbFile := path.Join(p.config.Datadir, "network", "payload_jobs.db")
+	if err := os.MkdirAll(filepath.Dir(dbFile), os.ModePerm); err != nil {
+		return fmt.Errorf("unable to setup database: %w", err)
+	}
+
+	if p.config.PayloadRetryDelay == 0 {
+		p.config.PayloadRetryDelay = 5 * time.Second
```
reinkrul, line 83: make this default a const

network/transport/v2/retry.go
```diff
@@ -0,0 +1,195 @@
...
+	}
+	var err error
+	p.db, err = bbolt.Open(dbFile, 0600, bbolt.DefaultOptions)
```
reinkrul, line 86: wouldn't it be more flexible to let the retry db be passed by the caller, instead of the retrier always making a database by itself?

Reasoning: maybe the payload_jobs bucket should be part of the DAG BBolt database. Because the DAG TXs and shouldn't be backup'd apart from each other (otherwise you might be missing payloads after copying/restoring), the same goes for the payload jobs: if you fail to restore/copy those, there might be private TXs that are never retried?

reinkrul (changes requested):

network/transport/v2/retry.go
dmeijboom: A uint16 is 16 bits thus two bytes so it makes sense right?

dmeijboom (changes requested):
`*.db` files should be in the gitignore

network/transport/v2/retry.go
reinkrul: Yeah, but when somebody decides to change it and doesn't realize the `2` is related to it, things will break? (edge case though)

reinkrul, line 164:
```suggestion
		// uint16 is 2 bytes long
		data := make([]byte, 2))
```

network/transport/v2/retry.go
```diff
@@ -0,0 +1,195 @@
...
+// The payload hashes as []byte should be added as job.
+func NewPayloadRetrier(config Config, callback jobCallBack) Retriable {
+	return &payloadRetrier{
+		config:   config,
+		callback: callback,
+	}
+}
+
+type payloadRetrier struct {
+	config   Config
```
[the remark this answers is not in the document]
dmeijboom: Fixed

network/transport/v2/retry.go
```diff
@@ -0,0 +1,195 @@
...
+func (p *payloadRetrier) Add(hash hash.SHA256Hash) error {
+	if err := p.writeCount(hash, 0); err != nil {
+		return err
+	}
+	p.retry(hash, 0)
+
+	return nil
+}
+
+var errContinue = errors.New("continue")
```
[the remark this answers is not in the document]
dmeijboom: Added a description and renamed it to make it more clear

network/transport/v2/retry.go
dmeijboom: It makes sense as the error will be `jobInProgress` anyway (previously called `errContinue`)

network/transport/v2/retry.go
woutslakhorst: One of the next big things to do is streamline the storage requirements for all engines/components. This relates to DB backup requirements. This DB would then be refactored as well, to get this piece of functionality forward, I chose for a simple local DB.

network/transport/v1/protocol.go
```diff
@@ -90,16 +91,19 @@ func (p protocolV1) UnwrapMessage(raw interface{}) interface{} {
...
 		time.Duration(p.config.CollectMissingPayloadsInterval)*time.Millisecond,
 		peerID)
+
+	return nil
 }
 
-func (p protocolV1) Start() {
+func (p protocolV1) Start() error {
 	p.protocol.Start()
+	return nil
```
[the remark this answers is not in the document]
woutslakhorst: the V1 logic protocol start func does not return an error.

network/transport/v1/protocol.go
```diff
@@ -90,16 +91,19 @@ func (p protocolV1) UnwrapMessage(raw interface{}) interface{} {
...
 
-func (p protocolV1) Configure(peerID transport.PeerID) {
+func (p protocolV1) Configure(peerID transport.PeerID) error {
 	p.protocol.Configure(
 		time.Duration(p.config.AdvertHashesInterval)*time.Millisecond,
 		time.Duration(p.config.AdvertDiagnosticsInterval)*time.Millisecond,
 		time.Duration(p.config.CollectMissingPayloadsInterval)*time.Millisecond,
 		peerID)
+
+	return nil
```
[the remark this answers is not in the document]
woutslakhorst: same

network/transport/v2/protocol.go
```diff
@@ -42,11 +43,17 @@ var _ grpc.Protocol = (*protocol)(nil)
 
 // Config specifies config for protocol v2
 type Config struct {
+	// Datadir from core.Config
+	Datadir string
```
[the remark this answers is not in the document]
woutslakhorst: The protocol configs do not use the core.Config. When storage is refactored this will be removed.

network/transport/v2/retry.go
```diff
@@ -0,0 +1,195 @@
...
+	go func() {
+		err := retry.Do(func() error {
+			count, existing, err := p.readCount(hash)
+			if err != nil {
+				return retry.Unrecoverable(err)
+			}
+			if existing {
+				if err := p.writeCount(hash, count+1); err != nil {
+					return retry.Unrecoverable(err)
+				}
```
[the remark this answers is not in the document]
woutslakhorst: There should be some protection against duplicate go procedures which is handled by the Schedule func. I'll add that.

network/transport/grpc/connection.go
```diff
@@ -99,7 +100,7 @@ type conn struct {
 }
 
 func (mc *conn) Peer() transport.Peer {
-	// Populated through createConnection and verifyOrSetPeerID
+	// Populated through createConnection and setPeerID
```
woutslakhorst, line 103: will revert change

network/transport/grpc/test.go
```diff
@@ -162,6 +163,10 @@ func (s StubConnection) verifyOrSetPeerID(_ transport.PeerID) bool {
 	panic("implement me")
 }
 
+func (s StubConnection) setNodeDID(_ did.DID) bool {
```
woutslakhorst, line 166: Obsolete, will remove

network/transport/v2/protocol.go
```diff
@@ -180,41 +222,59 @@ func (p *protocol) handlePrivateTx(msg *nats.Msg) error {
...
+	j := 0
+	for _, participant := range pal {
+		if !participant.Equals(nodeDID) {
+			pal[j] = participant
+			j++
+		}
+	}
+	pal = pal[:j]
+
 	conn := p.connectionList.Get(grpc.ByConnected(), grpc.ByNodeDID(pal[0]))
```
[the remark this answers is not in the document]
woutslakhorst: will make an issue for this

network/dag/publisher.go
```diff
@@ -150,30 +150,27 @@ func (s *replayingDAGPublisher) publish(ctx context.Context) {
...
-	}
-
 	payload, err := s.payloadStore.ReadPayload(ctx, transaction.PayloadHash())
 	if err != nil {
 		log.Logger().Errorf("Unable to read payload to publish DAG: (ref=%s) %v", transaction.Ref(), err)
 		return false
 	}
 
 	if payload == nil {
+		// We need to skip transactions with PAL header as it should be handled by the v2 protocol
```
dmeijboom, line 169: Nitpick: comment is not accurate anymore (`s/skip/process/` or something similar)

network/network.go
```diff
@@ -251,7 +255,9 @@ func (n *Network) Start() error {
 		return err
 	}
 	for _, prot := range n.protocols {
-		prot.Start()
+		if err = prot.Start(); err != nil {
+			return err
+		}
```
dmeijboom, line 260: I changed this to returning an error but I think we should revert it as it should just retry indefinitely

network/transport/grpc/connection.go
dmeijboom: Should be `verifyOrSetPeerID` I believe (could be part of the rebase/merge)

dmeijboom (changes requested):
Couple of (really) minor things left

network/transport/grpc/connection.go
reinkrul: I renamed it

network/network.go
woutslakhorst: I don't understand. This error is for starting the entire protocol....

dmeijboom (approved):

network/transport/v2/protocol.go
```diff
@@ -106,16 +122,40 @@ func (p protocol) UnwrapMessage(envelope interface{}) interface{} {
 	return envelope.(*Envelope).Message
 }
 
-func (p protocol) Configure(_ transport.PeerID) {
+func (p *protocol) Configure(_ transport.PeerID) error {
+	dbFile := path.Join(p.config.Datadir, "network", "payload_jobs.db")
```
reinkrul, line 126: wouldn't the database be passed into the protocol instead?

network/transport/v2/protocol.go
```diff
@@ -157,44 +203,96 @@ func (p *protocol) Start() {
...
 		}
 	}()
+
+	return
 }
 
 func (p *protocol) handlePrivateTx(msg *nats.Msg) error {
 	tx, err := dag.ParseTransaction(msg.Data)
 	if err != nil {
+		log.Logger().Errorf("failed to parse transaction from event: %v", err)
```
reinkrul, line 213: Will this be nack'd, so it stays there forever?

network/transport/v2/protocol.go
```diff
@@ -157,44 +203,96 @@ func (p *protocol) Start() {
...
 		return err
 	}
+	if err = p.payloadScheduler.Schedule(tx.Ref()); err != nil {
+		// this means the underlying DB is broken
+		log.Logger().Errorf("failed to add payload query retry job: %v", err)
```
reinkrul, line 218: add TX ref to log message for better anaysis

network/transport/v2/protocol.go
```diff
@@ -157,44 +203,96 @@ func (p *protocol) Start() {
...
+		return err
+	}
+	return nil
+}
+
+func (p *protocol) handlePrivateTxRetry(hash hash.SHA256Hash) {
+	if err := p.handlePrivateTxRetryErr(hash); err != nil {
+		log.Logger().Errorf("retry of TransactionPayloadQuery failed: %v", err)
```
reinkrul, line 226: add hash to log msg

network/transport/v2/protocol.go
```diff
@@ -157,44 +203,96 @@ func (p *protocol) Start() {
...
+}
+
+func (p *protocol) handlePrivateTxRetryErr(hash hash.SHA256Hash) error {
+	tx, err := p.graph.Get(context.Background(), hash)
+	if err != nil {
+		return fmt.Errorf("failed to retrieve transaction with ref:%s from the DAG: %w", hash.String(), err)
+	}
+
+	if tx == nil {
+		return fmt.Errorf("failed to find transaction with ref:%s in DAG", hash.String())
```
reinkrul, line 237: use `(tx=%s)` format

network/transport/v2/protocol.go
```diff
@@ -157,44 +203,96 @@ func (p *protocol) Start() {
...
+func (p *protocol) handlePrivateTxRetry(hash hash.SHA256Hash) {
+	if err := p.handlePrivateTxRetryErr(hash); err != nil {
+		log.Logger().Errorf("retry of TransactionPayloadQuery failed: %v", err)
+	}
+}
+
+func (p *protocol) handlePrivateTxRetryErr(hash hash.SHA256Hash) error {
+	tx, err := p.graph.Get(context.Background(), hash)
+	if err != nil {
+		return fmt.Errorf("failed to retrieve transaction with ref:%s from the DAG: %w", hash.String(), err)
```
reinkrul, line 233: use `(tx=%s)` format

network/transport/v2/protocol.go
```diff
@@ -157,44 +203,96 @@ func (p *protocol) Start() {
...
+
+	if payload != nil {
+		// stop retrying
+		log.Logger().Infof("Transaction payload already present, not querying (tx=%s)", hash)
+		return p.payloadScheduler.Finished(hash)
+	}
 
 	if len(tx.PAL()) == 0 {
-		return fmt.Errorf("PAL header is empty (ref=%s)", tx.Ref().String())
+		return fmt.Errorf("PAL header is empty (ref=%s)", hash.String())
```
reinkrul, line 253: remove from scheduler?

network/transport/v2/protocol.go
```diff
@@ -157,44 +203,96 @@ func (p *protocol) Start() {
...
 	}
 
-	return conn.Send(p, &Envelope{Message: &Envelope_TransactionPayloadQuery{
+	err = conn.Send(p, &Envelope{Message: &Envelope_TransactionPayloadQuery{
 		TransactionPayloadQuery: &TransactionPayloadQuery{
 			TransactionRef: tx.Ref().Slice(),
 		},
 	}})
+	if err != nil {
+		return fmt.Errorf("failed to send TransactionPayloadQuery message(ref=%s, DID=%s): %w", hash.String(), pal[0], err)
```
reinkrul, line 285: use `senderDID` instead of `pal[0]`

network/transport/v2/protocol.go
```diff
@@ -157,44 +203,96 @@ func (p *protocol) Start() {
...
+
 		return nil
 	}
 
-	conn := p.connectionList.Get(grpc.ByConnected(), grpc.ByNodeDID(pal[0]))
+	conn := p.connectionList.Get(grpc.ByConnected(), grpc.ByNodeDID(senderDID))
 
 	if conn == nil {
-		return fmt.Errorf("unable to retrieve payload, no connection found (ref=%s, DID=%s)", tx.Ref().String(), pal[0])
+		return fmt.Errorf("unable to retrieve payload, no connection found (ref=%s, DID=%s)", hash.String(), pal[0])
```
reinkrul, line 276: use `senderDID` instead of `pal[0]`

reinkrul (changes requested):

reinkrul (approved):

Issue #711, opened 2021-12-30
Nats consumer not setup correctly


woutslakhorst:
starting the node gives:

```
ERRO[0002] failed to setup NATS handler: nats: no stream matches subject 
```

When looking at https://github.com/nats-io/nats.go/blob/main/js_test.go, several are things are missing:

- A stream has to be setup. How to name a stream? Which subjects can the stream contain?
- Consumers have to be setup. They have to be unique. (Also over a cluster)
- Subscriptions have to be bound to the consumer/stream.
- The right options have to be set for our use-case.

A minimal example to have the server started and events delivered:

```
func (p *protocol) setupNatsHandler() error {
	_, js, err := p.eventsConnectionPool.Acquire(p.ctx)
	if err != nil {
		return err
	}

	if _, err = js.AddStream(&nats.StreamConfig{Name: "privates"}); err != nil { // todo stream name
		return err
	}

	// todo name
	todoId := make([]byte, 32)
	rand.Read(todoId)
	consumerId := hash.SHA256Sum(todoId).String()

	if _, err = js.AddConsumer(
		"privates",
		&nats.ConsumerConfig{
			Durable:        consumerId,
			DeliverSubject: events.PrivateTransactionsSubject,
			AckPolicy:      nats.AckExplicitPolicy,
		}); err != nil { // todo consumer config
		return err
	}

	if _, err = js.Subscribe(events.PrivateTransactionsSubject, func(msg *nats.Msg) {
		if err := p.handlePrivateTx(msg); err != nil {
			log.Logger().Errorf("failed to handle private transaction: %v", err)

			if err := msg.Nak(); err != nil {
				log.Logger().Errorf("failed to NACK private transaction event: %v", err)
			}

			return
		}

		if err := msg.Ack(); err != nil {
			log.Logger().Errorf("failed to ACK private transaction event: %v", err)
		}
	}, nats.AckExplicit(), nats.Bind("privates", consumerId)); err != nil { // todo set consumer name and binding!
		return err
	}

	return nil
}
```

Proposal for current state of the node is to remove the event queue from the private TX sending and use a direct call. Using this PR: #708. This will most likely require the V2 protocol to bind to the replayingDAGPublisher like the v1 protocol does.

dmeijboom:
It's true that there are some things missing (see: #717).

> A stream has to be setup. How to name a stream? Which subjects can the stream contain?

[...]

PR #712, opened 2021-12-30
make direct call to V2 handlePrivateTx from dag

woutslakhorst:
handles problem from #711

It calls the V2 protocol from a DAG subscriber to fetch a private TX.
As you can tell, the subscribers setup is dodgy and will have to be replaced with a proper event mechanism.
This was the purpose of Nats but that part is missing proper design for now.

- [x] cherry-pick #739 
- [x] cherry-pick #740 (splitted callbacks)
- [x] cherry-pick #744 (did doc creation)
- [x] cherry-pick #741 (publisher retry change)


dmeijboom:
I'll postpone reviewing this until #708 is merged 😅

vcr/ambassador.go
```diff
@@ -59,6 +59,11 @@ func (n ambassador) Configure() {
 // The VCR is used to verify the contents of the credential.
 // payload should be a json encoded vc.VerifiableCredential
 func (n ambassador) vcCallback(tx dag.Transaction, payload []byte) error {
+	if payload == nil {
+		// ignore first call, only act on second call (with payload)
+		return nil
+	}
```
dmeijboom, line 65: Why is this required?

vdr/ambassador.go
```diff
@@ -83,6 +83,11 @@ var thumbprintAlg = crypto.SHA256
 // The rules are based on the Nuts RFC006
 // payload should be a json encoded did.document
 func (n *ambassador) callback(tx dag.Transaction, payload []byte) error {
+	if payload == nil {
+		// ignore first call, only act on second call (with payload)
+		return nil
+	}
```
dmeijboom, line 89: Why is this required?

dmeijboom (changes requested):

woutslakhorst:
The e2e tests found a problem..... and they are correct. More things to think about....

vcr/ambassador.go
woutslakhorst: has been removed

vdr/ambassador.go
woutslakhorst: has been removed

dmeijboom (approved):
LGTM

test/predicates.go
```diff
@@ -41,6 +42,23 @@ func WaitFor(t *testing.T, p Predicate, timeout time.Duration, message string, m
 			assert.Fail(t, fmt.Sprintf(message, msgArgs...))
 			return false
 		}
-		time.Sleep(50 * time.Millisecond)
+		time.Sleep(10 * time.Millisecond)
+	}
+}
+
+func WaitForNoFail(t *testing.T, p Predicate, timeout time.Duration) bool {
```
reinkrul, line 49: Let's not wait too long ;-)

reinkrul (approved):
Much better now.

network/dag/publisher.go
```diff
@@ -77,7 +75,7 @@ func (s *replayingDAGPublisher) transactionAdded(ctx context.Context, transactio
 	s.emitEvent(TransactionAddedEvent, tx, nil)
 
 	// Received new transaction, add it to the subscription walker resume list, so it resumes from this transaction
-	// when the payload is received.
+	// when the payload is received. This block so it may only be added for TX where we know the payload will come.
```
reinkrul, line 78:
```suggestion
	// when the payload is received. This blocks so it may only be added for TX where we know the payload will come.
```

reinkrul (approved):

network/dag/publisher.go
```diff
@@ -134,6 +135,21 @@ func (s *replayingDAGPublisher) publish(ctx context.Context) {
...
 
+func remove(l *list.List, ref hash.SHA256Hash) {
+	current := l.Front()
+
+	for {
+		if current == nil {
+			return
+		}
+		if current.Value.(hash.SHA256Hash).Equals(ref) {
+			l.Remove(current)
```
reinkrul, line 146: could there be duplicate resume points? If so, we could loop over the entire list, trying to remove duplicates

woutslakhorst: Very unlikely the transactionAdded callback is called twice. Even so, doesn't hurt to remove the `return` here.

dmeijboom:
@woutslakhorst @reinkrul Looks like this works with the latest Demo EHR. Also verified it actually uses a private transaction. Setup with three nodes where node 1 performed an e-transfer to node 2 and node 3 wasn't able to decrypt the PAL header. Nice work 🎉

Issue #480, opened 2021-09-28
Network: properly shut down goroutines


reinkrul:
`Protocol.Start()` spawns goroutines for background processes, but there's no mechanism to stop them when the Nuts node process is shutting down. Nothing seems to be really going wrong, but it's not very nice either (since it may lead to some strange error messages when shutting down).

There needs to be a mechanism that stops the goroutines when `Protocol.Stop()` is called.

This might also be the case for the other network layers, so better check those as well.

PR #715, opened 2022-01-03
Network: properly shutdown goroutines in v1 protocol

reinkrul:
Thought this might be a problem in the current build failing now and then (deadlock), but on a second thought it's probably not the cause. Fixed it nonetheless.

Fixes https://github.com/nuts-foundation/nuts-node/issues/480


dmeijboom (approved):
LGTM

PR #722, opened 2022-01-07
VCR: Revert from issuing private VCs privately, because it's broken

reinkrul:


dmeijboom:
Wouldn't it be better to disable private transactions with a flag? For now this should be fine but it feels a bit unsafe to have a flag which can potentially "break privacy laws" as private data will be sent without encryption

stevenvegt (approved):
Not sure about the disabled tests, perhaps just skip them?
Not sure about the flag name, perhaps something like: `enablePrivateTxs`. Which is now default false, will default true en removed in the future?
If you want you can ignore this feedback en just merge the PR.

woutslakhorst (approved):

woutslakhorst:
As soon as private TXs work, this can be removed again

Issue #723, opened 2022-01-10
NodeDID not set on connection by connection manager


woutslakhorst:
For both incoming and outgoing connections the NodeDID is authenticated but the connection's peer is not updated with the new NodeDID value.

in `grpcConnectionManager.openOutboundStream` and `grpcConnectionManager.handleInboundStream` both call `authenticate` which results in a peer with NodeDID. This updated peer is then not set on the connection!

PR #724, opened 2022-01-10
Actually set the authenticated peer on the connection

dmeijboom:
Closes: https://github.com/nuts-foundation/nuts-node/issues/723

### DIFF
--- a/network/transport/grpc/connection_manager.go
+++ b/network/transport/grpc/connection_manager.go
@@ -316,13 +316,16 @@ func (s *grpcConnectionManager) openOutboundStream(connection Connection, protoc
 		return nil, fatalError{error: fmt.Errorf("peer sent invalid ID (id=%s)", peerID)}
 	}
 
-	//when bootstrap node, this instance has the AcceptUnauthenticated param
+	// When bootstrap node, this instance has the AcceptUnauthenticated param
 	peer := connection.Peer()
 	peerFromCtx, _ := grpcPeer.FromContext(clientStream.Context())
-	peer, err = s.authenticate(nodeDID, peer, peerFromCtx)
+
+	authenticatedPeer, err := s.authenticate(nodeDID, peer, peerFromCtx)
 	if err != nil {
 		return nil, fatalError{error: err}
 	}
+
+	connection.setPeer(authenticatedPeer)
 
 	if !connection.registerStream(protocol, clientStream) {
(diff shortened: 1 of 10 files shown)


woutslakhorst (approved):

reinkrul:
Note this is also fixed in https://github.com/nuts-foundation/nuts-node/pull/708

network/transport/grpc/connection_manager.go
```diff
@@ -316,14 +316,17 @@ func (s *grpcConnectionManager) openOutboundStream(connection Connection, protoc
...
 	peer := connection.Peer()
 	peerFromCtx, _ := grpcPeer.FromContext(clientStream.Context())
-	peer, err = s.authenticate(nodeDID, peer, peerFromCtx)
+
+	authenticatedPeer, err := s.authenticate(nodeDID, peer, peerFromCtx)
 	if err != nil {
 		return nil, fatalError{error: err}
 	}
 
+	connection.setPeer(authenticatedPeer)
```
reinkrul, line 328: assert in test

reinkrul (changes requested):

network/transport/grpc/connection_manager_test.go
```diff
@@ -303,6 +303,47 @@ func Test_grpcConnectionManager_openOutboundStreams(t *testing.T) {
...
+	t.Run("ok", func(t *testing.T) {
+		ctrl := gomock.NewController(t)
+
+		grpcPeer := &peer.Peer{}
+		peerInfo := transport.Peer{}
+		ctx := peer.NewContext(context.TODO(), grpcPeer)
+		nodeDID, _ := did.ParseDID("did:nuts:test")
+
+		authenticator := NewMockAuthenticator(ctrl)
+		authenticator.EXPECT().Authenticate(*nodeDID, *grpcPeer, peerInfo).Return(peerInfo, nil)
```
reinkrul, line 318: I'd return a different peerInfo with nodeDID populated ("authenticated" peer), because that's provided to `setPeer`. Because now `setPeer` can be fed the unauthenticated `Peer`, and the test will still succeed.

reinkrul (changes requested):

network/transport/grpc/connection_manager_test.go
dmeijboom: Good one, will fix!

reinkrul (approved):

PR #725, opened 2022-01-10
Network: fix node DID not being passed to v2

reinkrul:
Taken from https://github.com/nuts-foundation/nuts-node/pull/708/commits/b1fc658016b22bd13cc8b5b94dea465e82c79c33

The `nodeDIDResolver` passed to `v2.New()` (create v2 protocol instance) got passed the default, emtpy node DID resolver. After that, it's overwritten with a resolver that contains the actual node DID. But since it's overwritten, v2 doesn't get the new, populated node DID resolver. Leading to node DID not being populated when encountering a private TX. Simply moving the initialization be before the protocol init fixes this.

Not really unit testable, since it calls package functions (`v2.New`), which can't be mocked. It's also a private member of v2 protocol, so not accessible for assertion. Integration-testable by https://github.com/nuts-foundation/nuts-node/pull/708

### DIFF
--- a/network/network.go
+++ b/network/network.go
@@ -107,6 +107,7 @@ func NewNetworkInstance(
 		didDocumentResolver:    didDocumentResolver,
 		eventManager:           eventManager,
 		lastTransactionTracker: lastTransactionTracker{headRefs: make(map[hash.SHA256Hash]bool)},
+		nodeDIDResolver:        &transport.FixedNodeDIDResolver{},
 	}
 }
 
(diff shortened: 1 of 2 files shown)


network/network.go
```diff
@@ -148,6 +147,17 @@ func (n *Network) Configure(config core.ServerConfig) error {
...
 
+	// Resolve node DID
+	if n.config.NodeDID != "" {
+		configuredNodeDID, err := did.ParseDID(n.config.NodeDID)
+		if err != nil {
+			return fmt.Errorf("configured NodeDID is invalid: %w", err)
+		}
+		n.nodeDIDResolver = &transport.FixedNodeDIDResolver{NodeDID: *configuredNodeDID}
+	} else {
+		n.nodeDIDResolver = &transport.FixedNodeDIDResolver{}
```
woutslakhorst, line 158: leave this one in the initializer?

woutslakhorst (changes requested):

reinkrul:
Not blocking at this moment, also fixed in one of the succeeding PRs.

woutslakhorst (approved):

PR #730, opened 2022-01-11
Return session not found in IRMA signer if a session was not found

dmeijboom:
Hotfix for now. This should be 'redesigned' as currently we are checking every signer for a specific error (whether a session was found or not) until one either returns no error or another error. This will break when third-party libraries (such as IRMA) decide to raise other errors but with the same meaning.

### DIFF
--- a/auth/services/irma/signer.go
+++ b/auth/services/irma/signer.go
@@ -22,6 +22,7 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
+	"github.com/privacybydesign/irmago/server/irmaserver"
 	"os"
 	"strings"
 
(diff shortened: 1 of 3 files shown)


auth/services/irma/signer.go
```diff
@@ -112,6 +113,10 @@ func (v Service) StartSigningSession(rawContractText string) (contract.SessionPo
 func (v Service) SigningSessionStatus(sessionID string) (contract.SigningSessionResult, error) {
 	result, err := v.IrmaSessionHandler.GetSessionResult(sessionID)
 	if err != nil {
+		if _, ok := err.(*irmaserver.UnknownSessionError); ok {
```
reinkrul, line 116: untested?

reinkrul (changes requested):

reinkrul (approved):

PR #739, opened 2022-01-13
added robustness checks and tests

woutslakhorst:
This PR is just to make sure all listeners to the dag.Publisher can handle duplicate callbacks.

After merge create issue to handle ever growing processed maps.

### DIFF
--- a/vdr/ambassador.go
+++ b/vdr/ambassador.go
@@ -79,9 +79,10 @@ func (n *ambassador) Configure() {
 var thumbprintAlg = crypto.SHA256
 
 // callback gets called when new DIDDocuments are received by the network. All checks on the signature are already performed.
-// This method will check the integrity of the DID document related to the public key used to sign the network tr.
+// This method will check the integrity of the DID document related to the public key used to sign the network TX.
 // The rules are based on the Nuts RFC006
 // payload should be a json encoded did.document
+// Duplicates are handled as updates and will be merged. Merging two exactly the same DID Documents results in the original document.
 func (n *ambassador) callback(tx dag.Transaction, payload []byte) error {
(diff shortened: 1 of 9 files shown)


vdr/ambassador.go
```diff
@@ -79,9 +79,10 @@ func (n *ambassador) Configure() {
 var thumbprintAlg = crypto.SHA256
 
 // callback gets called when new DIDDocuments are received by the network. All checks on the signature are already performed.
-// This method will check the integrity of the DID document related to the public key used to sign the network tr.
+// This method will check the integrity of the DID document related to the public key used to sign the network TX.
 // The rules are based on the Nuts RFC006
 // payload should be a json encoded did.document
+// Duplicates are handled as updates and will be merged. Merging two exactly the same DID Documents results in the original document.
```
reinkrul, line 85: But it could still cause a conflict when an older version it published later on for the second time (e.g. A, B, C, then A again). But for VDR not an issue since it requires guaranteed order of DID document versions.

So that's another requirement/property of the design: for attached payloads, transaction-events MAY be published multiple times but NOT out-of-order?

woutslakhorst: SHOULD NOT be published multiple times and MAY NOT be processed out of order.

woutslakhorst: Actually it's a requirement that differs per component. The VCR doesn't care, it uses a tombstone set. The VDR requires the latest TX to point to the current TX, if not it'll mark it as a conflict. The network.blocks and lastTransactionTracker do require the correct order. But ordering is guaranteed by the LC's and thus the DAG walker?

reinkrul (approved):

PR #740, opened 2022-01-13
Network: split publisher events in TransactionAdded and PayloadAdded

reinkrul:


network/network.go
```diff
@@ -227,7 +227,7 @@ func (n *Network) Start() error {
 	n.startTime.Store(time.Now())
 
 	// Load DAG and start publishing
-	n.publisher.Subscribe(dag.AnyPayloadType, n.lastTransactionTracker.process)
+	n.publisher.Subscribe(dag.TransactionPayloadAddedEvent, dag.AnyPayloadType, n.lastTransactionTracker.process)
```
woutslakhorst, line 230: should be TransactionAddedEvent

network/dag/publisher.go
```diff
@@ -176,13 +179,22 @@ func (s *replayingDAGPublisher) publishTransaction(ctx context.Context, transact
...
 	}
 
-	for _, payloadType := range []string{transaction.PayloadType(), AnyPayloadType} {
-		receiver := s.subscribers[payloadType]
-		if receiver == nil {
-			continue
-		}
-		if err := receiver(transaction, payload); err != nil {
-			log.Logger().Errorf("Transaction subscriber returned an error (ref=%s,type=%s): %v", transaction.Ref(), transaction.PayloadType(), err)
+	// TODO: Now calls TransactionAddedEvent and TransactionPayloadAddedEvent after checken whether the payload is present.
```
woutslakhorst, line 182: Also, the TransactionAddedEvents may be blocked by a missing payload of a did.Document transaction

woutslakhorst (changes requested):
heads eligible to be prevs should listen to TransactionAddedEvent

network/network.go
reinkrul: As long as all previous payloads must be present in all cases `prevs` MUST refer to TXs which payload is guaranteed to be present. Otherwise the publisher will not directly publish (or maybe never, if the payload is forever lost) the TX (added by the local node through `CreateTransaction`). This is what `lastTransactionTracker` tracks.

Ofc might change in future or be phased out completely when TXs are guaranteed to have their payloads, due to atomic sync (where applicable)

reinkrul: As long as all previous payloads must be present in all cases `prevs` MUST refer to TXs which payload is guaranteed to be present. Otherwise the publisher will not directly publish (or maybe never, if the payload is forever lost) the TX (added by the local node through `CreateTransaction`). This is what `lastTransactionTracker` tracks. Thus, it needs to listen to the `TransactionPayloadEvent`.

Ofc might change in future or be phased out completely when TXs are guaranteed to have their payloads, due to atomic sync (where applicable)

woutslakhorst: So private TX will only be referred to by other TX of the same node, which is acceptable for now.

woutslakhorst (approved):

PR #742, opened 2022-01-14
Network: don't notify observers when TX isn't added to DAG

reinkrul:
DAG observers are currently notified when a TX was passed to `Add()` which already existed on the DAG. This never caused issues, but https://github.com/nuts-foundation/nuts-node/pull/741 assumes its `transactionAdded()` function is called only once per TX, because otherwise it emits duplicate `TransactionAddedEvent`s.

It would not be a big issue since all subscribers should be idempotent, but it might cause unneeded logging/confusion/load. Thus we should avoid if we can do so with relative ease.


woutslakhorst (approved):
Also works with dev

PR #748, opened 2022-01-14
Refactor VCR using a seperate issuer role.

stevenvegt:
Create a Issuer which implements the Issuer role from the W3C Credential specification.

Screenshot 2022-01-14 at 17 05 51

It introduces a api/v2 which introduces a `/issuer/` path.
Also it stores issued credentials in its own store.
Both v1 and v2 uses the new issuer.

The vulnarabilities found by codeQL are not really an issue since it is merely a debug statement. I addressed it in the https://github.com/sirupsen/logrus/issues/1307 issue

## TODO:
* [x] Issue credentials
* [x] Store issued credentials
* [x] Search for issued credentials

To keep this PR understandable, the following issuer related operations should be implemented in other PRs.
* [ ] Implement Revocations
* [ ] Store revocations
* [ ] Search revocations


docs/_static/vcr/v2.yaml
```diff
@@ -0,0 +1,245 @@
...
+          $ref: '../common/error_response.yaml'
+    delete:
+      summary: "Revoke an issued credential"
+      description: |
+        Revoke a credential.
+
+        error returns:
+        * 400 - Credential can't be revoked. Most likely due to a missing private key.
+        * 404 - Credential is not found
+        * 409 - Credential has already been revoked
```
reinkrul, line 85: did we already do this? Because revoking twice shouldn't really make a difference?

docs/_static/vcr/v2.yaml
```diff
@@ -0,0 +1,245 @@
...
+      responses:
+        "200":
+          description: Revocation has been processed locally. It has also been published to the network.
+          content:
+            application/vc+json;type=revocation:
+              schema:
+                $ref: '#/components/schemas/Revocation'
+        default:
+          $ref: '../common/error_response.yaml'
+  /internal/vcr/v2/verifier/vc:
```
reinkrul, line 99: who will use this API?

docs/_static/vcr/v2.yaml
```diff
@@ -0,0 +1,245 @@
...
+          description: Type definition for the credential.
+          type: string
+          example: "NutsOrganizationCredential"
+        issuer:
+          description: DID according to Nuts specification.
+          type: string
+          example: "did:nuts:B8PUHs2AUHbFF1xLLK4eZjgErEcMXHxs68FteY7NDtCY"
+        expirationDate:
+          description: rfc3339 time string until when the credential is valid.
+          type: string
```
reinkrul, line 148: `format: date-time`

docs/_static/vcr/v2.yaml
```diff
@@ -0,0 +1,245 @@
...
+          description: List of URIs
+          type: array
+          items:
+            type: string
+        id:
+          description: credential ID. A Nuts DID followed by a large number.
+          example: "did:nuts:B8PUHs2AUHbFF1xLLK4eZjgErEcMXHxs68FteY7NDtCY"
+          type: string
+        type:
+          description: List of type definitions for the credential. Always includes 'VerifiableCredential'
```
reinkrul, line 173: must always include, or is it added by the node?

docs/_static/vcr/v2.yaml
```diff
@@ -0,0 +1,245 @@
...
+        - credentialSubject
+        - proof
+      properties:
+        "@context":
+          description: List of URIs
+          type: array
+          items:
+            type: string
+        id:
+          description: credential ID. A Nuts DID followed by a large number.
```
reinkrul, line 169: is this required, if we might support other VCs in (near) future?

reinkrul (changes requested):

vcr/issuer/interface.go
```diff
@@ -0,0 +1,36 @@
...
+type Publisher interface {
+	PublishCredential(verifiableCredential vc.VerifiableCredential, public bool) error
+	PublishRevocation(revocation credential.Revocation) error
+}
+
+type keyResolver interface {
+	ResolveAssertionKey(issuerDID did.DID) (crypto.Key, error)
+}
+
+// Store defines the interface for a issuer store. An implementation stores all the issued credentials and the revocations.
```
woutslakhorst, line 29: I would describe it as the store for the "user" not the issuer.

vcr/issuer/networkPublisher.go
```diff
@@ -40,25 +45,20 @@ func (p networkPublisher) PublishCredential(verifiableCredential vc.VerifiableCr
 	}
 	issuerDID, _ := did.ParseDIDURL(verifiableCredential.Issuer.String())
 
+	key, err := p.keyResolver.ResolveAssertionKey(*issuerDID)
```
woutslakhorst, line 48: unprocessed error

docs/_static/vcr/v2.yaml
```diff
@@ -0,0 +1,245 @@
...
+        * 404 - Corresponding credential could not be found
+        * 500 - An error occurred while processing the request
+      operationId: "resolveIssuedVC"
+      parameters:
+        - name: resolveTime
+          in: query
+          description: a rfc3339 time string for resolving a VC at a specific moment in time
+          example: "2012-01-02T12:00:00Z"
+          schema:
+            type: string
```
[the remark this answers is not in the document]
woutslakhorst: > `format: date-time`

no please don't, language incompatibility

docs/_static/vcr/v2.yaml
woutslakhorst: again no

docs/_static/vcr/v2.yaml
```diff
@@ -0,0 +1,245 @@
...
+        default:
+          $ref: '../common/error_response.yaml'
+
+
+components:
+  schemas:
+    IssueVCRequest:
+      type: object
+      description: A request for issuing a new Verifiable Credential.
+      required:
```
woutslakhorst, line 133: I thought @context would be required as well?

vcr/issuer/keyresolver.go
```diff
@@ -0,0 +1,31 @@
...
+	vdr "github.com/nuts-foundation/nuts-node/vdr/types"
+)
+
+// vdrKeyResolver resolves private keys based upon the VDR document resolver
+type vdrKeyResolver struct {
+	docResolver vdr.DocResolver
+	keyResolver crypto.KeyResolver
+}
+
+func (r vdrKeyResolver) ResolveAssertionKey(issuerDID did.DID) (crypto.Key, error) {
```
woutslakhorst, line 17: No possibility to reuse the resolver in vdr/doc?

woutslakhorst (changes requested):
A first (small) review, some small changes required. Didn't review it all (yet).
First impression is that there will be 2 stores, 1 from your own issued Creds and 1 incoming from the network. Although this will duplicate some records, it'll help in making backups since only one of the two stores requires a backup. This will greatly reduce the size of the backup and the complexity in making it. 👍

reinkrul:
> A first (small) review, some small changes required. Didn't review it all (yet). First impression is that there will be 2 stores, 1 from your own issued Creds and 1 incoming from the network. Although this will duplicate some records, it'll help in making backups since only one of the two stores requires a backup. This will greatly reduce the size of the backup and the complexity in making it. 👍

These can probably have very similar, or the same implementations and/or API? Or same implementation, slightly different interfaces.

docs/_static/vcr/v2.yaml
```diff
@@ -0,0 +1,245 @@
...
+        "200":
+          description: "New VC has been created successfully. Returns the Verifiable Credential."
+          content:
+            application/vc+json:
+              schema:
+                $ref: '#/components/schemas/VerifiableCredential'
+        default:
+          $ref: '../common/error_response.yaml'
+
+  /internal/vcr/v2/issuer/vc/{id}:
```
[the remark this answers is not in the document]
stevenvegt: It only resolves VC issued by this node.
The store uses a leia backend and cannot be clustered. But we can implement other backends when the need for clustering arrives.

docs/_static/vcr/v2.yaml
stevenvegt: Seems strange to allow a revocation twice. This will publish the transaction again. I don't see any advantages to that over checking for existing revocation and failing if already revoked.

docs/_static/vcr/v2.yaml
stevenvegt: A verifier. So, someone who has received a VC and want to verify it.

vcr/issuer/interface.go
stevenvegt: User is imho to ambiguous, since the user can also be a holder or verifier. This store stores all the information for the issuer role, such as issued credentials and their revocations.

vcr/issuer/keyresolver.go
stevenvegt: The vdr does not return the private key. This is a combination of both the vdr and the private key store. It's goal is to encapsulate the key resolving proces for easier setup and testing.

woutslakhorst:
Also for another PR:

- [ ] update docs
- [ ] update workshop descriptions
- [ ] communicate on Slack

cmd/root.go
```diff
@@ -183,7 +184,10 @@ func CreateSystem() *core.System {
 		Updater:    vdrInstance,
 		Resolver:   docResolver,
 	}})
-	system.RegisterRoutes(&credAPI.Wrapper{CR: credentialInstance.Registry(), R: credentialInstance})
+	system.RegisterRoutes(&credAPIv1.Wrapper{CR: credentialInstance.Registry(), R: credentialInstance})
+	system.RegisterRoutes(&credAPIv2.Wrapper{
+		VCR: credentialInstance,
+	})
```
woutslakhorst, line 190: inconsistent indentation

docs/_static/vcr/v2.yaml
```diff
@@ -0,0 +1,281 @@
+openapi: "3.0.0"
+info:
+  title: Nuts Verifiable Credential API spec
+  description: API specification for Verifiable Credentials
```
woutslakhorst, line 4: Too generic; "API specification for creating, revoking and searching......"

docs/_static/vcr/v2.yaml
```diff
@@ -0,0 +1,281 @@
...
+        - name: issuer
+          in: query
+          description: the did of the issuer
+          example: did:nuts:123
+          required: true
+          schema:
+            type: string
+        - name: subject
+          in: query
+          description: the did of the issuer
```
woutslakhorst, line 67: incorrect

docs/_static/vcr/v2.yaml
```diff
@@ -0,0 +1,281 @@
...
+        "200":
+          description: "New VC has been created successfully. Returns the Verifiable Credential."
+          content:
+            application/vc+json:
+              schema:
+                $ref: '#/components/schemas/VerifiableCredential'
+        default:
+          $ref: '../common/error_response.yaml'
+
+  /internal/vcr/v2/issuer/vc/search:
```
woutslakhorst, line 39: The type of action and optionality of the query params suggests a List return. Both description and return value do not.

docs/_static/vcr/v2.yaml
```diff
@@ -0,0 +1,281 @@
...
+        default:
+          $ref: '../common/error_response.yaml'
+  /internal/vcr/v2/issuer/vc/{id}:
+    parameters:
+      - name: id
+        in: path
+        description: URL encoded ID.
+        required: true
+        example:
+          - "did:nuts:B8PUHs2AUHbFF1xLLK4eZjgErEcMXHxs68FteY7NDtCY#90382475609238467"
```
woutslakhorst, line 90: incorrect example

docs/_static/vcr/v2.yaml
woutslakhorst: It doesn't hurt and is safe to do it twice. But it's a client error since the calling software obvious has a bug (or poor design)

docs/_static/vcr/v2.yaml
```diff
@@ -0,0 +1,281 @@
...
+            application/vc+json;type=revocation:
+              schema:
+                $ref: '#/components/schemas/Revocation'
+        default:
+          $ref: '../common/error_response.yaml'
+  /internal/vcr/v2/verifier/vc:
+    post:
+      summary: Verifies a Verifiable Credential
+      description: |
+        Verifies a Verifiable Credential: It checks the signature, expiration, revocation status and if the issuer is trusted.
```
woutslakhorst, line 119: And validity of the issuer at the time of issuance?

docs/_static/vcr/v2.yaml
```diff
@@ -0,0 +1,281 @@
...
+      type: object
+      description: A request for issuing a new Verifiable Credential.
+      required:
+        - type
+        - issuer
+        - credentialSubject
+      properties:
+        "@context":
+          description: |
+            The resolvable context of the credentialSubject as URI. If omitted, the "https://nuts.nl/credentials/v1" context is used.
```
woutslakhorst, line 156: What if you add an unknown VC type with an external resolvable context?

docs/_static/vcr/v2.yaml
```diff
@@ -0,0 +1,281 @@
...
+            When set to false, the caller is responsible for distributing the VC to a holder. When the issuer is
+            also the holder, it then can be used to directly create a presentation (self issued).
+          type: boolean
+          default: true
+        visibility:
+            description: |
+              When publishToNetwork is true, the credential can be published publicly of privately to the holder.
+            type: string
+            enum: [ public, private ]
+            default: private
```
woutslakhorst, line 185: devs will forget this for NutsOrganizationCredentials. It's currently taken from the type settings. A different option would be to make this param mandatory.

woutslakhorst (changes requested):
- [ ] API spec not added to swagger docs
- [ ] Getting started not updated (New issue?)

vcr/issuer/networkPublisher.go
```diff
@@ -0,0 +1,132 @@
...
+		}
+		if err != nil {
+			return nil, fmt.Errorf("failed to determine credentialSubject.ID: %w", err)
+		}
+
+		// participants are not the issuer and the credentialSubject.id but the DID that holds the concrete endpoint for the NutsComm service
+		for _, vcp := range []did.DID{*issuer, *credentialSubjectID} {
+			serviceOwner, err := p.resolveNutsCommServiceOwner(vcp)
+			if err != nil {
+				return nil, fmt.Errorf("could not publish private credential: failed to resolve participating node (did=%s): %w", vcp.String(), err)
```
woutslakhorst, line 105: The `could not publish` part should be added by the caller of this method. This method can't possibly know it's used for a single purpose.

vcr/issuer/leiaStore.go
```diff
@@ -0,0 +1,82 @@
...
+	panic("implement me")
+}
+
+// createIndices creates the needed indices for the issued VC store
+// It allows faster searching on context, type issuer and subject values.
+func (s leiaStore) createIndices() error {
+	index := leia.NewIndex("issuedVCs",
+		leia.NewFieldIndexer("issuer"),
+		leia.NewFieldIndexer("type"),
+		leia.NewFieldIndexer("@context"),
```
woutslakhorst, line 78: Does the context have sufficient values opposed to types to make a difference? Not all fields need to be indexed.

vcr/issuer/leiaStore_test.go
```diff
@@ -0,0 +1,106 @@
...
+		testDir := io.TestDirectory(t)
+		sut, err := NewLeiaStore(testDir)
+
+		assert.NoError(t, err)
+		assert.IsType(t, &leiaStore{}, sut)
+	})
+
+	t.Run("error", func(t *testing.T) {
+		// the forward slash is the only invalid dirname and seems the simplest way of causing an error.
+		sut, err := NewLeiaStore("/")
```
woutslakhorst, line 25: you can also use the filename of this test

vcr/issuer/issuer.go
```diff
@@ -0,0 +1,121 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	validator, _ := credential.FindValidatorAndBuilder(*createdVC)
+	if err := validator.Validate(*createdVC); err != nil {
+		return nil, err
+	}
+
+	// TODO: Store credential in the store
```
woutslakhorst, line 49: remove todo

woutslakhorst (changes requested):

docs/_static/vcr/v2.yaml
```diff
@@ -0,0 +1,284 @@
...
+        proof:
+          description: one or multiple cryptographic proofs
+          type: object
+    ResolutionResult:
+      type: object
+      description: result of a Resolve operation.
+      required:
+        - verifiableCredential
+      properties:
+        revocationDate:
```
reinkrul, line 238: why not return the revocation here?

docs/_static/vcr/v2.yaml
```diff
@@ -0,0 +1,284 @@
...
+      description: Credential revocation record
+      required:
+        - issuer
+        - subject
+        - date
+      properties:
+        issuer:
+          $ref: '#/components/schemas/DID'
+        subject:
+          type: string
```
reinkrul, line 255: subject might also be an object instead of a string?

Oh nevermind. I see this is the VC ID? A bit confusing.

docs/_static/vcr/v2.yaml
```diff
@@ -0,0 +1,284 @@
...
+      description: Subject of a Verifiable Credential identifying the holder and expressing claims.
+    VCVerificationResult:
+      description: Contains the verifiable credential verification result.
+      type: object
+      required:
+        - validity
+      properties:
+        validity:
+          type: boolean
+          description: Indicates the validity of the signature, issuer and revokation state.
```
reinkrul, line 281:
```suggestion
          description: Indicates the validity of the signature, issuer and revocation state.
```

docs/_static/vcr/v2.yaml
```diff
@@ -0,0 +1,284 @@
...
+          content:
+            application/vc+json:
+              schema:
+                $ref: '#/components/schemas/VerifiableCredential'
+        default:
+          $ref: '../common/error_response.yaml'
+
+  /internal/vcr/v2/issuer/vc/search:
+    get:
+      summary: "Resolves verifiable credentials issued by this node which matches the search params"
```
reinkrul, line 43: searches?

docs/_static/vcr/v2.yaml
```diff
@@ -0,0 +1,284 @@
...
+      description: >
+        The resolutionResult contains the credential regardless of the validity.
+        The resolutionResult.currentStatus must be checked to see if the credential is valid.
+
+        error returns:
+        * 404 - Corresponding credential could not be found
+        * 500 - An error occurred while processing the request
+      operationId: "resolveIssuedVC"
```
reinkrul, line 51: searchIssuedVC?

docs/_static/vcr/v2.yaml
```diff
@@ -0,0 +1,284 @@
...
+        - credential
+      responses:
+        "200":
+          description: A list of matching credentials
+          content:
+            application/vc+json:
+              schema:
+                type: array
+                items:
+                    $ref: '#/components/schemas/ResolutionResult'
```
reinkrul, line 84: SearchResult?

vcr/api/v2/api.go
```diff
@@ -0,0 +1,126 @@
...
+
+// VerifyVC handles API request to verify a  Verifiable Credential.
+func (w *Wrapper) VerifyVC(ctx echo.Context) error {
+	requestedVC := VerifiableCredential{}
+
+	if err := ctx.Bind(&requestedVC); err != nil {
+		return err
+	}
+
+	if err := w.VCR.Validate(requestedVC, true, true, nil); err != nil {
```
reinkrul, line 121: why is `allowUntrusted` by default `true`?

vcr/issuer/interface.go
```diff
@@ -0,0 +1,43 @@
...
+	"github.com/nuts-foundation/nuts-node/vcr/credential"
+)
+
+// Publisher publishes new credentials and revocations to a channel. Used by a credential issuer.
+type Publisher interface {
+	PublishCredential(verifiableCredential vc.VerifiableCredential, public bool) error
+	PublishRevocation(revocation credential.Revocation) error
+}
+
+type keyResolver interface {
```
reinkrul, line 17: why not use the VDR `KeyResolver`?

vcr/issuer/interface.go
```diff
@@ -0,0 +1,43 @@
...
+	ResolveAssertionKey(issuerDID did.DID) (crypto.Key, error)
+}
+
+// Issuer is a role in the network for a party who issues credentials about a subject to a holder.
+type Issuer interface {
+	Issue(unsignedCredential vc.VerifiableCredential, publish, public bool) (*vc.VerifiableCredential, error)
+	Revoke(credentialID ssi.URI) error
+	CredentialResolver() StoreResolver
```
reinkrul, line 25: Feels like bad design to return an internal type/object, especially a store.

"Do not call friends of friends" (do not call methods on dependencies of a type's dependencies)

vcr/issuer/interface.go
```diff
@@ -0,0 +1,43 @@
...
+	// StoreCredential writes a VC to storage.
+	StoreCredential(vc vc.VerifiableCredential) error
+	// StoreRevocation writes a revocation to storage.
+	StoreRevocation(r credential.Revocation) error
+	StoreResolver
+}
+
+// StoreResolver defines the functions to resolve or search for credentials.
+// It is a separate interface from Store so when an object only needs resolving, it only needs the resolver.
+type StoreResolver interface {
```
reinkrul, line 40: It doesn't "resolve stores", so it should be named "CredentialResolver" or (judging the API), `CredentialSearcher`.

vcr/issuer/issuer.go
reinkrul, line 49: TODO?

vcr/issuer/issuer.go
```diff
@@ -0,0 +1,121 @@
...
+	return &issuer{
+		store:       store,
+		Publisher:   publisher,
+		keyResolver: resolver,
+	}
+}
+
+type issuer struct {
+	store       Store
+	Publisher   Publisher
```
reinkrul, line 31: why exported?

vcr/issuer/issuer.go
```diff
@@ -0,0 +1,121 @@
...
+	resolver := vdrKeyResolver{docResolver: docResolver, keyResolver: keyStore}
+	return &issuer{
+		store:       store,
+		Publisher:   publisher,
+		keyResolver: resolver,
+	}
+}
+
+type issuer struct {
+	store       Store
```
reinkrul, line 30: why exported?

vcr/issuer/issuer.go
```diff
@@ -0,0 +1,121 @@
...
+
+	// find issuer
+	issuer, err := did.ParseDID(credentialOptions.Issuer.String())
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse issuer: %w", err)
+	}
+
+	credentialID, _ := ssi.ParseURI(fmt.Sprintf("%s#%s", issuer.String(), uuid.New().String()))
+	unsignedCredential := vc.VerifiableCredential{
+		Context:           append(credentialOptions.Context, vc.VCContextV1URI()),
```
reinkrul, line 75: maybe I missed it, but I thought the default value is the Nuts context?

vcr/signature/proof/legacy.go
```diff
@@ -0,0 +1,75 @@
...
+		return nil, err
+	}
+
+	canonicalDocument, err := suite.GetCanonicalDocument(document)
+	if err != nil {
+		return nil, err
+	}
+
+	sums := append(hash.SHA256Sum(canonicalProof).Slice(), hash.SHA256Sum(canonicalDocument).Slice()...)
+	tbs := base64.RawURLEncoding.EncodeToString(sums)
```
reinkrul, line 49: Double-checked this is the right one, (v.s. (Raw)StdEncoding)?

vcr/issuer/keyresolver.go
reinkrul: Feels a bit like one encapsulation too many. It's only used once, in `networkPublisher`. So you could just make it a member function of that type, and use VDR's `KeyResolver.ResolveAssertionKeyID` which does 80% of this function (so it's sort of duplicate code). I'm not convinced eliminating `KeyResolver.Resolve()` in the unit tests justifies this type.

vcr/issuer/leiaStore.go
```diff
@@ -0,0 +1,82 @@
+package issuer
```
reinkrul, line 1: can we name it `leia_store.go`?

vcr/issuer/leiaStore.go
```diff
@@ -0,0 +1,82 @@
...
+	"fmt"
+	ssi "github.com/nuts-foundation/go-did"
+	"github.com/nuts-foundation/go-did/did"
+	"github.com/nuts-foundation/go-did/vc"
+	"github.com/nuts-foundation/go-leia/v2"
+	"github.com/nuts-foundation/nuts-node/vcr/credential"
+	"path"
+)
+
+// leiaStore implements the issuer Store interface. It is a simple and fast json store.
```
reinkrul, line 15:
```suggestion
// leiaStore implements the issuer Store interface. It is a simple and fast JSON store.
```

vcr/issuer/leiaStore.go
```diff
@@ -0,0 +1,82 @@
...
+// Note: It can not be used in a clustered setup.
+type leiaStore struct {
+	collection leia.Collection
+}
+
+// NewLeiaStore creates a new instance of leiaStore which implements the Store interface.
+func NewLeiaStore(dataDir string) (Store, error) {
```
reinkrul, line 22: more flexible to pass `dbPath`, since now it makes assumptions about the directory and filename ("vcr" and "issued-credentials").

Maybe even pass the leia instance, might make future merging/splitting of stores easier.

vcr/issuer/leiaStore.go
```diff
@@ -0,0 +1,82 @@
...
+	dbPath := path.Join(dataDir, "vcr", "issued-credentials.db")
+	store, err := leia.NewStore(dbPath, false)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create leiaStore: %w", err)
+	}
+	collection := store.Collection("issuedCredentials")
+	newLeiaStore := &leiaStore{collection: collection}
+	if err := newLeiaStore.createIndices(); err != nil {
```
reinkrul, line 30: must this always be called?

vcr/issuer/networkPublisher.go
```diff
@@ -0,0 +1,132 @@
+package issuer
```
reinkrul, line 1: `network_publisher.go`

vcr/issuer/networkPublisher.go
```diff
@@ -0,0 +1,132 @@
...
+	if err != nil {
+		return fmt.Errorf("invalid credential issuer: %w", err)
+	}
+
+	if len(verifiableCredential.CredentialSubject) == 0 {
+		return fmt.Errorf("missing credentialSubject")
+	}
+
+	// create participants list
+	participants, err := p.generateParticipants(verifiableCredential, public)
```
reinkrul, line 54: I find it odd to generate a participant list when the VC is public: because there should be no participant list (or an empty one) at all when `public == true`. So don't pass `public` to `generateParticipants`: only call it when `!public`.

vcr/signature/proof/legacy.go
```diff
@@ -0,0 +1,75 @@
...
+
+	// Don't use the suite's canonicalization method because it messes up the order of the fields:
+	// (this is one of the reasons this proof is deprecated)
+	// canonicalProof, err := suite.GetCanonicalDocument(p.asMap())
+	canonicalProof, err := json.Marshal(p)
+	if err != nil {
+		return nil, err
+	}
+
+	canonicalDocument, err := suite.GetCanonicalDocument(document)
```
reinkrul, line 43: but then why still canonicalizr the document here? Does it matter?

vcr/signature/signature.go
```diff
@@ -0,0 +1,51 @@
+package signature
+
+import (
+	"encoding/json"
+	ssi "github.com/nuts-foundation/go-did"
+	"github.com/nuts-foundation/nuts-node/crypto"
+	"github.com/nuts-foundation/nuts-node/crypto/hash"
+)
+
+// Suite is an interface which defines the methods a signature suite implementation should implement.
```
reinkrul, line 10: How do "signature suites" and "proof types" relate? Do we really need the 2 concepts or can there just be 1?

vcr/signature/signature.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+// Sign signs the provided doc and returns the signature bytes.
+func (l LegacyNutsSuite) Sign(doc []byte, key crypto.Key) ([]byte, error) {
+	sig, err := crypto.SignJWS(doc, detachedJWSHeaders(), key.Signer())
+	return []byte(sig), err
+}
+
+// GetCanonicalDocument canonicalizes the document by marshalling it to json
+func (l LegacyNutsSuite) GetCanonicalDocument(doc map[string]interface{}) ([]byte, error) {
```
reinkrul, line 29: make it a verb: "CanonicalizeDocument"

vcr/signature/signature.go
```diff
@@ -0,0 +1,51 @@
...
+	return json.Marshal(doc)
+}
+
+// GetDigest returns a digest for the doc by calculating the SHA256 hash.
+func (l LegacyNutsSuite) GetDigest(doc []byte) []byte {
```
reinkrul, line 34: make it a verb: "Digest" or "CalculateDigest"

vcr/types/interface.go
```diff
@@ -55,9 +52,11 @@ var ErrInvalidCredential = errors.New("invalid credential")
 // ErrInvalidPeriod is returned when the credential is not valid at the given time.
 var ErrInvalidPeriod = errors.New("credential not valid at given time")
 
-var vcDocumentType = "application/vc+json"
+// VcDocumentType holds the content type used in network documents which contain Verifiable Credentials
+var VcDocumentType = "application/vc+json"
```
reinkrul, line 56: const

And I think this is duplicate?

vcr/types/interface.go
```diff
@@ -117,6 +116,8 @@ type Resolver interface {
 
 // VCR is the interface that covers all functionality of the vcr store.
 type VCR interface {
+	Issuer() issuer.Issuer
```
reinkrul, line 119: maybe VCR should just implement `Issuer`  and delegate the calls? Now it's used as sort of "service registry" (just a holder of services to be used), while we normally call business functions on engine types (instead of getting one of its dependencies, and then call functions on it).

vcr/vcr.go
```diff
@@ -88,12 +94,18 @@ type vcr struct {
 	ambassador      Ambassador
 	network         network.Transactions
 	trustConfig     *trust.Config
+	issuerStore     issuer.Store
```
reinkrul, line 97: why do we need the issuer's store here? Maybe we initialize it in the VCR instance, but it should have no business storing it for later use?

reinkrul (changes requested):

docs/_static/vcr/v2.yaml
stevenvegt: What makes an issuer valid? Should we check for deactivation?

docs/_static/vcr/v2.yaml
stevenvegt: Currently nothing, but when we add support for JDON-LD, this context must be resolved. We can add a flag to the config to allow for resolving contexts.

vcr/issuer/issuer.go
stevenvegt: Not anymore. I want these api's to be more lowlevel. This allows for other context like schema.org or perhaps the KIK-V credentials.

vcr/issuer/leiaStore.go
stevenvegt: Cannot think about a use-case we want a store without it indices. Perhaps in a test condition, but the current tests do not demand it. We can always add a param later.

vcr/signature/proof/legacy.go
stevenvegt: This is the exact copy of the current implementation. Lets not mess with it. The current "canonicalization" step is just a json.Marshall.

vcr/signature/signature.go
stevenvegt: Must be multiple. We can have a JSONLD or a JWT proof with JsonWebSignature, or Ed25519Signature2018 or RsaSignature2018 suites.

vcr/types/interface.go
stevenvegt: Yes, but we have a nasty import cycle if I use the one from `vcr/types/interface.go`.
Is something we should fix in later iterations.

vcr/types/interface.go
stevenvegt: We could do that, but mind there will be a holder and issuer as well. This will be a lot of "empty" functions. We can also make these separate engines, but that;s also a bit weird. So, this is what I settled on. But the inclusion of the issuer.Issuer causes import cycles. So this is not optima.

docs/_static/vcr/v2.yaml
```diff
@@ -0,0 +1,284 @@
...
+      summary: "Resolves verifiable credentials issued by this node which matches the search params"
+      description: >
+        The resolutionResult contains the credential regardless of the validity.
+        The resolutionResult.currentStatus must be checked to see if the credential is valid.
+
+        error returns:
+        * 404 - Corresponding credential could not be found
+        * 500 - An error occurred while processing the request
+      operationId: "resolveIssuedVC"
+      parameters:
```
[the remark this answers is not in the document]
stevenvegt: The main use-case is finding a credential which you issued to revoke it. You probably know for which issuer and type you want to revoke. We can make more vars optional, but this probably requires more indices.

vcr/issuer/keyresolver.go
stevenvegt: It's used in both the NetworkPublisher and in the BuildVC method. Both are different parts in the process. Later it will be used in the buildRevocation and PublishRevocation. I think it saves some duplicate code and surely saves some testing code, since we only need 1 test for a failed assertionKey lookup instead of a failed doc lookup and a failed keyID lookup and a failed keyLookup. Now the errors are equal over all these usages.

vcr/issuer/interface.go
stevenvegt: VDR only has access to public keys embedded in the docs. This interface defines a service which can resolve private keys for a DID.

docs/_static/vcr/v2.yaml
```diff
@@ -0,0 +1,288 @@
...
+                $ref: '#/components/schemas/VerifiableCredential'
+        default:
+          $ref: '../common/error_response.yaml'
+
+  /internal/vcr/v2/issuer/vc/search:
+    get:
+      summary: "Searches for verifiable credentials issued by this node which matches the search params"
+      description: >
+        The resolutionResult contains the credential regardless of the validity.
+        The resolutionResult.currentStatus must be checked to see if the credential is valid.
```
woutslakhorst, line 46: It now returns a SearchVCResult, not a resolutionResult. It also doesn't contain a currentStatus jsut VC and/or revocation.

woutslakhorst (changes requested):

vcr/api/v2/api.go
```diff
@@ -118,7 +140,7 @@ func (w *Wrapper) VerifyVC(ctx echo.Context) error {
 		return err
 	}
 
-	if err := w.VCR.Validate(requestedVC, true, true, nil); err != nil {
+	if err := w.VCR.Validate(requestedVC, false, true, nil); err != nil {
```
reinkrul, line 143: does this then still work for `NutsAuthorizationCredential`s, which are never trusted?

vcr/issuer/network_publisher.go
```diff
@@ -22,7 +40,7 @@ type networkPublisher struct {
 	keyResolver     keyResolver
 }
 
-// VcDocumentType holds the content type used in network documents which contain Verifiable Credentials
+//VcDocumentType holds the content type used in network documents which contain Verifiable Credentials
```
reinkrul, line 43:
```suggestion
// VcDocumentType holds the content type used in network documents which contain Verifiable Credentials
```

vcr/vcr.go
```diff
@@ -97,8 +94,8 @@ type vcr struct {
 	ambassador      Ambassador
 	network         network.Transactions
 	trustConfig     *trust.Config
-	issuerStore     issuer.Store
-	issuer          issuer.Issuer
+	//issuerStore     issuer.Store
```
reinkrul, line 97: commented out code?

reinkrul (changes requested):

vcr/api/v2/api.go
stevenvegt: No, it will not work. Added extra options for this API request.

reinkrul (approved):

vcr/issuer/interface.go
```diff
@@ -0,0 +1,61 @@
...
+	ssi "github.com/nuts-foundation/go-did"
+	"github.com/nuts-foundation/go-did/did"
+	"github.com/nuts-foundation/go-did/vc"
+	"github.com/nuts-foundation/nuts-node/crypto"
+	"github.com/nuts-foundation/nuts-node/vcr/credential"
+)
+
+// Publisher publishes new credentials and revocations to a channel. Used by a credential issuer.
+type Publisher interface {
+	PublishCredential(verifiableCredential vc.VerifiableCredential, public bool) error
```
woutslakhorst, line 31: Undocumented interface methods

vcr/issuer/interface.go
```diff
@@ -0,0 +1,61 @@
...
+	PublishRevocation(revocation credential.Revocation) error
+}
+
+type keyResolver interface {
+	ResolveAssertionKey(issuerDID did.DID) (crypto.Key, error)
```
woutslakhorst, line 36: Undocumented interface methods

vcr/issuer/interface.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+// Issuer is a role in the network for a party who issues credentials about a subject to a holder.
+type Issuer interface {
+	Issue(unsignedCredential vc.VerifiableCredential, publish, public bool) (*vc.VerifiableCredential, error)
```
woutslakhorst, line 41: Undocumented interface methods

vcr/issuer/issuer.go
```diff
@@ -0,0 +1,142 @@
...
+// If publish is true, it publishes the credential to the network using the configured Publisher
+// Use the public flag to pass the visibility settings to the Publisher.
+func (i issuer) Issue(credentialOptions vc.VerifiableCredential, publish, public bool) (*vc.VerifiableCredential, error) {
+	createdVC, err := i.buildVC(credentialOptions)
+	if err != nil {
+		return nil, err
+	}
+
+	validator, _ := credential.FindValidatorAndBuilder(*createdVC)
+	if err := validator.Validate(*createdVC); err != nil {
```
woutslakhorst, line 63: untested: failing validation

vcr/signature/proof.go
```diff
@@ -0,0 +1,19 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *
+ */
+
+package signature
```
woutslakhorst, line 19: empty file?

woutslakhorst (changes requested):
only small stuff left

PR #749, opened 2022-01-14
Network: additional test for publisher with delayed payload

reinkrul:
Based on https://github.com/nuts-foundation/nuts-node/pull/712


woutslakhorst:
Covered by latest commit of #712

PR #750, opened 2022-01-17
Show PAL header in network get command

dmeijboom:
Todo:

- [x] Deduplicate PAL decryption logic
- [x] Find another way to get the node DID
- [x] Add unittests


docs/_static/network/v1.yaml
```diff
@@ -62,6 +62,36 @@ paths:
                 type: string
         default:
           $ref: '../common/error_response.yaml'
+  /internal/network/v1/transaction/{ref}/pal:
```
reinkrul, line 65: can't we just always try to decrypt it and enrich the returned transaction (when we `get` it)? Instead of having a specific supplemental API call for it?

reinkrul (changes requested):

docs/_static/network/v1.yaml
dmeijboom: Iirc the get transaction endpoint returns the raw data. I looked into modifying it but figured that creating a new endpoint specifically for the PAL header might be easier. I could also parse the transaction instead and add a `pal_decrypted` field to the same API?

woutslakhorst:
There's some lack of coverage. Someone should be appointed to this PR and finish it.

Issue #737, opened 2022-01-12
Private payload scheduler only supports direct exchange with 2 parties


woutslakhorst:
It can't handle 3 parties and also isn't be able to try all parties to retrieve the payload.

A node filter it's own node DID from the PAL and tries the first remaining other nodeDID.

reinkrul:
Fix is proposed by https://github.com/nuts-foundation/nuts-node/issues/720

PR #751, opened 2022-01-17
Network: broadcast TransactionPayloadQuery to all private TX participants

### LINKED ISSUE #720: Network: broadcast TransactionPayloadQuery to all participants in a private TX
Currently the last non-self PAL participant is selected, but this is vague, ambiguous and not spec'd. Best option would probably be to just broadcast the request to all (except the local node) participants in the PAL header.

reinkrul:
Fixes https://github.com/nuts-foundation/nuts-node/issues/737
Fixes #720 

Based on https://github.com/nuts-foundation/nuts-node/pull/712

- [x] Integration test reports "already connected to peer XYZ", check whether that issues already existed previously.

### DIFF
--- a/network/transport/v2/handlers.go
+++ b/network/transport/v2/handlers.go
@@ -54,7 +54,15 @@ func (p *protocol) handleTransactionPayloadQuery(peer transport.Peer, msg *Trans
 		}
 		epal := dag.EncryptedPAL(tx.PAL())
 
-		pal, _, err := p.decryptPAL(epal)
+		nodeDID, err := p.nodeDIDResolver.Resolve()
+		if err != nil {
+			return err
+		}
+		if nodeDID.Empty() {
+			return errors.New("node DID is not set")
+		}
+
+		pal, err := p.decryptPAL(epal, nodeDID)
 		if err != nil {
 			log.Logger().Errorf("Peer requested private transaction but decoding failed (peer=%s,tx=%s): %v", peer, tx.Ref(), err)
 			return p.send(peer, emptyResponse)
--- a/network/transport/v2/handlers.go
+++ b/network/transport/v2/handlers.go
@@ -54,15 +54,7 @@ func (p *protocol) handleTransactionPayloadQuery(peer transport.Peer, msg *Trans
 		}
 		epal := dag.EncryptedPAL(tx.PAL())
 
-		nodeDID, err := p.nodeDIDResolver.Resolve()
-		if err != nil {
-			return err
-		}
-		if nodeDID.Empty() {
-			return errors.New("node DID is not set")
-		}
-
-		pal, err := p.decryptPAL(epal, nodeDID)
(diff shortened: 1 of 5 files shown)


network/transport/v2/protocol.go
```diff
@@ -198,9 +198,17 @@ func (p *protocol) handlePrivateTxRetryErr(hash hash.SHA256Hash) error {
...
+		return err
+	}
+	if nodeDID.Empty() {
+		return errors.New("node DID is not set")
+	}
+
 	epal := dag.EncryptedPAL(tx.PAL())
 
-	pal, senderDID, err := p.decryptPAL(epal)
+	pal, err := p.decryptPAL(epal, nodeDID)
```
woutslakhorst, line 211: reason to duplicate nodeDID check (also in decryptPAL)

woutslakhorst (changes requested):

network/transport/v2/handlers.go
```diff
@@ -54,7 +54,15 @@ func (p *protocol) handleTransactionPayloadQuery(peer transport.Peer, msg *Trans
 		}
 		epal := dag.EncryptedPAL(tx.PAL())
 
-		pal, _, err := p.decryptPAL(epal)
+		nodeDID, err := p.nodeDIDResolver.Resolve()
```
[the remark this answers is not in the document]
reinkrul: Reverted it.

reinkrul:
@woutslakhorst ready for rereview.

network/transport/v2/protocol.go
```diff
@@ -214,19 +212,27 @@ func (p *protocol) handlePrivateTxRetryErr(hash hash.SHA256Hash) error {
...
-	err = conn.Send(p, &Envelope{Message: &Envelope_TransactionPayloadQuery{
-		TransactionPayloadQuery: &TransactionPayloadQuery{
-			TransactionRef: tx.Ref().Slice(),
-		},
-	}})
-	if err != nil {
-		return fmt.Errorf("failed to send TransactionPayloadQuery message(tx=%s, DID=%s): %w", hash.String(), senderDID, err)
+	if !sent {
+		return fmt.Errorf("unable to retrieve private TX payload, no connections to any of the participants (tx=%s, PAL=%v)", hash.String(), pal)
 	}
```
dmeijboom, line 237: Could be removed I think as there is already a log entry for when send fails

dmeijboom (approved):

network/transport/v2/protocol.go
woutslakhorst: The error return is logged as error, so the error formatting here is ok. The first part of the sentence is a bit of a duplicate. The wrapper adds `retry of TransactionPayloadQuery failed: %v`

woutslakhorst (approved):
With a single side note on the specific log message

PR #752, opened 2022-01-18
Network: do not log error when missing TXs for written payload

reinkrul:
This only occurs when a TX is created by the local node.

### DIFF
--- a/network/dag/publisher.go
+++ b/network/dag/publisher.go
@@ -51,13 +51,18 @@ type replayingDAGPublisher struct {
 }
 
 func (s *replayingDAGPublisher) payloadWritten(ctx context.Context, payloadHash interface{}) {
-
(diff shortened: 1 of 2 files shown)


dmeijboom:
Not sure if this is a good idea. Are we 100% this will never happen? Even with storage errors or something?

reinkrul:
> Not sure if this is a good idea. Are we 100% this will never happen? Even with storage errors or something?

If it happens, and the TX is added afterwards (which will happen in normal operation), the TX will still be published. If we suspect this will happen due to storage errors in production, we should have a mechanism to cope with it (auto healing/DLQ). Other than logging a message which I'm 100% sure nobody will read (or will just dismiss).

woutslakhorst (approved):

PR #754, opened 2022-01-18
pass ResolveMetadata for resolving VC issuer

### LINKED ISSUE #743: Issueing credentials from a DID that has a deactivated controller succeeds
But validating fails,

when issueing, the VCR does not pass any metadata to the DIDDoc.Resolve. This will bypass resolving the controllers.
It shoud also resolveControllers. This can be fixed by passing a metadata with resolve time set to now.

woutslakhorst:
fixes #743 

Passing resolveMetadata enables the controller active check.

### DIFF
--- a/vcr/vcr.go
+++ b/vcr/vcr.go
@@ -305,8 +305,9 @@ func (c *vcr) Issue(template vc.VerifiableCredential) (*vc.VerifiableCredential,
 	if err != nil {
 		return nil, fmt.Errorf("failed to parse issuer: %w", err)
 	}
-	// find did document/metadata for originating TXs
-	document, meta, err := c.docResolver.Resolve(*issuer, nil)
+	// find did document/metadata for originating TXs. ResolveMetadata is required to also check the validity of the issuer's controller.
+	now := timeFunc()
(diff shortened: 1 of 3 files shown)


vcr/vcr.go
```diff
@@ -305,8 +305,9 @@ func (c *vcr) Issue(template vc.VerifiableCredential) (*vc.VerifiableCredential,
 	if err != nil {
 		return nil, fmt.Errorf("failed to parse issuer: %w", err)
 	}
-	// find did document/metadata for originating TXs
-	document, meta, err := c.docResolver.Resolve(*issuer, nil)
+	// find did document/metadata for originating TXs. ResolveMetadata is required to also check the validity of the issuer's controller.
+	now := timeFunc()
```
reinkrul, line 309: is this equal to the issuance time of the VC?

woutslakhorst: give or take a few milliseconds. We could add issuance time to the VC builder to make sure they are the same at nanosecond level?

reinkrul (approved):

vcr/vcr.go
reinkrul: Would be nice (in a sense, more pure), but probably doesn't really matter much in practice.

PR #757, opened 2022-01-19
fix dag walking on duplicate did Doc update payload hash

woutslakhorst:
DID Document updates may have the same payloadHash. From payloadWritte in the dag.Publisher this may trigger a Dag walker starting at matching TXs for that payloadHash. It should not rewalk those!


woutslakhorst:
Tested with deve network to assure consistency

dmeijboom (approved):

Issue #736, opened 2022-01-11
The private payload scheduler stops after 100 retries


woutslakhorst:
see #708 

It's wise to gather some metrics/diagnostics/provide an api where an operator can act on when payload fetches are stuck as rety job. It'll happen and there might be something to do to fix it.

Something like the conflicted documents, we do not prevent things, we signal and give tools to fix.

PR #758, opened 2022-01-20
Add metrics for jobs that failed in the payload scheduler

dmeijboom:
Closes: #736


network/transport/v2/protocol.go
```diff
@@ -249,7 +249,12 @@ func (p *protocol) Stop() {
 }
 
 func (p protocol) Diagnostics() []core.DiagnosticResult {
-	return nil
+	failedJobs := p.payloadScheduler.GetFailedJobs()
+
+	return []core.DiagnosticResult{&core.GenericDiagnosticResult{
+		Title:   "failed_private_transactions",
```
reinkrul, line 255: "failed_private_transactions" isn't 100% accurate I think. Maybe add "retrieval"? Something like "private_transactions_failed_retrieval"?

network/transport/v2/scheduler.go
```diff
@@ -110,9 +114,23 @@ func (p *payloadScheduler) Run() error {
...
+	defer p.failedJobsLock.Unlock()
+
+	p.failedJobs = append(p.failedJobs, hash)
+}
+
+func (p *payloadScheduler) GetFailedJobs() []hash.SHA256Hash {
+	p.failedJobsLock.RLock()
+	defer p.failedJobsLock.RUnlock()
+
+	return p.failedJobs
```
reinkrul, line 128: access is still racy: when a `tx` is added after `GetFailedJobs` is called and returned: the diagnostics handler might be using the slice (e.g. marshalling to JSON) while another entry is being added.

For all diagnostics applies: best to work on copies of mutable data structures. Easiest is to make `GetFailedJobs()` return a copy. Then caller also can't mutate the internal state by altering the slice (although theoretically you can still alter the hashes inside it, since `hash.SHA256Hash` is just a slice)

network/transport/v2/scheduler.go
```diff
@@ -173,7 +191,8 @@ func (p *payloadScheduler) retry(hash hash.SHA256Hash, initialCount uint16) {
 			}),
 		)
 		if err != nil {
-			log.Logger().Errorf("Failed to pass payload query to network: %v", err)
+			log.Logger().Errorf("failed to pass payload query to network: %v", err)
+			p.addFailedJob(hash)
```
reinkrul, line 195: job is never removed? Is this the list of TXs that ultimately failed after many retries? Or TXs that are still being retried? If the first, the second is useful too, I think.

If they're retried no more, does the diagnostic survive restarts (since it's in-memory)?

reinkrul (changes requested):

network/transport/v2/scheduler.go
dmeijboom: > job is never removed? Is this the list of TXs that ultimately failed after many retries? Or TXs that are still being retried? If the first, the second is useful too, I think.

It's the list of txs that failed after 100 retries.

> If the first, the second is useful too, I think.

Could be true but wouldn't want to do that right now as it's specifically for debugging private transactions that were never fully processed.

> If they're retried no more, does the diagnostic survive restarts (since it's in-memory)?
 Write

I thought about this as well but it shouldn't matter as the retry DAG publisher should also retry these jobs right?

network/transport/v2/protocol.go
woutslakhorst: `payload_fetch_dlq`?

network/transport/v2/scheduler.go
woutslakhorst: If you query the DB on all jobs that have a retry count equal to max wouldn't that give you the failed jobs?

woutslakhorst (changes requested):

network/transport/v2/scheduler.go
dmeijboom: It most definitely would. Also makes for a more solid implementation.

network/transport/v2/scheduler.go
```diff
@@ -110,9 +114,27 @@ func (p *payloadScheduler) Run() error {
 	})
 }
 
+func (p *payloadScheduler) GetFailedJobs() (hashes []hash.SHA256Hash, err error) {
```
woutslakhorst, line 117: No coverage.

woutslakhorst (changes requested):
Add another diagnostic called `payload_fetch_delayed` (or something) where the count is > 20 and < 100. Those jobs are delayed for about 5 days. The max count is currently 100 days. That will take a while to show up....

reinkrul:
> Add another diagnostic called `payload_fetch_delayed` (or something) where the count is > 20 and < 100. Those jobs are delayed for about 5 days. The max count is currently ~90 days. That will take a while to show up....

I'd say more than an hour or so could count as delayed...?

reinkrul (approved):

woutslakhorst:
> > Add another diagnostic called `payload_fetch_delayed` (or something) where the count is > 20 and < 100. Those jobs are delayed for about 5 days. The max count is currently ~90 days. That will take a while to show up....
> 
> I'd say more than an hour or so could count as delayed...? Since in normal circumstances TX payload should be retrieved almost immediately (maybe some delay in case of (un)scheduled downtime).

Then > 10 retries would be helpful

dmeijboom:
> > > Add another diagnostic called `payload_fetch_delayed` (or something) where the count is > 20 and < 100. Those jobs are delayed for about 5 days. The max count is currently ~90 days. That will take a while to show up....
> > 
> > 
> > I'd say more than an hour or so could count as delayed...? Since in normal circumstances TX payload should be retrieved almost immediately (maybe some delay in case of (un)scheduled downtime).
> 
> Then > 10 retries would be helpful

Makes sense. I'll filter on 10 retries

woutslakhorst (approved):

Issue #621, opened 2021-11-22
Automatically resolve node DID


reinkrul:
Part of https://github.com/nuts-foundation/nuts-node/issues/603

To authenticate a connection, a node has to send its node DID to its peer. A node can only query the payload of private transactions when it authenticated its node DID to its peer.

We could simply introduce a configuration parameter for setting the node DID. This would work of course, but isn't very user-friendly. We could derive it from the private keys in crypto and the TLS certificate:

- Lookup private keys in crypto, distillate DID from the key IDs. These are either the vendor's DID(s) or its care organizations (when its primary vendor key is in secure offline storage).
-  Resolve corresponding DID documents, look for the `NutsComm` endpoints (follow references in care organization DID documents)
- Find the DID document that contains a `NutsComm` endpoint that matches one of the SANs in the TLS certificate

What if multiple are found?

woutslakhorst:
> Lookup private keys in crypto, distillate DID from the key IDs
won't work after key-rotation, 1st private key would be gone.

Another option:

- find all DID documents with a `NutsComm` absolute endpoint
- find private keys corresponding to those DIDs
- done

reinkrul:
> > Lookup private keys in crypto, distillate DID from the key IDs
> > won't work after key-rotation, 1st private key would be gone.
> 
> Another option:
> 
> * find all DID documents with a `NutsComm` absolute endpoint
> * find private keys corresponding to those DIDs
> * should be one
> * done

This works for probably 95% of the cases where the vendor key is simply in online storage. The rest last 5% will be competent enough to configure it, probably. I like it.

reinkrul:
I considered 2 approaches: first, just querying VDR and intersecting DID documents with the private key store.

The second involves using the "event stream" from DID document TXs, listening to events. This last approach isn't really feasible right now since we don't yet have DID document events (e.g. DIDDocumentCreated/Updated), thus would lead to another place where we'd have to parse DID documents. Even if we would have such events, persistence is going to be a problem when VDR storage is persisted (in future), because then those events probably don't fire at boot time. Thus needing the auto-resolved DID documents be persisted as well.

So for now, simplest approach is just querying VDR for docs with a NutsComm endpoint. The match will be cached and initialized lazy.

PR #760, opened 2022-01-20
Network: auto-resolve node DID

reinkrul:
When not configured and not in strict mode, allow the node DID to be auto-resolved. It's matched by intersecting DID documents with a NutsComm URL endpoint and the private keys in the keystore.

Fixes https://github.com/nuts-foundation/nuts-node/issues/621

### DIFF
--- a/crypto/interface.go
+++ b/crypto/interface.go
@@ -43,6 +43,8 @@ type KeyResolver interface {
 	Exists(kid string) bool
 	// Resolve returns a Key for the given KID. ErrKeyNotFound is returned for an unknown KID.
 	Resolve(kid string) (Key, error)
+	// List returns the KIDs of the private keys that are present in the KeyStore.
+	List() []string
 }
 
 // KeyStore defines the functions for working with private keys.
--- a/crypto/storage/fs.go
+++ b/crypto/storage/fs.go
@@ -118,7 +118,7 @@ func (fsc *fileSystemBackend) ListPrivateKeys() []string {
 	_ = filepath.Walk(fsc.fspath, func(path string, info fs.FileInfo, err error) error {
 		if !info.IsDir() && strings.HasSuffix(info.Name(), string(privateKeyEntry)) {
 			upper := len(info.Name()) - len(privateKeyEntry) - 1
-			if upper < len(info.Name()) {
+			if upper > 0 {
 				result = append(result, info.Name()[:upper])
 			}
 		}
--- a/crypto/storage/fs.go
+++ b/crypto/storage/fs.go
@@ -23,8 +23,10 @@ import (
 	"errors"
 	"fmt"
 	"github.com/nuts-foundation/nuts-node/crypto/util"
+	"io/fs"
 	"os"
 	"path/filepath"
+	"strings"
 )
 
 type entryType string
@@ -109,6 +111,20 @@ func (fsc *fileSystemBackend) SavePrivateKey(kid string, key crypto.PrivateKey) 
 	_, err = outFile.Write([]byte(pem))
 
 	return err
+}
+
+func (fsc *fileSystemBackend) ListPrivateKeys() []string {
+	var result []string
+	_ = filepath.Walk(fsc.fspath, func(path string, info fs.FileInfo, err error) error {
+		if !info.IsDir() && strings.HasSuffix(info.Name(), string(privateKeyEntry)) {
+			upper := len(info.Name()) - len(privateKeyEntry) - 1
+			if upper < len(info.Name()) {
+				result = append(result, info.Name()[:upper])
+			}
+		}
+		return nil
+	})
+	return result
 }
 
 func (fsc fileSystemBackend) readEntry(kid string, entryType entryType) ([]byte, error) {
--- a/crypto/storage/storage.go
+++ b/crypto/storage/storage.go
@@ -40,6 +40,8 @@ type Storage interface {
 	PrivateKeyExists(kid string) bool
 	// SavePrivateKey stores the key under the kid in the storage backend.
 	SavePrivateKey(kid string, key crypto.PrivateKey) error
+	// ListPrivateKeys returns the KIDs of the private keys that are present.
+	ListPrivateKeys() []string
 }
 
 // PublicKeyEntry is a public key entry also containing the period it's valid for.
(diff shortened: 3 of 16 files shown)


crypto/storage/storage.go
```diff
@@ -40,6 +40,8 @@ type Storage interface {
 	PrivateKeyExists(kid string) bool
 	// SavePrivateKey stores the key under the kid in the storage backend.
 	SavePrivateKey(kid string, key crypto.PrivateKey) error
+	// ListPrivateKeys returns the KIDs of the private keys that are present.
```
woutslakhorst, line 43: Maybe mention not to use for features that also work in strict mode?

dmeijboom (changes requested):

crypto/interface.go
```diff
@@ -43,6 +43,8 @@ type KeyResolver interface {
 	Exists(kid string) bool
 	// Resolve returns a Key for the given KID. ErrKeyNotFound is returned for an unknown KID.
 	Resolve(kid string) (Key, error)
+	// List returns the KIDs of the private keys that are present in the KeyStore.
```
woutslakhorst, line 46: Not to be used for features available in strictmode?

network/transport/nodedid.go
```diff
@@ -0,0 +1,83 @@
...
+// NewAutoNodeDIDResolver creates a new node DID resolver that tried to look it up by matching DID documents with a NutsComm endpoint set and the private key of the local node.
+func NewAutoNodeDIDResolver(keyResolver crypto.KeyResolver, docFinder types.DocFinder) NodeDIDResolver {
+	return &AutoNodeDIDResolver{
+		keyResolver: keyResolver,
+		docFinder:   docFinder,
+		mux:         &sync.Mutex{},
+	}
+}
+
+type AutoNodeDIDResolver struct {
```
woutslakhorst, line 37: why exporte?

network/transport/nodedid.go
```diff
@@ -0,0 +1,83 @@
...
+func (a *AutoNodeDIDResolver) Resolve() (did.DID, error) {
+	a.mux.Lock()
+	if !a.resolvedDID.Empty() {
+		result := a.resolvedDID
+		a.mux.Unlock()
+		return result, nil
+	}
+	defer a.mux.Unlock()
+
+	documents, err := a.docFinder.Find(doc.IsActive(), doc.ValidAt(time.Now()), doc.ByServiceType(NutsCommServiceType))
```
woutslakhorst, line 54: does that predicate only return absolute endpoints or also refs?

woutslakhorst (changes requested):

crypto/storage/vault.go
```diff
@@ -148,6 +148,11 @@ func (v vaultKVStorage) PrivateKeyExists(kid string) bool {
 	return ok
 }
 
+func (v vaultKVStorage) ListPrivateKeys() []string {
+	// not supported yet
+	return nil
```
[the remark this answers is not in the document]
reinkrul: Eeeh, I forgot about that 🤦

crypto/storage/fs.go
```diff
@@ -111,6 +113,20 @@ func (fsc *fileSystemBackend) SavePrivateKey(kid string, key crypto.PrivateKey)
 	return err
 }
 
+func (fsc *fileSystemBackend) ListPrivateKeys() []string {
+	var result []string
+	_ = filepath.Walk(fsc.fspath, func(path string, info fs.FileInfo, err error) error {
+		if !info.IsDir() && strings.HasSuffix(info.Name(), string(privateKeyEntry)) {
+			upper := len(info.Name()) - len(privateKeyEntry) - 1
+			if upper < len(info.Name()) {
```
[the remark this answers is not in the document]
reinkrul: The check I meant was:

```
			if upper > 0 {
				result = append(result, info.Name()[:upper])
			}
```

(if the file name is just the private key entry postfix).

crypto/storage/storage.go
reinkrul: `ListPrivateKeys()` can surely be used in strict mode (assuming Vault supports it). I see no reason why we should disallow it, as it just returns the key IDs, not the keys themselves.

network/transport/nodedid.go
reinkrul: I had to check it as well, it only returns absolute endpoints.

network/transport/nodedid.go
```diff
@@ -0,0 +1,83 @@
...
+
+// Resolve returns the auto-resolved node DID, or an empty DID if none could be found.
+func (a *AutoNodeDIDResolver) Resolve() (did.DID, error) {
+	a.mux.Lock()
+	if !a.resolvedDID.Empty() {
+		result := a.resolvedDID
+		a.mux.Unlock()
+		return result, nil
+	}
+	defer a.mux.Unlock()
```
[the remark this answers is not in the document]
reinkrul: Yes, and that's exactly why we should cache it. Otherwise the node DID might change during runtime. So it could present the node DID to one peer as `did:nuts:A` and to a second peer as `did:nuts:B`. Which would be very, very odd behavior imo.

Now it _shouldn't_ ever change, not even after a restart. But should it change (due to unforeseen flow/case), it's disallowed in strict mode (a.k.a. production usage).

crypto/storage/storage.go
woutslakhorst: > as it just returns the key IDs, not the keys themselves.
ah, right

network/transport/nodedid.go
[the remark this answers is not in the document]
dmeijboom: > Otherwise the node DID might change during runtime. So it could present the node DID to one peer as did:nuts:A and to a second peer as did:nuts:B. Which would be very, very odd behavior imo.

Why? The peer ID should be the 'main' identifier right? I agree that it shouldn't happen very often but it could happen and I think it should be supported.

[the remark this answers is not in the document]
reinkrul: If we allow it to be changed at runtime (because a NutsComm endpoint is registered on an existing, or new DID document), which one should be matched as the one being the correct node DID? The behavior will be unclear at best (last updated DID document = node DID) or unpredictable at worst (random matching DID document).

I suppose it _could_ change in dev/workshop envs (because people will be fiddling around/losing keys/DIDs), but in that case I still think predictable, simple behavior is better than a maybe more intelligent but unpredictable algorithm that (maybe unexpectedly) changes the node DID halfway?

In production environments it must be configured statically anyways.

woutslakhorst (approved):

dmeijboom (approved):

PR #762, opened 2022-01-21
Crypto: make vault config more clear

reinkrul:


docs/pages/production-configuration.rst
```diff
@@ -13,17 +13,23 @@ Persistence
...
+^^^^^
+
+This storage backend is the recommended way of storing secrets. It uses the `Vault KV version 1 store <https://www.vaultproject.io/docs/secrets/kv/kv-v1>`_.
+The prefix defaults to `kv` and can be configured using the `crypto.vault.pathprefix` option.
+There needs to be a KV Secrets Engine (v1) enabled under this prefix path.
+
+All private keys are stored under the path `<prefix>/nuts-private-keys/*`.
 Each key is stored under the kid, resulting in a full key path like `kv/nuts-private-keys/did:nuts:123#abc`.
-A Vault token must be provided by either configuring it using the config `crypto.vault.token` or setting the VAULT_TOKEN environment variable.
+A Vault token must be provided by either configuring it using the `crypto.vault.token` configuration option.
```
woutslakhorst, line 27: The word `either` is no longer needed. Does `VAULT_TOKEN` not work?

reinkrul, line 27:
```suggestion
A Vault token must be provided by either configuring it using the config `crypto.vault.token` or setting the VAULT_TOKEN environment variable.
```

reinkrul: I assumed it was an incorrect example on how to configure it using the Nuts node's env options. But it's actually from the vault client lib itself, so it was correct. I reverted it.

Although I'm not sure whether we'd want people configuring the vault client using env options parsed by the client lib instead of them being passed through the Nuts node config, since we then lose control over it. But I guess OK for now.

dmeijboom (approved):
LGTM but I wasn't aware we still use the v1 kv engine. Why aren't we using the v2 kv engine? Note that keys/paths are not encrypted or obfuscated in the v1 kv engine.

woutslakhorst (approved):

reinkrul:
> LGTM but I wasn't aware we still use the v1 kv engine. Why aren't we using the v2 kv engine? Note that keys/paths are not encrypted or obfuscated in the v1 kv engine.

I wondered that too, according to @stevenvegt we don't need the versioning feature in the v2 engine. Using v2 (and the default `secrets` path) would have the advantage that it's present by default when starting a new Vault instance. But at the same time, people who are running Vault in production will (hopefully) know how to configure it.

I'm not against switching to v2, but not part of this PR which just makes the documentation a bit better. I'll make an issue for it.

PR #771, opened 2022-01-25
VDR: Fix resolving disabled DID docs

stevenvegt:
Better fix for #743.
Instead of providing an empty resolverMetadata (or a time), make checking for disabled controllers default.

### DIFF
--- a/vdr/doc/resolvers.go
+++ b/vdr/doc/resolvers.go
@@ -61,6 +61,7 @@ func (d Resolver) resolve(id did.DID, metadata *types.ResolveMetadata, depth int
 		return nil, nil, err
 	}
 
+	// has the doc controllers, should we check for controller deactivation?
 	if len(doc.Controller) > 0 && (metadata == nil || !metadata.AllowDeactivated) {
 		// also check if the controller is not deactivated
 		// since ResolveControllers calls Resolve and propagates the metadata
@@ -68,6 +69,7 @@ func (d Resolver) resolve(id did.DID, metadata *types.ResolveMetadata, depth int
 		if err != nil {
 			return nil, nil, err
 		}
+		// doc should have controllers, but no results, so they are not active, return error:
 		if len(controllers) == 0 {
 			return nil, nil, types.ErrNoActiveController
 		}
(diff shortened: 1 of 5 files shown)


reinkrul:
Can you link the issue?

vdr/doc/resolvers.go
```diff
@@ -61,7 +61,7 @@ func (d Resolver) resolve(id did.DID, metadata *types.ResolveMetadata, depth int
 		return nil, nil, err
 	}
 
-	if metadata != nil && !metadata.AllowDeactivated {
+	if len(doc.Controller) > 0 && (metadata == nil || !metadata.AllowDeactivated) {
```
reinkrul, line 64: shouldn't this be:

```
if len(doc.Controller) > 0 && (metadata == nil && !metadata.AllowDeactivated)
```

?

Because now you can't resolve the DID document with `AllowDeactivated == true` when the document has controllers, at all?

vdr/doc/resolvers_test.go
```diff
@@ -190,30 +190,46 @@ func TestResolver_Resolve(t *testing.T) {
...
-		store.EXPECT().Resolve(*id123, resolveMD).Return(&docA, &types.DocumentMetadata{}, nil)
-
-		doc, _, err := resolver.Resolve(*id456, resolveMD)
-
-		assert.Error(t, err)
-		assert.Equal(t, types.ErrNoActiveController, err)
-		assert.Nil(t, doc)
-	})
-
-	t.Run("docA is controller of docB and docA is deactivated but is allowed", func(t *testing.T) {
```
reinkrul, line 206: this is still a valid test?

reinkrul (changes requested):

vdr/doc/resolvers.go
stevenvegt: `(metadata == nil && !metadata.AllowDeactivated)` will resolve in a nullpointer exception.

Why would it not be possible to resolve the docuent when `allowDeactivated == true`? at line 59 the document got resolved, regardless of the `allowDeactivated` option.

vdr/doc/resolvers_test.go
stevenvegt: Yes, it still there but now moved to line 221.

PR #780, opened 2022-01-27
Network: improve diagnostics on outbound connector stats

reinkrul:
Issues with current diagnostics:

- It shows `outbound_connectors` for active connections (unnecessary information)
- It does not actually show outbound peer address
- It shows the peer info, which does not add information (because only relevant when actually connected, and then you'll find it at the `peers` diagnostic info)


dmeijboom (changes requested):
One (minor) change 😅

network/transport/grpc/connection_list.go
```diff
@@ -128,10 +128,15 @@ func (c *connectionList) Diagnostics() []core.DiagnosticResult {
...
 	var peers []transport.Peer
+	// Only add peer to "outbound_connectors" list if not connected
 	for _, curr := range c.list {
-		connectors = append(connectors, curr.stats())
 		if curr.IsConnected() {
 			peers = append(peers, curr.Peer())
+		} else {
+			connector := curr.outboundConnector()
+			if connector != nil {
+				connectors = append(connectors, curr.outboundConnector().stats())
```
reinkrul, line 138:
```suggestion
				connectors = append(connectors, connector.stats())
```

dmeijboom (approved):
LGTM

PR #783, opened 2022-01-27
Refactor dag, payloadStore and publisher into abstraction

woutslakhorst:
State is now responsible for TX isolation and callbacks.

This is in preparation for adding more storage formats that must reuse the same transaction (XOR+IBLT).


network/dag/state.go
```diff
@@ -0,0 +1,236 @@
...
+	newState := &state{
+		db:           db,
+		graph:        graph,
+		payloadStore: payloadStore,
+		txVerifiers:  verifiers,
+	}
+
+	publisher := NewReplayingDAGPublisher(payloadStore, graph).(*replayingDAGPublisher)
+	newState.RegisterObserver(func(ctx context.Context, transaction Transaction, payload []byte) {
+		publisher.publishMux.Lock()
```
reinkrul, line 82: publisher and state are a bit too tangled here: `state` calls functions on `publisher`, but knows it has to use some certain mutex of it to make them safe. I'd say publisher has to register those observers then, or let `transactionAdded` and `payloadWritten` lock themselves, as required?

network/dag/state.go
```diff
@@ -0,0 +1,236 @@
...
+			payloadHash := hash.SHA256Sum(payload)
+			if err := s.payloadStore.WritePayload(contextWithTX, payloadHash, payload); err != nil {
+				return err
+			}
+		}
+		if err := s.graph.Add(contextWithTX, transaction); err != nil {
+			return err
+		}
+
+		// notify observers
```
reinkrul, line 116: luckily this comment is here :'-)

network/dag/interface.go
```diff
@@ -32,38 +33,47 @@ const AnyPayloadType = "*"
...
+	Add(ctx context.Context, transactions Transaction, payload []byte) error
 	// FindBetween finds all transactions which signing time lies between startInclude and endExclusive.
 	// It returns the transactions in DAG walking order.
 	FindBetween(ctx context.Context, startInclusive time.Time, endExclusive time.Time) ([]Transaction, error)
-	// Get retrieves a specific transaction from the DAG. If it isn't found, nil is returned.
-	Get(ctx context.Context, ref hash.SHA256Hash) (Transaction, error)
 	// GetByPayloadHash retrieves all transactions that refer to the specified payload.
 	GetByPayloadHash(ctx context.Context, payloadHash hash.SHA256Hash) ([]Transaction, error)
+	// GetTransaction returns the transaction from local storage
+	GetTransaction(ctx context.Context, hash hash.SHA256Hash) (Transaction, error)
```
reinkrul, line 52: stick these functions in a DAG interface (like we had), since the interface applies to both `bboltDAG` and `state` (since you're delegating the calls).

network/dag/state.go
```diff
@@ -0,0 +1,236 @@
...
+
+	payloadStore := NewBBoltPayloadStore(db)
+	newState := &state{
+		db:           db,
+		graph:        graph,
+		payloadStore: payloadStore,
+		txVerifiers:  verifiers,
+	}
+
+	publisher := NewReplayingDAGPublisher(payloadStore, graph).(*replayingDAGPublisher)
```
reinkrul, line 80: this casting feels bad

network/dag/state.go
```diff
@@ -0,0 +1,236 @@
...
+		notifyObservers(ctx, s.observers, nil, data)
+	}
+	return err
+}
+
+func (s *state) PayloadHashes(ctx context.Context, visitor func(payloadHash hash.SHA256Hash) error) error {
+	return s.graph.PayloadHashes(ctx, visitor)
+}
+
+func (s *state) ReadMany(ctx context.Context, consumer func(context.Context, PayloadReader) error) error {
```
reinkrul, line 165: should then be named `ReadManyPayloads`?

network/dag/state.go
```diff
@@ -0,0 +1,236 @@
...
+}
+
+// NewState returns a new State. The State is used as entry point, it's methods will start transactions and will notify observers from within those transactions.
+func NewState(dataDir string, verifiers ...Verifier) (State, error) {
+	dbFile := path.Join(dataDir, "network", "data.db")
+	if err := os.MkdirAll(filepath.Dir(dbFile), os.ModePerm); err != nil {
+		return nil, fmt.Errorf("unable to create BBolt database: %w", err)
+	}
+
+	// for tests we set NoSync to true, this option can only be set through code
```
reinkrul, line 63: but it's not actually set?

network/network.go
```diff
@@ -68,9 +62,7 @@ type Network struct {
 	lastTransactionTracker lastTransactionTracker
 	protocols              []transport.Protocol
 	connectionManager      transport.ConnectionManager
-	graph                  dag.DAG
-	publisher              dag.Publisher
-	payloadStore           dag.PayloadStore
+	txState                dag.State
```
reinkrul, line 65: why `txState` (vs. just `state` or `dagState`, since the package is called DAG)?

reinkrul (changes requested):
In general I like the change, because it makes access to the DAG (TXs + payload + events) simpler from caller point of view. But I think the refactor isn't complete yet, since there are some oddities in the `state` implementation (using mutexes of dependencies, casting).

network/dag/interface.go
woutslakhorst: the bboltDAG impl is now only a concrete implementation. It no longer implements an interface. There's also no longer a need to mock it since all other components call the State interface.

network/dag/state.go
woutslakhorst: removed it since influence on test performance was minimal

network/network.go
woutslakhorst: renamed it to `state` in all packages outside `dag`

woutslakhorst:
Have some random failing test on missing lock value....

reinkrul (approved):

PR #784, opened 2022-01-27
Core: properly shutdown node on OS interrupt

reinkrul:
Currently, the Nuts node isn't shut down at all. This risks data corruption and sending avoidable errors to network peers.

This PR introduces signal trapping to instruct the engines to shut down. It also has some fixes for existing problems with the shutdown.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -23,13 +23,12 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+	"go.uber.org/atomic"
 	"io"
 	"net/http"
 	"os"
-	"sync"
-
-	"github.com/sirupsen/logrus"
-	"github.com/spf13/cobra"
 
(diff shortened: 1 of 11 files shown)


cmd/root.go
```diff
@@ -135,13 +139,31 @@ func startServer(system *core.System) error {
...
+		shutdownOnce.Do(func() {
+			logrus.Info("Shutting down...")
+			echoServer.Shutdown()
+			if err := system.Shutdown(); err != nil {
+				logrus.Error("Error shutting down system:", err)
+			}
+			logrus.Info("Shutdown complete. Goodbye!")
+		})
+	}
+	defer shutdown()
```
dmeijboom, line 154: I think it's cleaner to create a new cancellable context with the original context as parent. When the parent context is cancelled it will be propagated to the child and otherwise we will cancel the child context ourselves anyway. Then we can also remove `defer shutdown()` and the `sync.Once` stuff.

dmeijboom (changes requested):

dmeijboom (approved):

Issue #747, opened 2022-01-14
Retry doesn't stop when node does not have a nodeDID


woutslakhorst:
check if a VC can be created for this node, that should already give a problem since that would require a did Node?

This is annoying since other nodes that do not have a did node set will get this error even if the TX is not for them...

```
network-node-three-1  | time="2022-01-14T15:04:00Z" level=error msg="retry of TransactionPayloadQuery failed: failed to decrypt PAL header (tx=f4045fcbd996919ed6171fcbf9403bbce1fcdf982c2ef169aefbd8d259635b04): node DID is not set" module=Network
```

dmeijboom:
For the Node DID we currently use the `transport.NodeDIDResolver` interface. There is no guarantee (even thought it's currently the case) that repeated calls on `NodeDIDResolver.Resolve()` will return the same result.

We also discussed resolving the Node DID automatically somehow. This means that a new implementation which automatically resolves the Node DID can actually return a Node DID later on even though previous calls returned an error or nil.

woutslakhorst:
It's not a requirement to set a nodeDID. So "private" network users get bothered by lots of error messages....

dmeijboom:
> It's not a requirement to set a nodeDID. So "private" network users get bothered by lots of error messages.... But how to distinguish between not willing to set and not set yet?
> 
> Would that require replaying the DAG when changing your nodeDID?

I don't think there is a way to distinguish between purposely not setting your node DID or forgetting to set it. However, we could try rate limiting the `Resolve()` calls? An alternative would be to add a flag which disables resolving the node DID for your own node.

woutslakhorst:
*not set yet* could actually be detectable: If you have DID documents under your control and those have a NutsComm endpoint but you didn't set your nodeDID and auto-detect fails => startup WARNING/ERROR/failure?

Then the retry job can just ignore all jobs when no nodeDID is known.

reinkrul:
check docs

PR #786, opened 2022-01-28
Don't start the payloadscheduler when the node DID was not set

dmeijboom:
Closes: #747


network/transport/v2/protocol.go
```diff
@@ -140,12 +140,22 @@ func (p *protocol) Configure(_ transport.PeerID) error {
 func (p *protocol) Start() (err error) {
 	p.ctx, p.cancel = context.WithCancel(context.Background())
 
-	// load old payload query jobs
-	if err = p.payloadScheduler.Run(); err != nil {
-		return fmt.Errorf("failed to start retrying TransactionPayloadQuery: %w", err)
+	nodeDID, err := p.nodeDIDResolver.Resolve()
+	if err != nil {
+		log.Logger().WithError(err).Errorf("failed to resolve node DID")
```
reinkrul, line 145: I think we (normally/should) always start log messages with a capital (errors start lowercase)

network/transport/v2/protocol.go
```diff
@@ -140,12 +140,22 @@ func (p *protocol) Configure(_ transport.PeerID) error {
...
+	}
+
+	if nodeDID.Empty() {
+		log.Logger().Warn("not starting the payload scheduler as node DID is not set")
+	} else {
+		// load old payload query jobs
+		if err = p.payloadScheduler.Run(); err != nil {
```
reinkrul, line 152: but this will still add the jobs to the scheduler? I thought we also discussed we shouldn't add them add all, to avoid writing lots of jobs that will never be ran/cleaned up when a node is just reading?

reinkrul (changes requested):

network/transport/v2/protocol.go
reinkrul: Also: when using the "automatic node DID resolver" (default when not set and not in strict mode) the node DID might be available later when a DID document (with nutscomm endpoint) is created, without restarting. So technically more correct would be to start the scheduler, but not adding any jobs to it (when node DID is empty).

That would allow the node to start receiving private TXs after getting a node DID set, without restarting.

Might sound like an edge case, but I think it might happen in workshops/demos? (otherwise we'll have to make sure the manuals state you'll have to restart the node).

dmeijboom: > but this will still add the jobs to the scheduler? I thought we also discussed we shouldn't add them add all, to avoid writing lots of jobs that will never be ran/cleaned up when a node is just reading?

Oops. Indeed. Will fix

dmeijboom: > Might sound like an edge case, but I think it might happen in workshops/demos? (otherwise we'll have to make sure the manuals state you'll have to restart the node).

I'll add it to the manuals. This way the behaviour is at least deterministic.

woutslakhorst: > Actually, this isn't great. Does it make sense to only do this when strict mode is enabled? You'd still get error messages on development. Not sure if that's an issue.

I also want my dev environments to be clean of errors.

network/transport/v2/protocol.go
```diff
@@ -140,12 +140,22 @@ func (p *protocol) Configure(_ transport.PeerID) error {
 func (p *protocol) Start() (err error) {
 	p.ctx, p.cancel = context.WithCancel(context.Background())
 
-	// load old payload query jobs
-	if err = p.payloadScheduler.Run(); err != nil {
-		return fmt.Errorf("failed to start retrying TransactionPayloadQuery: %w", err)
+	nodeDID, err := p.nodeDIDResolver.Resolve()
+	if err != nil {
+		log.Logger().WithError(err).Errorf("Failed to resolve node DID")
```
woutslakhorst, line 145: nitpicky: `Errorf` can be without the `f`

woutslakhorst (approved):

reinkrul:
Alternatively we could try resolving the node DID in a goroutine if we couldn't start the scheduler, and start it when we _can_ resolve it:

```
payloadSchedulerStarted := startPayloadScheduler(p.nodeDIDResolver, p.payloadScheduler)
	if !payloadSchedulerStarted {
		// If we couldn't start the payload scheduler at this time (because the node DID wasn't configured),
		// retry at a set interval.
		go func(resolver transport.NodeDIDResolver, scheduler Scheduler) {
			ticker := time.NewTicker(payloadSchedulerStartInterval)
			c := ticker.C
			done := p.ctx.Done()
			for {
				select {
					case <-c:
						if startPayloadScheduler(resolver, scheduler) {
							ticker.Stop()
							return
						}
					case <-done:
						// Shutdown() called
						ticker.Stop()
						return
				}
			}
		}(p.nodeDIDResolver, p.payloadScheduler)
	}
```

Thoughts?

woutslakhorst:
> Alternatively we could try resolving the node DID in a goroutine if we couldn't start the scheduler, and start it when we _can_ resolve it:
> 
> ...
> 
> Thoughts?

Although it would work, it's still very specific for development and rather annoying to test. Maybe if it's wrapped within a `DelayedInitializer` format which adds some abstraction and testability. Then that construct can be used for all initializers that may be started later?

woutslakhorst:
Discussed: lets not use lazy loading since this could hide problems.

PR #790, opened 2022-01-28
Implement revocation for v2 issuer api

stevenvegt:


stevenvegt:
Closed in favour of #831

PR #791, opened 2022-01-31
added static config for event streams and subscribers

### LINKED ISSUE #759: Nats stream design
In our pursuit to use Nats for various events, there are some Nats Jetstream concepts that require a little bit of design.
Basically there are producers, consumers, streams and servers. All have lots of options to consider. This issue is about the choices for the streams.

Streams are defined as the storage settings of the event system. Each stream has its own storage options. Multiple events can be grouped on a stream if they have the same requirements for storage (and still then, it's an option to do it different). Events in the Nuts node are to be used to decouple engines better and to enable events to flow from the node to other systems. Based on those requirements, two main streams can be identified:

- TRANSACTIONS containing events with and without payload in various states throughout the system.
- DATA containing VC's and DID Documents that can be used for service discovery or can trigger external systems.

Based on the limited knowledge, these streams can have the following config:

- Storage: File
- MaxAge: one week (retention based on time rather than size)

Any downtime longer than a week would require a *replay* of transactions from storage!
Streams MUST be available before producers and consumers connect. The event engine could register required streams from the `Start()` method.

Nats documentation on streams: https://docs.nats.io/nats-concepts/jetstream/streams

woutslakhorst:
closes #759

Nats is very picky about the setup of streams and its consumers. If settings do not match, it'll return an error.
This PR adds static config for both a set of streams and the consumer setup.

I also added an integration test to demonstrate current settings work.


events/interface.go
```diff
@@ -21,6 +21,8 @@ package events
 
 // Event is the event manager interface that gives access to all event related functions
 type Event interface {
+	// GetStream returns a predefined stream with file storage
+	GetStream(streamName string) (Stream, bool)
```
reinkrul, line 25: document `bool` return

events/stream.go
```diff
@@ -100,11 +116,12 @@ func NewDisposableStream(name string, subjects []string, maxMessages int64) Stre
 	}, []nats.SubOpt{
 		nats.AckNone(),
```
reinkrul, line 117: wasn't it `ExplicitAck`?

events/stream.go
```diff
@@ -100,11 +116,12 @@ func NewDisposableStream(name string, subjects []string, maxMessages int64) Stre
...
 		nats.DeliverNew(),
```
reinkrul, line 118: not consistent with linked issue

reinkrul (changes requested):

events/stream.go
woutslakhorst: Accepted proposal for New in issue

events/stream.go
woutslakhorst: Had some troubles when testing which returned weird errors. This eventually had to do with the fact a relative testDir was used instead of a tmp testDir.....

events/interface.go
woutslakhorst: Changed to using nil return for existence.

reinkrul (approved):

PR #796, opened 2022-02-01
Network: fix startup/verify order of DAG state

reinkrul:
Order of starting publisher/verifying DAG was incorrect. Caused VDR to be empty when the DAG was verified, thus keys were missing for checking signatures.

Fixed and introduced an integration test to assert starting a node with existing DAG works properly. Also found a bug in VCR due to leia store not being closed on shutdown.


woutslakhorst (approved):

woutslakhorst:
You forgot to regenerate mocks

Issue #821, opened 2022-02-16
VCR: Switch to JSON-LD for VCs


woutslakhorst:
JSON-LD offers some benefits, mostly in flexibility. To support JSON-LD, the go-leia indexing lib must support JSON-LD natively. When it does, queries at the API layer can be passed to the DB and the DB itself can figure out which collection to use by matching the context from the query with the contexts used in a collection.

It would also allow for queries to be syntactically different, but semantically the same.

* [x] #943
* [x] Go-leia should define a new type of store: `LDStore`
* [x] Concepts needs to be refactored to indices (which canonicalized paths to use as index is still needed)
* [x] VCR search APIs must pass the JSON-LD query directly to the DB
* [x] Internal VC searches (auth) could use its own custom context (eg: nuts_organization_search_context) backed by hardcoded structs for easy access to the DB.
* [x] Migrate Existing VCs to new DB
   Existing VCs in the current DB need to be copied to the new DB. Newly issued VCs need to be added to both DBs

woutslakhorst:
All done

PR #824, opened 2022-02-18
added holder V2 search api

### LINKED ISSUE #822: Redefine VC search API
For both issuer and holder.

Searching VCs will be done by POSTing a JSON_LD document. The document will eventually be expanded and the canonicalized paths will be used to search in the DB. All nodes/values in the JSON-LD document must be present. Boolean OR and optional logic is not supported.

For this issue it's only needed to define the OAS spec and use hardcoded contexts that cover all current use-cases. The API will detect the context and use the old search API to find the correct documents. Technically it's a small change. A lot of documentation will need to be added though in order for users to transition to using JSON-LD.

#821 is the follow-up for the second stage of refactoring.

woutslakhorst:
closes #822 

The API requires a JSON-LD document to be sent. It also supports the `trusted` query param as defined in the v1 API.
Schema IRI's are placeholder. Any suggestions are welcome.

after acceptance: new issues need to be created:
- update Node docs on how to use this API (and how to use JSON-LD, this might require published context)
- update workshop docs
- update demo registry
- update demo ehr


woutslakhorst:
@stevenvegt queryselector changed to using types.

docs/_static/vcr/v2.yaml
```diff
@@ -10,6 +10,72 @@ info:
 servers:
   - url: http://localhost:1323
 paths:
+  /internal/vcr/v2/holder/vc/search:
```
reinkrul, line 13: How it matching performed? Do you need to pass the exact VC (kinda defeats the purpose of the search function)?

vcr/api/v2/holder.go
```diff
@@ -0,0 +1,169 @@
...
+			return c
+		}
+	}
+	return ""
+}
+
+// SearchVCs checks the context used in the JSON-LD query, based on the contents it maps to a non-JSON-LD query
+// After V1, this needs to be remapped to a DB search that supports native JSON-LD
+func (w *Wrapper) SearchVCs(ctx echo.Context, params SearchVCsParams) error {
+	var rawRequest typeHelper
```
reinkrul, line 58: it's OK, but just could even be an anonymous type inside this function! (since it's only used here).

vcr/api/v2/holder.go
```diff
@@ -0,0 +1,169 @@
...
+}
+
+// searchType returns the first type that is in this list:
+// - NutsOrganizationCredential
+// - NutsAuthorizationCredential
+func (ch typeHelper) searchType() string {
+	for _, c := range ch.Type {
+		switch c {
+		case credential.NutsOrganizationCredentialType:
+			fallthrough
```
reinkrul, line 47: fallthrough seems incorrect?

vcr/api/v2/holder.go
```diff
@@ -0,0 +1,169 @@
...
+		return err
+	}
+
+	return ctx.JSON(http.StatusOK, results)
+}
+
+func (w *Wrapper) searchAuths(ctx echo.Context, allowUntrusted bool, body []byte) error {
+	// some helper structs that reflect current required query params
+	type helper struct {
+		CredentialSubject struct {
```
reinkrul, line 130: this will all go when v1 API is removed, right?

vcr/types/interface.go
```diff
@@ -64,10 +65,18 @@ type ConceptFinder interface {
 	// It also returns untrusted credentials when allowUntrusted == true
 	Get(conceptName string, allowUntrusted bool, subject string) (concept.Concept, error)
 
-	// Search for matching concepts based upon a query. It returns an empty list if no matches have been found.
+	// SearchConcept returns matching concepts based upon a query. It returns an empty list if no matches have been found.
+	// It also returns untrusted credentials when allowUntrusted == true
+	// a context must be passed to prevent long-running queries
+	SearchConcept(ctx context.Context, conceptName string, allowUntrusted bool, query map[string]string) ([]concept.Concept, error)
```
reinkrul, line 71: is the `ConceptFinder` something that will be deprecated and replaced by JSON-LD schemas in the future?

docs/_static/vcr/v2.yaml
woutslakhorst: > Each field in the request body must be present in the VC in order for it to be passed as result.
So the less you enter in the request body, the more it finds.

vcr/api/v2/holder.go
woutslakhorst: if that type matches, it should return it, same goes for the other credential.
Expensive way of checking if `slice.Contains(x) => return x`

vcr/api/v2/holder.go
woutslakhorst: the underlying store (go-leia) must create indices based on JSON-LD instead of the current templates. Then the JSON-LD documents can be routed straight into go-leia without any conversion.
This PR is to provide a stable API for devs. Then in a next version we can change the underlying DB.

woutslakhorst: see #821 for v2 milestone

vcr/types/interface.go
woutslakhorst: yup

reinkrul (approved):

vcr/api/v2/holder.go
```diff
@@ -30,55 +29,38 @@ import (
...
 	}
 
-	err = json.Unmarshal(bodyBytes, &rawRequest)
-	if err != nil {
-		return core.InvalidInputError("failed to unmarshall JSON request body: %w", err)
-	}
+	//err = json.Unmarshal(bodyBytes, &rawRequest)
+	//if err != nil {
+	//	return core.InvalidInputError("failed to unmarshall JSON request body: %w", err)
+	//}
```
stevenvegt, line 44: Remove disabled code

stevenvegt (approved):
LGTM.
Remove disabled code in holder.go#41

PR #830, opened 2022-02-23
Private payload stores

dmeijboom:
As private credentials can contain (and often _will_ contain) sensitive data we can't store them in the same database as the public credentials. Creating a backup is also an issue as we would have to filter on private transactions or something similar.

This PR implements a solution for both problems. It works by adding a `payloadHeader` structure which either contains a byte slice (for public transations) or magic bytes (to indicate it's a private transaction) and a store ID (which is a sha256 hash). This also makes it backwards compatible as the format for public and previously created private transactions doesn't change.

This means that `WritePayload` requires the knowledge of the Store ID but `ReadPayload` doesn't. It also means that for public transactions reading/writing will stay `O(1)` but becomes `O(2)` for private transactions. 

Todo:
- [ ] Write unittests
- [ ] Make sure the Store ID is correct (I'm currently using `tx.PAL()[0]` but I'm sure that's not right..
- [ ] Make it work


reinkrul:
Will be replaced by new storage engine in v4

PR #825, opened 2022-02-21
V1 Credential revocation

stevenvegt:

PR #831, opened 2022-02-24
V1 Credential revocation

stevenvegt:
New PR based on #825.
Includes:
* Credential revocation with JSON-LD proofs
* Addition of `CredentialRevocation` in the Nuts JSON-LD context
* Introduction of the validator with its own validator store
* Credential validation on the v2 api checks for revocations

Other:
* Removes dependency on deprecated `github.com/pkg/errors`
* Cleanup of unused crypto code in vcr.go


vcr/api/v2/api.go
```diff
@@ -87,6 +86,14 @@ func (w Wrapper) IssueVC(ctx echo.Context) error {
 		public = *issueRequest.Visibility == IssueVCRequestVisibilityPublic
 	}
 
+	if issueRequest.Type == "" {
```
reinkrul, line 89: kinda out-of-scope change?

vcr/api/v2/api.go
```diff
@@ -87,6 +86,14 @@ func (w Wrapper) IssueVC(ctx echo.Context) error {
...
+		return core.InvalidInputError("missing credential type")
+	}
+
+	if issueRequest.CredentialSubject == nil {
```
reinkrul, line 93: what if the element is present, but is an empty list? Or an empty map?

vcr/api/v2/api.go
```diff
@@ -103,8 +110,16 @@ func (w Wrapper) IssueVC(ctx echo.Context) error {
...
+	credentialID, err := ssi.ParseURI(id)
+	if err != nil {
+		return core.InvalidInputError("invalid credential id: %w", err)
+	}
+
+	revocation, err := w.VCR.Revoke(*credentialID)
+	if err != nil {
+		return err
+	}
+	return ctx.JSON(200, revocation)
```
reinkrul, line 122:
```suggestion
	return ctx.JSON(http.StatusOK, revocation)
```

vcr/issuer/issuer.go
```diff
@@ -139,9 +140,58 @@ func (i issuer) buildVC(credentialOptions vc.VerifiableCredential) (*vc.Verifiab
...
+
+	ldProof := proof.NewLDProof(proof.ProofOptions{Created: time.Now()})
+	signingResult, err := ldProof.Sign(revocationAsMap, signature.JSONWebSignature2020{ContextLoader: i.contextLoader}, assertionKey)
+	if err != nil {
+		return nil, err
+	}
+
+	signingResultAsMap, ok := signingResult.(proof.SignedDocument)
+	if !ok {
+		return nil, errors.New("unable to cast signing result to SignedDocument")
```
reinkrul, line 187: `LDProof.Sign()` always returns a `SignedDocument`, and otherwise it isn't a recoverable error (by retrying/modifying input). So I'd say don't check the cast and let it blow up if it fails.

vcr/issuer/issuer.go
```diff
@@ -139,9 +140,58 @@ func (i issuer) buildVC(credentialOptions vc.VerifiableCredential) (*vc.Verifiab
...
+	}
+	b, _ = json.Marshal(signingResultAsMap)
+	fmt.Println(string(b))
```
reinkrul, line 190: letover debugging

vcr/issuer/issuer.go
```diff
@@ -139,9 +140,58 @@ func (i issuer) buildVC(credentialOptions vc.VerifiableCredential) (*vc.Verifiab
...
 }
 
-func (i issuer) Revoke(credentialID ssi.URI) error {
-	//TODO implement me
-	panic("implement me")
+func (i issuer) Revoke(credentialID ssi.URI) (*credential.Revocation, error) {
+	// first find it using a query on id.
+	credentialToRevoke, err := i.store.GetCredential(credentialID)
+	if err != nil {
+		return nil, fmt.Errorf("could not revoke: %w", err)
```
reinkrul, line 147: Add credential ID?
```suggestion
		return nil, fmt.Errorf("could not revoke (id=%s): %w", credentialID, err)
```

vcr/verifier/interface.go
```diff
@@ -0,0 +1,42 @@
...
+
+// Store defines the interface for a store for a verifier.
+// The store is filled with public information such as revoked credentials,
+// as well as local defined trust relations between issuer and credential type.
+type Store interface {
+	// GetRevocation find the revocation by the credential ID
+	// Returns an ErrNotFound when the revocation is not in the store
+	GetRevocation(id ssi.URI) (*credential.Revocation, error)
+	// StoreRevocation writes a revocation to storage.
+	StoreRevocation(r credential.Revocation) error
```
reinkrul, line 39: API naming is a bit assymetric: Get v.s. Set, Store v.s. Read/Retrieve

vcr/verifier/interface.go
```diff
@@ -0,0 +1,42 @@
...
+	// Close closes and frees the underlying resources the store uses.
+	Close() error
```
reinkrul, line 41: you could let it implement `io.Closer` (same func)

vcr/verifier/interface.go
```diff
@@ -0,0 +1,42 @@
...
+	// if it has been revoked
+	// if the issuer is registered as trusted
+	Verify(credential vc.VerifiableCredential, allowUntrusted bool, checkSignature bool, validAt *time.Time) error
+	// Validate checks the verifiable credential technical correctness
+	Validate(credentialToVerify vc.VerifiableCredential, at *time.Time) error
+	// IsRevoked checks if the credential is revoked
+	IsRevoked(credentialID ssi.URI) (bool, error)
+	// CheckAndStoreRevocation accepts a signed document, checks the signature and type.
+	// If all is valid, it stores the revocation.
+	CheckAndStoreRevocation(document proof.SignedDocument) error
```
reinkrul, line 25: This function feels a bit misplaced here. Maybe because it combines both verifying the revocation (probably OK) and storing it, a responsibility of the `Store` (below), causing it to just pass it on? Hence the long name.

Possible improvement, 2 options I see (although not fantastic either):

- Split `CheckAndStoreRevocation` in `VerifyRevocation` and `RegisterRevocation`
- Just provide `VerifyRevocation` here and have the caller store it directly in the store

vcr/verifier/interface.go
```diff
@@ -0,0 +1,42 @@
...
+}
+
+// ErrNotFound is returned when a credential or revocation can not be found based on its ID.
+var ErrNotFound = errors.New("not found")
```
reinkrul, line 29: Better use `types.ErrNotFound`, but that might cause a cyclic import. I refactored away most of the `types` package in my PR (to avoid that cylic import)

vcr/verifier/verifier.go
```diff
@@ -135,6 +126,14 @@ func (v verifier) Verify(credentialToVerify vc.VerifiableCredential, allowUntrus
 		return err
 	}
 
+	revoked, err := v.IsRevoked(*credentialToVerify.ID)
+	if revoked {
+		return errors.New("credential is revoked")
+	}
+	if err != nil {
```
reinkrul, line 133: as a matter of principle, better to check `err` first, then the regular return values?

vcr/verifier/verifier.go
```diff
@@ -145,3 +144,58 @@ func (v verifier) Verify(credentialToVerify vc.VerifiableCredential, allowUntrus
...
+}
+
+func (v *verifier) CheckAndStoreRevocation(document proof.SignedDocument) error {
+	asBytes, err := json.Marshal(document)
+	revocation := credential.Revocation{}
+	if err := json.Unmarshal(asBytes, &revocation); err != nil {
+		return err
+	}
+
+	if err := credential.ValidateRevocation(revocation); err != nil {
```
reinkrul, line 166: should `revocation.Type` be checked?

vcr/verifier/verifier.go
```diff
@@ -145,3 +144,58 @@ func (v verifier) Verify(credentialToVerify vc.VerifiableCredential, allowUntrus
...
+	ldProof := proof.LDProof{}
+	if err := document.UnmarshalProofValue(&ldProof); err != nil {
+		return err
+	}
+	err = ldProof.Verify(document.DocumentWithoutProof(), signature.JSONWebSignature2020{ContextLoader: v.contextLoader}, pk)
+	if err != nil {
+		return fmt.Errorf("unable to verify revocation signature: %w", err)
+	}
+
+	if err := v.store.StoreRevocation(revocation); err != nil {
```
reinkrul, line 197: Since `IsRevoked` is only supplied the credential ID (which can be whatever URI), can I (issuer `did:nuts:rein`) revoke you credentials if I specify IDs of credentials issued by you (issuer `did:nuts:steven`)? Shouldn't `IsRevoked` be also supplied with the issuer URI? Then you're safe, even when issuers accidentally (or on purpose) issue VCs with colliding IDs.

reinkrul (changes requested):

vcr/api/v2/api.go
stevenvegt: Yes it is.

vcr/issuer/issuer.go
stevenvegt: Will result in nasty stacktraces, but that is probably a good thing.

vcr/verifier/verifier.go
stevenvegt: The revocation validation step checks if the issuer of the revocation is also the issuer of the revoked credential. So, the answer should be no.

vcr/api/v2/api.go
stevenvegt: Yes, that is not trivial, we could check for all cases but that is a bit verbose. I think it will fail eventually, but this is meant as a reminder that a credentialSubject is required. So if you provide a `"credentialSubject": []`. This check doesn't fail, but you can say that is was a conscious decision.

reinkrul (approved):

PR #839, opened 2022-02-28
Adds TransactionListQuery and response to v2 protocol

### LINKED ISSUE #832: V2: implement TransactionList
Including both TransactionListQuery and TransactionList messages.

part of #666

woutslakhorst:
Closes #832 

There are some todo's in there. They are a reminder to change after the entire v2 protocol is implemented and tested.

During implementation, some questions regarding the protocol came up:

- [x] Do we always need to respond on a TransactionListQuery? If we have nothing to send, can't we just ignore the message? This might make stuff easier?
- [x] Nothing is specced on what to do when a transaction (or payload) is missing. The current implementation stops adding transactions to the message based on LC values. It sends what it can. This also links to the next point.
- [x] How to mark the conversation as done? If we do not know for sure how many chunks are to be expected.
- [x] What to do with received transactions where the payload is missing? This is also not specced

TODO:
- [x] integration test with lots of TXs.

### DIFF
--- a/network/transport/v2/conversation.go
+++ b/network/transport/v2/conversation.go
@@ -111,11 +111,11 @@ func (cMan *conversationManager) done(cid conversationID) {
 }
 
 // startConversation sets a conversationID on the envelope and stores the conversation
-func (cMan *conversationManager) startConversation(envelope checkable) (newConversation conversation) {
+func (cMan *conversationManager) startConversation(envelope checkable) conversation {
 	cid := newConversationID()
 
 	envelope.setConversationID(cid)
-	newConversation = conversation{
+	newConversation := conversation{
 		conversationID:   cid,
 		createdAt:        time.Now(),
 		conversationData: envelope,
@@ -126,7 +126,7 @@ func (cMan *conversationManager) startConversation(envelope checkable) (newConve
 
 	cMan.conversations[cid.String()] = newConversation
 
-	return
+	return newConversation
 }
 
 func (cMan *conversationManager) check(envelope isEnvelope_Message) error {
--- a/network/transport/v2/conversation.go
+++ b/network/transport/v2/conversation.go
@@ -50,6 +50,8 @@ type conversation struct {
 	conversationID   conversationID
 	createdAt        time.Time
 	conversationData checkable
+	// additionalInfo can be used to check if a conversation is done
+	additionalInfo map[string]interface{}
 }
 
 type checkable interface {
@@ -64,13 +66,13 @@ type conversationable interface {
 
 type conversationManager struct {
 	mutex         sync.RWMutex
-	conversations map[string]conversation
+	conversations map[string]*conversation
 	validity      time.Duration
 }
 
 func newConversationManager(validity time.Duration) *conversationManager {
 	return &conversationManager{
-		conversations: map[string]conversation{},
+		conversations: map[string]*conversation{},
 		validity:      validity,
 	}
 }
@@ -111,14 +113,15 @@ func (cMan *conversationManager) done(cid conversationID) {
 }
 
 // startConversation sets a conversationID on the envelope and stores the conversation
-func (cMan *conversationManager) startConversation(envelope checkable) conversation {
+func (cMan *conversationManager) startConversation(envelope checkable) *conversation {
 	cid := newConversationID()
 
 	envelope.setConversationID(cid)
-	newConversation := conversation{
+	newConversation := &conversation{
 		conversationID:   cid,
 		createdAt:        time.Now(),
 		conversationData: envelope,
+		additionalInfo:   map[string]interface{}{},
 	}
 
 	cMan.mutex.Lock()
--- a/network/transport/v2/conversation.go
+++ b/network/transport/v2/conversation.go
@@ -21,7 +21,7 @@ package v2
 
 import (
 	"context"
-	"encoding/hex"
+	"errors"
 	"fmt"
 	"sync"
 	"time"
@@ -32,40 +32,34 @@ import (
 
 var maxValidity = 30 * time.Second
 
-type conversationID uuid.UUID
+type conversationID string
 
 func newConversationID() conversationID {
-	return conversationID(uuid.New())
-}
-
-func parseConversationID(bytes []byte) (cid conversationID, err error) {
-	var u uuid.UUID
-	u, err = uuid.FromBytes(bytes)
-	if err != nil {
-		return
-	}
-	cid = conversationID(u)
-	return
+	return conversationID(uuid.New().String())
 }
 
 func (cid conversationID) slice() []byte {
-	// error not possible when marshalling
-	bytes, _ := uuid.UUID(cid).MarshalBinary()
-	return bytes
+	return []byte(cid)
 }
 
 func (cid conversationID) String() string {
-	return hex.EncodeToString(cid.slice())
+	return string(cid)
 }
 
 type conversation struct {
 	conversationID   conversationID
 	createdAt        time.Time
-	conversationData conversationData
+	conversationData checkable
 }
 
-type conversationData interface {
+type checkable interface {
+	conversationable
 	checkResponse(envelope isEnvelope_Message) error
+}
+
+type conversationable interface {
+	setConversationID(cid conversationID)
+	conversationID() []byte
 }
 
 type conversationManager struct {
@@ -116,22 +110,15 @@ func (cMan *conversationManager) done(cid conversationID) {
 	delete(cMan.conversations, cid.String())
 }
 
-// conversationFromEnvelope sets a conversationID on the envelope and stores the conversation
-func (cMan *conversationManager) conversationFromEnvelope(envelope isEnvelope_Message) (newConversation conversation) {
+// startConversation sets a conversationID on the envelope and stores the conversation
+func (cMan *conversationManager) startConversation(envelope checkable) (newConversation conversation) {
 	cid := newConversationID()
 
-	switch t := envelope.(type) {
-	case *Envelope_TransactionListQuery:
-		t.TransactionListQuery.ConversationID = cid.slice()
-		newConversation = conversation{
-			conversationID: cid,
-			createdAt:      time.Now(),
-			conversationData: transactionListConversation{
-				msg: t,
-			},
-		}
-	default:
-		return
+	envelope.setConversationID(cid)
+	newConversation = conversation{
+		conversationID:   cid,
+		createdAt:        time.Now(),
+		conversationData: envelope,
 	}
 
 	cMan.mutex.Lock()
@@ -143,19 +130,13 @@ func (cMan *conversationManager) conversationFromEnvelope(envelope isEnvelope_Me
 }
 
 func (cMan *conversationManager) check(envelope isEnvelope_Message) error {
-	var cidBytes []byte
-
-	switch t := envelope.(type) {
-	case *Envelope_TransactionList:
-		cidBytes = t.TransactionList.ConversationID
-	default:
-		return fmt.Errorf("invalid response msg type: %s", t)
+	otherEnvelope, ok := envelope.(conversationable)
+	if !ok {
+		return errors.New("expecting envelope to contain be a conversation response type message")
 	}
 
-	cid, err := parseConversationID(cidBytes)
-	if err != nil {
-		return fmt.Errorf("failed to parse conversationID: %w", err)
-	}
+	cidBytes := otherEnvelope.conversationID()
+	cid := conversationID(cidBytes)
 
 	cMan.mutex.RLock()
 	defer cMan.mutex.RUnlock()
@@ -167,15 +148,22 @@ func (cMan *conversationManager) check(envelope isEnvelope_Message) error {
 	}
 }
 
-type transactionListConversation struct {
-	msg *Envelope_TransactionListQuery
+func (envelope *Envelope_TransactionListQuery) setConversationID(cid conversationID) {
+	envelope.TransactionListQuery.ConversationID = cid.slice()
 }
 
-func (c transactionListConversation) checkResponse(envelope isEnvelope_Message) error {
-	// envelope type already checked in cMan.check()
-	otherEnvelope := envelope.(*Envelope_TransactionList)
+func (envelope *Envelope_TransactionListQuery) conversationID() []byte {
+	return envelope.TransactionListQuery.ConversationID
+}
 
-	payloadRequest := c.msg.TransactionListQuery
+func (envelope *Envelope_TransactionListQuery) checkResponse(other isEnvelope_Message) error {
+	// envelope type already checked in cMan.check()
+	otherEnvelope, ok := other.(*Envelope_TransactionList)
+	if !ok {
+		return errors.New("checking wrong envelope type")
+	}
+
+	payloadRequest := envelope.TransactionListQuery
 	payloadResponse := otherEnvelope.TransactionList
 
 	// as map for easy finding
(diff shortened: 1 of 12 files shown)


reinkrul:
> How to mark the conversation as done? If we do not know for sure how many chunks are to be expected.

If this is something that's required, we could add a bool flag `isLast`, which is `true` for the last respond message in the conversation?

woutslakhorst:
> > How to mark the conversation as done? If we do not know for sure how many chunks are to be expected.
> 
> If this is something that's required, we could add a bool flag `isLast`, which is `true` for the last respond message in the conversation?

Or we just don't care. Any response can only list transactions we requested. Any duplicates wouldn't matter since they would not be processed again. After 30 seconds nothing may be sent anyway....

network/transport/v2/handlers.go
```diff
@@ -151,10 +158,114 @@ func (p protocol) handleGossip(peer transport.Peer, msg *Gossip) error {
...
+	ctx := context.Background()
+	for _, tx := range msg.Transactions {
+		transactionRef := hash.FromSlice(tx.Hash)
+		transaction, err := dag.ParseTransaction(tx.Data)
+		if err != nil {
+			return fmt.Errorf("received transaction is invalid (peer=%s, ref=%s): %w", peer, transactionRef, err)
+		}
+
+		present, err := p.state.IsPresent(ctx, transactionRef)
+		if !present {
```
reinkrul, line 193: `err` being ignored

network/transport/v2/handlers.go
```diff
@@ -151,10 +158,114 @@ func (p protocol) handleGossip(peer transport.Peer, msg *Gossip) error {
...
 	p.gManager.GossipReceived(peer.ID, refs...)
 	// TODO compare xor
 	// TODO send new message
 
 	return nil
 }
+
+func (p *protocol) handleTransactionList(peer transport.Peer, envelope *Envelope_TransactionList) error {
+	msg := envelope.TransactionList
+	cid, err := parseConversationID(msg.ConversationID)
```
reinkrul, line 172: If a brand new node is spawned, is 30 seconds (max. conversation lifetime) enough for the DAG to be sync'd?

network/transport/v2/handlers.go
```diff
@@ -151,10 +158,114 @@ func (p protocol) handleGossip(peer transport.Peer, msg *Gossip) error {
...
+		transactionRef := hash.FromSlice(tx.Hash)
+		transaction, err := dag.ParseTransaction(tx.Data)
+		if err != nil {
+			return fmt.Errorf("received transaction is invalid (peer=%s, ref=%s): %w", peer, transactionRef, err)
+		}
+
+		present, err := p.state.IsPresent(ctx, transactionRef)
+		if !present {
+			// TODO does this always trigger fetching missing payloads? (through observer on DAG) Prolly not for v2
+			if err = p.state.Add(ctx, transaction, tx.Payload); err != nil {
```
reinkrul, line 195: Who checks whether `tx.PayloadHash == hash(tx.Payload)`?

network/transport/v2/handlers.go
```diff
@@ -151,10 +158,114 @@ func (p protocol) handleGossip(peer transport.Peer, msg *Gossip) error {
...
+	for _, tx := range msg.Transactions {
+		transactionRef := hash.FromSlice(tx.Hash)
+		transaction, err := dag.ParseTransaction(tx.Data)
+		if err != nil {
+			return fmt.Errorf("received transaction is invalid (peer=%s, ref=%s): %w", peer, transactionRef, err)
+		}
+
+		present, err := p.state.IsPresent(ctx, transactionRef)
+		if !present {
+			// TODO does this always trigger fetching missing payloads? (through observer on DAG) Prolly not for v2
```
reinkrul, line 194: does this also work for private TXs?

network/transport/v2/handlers.go
```diff
@@ -151,10 +158,114 @@ func (p protocol) handleGossip(peer transport.Peer, msg *Gossip) error {
...
+	}
+	// TODO convert to trace logging
+	log.Logger().Infof("handling transactionListQuery from peer (peer=%s, conversationID=%s)", peer.ID.String(), cid.String())
+
+	for i, refBytes := range msg.Refs {
+		requestedRefs[i] = hash.FromSlice(refBytes)
+	}
+
+	if len(requestedRefs) == 0 {
+		log.Logger().Warnf("peer sent request for 0 transactions (peer=%s)", peer.ID)
```
reinkrul, line 223: conversation is done?

network/transport/v2/handlers.go
```diff
@@ -151,10 +158,114 @@ func (p protocol) handleGossip(peer transport.Peer, msg *Gossip) error {
...
+
+	ctx := context.Background()
+
+	// first retrieve all transactions, this is needed to sort them on LC value
+	for _, ref := range requestedRefs {
+		transaction, err := p.state.GetTransaction(ctx, ref)
+		if err != nil {
+			return err
+		}
+		// If a transaction is not present, we stop any further transaction gathering.
```
reinkrul, line 235: How is `requestedRefs` ordered? Depending on order (LC ascending/descending, random order, some other order) it makes sense to stop gathering TXs or not. If the order is random, there could be a very recent TX halfway that we  just don't have yet (while the rest could still be sent back).

Should it be spec'd?

network/transport/v2/handlers.go
```diff
@@ -151,10 +158,114 @@ func (p protocol) handleGossip(peer transport.Peer, msg *Gossip) error {
...
+		if transaction == nil {
+			log.Logger().Warnf("peer requested transaction we don't have (peer=%s, node=%s, ref=%s)", peer.ID, peer.NodeDID.String(), ref.String())
+			break
+		}
+		unsorted = append(unsorted, transaction)
+	}
+
+	// now we sort on LC value
+	sort.Slice(unsorted, func(i, j int) bool {
```
reinkrul, line 244: and now `unsorted` is actually sorted

network/transport/v2/handlers.go
```diff
@@ -151,10 +158,114 @@ func (p protocol) handleGossip(peer transport.Peer, msg *Gossip) error {
...
+	if len(requestedRefs) == 0 {
+		log.Logger().Warnf("peer sent request for 0 transactions (peer=%s)", peer.ID)
+		return nil
+	}
+
+	ctx := context.Background()
+
+	// first retrieve all transactions, this is needed to sort them on LC value
+	for _, ref := range requestedRefs {
+		transaction, err := p.state.GetTransaction(ctx, ref)
```
reinkrul, line 231: do we reuse the BBolt TX here? Would that benefit performance?

network/transport/v2/handlers.go
```diff
@@ -151,10 +158,114 @@ func (p protocol) handleGossip(peer transport.Peer, msg *Gossip) error {
...
+
+		// do not add private TX payloads
+		if len(transaction.PAL()) == 0 {
+			payload, err := p.state.ReadPayload(ctx, transaction.PayloadHash())
+			if err != nil {
+				return err
+			}
+			// TODO we abort here as well, since there's no mechanism for missing payloads on public transactions in v2 protocol
+			if payload == nil {
+				log.Logger().Warnf("peer requested transaction with missing payload (peer=%s, node=%s, ref=%s)", peer.ID, peer.NodeDID.String(), transaction.Ref().String())
```
reinkrul, line 262: should we sent an error back to the peer then, instead of just responding with what we collected up until now?

If you return an error from the handler, it is returned to the peer.

network/transport/v2/senders.go
```diff
@@ -46,3 +47,84 @@ func (p *protocol) sendGossipMsg(id transport.PeerID, refs []hash.SHA256Hash) er
...
+		return errors.New("no connection available")
+	}
+
+	// there shouldn't be more than a ~650 in there, this will fit in a single message
+	refsAsBytes := make([][]byte, len(refs))
+	for i, ref := range refs {
+		refsAsBytes[i] = ref.Slice()
+	}
+
+	envelop := &Envelope_TransactionListQuery{
```
reinkrul, line 64: envelope

network/transport/v2/senders.go
```diff
@@ -46,3 +47,84 @@ func (p *protocol) sendGossipMsg(id transport.PeerID, refs []hash.SHA256Hash) er
...
+		TransactionListQuery: &TransactionListQuery{
+			Refs: refsAsBytes,
+		},
+	}
+
+	conversation := p.cMan.conversationFromEnvelope(envelop)
```
reinkrul, line 70: must say `conversationFromEnvelope` is a bit of an oddball func (switching on types, setting fields). Maybe nicer to make the applicable messages implement an interface, so it can be solved a bit safer/generic?

```go
type Conversatable interface { // or whatever
   SetConversationID(cid conversationID)
}

// Provide this for every message that supports a conversation ID
func (envelope *Envelope_TransactionListQuery) SetConversationID(cid conversationID) {
	envelope.TransactionListQuery.ConversationID = cid.slice()
}
```

network/transport/v2/senders.go
```diff
@@ -46,3 +47,84 @@ func (p *protocol) sendGossipMsg(id transport.PeerID, refs []hash.SHA256Hash) er
...
+	log.Logger().Infof("requesting transactions from peer (peer=%s, conversationID=%s)", id, conversation.conversationID.String())
+
+	return conn.Send(p, &Envelope{Message: envelop})
+}
+
+func (p *protocol) sendTransactionList(id transport.PeerID, conversationID conversationID, transactions []*Transaction) error {
+	conn := p.connectionList.Get(grpc.ByConnected(), grpc.ByPeerID(id))
+	if conn == nil {
+		// shouldn't happen
+		return errors.New("no connection available")
```
reinkrul, line 81: maybe make this error a constant/var

network/transport/v2/senders.go
```diff
@@ -46,3 +47,84 @@ func (p *protocol) sendGossipMsg(id transport.PeerID, refs []hash.SHA256Hash) er
...
+// chunkTransactionList splits a large set of transactions into smaller sets. Each set adheres to the maximum message size.
+func chunkTransactionList(transactions []*Transaction) [][]*Transaction {
+	chunked := make([][]*Transaction, 0)
+
+	currentSize := 0
+	newSize := 0
+	startIndex := 0
+	endIndex := 0
+
+	// TODO to be tested in practise
```
reinkrul, line 107: Using gRPC/Protobuf reflect you can calculate the would-be message size (maybe for a test with some small/large random TXs)

reinkrul (changes requested):

network/transport/v2/handlers.go
woutslakhorst: If it can't do it within this conversation, the protocol should just continue in the next gossip/state sequence with a new conversation.

network/transport/v2/handlers.go
woutslakhorst: There's a TODO for that: state.go#93. Will implement.

network/transport/v2/handlers.go
woutslakhorst: Yes because that is handled through the observers. From the publisher, the private payload fetcher is triggered.

network/transport/v2/handlers.go
woutslakhorst: Added to State.Add

network/transport/v2/handlers.go
woutslakhorst: this is the node being queried, it does not have the conversation in its storage. The other node has it.

network/transport/v2/handlers.go
woutslakhorst: order is unknown, but this query is sent as a direct result of the state this node has. So logically it's almost impossible to miss anything. To make it a bit more resilient, the node can just send whatever it has.

network/transport/v2/handlers.go
woutslakhorst: I opted for not using a long running TX. The main performance bottleneck would be retrieving pages from disk. A read transaction would block any write transactions if that write needs to reorganize pages. Then reads block writes that block reads in other processes. By keeping these TXs small, the writes can peep inbetween.

network/transport/v2/handlers.go
woutslakhorst: Another node might have the payload, is this way nodes can share the burden. The sync with the other node will continue where this one was left. In a pure v2 protocol implementation, we wouldn't accept transactions without payloads anymore.

network/transport/v2/senders.go
woutslakhorst: that would require message chunking based on trial and error. (does this fit, no too large, this then?, still too large)

network/transport/v2/senders.go
woutslakhorst: refactored

network/transport/v2/conversation.go
```diff
@@ -32,42 +32,36 @@ import (
...
 func (cid conversationID) slice() []byte {
-	// error not possible when marshalling
-	bytes, _ := uuid.UUID(cid).MarshalBinary()
-	return bytes
+	return []byte(cid)
 }
 
 func (cid conversationID) String() string {
-	return hex.EncodeToString(cid.slice())
+	return string(cid)
```
gerardsn, line 46: isn't cid already a string?

gerardsn:
> > > How to mark the conversation as done? If we do not know for sure how many chunks are to be expected.
> > 
> > 
> > If this is something that's required, we could add a bool flag `isLast`, which is `true` for the last respond message in the conversation?
> 
> Or we just don't care. Any response can only list transactions we requested. Any duplicates wouldn't matter since they would not be processed again. After 30 seconds nothing may be sent anyway....

What happens if a node takes longer than the gossip interval to complete reconciliation? Could it end up sending a bunch of similar requests? Is sending a new request to the same peer desirable while there is still an open request desirable behavior?

Can we remove received Txs from the list of requested refs? Once the list of requested refs is empty the conversation can be marked done. This fails when the peer does not have all Txs, but it should not receive a request for Txs it doesn't have anyway.

gerardsn (commented):
some minor things

woutslakhorst:
> What happens if a node takes longer than the gossip interval to complete reconciliation? Could it end up sending a bunch of similar requests? Is sending a new request to the same peer desirable while there is still an open request desirable behavior?

We could say that a node should not initiate a conversation if one is already active on a connection. This would also require conversations to be ended correctly (otherwise it'll have to wait for 30 seconds when it could have been faster)?

network/transport/v2/handlers.go
```diff
@@ -151,10 +158,113 @@ func (p protocol) handleGossip(peer transport.Peer, msg *Gossip) error {
...
+	unsorted := make([]dag.Transaction, 0)
+
+	cid := conversationID(msg.ConversationID)
+
+	// TODO convert to trace logging
+	log.Logger().Infof("handling transactionListQuery from peer (peer=%s, conversationID=%s)", peer.ID.String(), cid.String())
+
+	for i, refBytes := range msg.Refs {
+		requestedRefs[i] = hash.FromSlice(refBytes)
+	}
```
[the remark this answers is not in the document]
woutslakhorst: The gossip part handles duplication. At this point we trust everything is done well. Would be extra logic at each sender/response that doesn't do anything....

reinkrul (approved):

woutslakhorst:
@gerardsn I'll make an issue for the non-overlapping conversation issue.

woutslakhorst:
Latest addition includes marking the TransactionListQuery as done.

gerardsn (approved):
LGTM

Issue #835, opened 2022-02-28
Very slow start due to DAG persistent replay


reinkrul:
Since stores were made persistent (VDR and VCR) startup with a filled DAG became very slow. This is because on startup, the DAG is replayed, causing all DID documents and VCs, etc to be written to storage, although they're already there.

This is due to every database flush taking approx. 50ms (measured on fast development system), which happens on every write.

Solution would be to allow DAG replay to be disabled. Ultimately the performance bottleneck will also have to be resolved (https://github.com/nuts-foundation/nuts-node/issues/836).

reinkrul:
Fixed by reverting persisted VDR

PR #840, opened 2022-02-28
Network: replace lastTransactionTracker with calls to DAG/PayloadStore

reinkrul:
In preparation of optional DAG replay (https://github.com/nuts-foundation/nuts-node/issues/835).

The `lastTransactionTracker` tracked the last TXs on the DAG with payload, to avoid locally creating a TX which prevs payload(s) were missing, leading to a locally unprocessable transaction. 

This tracker subscribed to the DAG but only kept track in-memory, so to allow disabling DAG replay it had to either start persisting its internal state or replace it altogether.

### DIFF
--- a/network/dag/interface.go
+++ b/network/dag/interface.go
@@ -57,6 +57,9 @@ type State interface {
 	// RegisterObserver allows observers to be notified when a transaction is added to the DAG.
 	// If the observer needs to be called within the transaction, transactional must be true.
 	RegisterObserver(observer Observer, transactional bool)
+	// GetHeads returns the heads of the DAG which payload is present. If the payload of the last DAG heads is missing,
+	// their prevs are inspected as possible heads to be returned by this function.
+	GetHeads(ctx context.Context) ([]hash.SHA256Hash, error)
 	// Subscribe lets an application subscribe to a specific type of transaction. When a new transaction is received
 	// the `receiver` function is called. If an asterisk (`*`) is specified as `payloadType` the receiver is subscribed
 	// to all payload types.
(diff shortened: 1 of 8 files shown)


reinkrul:
> I think this solution is overly complicated. The only reason is that you don't want to refer to transactions for which a payload is missing. Missing payloads will be extremely common due to private transactions. If only sender and receiver of private transactions can refer to those transactions, that will also leak for who that transaction was.
> 
> I think we should switch to referring to TXs even if there payload is missing. The transaction is still valid and will always be there, so why not refer to it. It'll reduce the complexity of the GetHeads function to `s.graph.Heads`

Problem is we require for some TXs (containing DID documents) that the payload is present, otherwise we won't publish any (succeeding) TXs from that point on. But since v1 doesn't atomically sync TX + payload, a node could publish a TX with a DID document, while never publishing the payload. The publisher will never publish that, and succeeding TXs. This happened during the last hackaton, preventing any new TXs to be processed.

Now I think of it, this PR doesn't fix this case either: a node could first publish a TX (type = DID Document) without payload, and then a subsequent TX (referencing it) with payload. Other nodes will then errornuously use this last TX with payload as prev, creating a TX that will never be published.

Since this is caused by non-atomicity for blocking TXs in v1, this isn't a problem any more in v2. But we have to fix it in the mean time.

I think the best/easiest solution for now is to persist the state of `lastTransactionTracker`.

woutslakhorst:
bbolt vdr implementation caused some of the delay but appears to be broken: #846 
Fixing that might solve some of the issues.

reinkrul:
Not required for now.

PR #844, opened 2022-03-01
VCR: API to create Verifiable Presentations

reinkrul:
Based on https://github.com/nuts-foundation/nuts-node/pull/831


stevenvegt (approved):
LGTM.
So, no verification?
This is going to clash with #831 since that one also moves around some of the interfaces.

PR #856, opened 2022-03-03
VCR: Support verifying VPs

reinkrul:
Based on https://github.com/nuts-foundation/nuts-node/pull/844

### DIFF
--- a/docs/_static/vcr/v2.yaml
+++ b/docs/_static/vcr/v2.yaml
@@ -222,7 +222,7 @@ paths:
         * If the issuers of the verifiable credentials were not deactivated at time of issuing
 
         error returns:
-        * 400 - One or more of the given parameters are invalid
+        * 400 - A parameter or the verifiable presentation is invalid
         * 500 - An error occurred while processing the request
       operationId: "verifyVP"
       tags:
--- a/docs/_static/vcr/v2.yaml
+++ b/docs/_static/vcr/v2.yaml
@@ -210,6 +210,38 @@ paths:
                 $ref: '#/components/schemas/VCVerificationResult'
         default:
           $ref: '../common/error_response.yaml'
+  /internal/vcr/v2/verifier/vp:
+    post:
+      summary: Verifies a Verifiable Presentation
+      description: |
+        Verifies a Verifiable Presentation. It checks:
+        * Signature of the verifiable presentation and the verifiable credentials
+        * Expiration
+        * Revocation status
+        * If the issuers of the verifiable credentials are trusted
+        * If the issuers of the verifiable credentials were not deactivated at time of issuing
+
+        error returns:
+        * 400 - One or more of the given parameters are invalid
+        * 500 - An error occurred while processing the request
+      operationId: "verifyVP"
+      tags:
+        - credential
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/VPVerificationRequest'
+      responses:
+        "200":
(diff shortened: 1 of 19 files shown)


docs/_static/vcr/v2.yaml
```diff
@@ -210,6 +210,38 @@ paths:
...
+      description: |
+        Verifies a Verifiable Presentation. It checks:
+        * Signature of the verifiable presentation and the verifiable credentials
+        * Expiration
+        * Revocation status
+        * If the issuers of the verifiable credentials are trusted
+        * If the issuers of the verifiable credentials were not deactivated at time of issuing
+
+        error returns:
+        * 400 - One or more of the given parameters are invalid
```
woutslakhorst, line 225: I would expect one of the above rules to fail since the call has no params.

vcr/api/v2/api.go
```diff
@@ -229,3 +231,34 @@ func (w *Wrapper) CreateVP(ctx echo.Context) error {
 	}
 	return ctx.JSON(http.StatusOK, vp)
 }
+
+// VerifyVP handles API request to verify a Verifiable Presentation.
+func (w *Wrapper) VerifyVP(ctx echo.Context) error {
```
woutslakhorst, line 236: The number of checks do not represent the number of 400 errors that are returned by this code. It seems some are returned as 500s?

vcr/api/v2/api_test.go
```diff
@@ -540,6 +541,63 @@ func TestWrapper_CreateVP(t *testing.T) {
...
+			return nil
+		})
+		testContext.mockVerifier.EXPECT().VerifyVP(vp, false, nil).Return(vp.VerifiableCredential, nil)
+		testContext.echo.EXPECT().JSON(http.StatusOK, vp.VerifiableCredential)
+
+		err := testContext.client.VerifyVP(testContext.echo)
+
+		assert.NoError(t, err)
+	})
+	t.Run("error - verification failed", func(t *testing.T) {
```
woutslakhorst, line 585: Missing tests on other error conditions as stated in OAS

vcr/api/v2/api.go
woutslakhorst: This could simply be fixed by mapping the validationError to a status code.

woutslakhorst (changes requested):
Also some linter errors

docs/_static/vcr/v2.yaml
reinkrul: I'll change the 400 description to parameters or VP.

Also, 400 is returned when the VP itself can't be verified (e,g. invalid signature, not trusted), which is ambiguous. So i'll change that to 422

vcr/api/v2/api.go
reinkrul: 400 is returned when `Bind()` fails (see `github.com/labstack/echo/v4@v4.6.3/bind.go:73`)

vcr/api/v2/api_test.go
reinkrul: Added tests for failed VP validation, 400 and 500 are returned by Echo

stevenvegt (approved):
LGTM

reinkrul:
> LGTM

Thanks, but wasn't even finished yet ;-)

woutslakhorst (approved):
only question would be the 422 vs 200. A valid response with an unverified VP, would that be a 200 or 422?

A use may try the call again, the outcome would be the same but it could. 4xx return calls say that the same call may not be done again.

PR #866, opened 2022-03-10
Replace murmur3 dependency

gerardsn:
The currently used implementation of murmur3 hash [github.com/spaolacci/murmur3](https://github.com/spaolacci/murmur3) has some issues due to changed behavior in Go and the project appears to be abandoned. This PR replaces the dependency with a fork that is still actively maintained [github.com/twmb/murmur3](https://github.com/twmb/murmur3).

### DIFF
--- a/crl/validator.go
+++ b/crl/validator.go
@@ -33,7 +33,7 @@ import (
 	"sync"
 	"time"
 
-	"github.com/spaolacci/murmur3"
+	"github.com/twmb/murmur3"
 )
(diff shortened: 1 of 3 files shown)


reinkrul:
Looks ok, although the replacement lib does look pretty dead, but less dead than the first one. But I guess if it works and is compatible with the latest Go versions, it's ok.

reinkrul (approved):

PR #895, opened 2022-03-18
Replace ParseURI with MustParseURI where returned error is ignored

reinkrul:
Declutters (test) code.

### DIFF
--- a/vdr/doc/resolvers.go
+++ b/vdr/doc/resolvers.go
@@ -210,7 +210,9 @@ func (r KeyResolver) ResolveKeyAgreementKey(id did.DID) (crypto.PublicKey, error
 func ExtractAssertionKeyID(doc did.Document) (ssi.URI, error) {
 	keys := doc.AssertionMethod
 	for _, key := range keys {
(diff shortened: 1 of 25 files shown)


vdr/doc/resolvers.go
```diff
@@ -210,9 +210,7 @@ func (r KeyResolver) ResolveKeyAgreementKey(id did.DID) (crypto.PublicKey, error
 func ExtractAssertionKeyID(doc did.Document) (ssi.URI, error) {
 	keys := doc.AssertionMethod
 	for _, key := range keys {
-		kid := key.ID.String()
-		u, _ := ssi.ParseURI(kid)
-		return *u, nil
+		return ssi.MustParseURI(key.ID.String()), nil
```
woutslakhorst, line 213: @reinkrul  Could this possibly open up an attack by creating wrong key IDs?

reinkrul: Probably safer to revert it. IDs should've been validated, but panicing if one slips through is not a very good idea either. Then better just return the first valid KID.

woutslakhorst (approved):

PR #898, opened 2022-03-18
VCR: return application/json from v2 api instead of application/ld+json

reinkrul:
To aid code generation


woutslakhorst (commented):
What were the consequences for code generation? Since ld+json is the correct content type

reinkrul:
It only generates `SearchVCWithRequestBody()` which you have to feed a `[]byte` body, since OpenAPI doesn't support JSON-LD. Which you can still use, but you have some manual (un)marshalling to do.

But since we don't use `ld+json` anywhere else, I don't see a point why we would be using it here. Otherwise we should try to return `ld+json` every where we return a VC/VP (and maybe DID documents as well)?

And since JSON-LD is just JSON (but the reverse might not be true), we can just return JSON content type?

woutslakhorst (approved):

PR #899, opened 2022-03-21
VCR: SearchVC() unmarshal query as VC instead of map

reinkrul:
Required to properly support variants of data types (single value v.s. list) of VCs.

### DIFF
--- a/vcr/api/v2/holder_test.go
+++ b/vcr/api/v2/holder_test.go
@@ -40,7 +40,7 @@ var organizationQuery = `
 {
 	"query": {
 		"@context": ["https://www.w3.org/2018/credentials/v1","https://nuts.nl/credentials/v1"],
-		"type": "NutsOrganizationCredential",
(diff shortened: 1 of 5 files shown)


vcr/api/v2/holder_test.go
```diff
@@ -40,7 +40,7 @@ var organizationQuery = `
 {
 	"query": {
 		"@context": ["https://www.w3.org/2018/credentials/v1","https://nuts.nl/credentials/v1"],
-		"type": ["VerifiableCredential", "NutsOrganizationCredential"],
+		"type": "NutsOrganizationCredential",
```
woutslakhorst, line 43: This would be an incorrect query. Without `VerifiableCredential` as type, the default VC fields will not be used in the query.

woutslakhorst (changes requested):

woutslakhorst (approved):

PR #900, opened 2022-03-21
VCR: Verify Nuts LD context is present for Nuts and set it by default

reinkrul:
As specified by OAS:

> The resolvable context of the credentialSubject as URI. If omitted, the "https://nuts.nl/credentials/v1" context is used. Note: it is not needed to provide the "https://www.w3.org/2018/credentials/v1" context here.

### DIFF
--- a/vcr/api/v2/api.go
+++ b/vcr/api/v2/api.go
@@ -21,6 +21,7 @@ package v2
 import (
 	"encoding/json"
 	"errors"
+	"github.com/nuts-foundation/nuts-node/vcr/credential"
 	"github.com/nuts-foundation/nuts-node/vcr/verifier"
 	"net/http"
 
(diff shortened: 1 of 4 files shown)


vcr/api/v2/api.go
```diff
@@ -96,6 +98,12 @@ func (w Wrapper) IssueVC(ctx echo.Context) error {
 		public = *issueRequest.Visibility == IssueVCRequestVisibilityPublic
 	}
 
+	// Set default context, if not set
+	if issueRequest.Context == nil {
```
woutslakhorst, line 102: `https://www.w3.org/2018/credentials/v1` must be added as well

woutslakhorst (changes requested):

vcr/api/v2/api.go
reinkrul: The default VC context is added by the issuer (also stated in the OAS) (see `vcr/issuer/issuer.go:109`).

reinkrul: 🤷‍♂️

PR #906, opened 2022-03-22
VCR: Allow searching on VC ID in v2 API

reinkrul:
In v1 we had a `Resolve(ID) VerifiableCredential` function, but in v2 we don't. But since IDs can be used by multiple issuers, its not a bad idea to yield a list when matching VCs by ID. So the v2 `SearchVC` should just be used.

This PR adds support to the v2 API to search on ID.

### DIFF
--- a/vcr/api/v2/holder.go
+++ b/vcr/api/v2/holder.go
@@ -63,6 +63,7 @@ func (w *Wrapper) SearchVCs(ctx echo.Context) error {
 func (w *Wrapper) searchOrgs(ctx echo.Context, allowUntrusted bool, body []byte) error {
 	// some helper structs
 	type helper struct {
+		ID                string `json:"id"`
 		CredentialSubject struct {
 			ID           string `json:"id"`
 			Organization struct {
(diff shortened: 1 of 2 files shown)


woutslakhorst (changes requested):
as stated by https://nuts-foundation.gitbook.io/drafts/rfc/rfc011-verifiable-credential#3.4-identifiers

IDs are unique, so adding the `resolve` would be a better fit. Searching will automatically be supported when moving to semantic search

reinkrul:
> as stated by https://nuts-foundation.gitbook.io/drafts/rfc/rfc011-verifiable-credential#3.4-identifiers
> 
> IDs are unique, so adding the `resolve` would be a better fit. Searching will automatically be supported when moving to semantic search

Why must they be unique? There's no way to guarantee other issuers don't issue the same ID?

I'd say the combination issuer + ID must be unique (actually just like in X.509)

reinkrul:
> > > as stated by https://nuts-foundation.gitbook.io/drafts/rfc/rfc011-verifiable-credential#3.4-identifiers
> > > IDs are unique, so adding the `resolve` would be a better fit. Searching will automatically be supported when moving to semantic search
> > 
> > 
> > Why must they be unique? There's no way to guarantee other issuers don't issue the same ID?
> > I'd say the combination issuer + ID must be unique (actually just like in X.509)
> 
> IDs are currently required to be DID+ # + some ID, that makes them unique

(just checked, the check on ID format is missing in the current code base)

When we support other VCs types, this requirement might not apply. So why not just make them non-unique? There's no APIs that will break.

woutslakhorst:
They must be unique for the oauth flow since they are referenced by ID in the access token. Also revocation depends on the IDs.

reinkrul:
I created a separate PR to add `Resolve(id)` to the v2 API: https://github.com/nuts-foundation/nuts-node/pull/913

reinkrul:
> If a resolve is added, is this still needed? It'll be removed for v2

It would still be nice to be able to search on ID? (even though it will probably be supported when we support actual JSON-LD  searching)

reinkrul:
Probably irrelevant, and will be succeeded by actual JSON-LD-backed storage.

PR #922, opened 2022-03-28
Network V2: track DAG state with trees

gerardsn:
this PR adds
- tracking of DAG state in xor & iblt trees.
- persistent and transactional storage for the trees
- missing copyright notices

I could use some feedback on what to test in `state` and the TODO in `bbolt_tree`

### DIFF
--- a/network/dag/bbolt_tree.go
+++ b/network/dag/bbolt_tree.go
@@ -1,38 +1,141 @@
+/*
+ * Copyright (C) 2022 Nuts community
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *
+ */
+
 package dag
 
 import (
+	"context"
 	"encoding/binary"
 	"fmt"
-	"github.com/nuts-foundation/nuts-node/network/dag/tree"
 	"go.etcd.io/bbolt"
+	"time"
+
+	"github.com/nuts-foundation/nuts-node/crypto/hash"
+	"github.com/nuts-foundation/nuts-node/network/dag/tree"
+	"github.com/nuts-foundation/nuts-node/network/log"
+	"github.com/nuts-foundation/nuts-node/network/storage"
 )
 
-const (
-	ibltBucketName = "treeIBLT"
-	xorBucketName  = "treeXOR"
-	// BucketFillPercent can be much higher than default 0.5 since the keys (unique node ids)
-	// should be added to the bucket in monotonic increasing order, and the values are of fixed size.
-	BucketFillPercent = 0.9
-)
+// treeBucketFillPercent can be much higher than default 0.5 since the keys (unique node ids)
+// should be added to the bucket in monotonic increasing order, and the values are of fixed size.
+const treeBucketFillPercent = 0.9
 
 type bboltTree struct {
-	db *bbolt.DB
+	db                *bbolt.DB
+	bucketFillPercent float64
+	bucketName        string
+	tree              tree.Tree
 }
 
-// newBBoltTreeStore returns an instance of a BBolt based tree store. Buckets managed by this store are filled to BucketFillPercent
-func newBBoltTreeStore(db *bbolt.DB) *bboltTree {
-	return &bboltTree{db: db}
+// newBBoltTreeStore returns an instance of a BBolt based tree store. Buckets managed by this store are filled to treeBucketFillPercent
+func newBBoltTreeStore(db *bbolt.DB, bucketName string, tree tree.Tree) *bboltTree {
+	return &bboltTree{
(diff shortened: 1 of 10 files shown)


network/dag/bbolt_tree.go
```diff
@@ -1,38 +1,143 @@
...
+		})
+		if err != nil {
+			log.Logger().Errorf("failed to add transaction to %s: %s", store.bucketName, err.Error())
+		}
+	}
+}
+
+// migrate builds a tree by walking over the dag and adding all Transaction references to the tree without checking for validity.
+// The tree is stored on disk once it is in sync with the dag.
+// Should only be called on an empty tree.
```
reinkrul, line 106: There's a check for that (and it's always called by Start)

network/dag/bbolt_tree.go
```diff
@@ -1,38 +1,143 @@
...
+// Returns an error the bucket does not exist, or if data in the bucket doesn't match the tree's Data prototype.
+func (store *bboltTree) read(ctx context.Context) error {
+	return storage.BBoltTXUpdate(ctx, store.db, func(_ context.Context, tx *bbolt.Tx) error {
 		// get bucket
-		bucket := tx.Bucket([]byte(bucketName))
+		bucket := tx.Bucket([]byte(store.bucketName))
 		if bucket == nil {
-			return fmt.Errorf("bucket '%s' not found", bucketName)
+			// should only happen once for a new tree/bucket.
+			log.Logger().Warnf("tree bucket '%s' does not exist", store.bucketName)
```
reinkrul, line 139: how is this log useful?

network/dag/tree/tree.go
```diff
@@ -14,21 +33,29 @@ type Data interface {
...
 type Tree interface {
 	// Insert a transaction reference at the specified clock value.
 	// The result of inserting the same ref multiple times is undefined.
-	Insert(ref hash.SHA256Hash, clock uint32) error
+	Insert(ref hash.SHA256Hash, clock uint32)
+	// InsertGetDirty inserts a transaction reference like Insert and also returns the dirty leaves like GetUpdates.
+	// This is an atomic operation to make sure no ResetUpdate can happen in between from calling logic.
+	InsertGetDirty(ref hash.SHA256Hash, clock uint32) map[uint32][]byte
+	// Delete a transaction reference without checking if ref is in the Tree
+	Delete(ref hash.SHA256Hash, clock uint32)
```
reinkrul, line 58: >without checking if ref is in the Tree

What happens if it's not present? If nothing happens, you can leave ouyt this comment imo

network/dag/tree/tree.go
```diff
@@ -101,29 +132,49 @@ func (t *tree) Load(leaves map[uint32][]byte) error {
 		if err != nil {
 			return err
 		}
-		err = t.updateOrCreatePath(split, func(n *node) error {
-			return n.data.Add(data)
+		t.updateOrCreatePath(split, func(n *node) {
+			n.data.Add(data)
```
reinkrul, line 136: error is ignored?

reinkrul (changes requested):
Is there some way the tests (preferably integration-level tests) can assert the correctness of the implementation? Say, do some edge case TXs and check whether all nodes have the same state (starting at the `xor` value of the tree)?

network/dag/bbolt_tree.go
gerardsn: the bucket should always exist, except during migration. Maybe change this back to return an error in V3?

network/dag/tree/tree.go
gerardsn: `Delete` puts -ref in the tree if ref was not in the tree. Deleting a ref from an Xor tree that is not in the tree has the same as effect as `Insert`ing it.

network/dag/tree/tree.go
gerardsn: No error can occur here since `n.data` and `data` are of the same concrete type.

changed to explicitly ignore error.

network/dag/bbolt_tree.go
```diff
@@ -1,38 +1,143 @@
...
+				<-c.Done()
+				err := c.Err()
+				if err == context.DeadlineExceeded {
+					log.Logger().Warnf("deadline exceeded - rollback transaction %s from %s", transaction.Ref(), store.bucketName)
+					store.tree.Delete(transaction.Ref(), transaction.Clock())
+				}
+			}()
+			tx.OnCommit(func() {
+				store.tree.ResetUpdate()
+				cancel()
```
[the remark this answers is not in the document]
gerardsn: The cancel() will Delete() the transaction from the tree and (eventually from) its storage, so the tree will not be corrupt. 

However, the tree will now be out of sync with the dag. Since the tree is used for set reconciliation, the node should request the 'missing' transaction from its peers, but the Tx will be dropped (not offered to the dag observers again) since it is already in the dag. 

It is even possible that both trees and the dag in `state` all have a different state.

Any suggestions on what to do with this?
- We could add metrics to log this?
- How likely is this to happen with the current timeout of 10 sec?

any thoughts on this @woutslakhorst

gerardsn:
> Is there some way the tests (preferably integration-level tests) can assert the correctness of the implementation? Say, do some edge case TXs and check whether all nodes have the same state (starting at the `xor` value of the tree)?

I'm happy to write some tests if you can help me get started.

gerardsn:
> > > Is there some way the tests (preferably integration-level tests) can assert the correctness of the implementation? Say, do some edge case TXs and check whether all nodes have the same state (starting at the `xor` value of the tree)?
> > 
> > 
> > I'm happy to write some tests if you can help me get started.
> 
> There's an integration test with paging where you can add a check on the diagnostics?

Added the test as suggested.

reinkrul (approved):

Issue #931, opened 2022-03-30
No way to test API error mappings with Open API spec


woutslakhorst:
The OAS files specify the expected return status codes. These are returned manually form methods in an API or by the `ResolveStatusCode` method on the API. Some of these errors come from way deep and we do not have a good method for testing/checking this.

reinkrul:
The only way is (currently) to build a system test: spin up the node and then run some tests against the HTTP interface.

PR #932, opened 2022-03-30
map error codes to status codes

### LINKED ISSUE #915: wrong status code for issuing private credential with missing NutsComm
```
{
    "detail": "unable to publish the issued credential: failed to resolve participating node (did=did:nuts:QmPDxwLnbys6PudTXkPLxjojjhyBqXFfKeySRMRBr7D): could not resolve NutsComm service owner: service not found in DID Document",
    "status": 500,
    "title": "IssueVC failed"
}
```

500 should be a 412? (precondition failed)

woutslakhorst:
fixes #915

The mappings from err to status code was missing. See also #931
Also fixed incorrect status code on `issueVC` and fixed a randomly failing integration test


reinkrul (approved):

stevenvegt (changes requested):
Add an extra tests to each test condition like this:
`assert.Equal(t, http.StatusNotFound, ctx.client.ResolveStatusCode(err))`

stevenvegt (approved):

PR #933, opened 2022-03-30
add revocation status to searchIssuedVcs

### LINKED ISSUE #930: Missing revocation status for searchIssuedVCs
a todo in vcr/v2/api.go

woutslakhorst:
closes #930

### DIFF
--- a/vcr/api/v2/api.go
+++ b/vcr/api/v2/api.go
@@ -21,9 +21,10 @@ package v2
 import (
 	"encoding/json"
 	"errors"
+	"net/http"
+
 	"github.com/nuts-foundation/nuts-node/vcr/credential"
 	"github.com/nuts-foundation/nuts-node/vcr/verifier"
-	"net/http"
 
 	"time"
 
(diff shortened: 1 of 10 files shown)


vcr/api/v2/api.go
```diff
@@ -160,14 +161,19 @@ func (w *Wrapper) SearchIssuedVCs(ctx echo.Context, params SearchIssuedVCsParams
...
-		result[i] = SearchVCResult{VerifiableCredential: resolvedVC}
-	}
 	if err != nil {
 		return err
 	}
+	result := make([]SearchVCResult, len(foundVCs))
+	for i, resolvedVC := range foundVCs {
+		var revocation *Revocation
+		revocation, err = w.VCR.Verifier().GetRevocation(*resolvedVC.ID)
+		if err != nil && !errors.Is(err, verifier.ErrNotFound) {
```
reinkrul, line 171: if a VC is revoked twice, GetRevocation returns an error, causing SearchResults to return an error (and no way to execute it without it erroring out). Is that OK? Should GetRevocation return a list, or the most recent (active) revocation instead in case there's multiple?

reinkrul (changes requested):

vcr/api/v2/api.go
stevenvegt: I don't like double revocations. I would say only the first one should be registred.

woutslakhorst: Although you can't revoke an already revoked VC (409 on API). This could still happen theoretically. It would be fine to return the first one.

woutslakhorst:
Added #937 

Changed behaviour of GetRevocation to not return an error when multiple exist. In a cluster setting this would be possible.

vcr/api/v2/api.go
reinkrul: This is actually interesting (out of scope for this PR); what if you revoke it with an incorrect revocation date (say, 5 years in future). Does that make the revocation effective after 5 years or immediate? Rationale: if you make a mistake during a revocation that causes the revocation not to be effective, can you issue another revocation that **is** effective?

reinkrul (approved):

vcr/api/v2/api.go
reinkrul: @woutslakhorst ?

PR #935, opened 2022-03-31
Network: only return error when node DID is invalid in strict mode

### LINKED ISSUE #934: Node exits during startup with unhelpful error message
After an update of my local Nuts node, the node exists during startup giving me the cryptic message: `DID document does not contain a keyAgreement key`.
This is not very helpful. Perhaps the node should not exit, of if it is really a deal breaker, explain how to fix it.

reinkrul:
Fixes #934 

Rationale: the linked issue occurs on nodes that auto-resolve the node DID (typically outdated nodes or nodes with outdated DID documents), which is only allowed in non-strict mode. In strict mode (aka production) node DID **must** be configured, and then the error should be blocking.

Letting the node start with auto-resolved node DID which is invalid allows operators in dev/test environments to fix the issue using the API/CLI.

### DIFF
--- a/network/network.go
+++ b/network/network.go
@@ -59,6 +59,7 @@ var defaultBBoltOptions = bbolt.DefaultOptions
 // Network implements Transactions interface and Engine functions.
 type Network struct {
 	config                 Config
+	strictMode             bool
 	lastTransactionTracker lastTransactionTracker
 	protocols              []transport.Protocol
 	connectionManager      transport.ConnectionManager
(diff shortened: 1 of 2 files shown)


woutslakhorst (commented):
The error for the mentioned issue would now be:

`invalid NodeDID configuration: DID document does not contain a keyAgreement key, register a keyAgreement key (did=%s)`

Most will not link this to a missing setting in the nuts.yaml

reinkrul:
I'd rather not point operators to specific configuration means (environment/CLI flags/file-based) since it's highly subjective to the environment the node runs in, and might change (and then the log message is invalid).

This should not happen in practice, since a keyAgreement key is registered when a new DID document is created. Maybe we should add a documentation section on important concepts and refer to the documentation (in general)?

woutslakhorst:
> This should not happen in practice, since a keyAgreement key is registered when a new DID document is created. Maybe we should add a documentation section on important concepts and refer to the documentation (in general)?

Yes true, only early adopters on development have issues with it now. Impact is minimum.

woutslakhorst (approved):

PR #947, opened 2022-04-05
Network test fix: assert on number of completed rollback TXs

reinkrul:
This way it's not influenced by timing or externally started goroutines

### DIFF
--- a/network/dag/bbolt_tree.go
+++ b/network/dag/bbolt_tree.go
@@ -50,6 +50,7 @@ type bboltTree struct {
 	bucketName             string
 	tree                   tree.Tree
 	activeRollbackRoutines *uint32
+	numRollbacks           *uint32
 }
 
 // newBBoltTreeStore returns an instance of a BBolt based tree store. Buckets managed by this store are filled to treeBucketFillPercent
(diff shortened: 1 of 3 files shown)


network/dag/bbolt_tree.go
```diff
@@ -60,6 +61,7 @@ func newBBoltTreeStore(db *bbolt.DB, bucketName string, tree tree.Tree) *bboltTr
 		bucketName:             bucketName,
 		tree:                   tree,
 		activeRollbackRoutines: new(uint32),
```
woutslakhorst, line 63: is this one still used?

gerardsn: https://github.com/nuts-foundation/nuts-node/blob/55fe9d4576f0d80f615a7a894b5d70c81b98992c/network/dag/bbolt_tree_test.go#L169-L171

it is still used here. But this is currently always zero and cannot be replaced by `store.numRollbacks`

gerardsn (changes requested):

network/dag/bbolt_tree.go
reinkrul: I intended to remove `activeRollbackRoutines`. Will fix.

reinkrul: I thought it wasn't required any more, but it still was. I put it back.

woutslakhorst (approved):

PR #948, opened 2022-04-05
replaced go-leia lib with V3

woutslakhorst:
this is just the lib replacement.

go-leia V3 longer allows aliases.

### DIFF
--- a/vcr/api/v1/api.go
+++ b/vcr/api/v1/api.go
@@ -20,10 +20,11 @@
 package v1
 
 import (
-	"github.com/labstack/echo/v4"
-	ssi "github.com/nuts-foundation/go-did"
 	"net/http"
 	"time"
+
+	"github.com/labstack/echo/v4"
+	ssi "github.com/nuts-foundation/go-did"
 
 	"github.com/nuts-foundation/nuts-node/core"
 	"github.com/nuts-foundation/nuts-node/vcr"
@@ -82,6 +83,22 @@ func (w *Wrapper) Search(ctx echo.Context, conceptName string, requestParams Sea
 	untrusted := false
 	if requestParams.Untrusted != nil {
 		untrusted = *requestParams.Untrusted
+	}
+
+	// for backwards compatibility
+	for k, v := range params {
+		switch k {
+		case "subject":
+			params[credential.CredentialSubjectPath] = v
+			delete(params, k)
+		case "organization.name":
+			params["credentialSubject.organization.name"] = v
+			delete(params, k)
+		case "organization.city":
+			params["credentialSubject.organization.city"] = v
+			delete(params, k)
+			// default change nothing
+		}
 	}
 
 	results, err := w.VCR.SearchConcept(ctx.Request().Context(), conceptName, untrusted, params)
--- a/vcr/issuer/leia_store.go
+++ b/vcr/issuer/leia_store.go
@@ -23,11 +23,13 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+
 	ssi "github.com/nuts-foundation/go-did"
 	"github.com/nuts-foundation/go-did/did"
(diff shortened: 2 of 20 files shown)


vcr/api/v1/api.go
```diff
@@ -84,6 +85,22 @@ func (w *Wrapper) Search(ctx echo.Context, conceptName string, requestParams Sea
 		untrusted = *requestParams.Untrusted
 	}
 
+	// for backwards compatibility
```
reinkrul, line 88:
```suggestion
	// for backwards compatibility: map old concept search parameters to JSON-LD IRI paths
```

vcr/issuer/leia_store.go
```diff
@@ -39,36 +41,35 @@ type leiaIssuerStore struct {
 
 // NewLeiaIssuerStore creates a new instance of leiaIssuerStore which implements the Store interface.
 func NewLeiaIssuerStore(dbPath string) (Store, error) {
-	store, err := leia.NewStore(dbPath, false)
+	store, err := leia.NewStore(dbPath)
 	if err != nil {
 		return nil, fmt.Errorf("failed to create leiaIssuerStore: %w", err)
 	}
-	collection := store.Collection("issuedCredentials")
+	collection := store.JSONCollection("issuedCredentials")
```
reinkrul, line 48: why don't we treat revocations as JSONLD?

vcr/issuer/leia_store.go
```diff
@@ -39,36 +41,35 @@ type leiaIssuerStore struct {
...
-	doc := leia.DocumentFromBytes(vcAsBytes)
-	return s.issuedCredentials.Add([]leia.Document{doc})
+	return s.issuedCredentials.Add([]leia.Document{vcAsBytes})
 }
 
 func (s leiaIssuerStore) SearchCredential(jsonLDContext ssi.URI, credentialType ssi.URI, issuer did.DID, subject *ssi.URI) ([]vc.VerifiableCredential, error) {
-	query := leia.New(leia.Eq("issuer", issuer.String())).
-		And(leia.Eq("type", credentialType.String())).
-		And(leia.Eq("@context", jsonLDContext.String()))
+	query := leia.New(leia.Eq(leia.NewJSONPath("issuer"), leia.MustParseScalar(issuer.String()))).
```
reinkrul, line 65: `MustParseScalar` will never fail here? (because they're strings?). Not wanting to cause a panic here at some point in future.

reinkrul (changes requested):
Note in general: we should probably make sure that part of the future JSON-LD search feature, we should return an error when a non-searchable (because non-indexed?) field is supplied. Otherwise, behavior of integrations could be unexpected (an EHR expecting organizations/records with certain properties to be returned, but actually never returned due to the field(s) not being indexed).

vcr/issuer/leia_store.go
woutslakhorst: it can only fail if the given argument is not a string, bool or float64

vcr/issuer/leia_store.go
woutslakhorst: This is just the lib update. To JSON-LD happens in the next PR. It has not been decided yet to migrate the other DBs as well. Will be in separate PRs if so

woutslakhorst:
> Note in general: we should probably make sure that part of the future JSON-LD search feature, we should return an error when a non-searchable (because non-indexed?) field is supplied. Otherwise, behavior of integrations could be unexpected (an EHR expecting organizations/records with certain properties to be returned, but actually never returned due to the field(s) not being indexed).

non-indexed fields can still be used. The search will use a full-table scan. Fields that are not in a context is another story.

reinkrul (approved):

reinkrul:
> 

Does that mean unmarshal every document and check if it matches? If yes, that sounds extremely slow (and maybe shouldn't be possible)?

PR #950, opened 2022-04-05
Network: v2 TransactionRangeQuery and response

reinkrul:


network/transport/v2/conversation.go
```diff
@@ -159,6 +159,19 @@ func (envelope *Envelope_TransactionListQuery) conversationID() []byte {
...
+func (envelope *Envelope_TransactionRangeQuery) setConversationID(cid conversationID) {
+	envelope.TransactionRangeQuery.ConversationID = cid.slice()
+}
+
+func (envelope *Envelope_TransactionRangeQuery) conversationID() []byte {
+	return envelope.TransactionRangeQuery.ConversationID
+}
+
+func (envelope *Envelope_TransactionRangeQuery) checkResponse(other isEnvelope_Message) error {
+	// TODO: Not sure what to do here
```
woutslakhorst, line 171: according to spec every TX received in the TransactionList must have a LC value in the requested range.

network/dag/bbolt_dag.go
```diff
@@ -274,6 +274,44 @@ func (dag *bboltDAG) FindBetween(ctx context.Context, startInclusive time.Time,
 	return result, err
 }
 
+func (dag *bboltDAG) findBetweenLC(ctx context.Context, startInclusive uint32, endExclusive uint32) ([]Transaction, error) {
```
woutslakhorst, line 277: overlapping code with `Walk`, can this be reduced?

network/dag/bbolt_dag.go
```diff
@@ -274,6 +274,44 @@ func (dag *bboltDAG) FindBetween(ctx context.Context, startInclusive time.Time,
...
+		}
+
+		// Initiate a cursor and start from the given lamport clock, collect the transactions until the LC upper bound is reached.
+		// This works because the clock index is sorted by the clock value.
+		cursor := clocks.Cursor()
+	outer:
+		for _, list := cursor.Seek(clockToBytes(startInclusive)); list != nil; _, list = cursor.Next() {
+			parsed := parseHashList(list)
+
+			for _, next := range parsed {
```
woutslakhorst, line 293: these need to be sorted as well on byteOrder (specced) as in `walk`

network/transport/v2/handlers.go
```diff
@@ -139,6 +142,31 @@ func (p *protocol) handleTransactionPayload(msg *TransactionPayload) error {
 	return p.payloadScheduler.Finished(ref)
 }
 
+func (p protocol) handleTransactionRangeQuery(peer transport.Peer, envelope *Envelope_TransactionRangeQuery) error {
+	msg := envelope.TransactionRangeQuery
+	cid := conversationID(msg.ConversationID)
+	if err := p.cMan.check(envelope); err != nil {
```
woutslakhorst, line 148: Only the response needs to be checked. All TXs in the response must have a LC value within the requested range.

network/transport/v2/handlers.go
```diff
@@ -139,6 +142,31 @@ func (p *protocol) handleTransactionPayload(msg *TransactionPayload) error {
...
+	}
+
+	ctx := context.Background()
+	txs, err := p.state.FindBetweenLC(ctx, envelope.TransactionRangeQuery.Start, envelope.TransactionRangeQuery.End)
+	if err != nil {
+		return err
+	}
+
+	transactionList, err := p.collectTransactionList(ctx, txs)
+	if err != nil {
```
woutslakhorst, line 163: The handleTransactionListQuery logs a warning here.

network/transport/v2/handlers_test.go
```diff
@@ -573,6 +572,32 @@ func TestProtocol_handleTransactionListQuery(t *testing.T) {
...
 }
 
+func TestProtocol_handleTransactionRangeQuery(t *testing.T) {
+	payload := []byte("Hello, World!")
+	tx1, _, _ := dag.CreateTestTransaction(0)
+	tx2, _, _ := dag.CreateTestTransaction(1, tx1)
+	lcStart := uint32(1)
+	lcEnd := uint32(5)
+
+	t.Run("ok", func(t *testing.T) {
```
woutslakhorst, line 582: missing error tests.

network/transport/v2/senders.go
```diff
@@ -107,6 +108,20 @@ func (p *protocol) sendTransactionList(id transport.PeerID, conversationID conve
 	return nil
 }
 
+func (p *protocol) sendTransactionRangeQuery(id transport.PeerID, conversationID conversationID, lcStart uint32, lcEnd uint32) error {
```
woutslakhorst, line 111: as with `sendTransactionListQuery`, this method is responsible for creating the conversationID.

woutslakhorst (changes requested):

network/transport/v2/handlers.go
reinkrul: But shouldn't TransactionRangeQuery only be sent after receiving a TransactionSet? Is that something we'd want to make explicit?

network/transport/v2/senders.go
reinkrul: Seems this isn't spec'd. I'd expect it to be part of an existing conversation, because it is send after receiving `TransactionSet`? So it should use that conversationID?

6.2.4 states:
> The message MUST contain a conversationID.

But is it a new conversation, or an existing (and, which?)


https://nuts-foundation.gitbook.io/drafts/rfc/rfc017-distributed-network-grpc-v2#6.2.4-transaction-range-query

network/dag/bbolt_dag.go
reinkrul: Couldn't find it in RFC017, can you point me to where this is spec'd?

reinkrul:
Should review/merge https://github.com/nuts-foundation/nuts-node/pull/965 first

network/transport/v2/handlers.go
woutslakhorst: The idea is that any tx received is a result of a request of the receiving node. How it came to that request is less relevant. The reason behind this is to limit unwanted transactions from the past (because this is less performant). If those weren't requested an evil node could push old txs to a node resulting in poor performance.

expanding conversations to be very strict would make it more complex while adding no extra security (I think)

network/transport/v2/senders.go
woutslakhorst: New, only requests for TX (+payload) start a conversation a `TransactionSet` does not.

gerardsn: This should probably be a new conversation. A standalone message allows faster synchronization of new nodes. Making this part of the set reconciliation loop (State / TransactionSet) would require a lot of messages to finally realize that the next step is to request the next page from a peer.

woutslakhorst: The idea for conversations is that they span only a single req/resp. So no deep conversations. A response may contain multiple messages of the same type.

network/transport/v2/handlers.go
reinkrul: Removed the log message there: I left it there because then the peer ID would get logged. But I think we need to log handler errors on a central spot including all metadata (peer ID and such): https://github.com/nuts-foundation/nuts-node/issues/973

woutslakhorst (approved):

PR #962, opened 2022-04-07
Feature/json ld supporting logic

woutslakhorst:
Basic logic for handling JSON-LD expanded form.

### DIFF
--- a/jsonld/document.go
+++ b/jsonld/document.go
@@ -49,20 +49,20 @@ func (p Path) Tail() Path {
 	return p[1:]
 }
 
-// Expanded represents a JSON-LD document in expanded form.
+// Document represents a JSON-LD document in expanded form.
 // The expanded form is ideal for traversing and finding values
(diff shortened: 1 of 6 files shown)


jsonld/document.go
```diff
@@ -0,0 +1,113 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *
+ */
+
+package jsonld
+
+// Path is used to traverse an expanded JSON-LD document
+// Its elements are IRIs
+type Path []string
```
reinkrul, line 24: I remember this from go-leia?

woutslakhorst: indeed, same stuff is required

reinkrul: At some point in future, there will be a library.

reinkrul (approved):

PR #963, opened 2022-04-07
Helm Fixes for GKE deployment

rolandgroen:
- Make PV and PVC controllable by values.yaml, GKE does not need a VPC. Config for PV and VPC are split up.
 - Added storageClassName to PV and PVC configuration, GKE needs it.
 - Deployment, added startupProbe as the downloading and writing of the IRMA schema on the PV cause container creation timeouts.

### DIFF
--- a/charts/nuts-node/templates/nuts-data-pv.yaml
+++ b/charts/nuts-node/templates/nuts-data-pv.yaml
@@ -14,7 +14,7 @@ spec:
     storage: {{ .Values.nuts.data.persistedVolume.capacity | default "100Mi" }}
   accessModes: {{ required "Please define `nuts.data.persistedVolume.accessModes` in `values.yaml`" .Values.nuts.data.persistedVolume.accessModes }}
   hostPath:
-    path: {{ .Values.nuts.data.persistedVolume.mountPath  }}
+    path: {{ .Values.nuts.data.persistedVolume.mountPath | default "/opt/nuts/data" }}
 ---
 {{ end }}
 {{ if .Values.nuts.data.persistedVolumeClaim.enabled }}
--- a/charts/nuts-node/templates/nuts-data-pv.yaml
+++ b/charts/nuts-node/templates/nuts-data-pv.yaml
@@ -1,3 +1,4 @@
+{{ if .Values.nuts.data.persistedVolume.enabled }}
 apiVersion: v1
 kind: PersistentVolume
 metadata:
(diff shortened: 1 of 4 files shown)


charts/nuts-node/templates/nuts-data-pv.yaml
```diff
@@ -8,22 +9,26 @@ metadata:
...
 apiVersion: v1
 kind: PersistentVolumeClaim
 metadata:
   name: nuts-data-pvc
   annotations:
     "helm.sh/resource-policy": keep
 spec:
-  storageClassName: "" # Empty string must be explicitly set otherwise default StorageClass will be set
-  accessModes: {{ required "Please define `nuts.data.persistedVolume.accessModes` in `values.yaml`" .Values.nuts.data.persistedVolume.accessModes }}
+  storageClassName: {{ .Values.nuts.data.persistedVolumeClaim.storageClassName | default "" }}
```
stevenvegt, line 28: If GKE needs a storage class name, can we provide a default or warn the user about a missing name?

stevenvegt:
@JorisHeadease Can you take a look at this?

charts/nuts-node/templates/nuts-data-pv.yaml
```diff
@@ -8,22 +9,26 @@ metadata:
...
 spec:
   persistentVolumeReclaimPolicy: Retain
-  storageClassName: ""
+  storageClassName: {{ .Values.nuts.data.persistedVolume.storageClassName | default "" }}
   capacity:
     storage: {{ .Values.nuts.data.persistedVolume.capacity | default "100Mi" }}
   accessModes: {{ required "Please define `nuts.data.persistedVolume.accessModes` in `values.yaml`" .Values.nuts.data.persistedVolume.accessModes }}
   hostPath:
-    path: {{ .Values.nuts.data.persistedVolume.mountPath | default "/opt/nuts/data" }}
+    path: {{ .Values.nuts.data.persistedVolume.mountPath  }}
```
JorisHeadease, line 17: Is there a specific reason why the default is removed?

woutslakhorst: @rolandgroen ?

stevenvegt:
ping @rolandgroen

charts/nuts-node/templates/nuts-data-pv.yaml
rolandgroen: For GKE empty ("") means the default storage class. Currently the default storage class causes problems, because the default is read only. I consider this to be specific to GKE and something the deployer will run into at an early stage. Also, GKE might change this behaviour in the future, or is there a possibility to change the default storage class that we have missed somehow.. 

In general, this PR is about the ability to change the storage class. Without this ability, it is impossible to deploy to GKE. A warning should only be in place if all platforms would suffer from this.

stevenvegt (approved):

PR #965, opened 2022-04-08
Network: remove Transaction.Hash from v2 protobuf definition

reinkrul:
This also introduces the need to re-parse the transactions, because now we work on the actual ref (hash) of the TX. Future implemented messages also need this (`TransactionRangeQuery`). Introduced `ParseTransactions` that cached the value during handling of the response.


network/transport/v2/conversation.go
```diff
@@ -194,3 +196,26 @@ func (envelope *Envelope_TransactionList) setConversationID(cid conversationID)
 func (envelope *Envelope_TransactionList) conversationID() []byte {
 	return envelope.TransactionList.ConversationID
 }
+
+// ParseTransactions parses the transactions from the message and caches them in the handlerData,
+// so subsequent calls to ParseTransactions will not parse the transactions again.
+func (envelope Envelope_TransactionList) ParseTransactions(data handlerData) ([]dag.Transaction, error) {
```
woutslakhorst, line 202: could be private?

network/transport/v2/conversation.go
```diff
@@ -32,6 +33,11 @@ import (
 
 var maxValidity = 30 * time.Second
 
+// handlerData contains contextual data produced and consumed by a message handler.
+// It is used to cache expensive data (e.g. parsed transactions) that are used multiple times within a handler,
+// but cannot be shared directly due to interface incompatibility.
+type handlerData map[interface{}]interface{}
```
woutslakhorst, line 39: why interface{} as key?

woutslakhorst (commented):
Just 2 questions

network/transport/v2/conversation.go
reinkrul: So handlers can define a key type that is completely private (struct type scoped to function), so it can't be altered (or at least makes it hard) outside of it. (see `parseTransactions`).

woutslakhorst (approved):

PR #966, opened 2022-04-08
Update v1.yaml

jelmerterwal:
problem with generated client missing 'capabillityDelegation'


reinkrul:
> problem with generated client missing 'capabillityDelegation'

Although it's technically missing when looking at the DID core spec, we don't ever use it. What problem did you encounter, exactly?

woutslakhorst:
@jelmerterwal is your code generator/parsing set to be very strict? eg: no additional fields may exist? If so it might be better to change that since DIDs/VCs may contain fields you don't know.

docs/_static/vdr/v1.yaml
```diff
@@ -272,6 +258,26 @@ components:
...
+          description: List of KIDs that may sign JWTs, JWSs and VCs
+          type: array
+          items:
+            type: string
+        authentication:
+          description: List of KIDs that may alter DID documents that they control
+          type: array
+          items:
+            type: string
+        capabilityDelegation:
```
woutslakhorst, line 271: We don't use this one, so why should it be added?

reinkrul: Because it can be returned (as @jelmerterwal found out). Maybe not when using the high-level APIs of VDR/DIDMan, but you can just update a DID Document, adding a `capabilityDelegation` (for whatever reason, maybe a Bolt?).
You actually might be able to add whatever to the DID Document (not specified in the DID core spec), but I think we should at least try to be complete when it comes to the basic properties of a DID document?

Else, we should ditch this spec altogether and point to either libraries or schemas people can use to parse/create DID documents, imo.

woutslakhorst: I'm trying to rule out strict parsers

woutslakhorst (approved):

docs/_static/vdr/v1.yaml
reinkrul: I agree: "be lenient about what you accept, and strict about what you return" (or something in those words).

PR #967, opened 2022-04-08
DIDMan: AddEndpoint OAS doesn't match returned type

jelmerterwal:
Wrong field name. When using generated code it crashes


woutslakhorst:
@jelmerterwal could you elaborate why `endpoint` can't be used as name?

reinkrul:
I checked the OAS v.s. the actual response data, what is spec'd is `{id,type,endpoint}` but what is returned is a `did.Service`, which is `{id,type,serviceEndpoint}`. Best solution is probably to adjust the OAS to reflect `did.Service` (as @jelmerterwal suggested). But then it's asymmetric with the request body (`endpoint v.s. serviceEndpoint`). Best would probably be to take a `did.Service` as request body by the time we release v2 of the DIDMan API.

reinkrul (approved):

PR #955, opened 2022-04-06
Adds configurable json-ld contexts

stevenvegt:
Closes #943

PR #971, opened 2022-04-11
added jsonld engine

woutslakhorst:
This PR adds the jsonld engine. That will be the place for all JSON-LD related logic. All under vcr/signature will have to be moved there. Waiting for #955 to be completed before that can be done.

Usage of the context manager in notary/didman/auth will be added later.


jsonld/transformer.go
```diff
@@ -0,0 +1,56 @@
...
+	"encoding/json"
+
+	"github.com/nuts-foundation/go-did/vc"
+	"github.com/piprate/json-gold/ld"
+)
+
+// JSONLdBase is the base URL used for IRIs that are unknown when expanding a JSON-LD document.
+const JSONLdBase = ""
+
+type transformer struct {
```
reinkrul, line 32: really feels like a helper type, is it required that it is a type itself or should it just be some util funcs?

jsonld/transformer.go
```diff
@@ -0,0 +1,56 @@
...
+
+import (
+	"encoding/json"
+
+	"github.com/nuts-foundation/go-did/vc"
+	"github.com/piprate/json-gold/ld"
+)
+
+// JSONLdBase is the base URL used for IRIs that are unknown when expanding a JSON-LD document.
+const JSONLdBase = ""
```
reinkrul, line 30: empty?

reinkrul (changes requested):

jsonld/transformer.go
woutslakhorst: I was going to abstract the contextLoader. But given the other PRs, the contextLoader is required at different levels. Replaced it with a simple Reader type.

jsonld/transformer.go
woutslakhorst: When expanding you need to pass a default IRI prefix. This const represents that default. Better to make a const than to have magic double quotes everywhere.

reinkrul (approved):

PR #972, opened 2022-04-11
Adding documents to JSON-LD store

woutslakhorst:
documents are added via ambassador.

- [ ] Organizations/Authorizations won't be indexed due to missing JSON-LD context.


vcr/assets/assets/default.index.yaml
```diff
@@ -0,0 +1,18 @@
+indices:
+  - name: index_id
+    parts:
+      - iriPath: []
```
reinkrul, line 4: why's this empty?

vcr/assets/assets/default.index.yaml
```diff
@@ -0,0 +1,18 @@
...
+  - name: index_issuer
+    parts:
+      - iriPath: ["https://www.w3.org/2018/credentials#issuer"]
+  - name: index_subject
+    parts:
+      - iriPath: ["https://www.w3.org/2018/credentials#credentialSubject"]
+  - name: index_organization
+    parts:
+      - iripath: ["https://www.w3.org/2018/credentials#credentialSubject", "http://schema.org/organization", "http://schema.org/legalname"]
```
reinkrul, line 13: inconsitent capitalization of `iripath`

vcr/assets/assets/default.index.yaml
```diff
@@ -0,0 +1,18 @@
...
+        tokenizer: whitespaceOrExact
+        transformer: cologne
+      - iripath: ["https://www.w3.org/2018/credentials#credentialSubject", "http://schema.org/organization", "http://schema.org/city"]
```
reinkrul, line 16: inconsitent capitalization of `iripath`

reinkrul (changes requested):

vcr/assets/assets/default.index.yaml
woutslakhorst: The root element has no path. In the root it looks for @id.

vcr/assets/assets/default.index.yaml
```diff
@@ -0,0 +1,18 @@
+indices:
+  - name: index_id
+    parts:
+      - iriPath: []
+  - name: index_issuer
+    parts:
+      - iriPath: ["https://www.w3.org/2018/credentials#issuer"]
+  - name: index_subject
```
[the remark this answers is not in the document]
woutslakhorst: For a given path it looks for @id/@value at that location. It will have more children but it will also have an @id

reinkrul (approved):

PR #983, opened 2022-04-13
VCR V2 API JSON-LD usage

woutslakhorst:
Also fixed the Nuts specific context for org/auth credentials.

### DIFF
--- a/vcr/api/v2/registry.go
+++ b/vcr/api/v2/registry.go
@@ -46,7 +46,13 @@ func (w *Wrapper) ResolveVC(ctx echo.Context, id string) error {
 // SearchVCs checks the context used in the JSON-LD query, based on the contents it maps to a non-JSON-LD query
 // After V1, this needs to be remapped to a DB search that supports native JSON-LD
 func (w *Wrapper) SearchVCs(ctx echo.Context) error {
-	var request SearchVCRequest
+	// use different struct for unmarshalling, we don't want default values for required params
+	type searchVCRequest struct {
+		Query         map[string]interface{} `json:"query"`
+		SearchOptions *SearchOptions         `json:"searchOptions,omitempty"`
+	}
+
+	var request searchVCRequest
 	err := ctx.Bind(&request)
 	if err != nil {
 		return core.InvalidInputError("failed to parse request body: %w", err)
@@ -57,7 +63,7 @@ func (w *Wrapper) SearchVCs(ctx echo.Context) error {
 		untrusted = *request.SearchOptions.AllowUntrustedIssuer
 	}
 
-	if len(request.Query.CredentialSubject) > 1 {
+	if credentials, ok := request.Query["credentialSubject"].([]interface{}); ok && len(credentials) > 1 {
 		return core.InvalidInputError("can't match on multiple VC subjects")
 	}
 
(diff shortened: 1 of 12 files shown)


vcr/api/v2/registry.go
```diff
@@ -61,110 +61,94 @@ func (w *Wrapper) SearchVCs(ctx echo.Context) error {
...
+		switch k {
+		case "@id":
+			fallthrough
+		case "@value":
+			results = append(results, vcr.SearchTerm{
+				IRIPath: currentPath,
+				Value:   v,
+			})
+		case "@list":
+			// TODO: not supported... This would require OR type queries
```
reinkrul, line 147: maybe return an error indicating the query isn't supported? Otherwise it might lead to the querier falsely assuming there are no VCs with the given parameters?

vcr/api/v2/registry.go
```diff
@@ -61,110 +61,94 @@ func (w *Wrapper) SearchVCs(ctx echo.Context) error {
...
+	for _, st := range searchTerms {
+		switch value := st.Value.(type) {
+		case string:
+			if value != "" && value != "0001-01-01T00:00:00Z" {
+				searchTerms[i] = st
+				i++
+			}
+		case float64:
+			if value != 0.0 {
+				searchTerms[i] = st
```
reinkrul, line 100: untested

vcr/api/v2/registry.go
```diff
@@ -61,110 +61,94 @@ func (w *Wrapper) SearchVCs(ctx echo.Context) error {
...
-			query.AddClause(concept.Prefix("credentialSubject.organization.city", subjectQuery.Organization.City))
-		}
-		if subjectQuery.ID != "" {
-			query.AddClause(concept.Eq("credentialSubject.id", subjectQuery.ID))
+
+	return ctx.JSON(http.StatusOK, results)
+}
+
+// filter default required values that are empty in the query
+func filter(searchTerms []vcr.SearchTerm) []vcr.SearchTerm {
```
reinkrul, line 89: bit of an oddball function. Maybe edges; but I can't search for credentials with a numeric value of `0` in it, or a string having the value `0001-01-01T00:00:00Z`? The latter is a bit far fetched, but indicates a typing problem imo (fix marshalling with checking `IsZero()` for dates?)

vcr/api/v2/registry.go
```diff
@@ -61,110 +61,94 @@ func (w *Wrapper) SearchVCs(ctx echo.Context) error {
...
 		}
+		// TODO bool problem, luckily there are no default bool fields in a VC
 	}
+	return searchTerms[:i]
+}
 
-	results, err := w.VCR.SearchLegacy(ctx.Request().Context(), query, allowUntrusted, nil)
-	if err != nil {
-		return err
+func flatten(document interface{}, currentPath []string) []vcr.SearchTerm {
```
reinkrul, line 109: flattening is also a JSON-LD operation, so shouldn't it be part of the JSON-LD engine? (and transform it to VCR SearchTerms here)

reinkrul (changes requested):

vcr/api/v2/registry.go
woutslakhorst: JSON-LD flatten is something different. The API needs to transform a JDON-LD document to a slice of `[]SearchTerm`, that logic is only used at this location.

vcr/api/v2/registry.go
woutslakhorst: is removed

vcr/api/v2/registry.go
woutslakhorst: actually added every item in the list as required match. This would still be according to spec (all must match)

vcr/api/v2/registry.go
woutslakhorst: fixed by not marshalling/unmarshalling through VC but through map[string]interface{}

reinkrul (approved):

PR #992, opened 2022-04-15
Network: fix peer ID being empty in connection observer due to race condition

reinkrul:


woutslakhorst (commented):
Since this is a bug fix, wouldn't it be better to prove it with a separate test. The altered test is already quite big.

woutslakhorst (approved):

PR #1001, opened 2022-04-19
Network: fix message overhead for chunked TransactionList

reinkrul:
Asserts the (hardcoded) overheads against the actual message size.


network/transport/v2/senders.go
```diff
@@ -133,10 +136,10 @@ func chunkTransactionList(transactions []*Transaction) [][]*Transaction {
 	endIndex := 0
 
 	// TODO to be tested in practise
```
woutslakhorst, line 138: TODO may be removed

woutslakhorst (approved):

PR #1003, opened 2022-04-19
Network: send state when missing prevs

### LINKED ISSUE #999: V2 protocol: send State message when missing prevs in List/Range message
When ErrPreviousTransactionMissing is returned when adding a transaction to the DAG, the node must send a state message.

gerardsn:
fixes #999

### DIFF
--- a/network/transport/v2/handlers.go
+++ b/network/transport/v2/handlers.go
@@ -248,7 +248,8 @@ func (p *protocol) handleTransactionList(peer transport.Peer, envelope *Envelope
 			if err != nil {
 				if errors.Is(err, dag.ErrPreviousTransactionMissing) {
 					p.cMan.done(cid)
-					xor, clock := p.state.XOR(context.Background(), math.MaxUint32)
+					log.Logger().Warnf("ignoring remainder of TransactionList due to missing prevs (conversation=%s, Tx with missing prevs=%s)", cid, tx.Ref())
+					xor, clock := p.state.XOR(ctx, math.MaxUint32)
 					return p.sender.sendState(peer.ID, xor, clock)
 				}
 				return fmt.Errorf("unable to add received transaction to DAG (tx=%s): %w", tx.Ref(), err)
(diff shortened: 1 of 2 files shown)


network/transport/v2/handlers.go
```diff
@@ -244,7 +244,13 @@ func (p *protocol) handleTransactionList(peer transport.Peer, envelope *Envelope
 		}
 		if !present {
 			// TODO does this always trigger fetching missing payloads? (through observer on DAG) Prolly not for v2
-			if err = p.state.Add(ctx, tx, envelope.TransactionList.Transactions[i].Payload); err != nil {
+			err = p.state.Add(ctx, tx, envelope.TransactionList.Transactions[i].Payload)
+			if err != nil {
+				if errors.Is(err, dag.ErrPreviousTransactionMissing) {
+					p.cMan.done(cid)
```
reinkrul, line 250: If the peer sent a chunked list, it causes errors. Do we expect this situation to occur often? Then the error message (TransacitonList for finished conversation) could be cumbersome.

network/transport/v2/handlers.go
```diff
@@ -244,7 +244,13 @@ func (p *protocol) handleTransactionList(peer transport.Peer, envelope *Envelope
...
+					xor, clock := p.state.XOR(context.Background(), math.MaxUint32)
```
reinkrul, line 251: context -> ctx?

reinkrul (changes requested):

network/transport/v2/handlers.go
gerardsn: Chunked lists should still send all messages in the correct order. A Tx with a missing prevs cannot be processed, and Txs received in later chucks are likely to depend on the Tx that is skipped here. This results in a chain of Txs that cannot be processed. 

I'd prefer to cancel the request at this point, but ignoring all future messages is the closest we have right now.

gerardsn: Added a logger to keep track of this

woutslakhorst (approved):

reinkrul (approved):

Issue #998, opened 2022-04-19
Auth: explicitly fail when drawing up a contract for a care organization with multiple organization VCs


reinkrul:
It's currently unfeasible to fix the issue with a care organization having multiple organization VCs (https://github.com/nuts-foundation/nuts-node/issues/987) due to time constraints (for v2 release). Also is the use case not 100% clear yet and it's much easier to implement when we introduce logical identifiers for care organizations (then the VC will have to be provided in the API call anyways).

For now, we will only block when a care organization has multiple (non-matching!) VCs, and return a helpful error message.

PR #1014, opened 2022-04-21
Auth: fail DrawUpContract when DID has multiple non-matching VCs

reinkrul:
Fixes https://github.com/nuts-foundation/nuts-node/issues/998


reinkrul:
Already on master.

woutslakhorst (approved):

PR #1016, opened 2022-04-21
moved State.IsPresent check to within State.Add for transactional context

### LINKED ISSUE #1015: TransactionList handler does not check duplicates within transactional context
The `IsPresent` is done outside of the bbolt transaction of `Add`. When syncing with 2 nodes simultaneous, this causes all TXs to be added twice. The DAG doesn't mind but the XOR and IBLT trees do...

woutslakhorst:
fixes #1015

### DIFF
--- a/network/network_integration_test.go
+++ b/network/network_integration_test.go
@@ -131,10 +131,10 @@ func TestNetworkIntegration_V2(t *testing.T) {
 
 		// start nodes with v1 disabled, we rely on the gossip protocol
 		bootstrap := startNode(t, "integration_bootstrap", testDirectory, func(cfg *Config) {
-			cfg.ProtocolV1.AdvertHashesInterval = 24 * 60 * 60 * 1000
+			cfg.Protocols = []int{2}
 		})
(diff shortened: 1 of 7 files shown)


network/network_integration_test.go
```diff
@@ -203,6 +204,56 @@ func TestNetworkIntegration_V2(t *testing.T) {
...
+
+	t.Run("Parallel node sync", func(t *testing.T) {
+		key := nutsCrypto.NewTestKey("key")
+		expectedDocLogSize := 0
+
+		testDirectory := io.TestDirectory(t)
+		resetIntegrationTest()
+
+		node1 := startNode(t, "integration_node1", testDirectory, func(cfg *Config) {
+			cfg.ProtocolV1.AdvertHashesInterval = 24 * 60 * 60 * 1000
```
reinkrul, line 216: you can (since recently) disable v1 using:

`cfg.Protocols = []int{2}` (just enables v2)

reinkrul (approved):
trivial improvement in test

PR #1027, opened 2022-04-22
Nats events for new transactions

### LINKED ISSUE #927: Publish Nats events for transactions
- TX+payload, to be used by VCR and VDR instead of replaying DAG. Published by DAG.
- TX only (private), to be used by private payload scheduler. Published by DAG

Needed for removing ReplayingPublisher in V3.

woutslakhorst:
Closes #927 

transactions are emitted for both public and private transactions.


network/dag/interface.go
```diff
@@ -133,16 +135,17 @@ type Visitor func(ctx context.Context, transaction Transaction) bool
 // PayloadStore defines the interface for types that store and read transaction payloads.
 type PayloadStore interface {
 	PayloadReader
-	PayloadWriter
-	// ReadManyPayloads allows the caller read many payloads in an optimized fashion.
+	// WritePayload writes contents for the specified payload, identified by the given hash.
+	WritePayload(ctx context.Context, payloadHash hash.SHA256Hash, data []byte) error
```
reinkrul, line 139: the inconsistency we now have with PayloadWriter.WritePayload reeks like bad design? PayloadStore composes from PayloadReader and not PayloadWriter? Maybe just pass in the TX here as well?

network/dag/publisher.go
```diff
@@ -62,6 +62,7 @@ func (s *replayingDAGPublisher) ConfigureCallbacks(state State) {
 		if payload != nil {
 			s.payloadWritten(ctx, transaction, payload)
 		}
+		return nil
```
reinkrul, line 65: would it make sense to now return the errors that can occur in payloadWritten (but are just logged)

network/dag/bbolt_tree.go
```diff
@@ -82,7 +82,7 @@ func (store *bboltTree) isEmpty() bool {
 
 // dagCallback inserts a transaction reference to the in-memory tree and to persistent storage.
 // The tree is not aware of previously seen transactions, so it should be transactional with updates to the dag.
-func (store *bboltTree) dagObserver(ctx context.Context, transaction Transaction, _ []byte) {
+func (store *bboltTree) dagObserver(ctx context.Context, transaction Transaction, _ []byte) error {
```
reinkrul, line 85: why this change?

reinkrul (changes requested):

network/dag/interface.go
woutslakhorst: The bad design imho was that the business logic and db impl used the same interface. A renaming of interfaces/method to make this clearer could be a solution.

network/dag/bbolt_tree.go
woutslakhorst: This is due to nats events being added as observer. The events will be the main mechanism of transfering TX content between engines. This is an operation that MUST succeed when a bbolt transaction commits. Therefore the event is published from within the bbolt transaction. If the event publishing fails, the transactino fails.

network/dag/interface.go
reinkrul: Probably, but we can also leave it to be refactored part of the storage layer?

reinkrul (approved):

PR #1043, opened 2022-04-26
Network: extend TransactionList message to detect when all chunks are received

gerardsn:


network/transport/v2/handlers.go
```diff
@@ -266,28 +263,10 @@ func (p *protocol) handleTransactionList(peer transport.Peer, envelope *Envelope
...
-		}
-		newRefs = newRefs[:i]
-		conversation.set("refs", newRefs)
+	}
 
-		// if len == 0, mark as done
-		if len(newRefs) == 0 {
-			p.cMan.done(cid)
-		}
+	if msg.MessageNumber == msg.TotalMessages {
```
reinkrul, line 268:
```suggestion
	if msg.MessageNumber >= msg.TotalMessages {
```

for added safety

network/transport/v2/handlers.go
```diff
@@ -236,12 +236,10 @@ func (p *protocol) handleTransactionList(peer transport.Peer, envelope *Envelope
...
 
 	// TODO convert to trace logging
-	log.Logger().Infof("handling handleTransactionList from peer (peer=%s, conversationID=%s)", peer.ID.String(), cid.String())
+	log.Logger().Infof("handling handleTransactionList from peer (peer=%s, conversationID=%s, message=%d/%d)", peer.ID.String(), cid.String(), msg.MessageNumber, msg.TotalMessages)
 
 	// check if response matches earlier request
-	var conversation *conversation
-	var err error
-	if conversation, err = p.cMan.check(envelope, data); err != nil {
+	if _, err := p.cMan.check(envelope, data); err != nil {
```
reinkrul, line 242: we could add an additional check in the checkResponse method for this message; `currentMsg.number == prevMsg.number + 1` (store message & total number in conversation)

reinkrul, line 242: we could add an additional check in the checkResponse method for this message; `currentMsg.number == prevMsg.number + 1` (store message & total number in conversation)

reinkrul (changes requested):

network/transport/v2/handlers.go
gerardsn: GRPC should deliver the messages in the correct order, so if the numbering is off this would indicate a mistake in another implementation. If this is by mistake there is no real harm, and if this is on purpose (repeatedly using the same low number) these checks don't really solve anything since there is no limit on `TotalMessages`

woutslakhorst: and it would trigger missing prevs and thus log an error indicating the other node is doing something wrong.

woutslakhorst (changes requested):
`properties` can be removed from the conversation struct. The `get` and `set` methods as well I think.

gerardsn:
> `properties` can be removed from the conversation struct. The `get` and `set` methods as well I think.

these are still used by the `TransactionListQuery` and subsequent `TransactionList` to validate that received transactions match requested transactions

woutslakhorst (approved):

reinkrul:
> `properties` can be removed from the conversation struct. The `get` and `set` methods as well I think.

I introduced `get` and `set` as syntactic sugar when I added a mutex protecting the conversation properties (we had a race condition there), so we don't forget to lock it when accessing properties.

reinkrul (approved):

Issue #861, opened 2022-03-07
Network: Prevent new conversations to start if one is already running


woutslakhorst:
This prevents overlapping conversations on a single connection. Overlapping conversations will mostly be duplicate TXs, so we have to prevent that. This also means each conversation needs to be marked as done when completed. TransactionList queries are already marked as done when each ref that has been requested is received.

woutslakhorst:
When doing an initial sync this creates an additional ~15 messages.

woutslakhorst:
done through #1046

PR #1046, opened 2022-04-28
Network: limit node to 1 active conversation per peer

gerardsn:
#861 aims to limit the number of conversations to 1, but this would be too strict. 

The `TransactionSet` for example can be followed by `TransactionListQuery` and `TransactionRangeQuery`. It should also be possible request newly Gossiped transactions from multiple peers at the same time. Limiting a node to 1 active range query means that it can only perform set reconciliation with 1 peer at a time.

### DIFF
--- a/network/transport/v2/conversation.go
+++ b/network/transport/v2/conversation.go
@@ -26,10 +26,10 @@ import (
 	"sync"
 	"time"
 
-	"github.com/nuts-foundation/nuts-node/network/dag"
-
 	"github.com/google/uuid"
 	"github.com/nuts-foundation/nuts-node/crypto/hash"
+	"github.com/nuts-foundation/nuts-node/network/dag"
+	"github.com/nuts-foundation/nuts-node/network/transport"
 )
 
 var maxValidity = 30 * time.Second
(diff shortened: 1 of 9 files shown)


woutslakhorst (commented):
A busy node could drop the request. Would it be better to limit it to 1 connection? This would reduce the risk of one node not answering at the cost of some duplicates. Duplicates come from nodes that have time to send them.

Also the TransactionListQuery should be limited as well, since that could also be a lot of transactions.

gerardsn:
> A busy node could drop the request. Would it be better to limit it to 1 connection? This would reduce the risk of one node not answering at the cost of some duplicates. Duplicates come from nodes that have time to send them.
> 
> Also the TransactionListQuery should be limited as well, since that could also be a lot of transactions.

Limiting this to a single conversation per peer probably works better, but is not possible if we follow the specs since a `TransactionSet` can be followed by:
- `State` (iblt decode unsuccessful)
- `TransactionListQuery` (decode provides missing tx)
- `TransactionRangeQuery` (decode provides no missing tx, but peer has more pages)
- `TransactionListQuery` and `TransactionRangeQuery` (decode provides missing tx, and peer has more pages)

All of these will start a new conversation while the `State` message leading up to the `TransactionSet` is still active. The biggest issue here is that the last option requires 2 new conversations. 

We can probably change the specs to remove the last option since that option shouldn't occur too often.
 (In that case the size of the IBLT or page size may have to change as well to make it impossible for the IBLT to decode an entire page.)

gerardsn:
> Hmmm, it's becoming quite complicated...

See [drawio](https://drive.google.com/file/d/1jAZHiaXBGAxeRsHgfpmbCNdeHpW9V3B5/view?usp=sharing) for summary of the issue. I will change the PR to 1 conversation per peer and make some minor changes to the flow in the protocol and see how that functions in the current network.

woutslakhorst (approved):

PR #1047, opened 2022-04-28
ASync handling of network message

### LINKED ISSUE #1028: Make GRPC handlers aSync
messages over the network are now handled synchronously. This means that the sender is waiting for the Ack while the receiver is processing the message. Longer processing means a longer (b)lock. The handler should accept the message, check if it can process it (message type) and then immediately return. Further (more intensive) processing should be done by a separate routine.

woutslakhorst:
closes #1028 

It also features a sink when handling transactionList messages. This may conflict with some MUST rules in the specs on specific answers. It might be better to change those to SHOULD with a reasoning in the appendix.


network/transport/v2/handlers.go
```diff
@@ -47,32 +48,66 @@ func (p protocol) Handle(peer transport.Peer, raw interface{}) error {
...
-		return p.handleTransactionRangeQuery(peer, msg)
+		return p.handleASynchWithLock(peer, envelope, p.handleTransactionRangeQuery)
 	case *Envelope_State:
-		return p.handleState(peer, msg.State)
+		// no DB interaction so no lock required
+		return handleASync(peer, envelope, p.handleState)
 	case *Envelope_TransactionSet:
-		return p.handleTransactionSet(peer, msg)
+		// no DB interaction so no lock required
+		return handleASync(peer, envelope, p.handleTransactionSet)
```
gerardsn, line 120: `p.state.XOR()` and `p.state.IBLT()` currently use a cursor to get the highest LC from the DB when the requested clock value is `math.MaxUint32`. State should probably keep track of the highest LC to reduce DB interactions.

gerardsn (changes requested):

network/transport/v2/protocol.go
```diff
@@ -103,6 +104,8 @@ type protocol struct {
 	cMan              *conversationManager
 	gManager          gossip.Manager
 	sender            messageSender
+	listHandler       *transactionListHandler
+	mLock             sync.RWMutex
```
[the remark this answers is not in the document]
woutslakhorst: I'll rename to `handlerMutex`

network/transport/v2/transactionlist_handler.go
```diff
@@ -0,0 +1,140 @@
...
+type transactionListHandler struct {
+	ctx context.Context
+	ch  chan peerEnvelope
+	fn  handleFunc
+}
+
+// newTransactionListHandler creates a new transactionListHandler.
+// The passed context is used to stop the go routine when cancelled.
+func newTransactionListHandler(ctx context.Context, fn handleFunc) *transactionListHandler {
+	ch := make(chan peerEnvelope, 100)
```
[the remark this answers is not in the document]
woutslakhorst: blocking, I'll fix with a select statement

network/transport/v2/handlers.go
```diff
@@ -47,32 +48,66 @@ func (p protocol) Handle(peer transport.Peer, raw interface{}) error {
...
-func (p protocol) handle(peer transport.Peer, envelope *Envelope) error {
-	switch msg := envelope.Message.(type) {
+type handleFunc func(peer transport.Peer, envelope *Envelope) error
+
+func (p *protocol) handleASynchWithLock(peer transport.Peer, envelope *Envelope, f handleFunc) error {
+	go func() {
+		if !p.mLock.TryRLock() {
+			log.Logger().Debugf("Sink active, ignoring %T (peer=%s)", envelope.Message, peer)
+			return
+		}
```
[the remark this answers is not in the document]
woutslakhorst: could, but the `Unlock` remains within the go routine. That places the lock and unlock at different places. If the go routine is not scheduled for processing, the lock would remain. This would increase the chance for a long running lock.

network/transport/v2/handlers.go
```diff
@@ -47,32 +48,66 @@ func (p protocol) Handle(peer transport.Peer, raw interface{}) error {
 	return nil
 }
 
-func (p protocol) handle(peer transport.Peer, envelope *Envelope) error {
-	switch msg := envelope.Message.(type) {
+type handleFunc func(peer transport.Peer, envelope *Envelope) error
+
+func (p *protocol) handleASynchWithLock(peer transport.Peer, envelope *Envelope, f handleFunc) error {
```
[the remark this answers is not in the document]
woutslakhorst: The problem this fix addresses is that a node can be too busy processing transactions to do anything else. The quickest way for this node to be active again is to complete its task. Doing two or three things at once will probably take longer in total.

If a node is busy processing, other nodes should probably look to other nodes for their new transactions. This would argue for not limiting the number of total conversations (maybe just per peer).

If a node is processing a transactionRangeQuery, it already has a RLock so the Lock has to wait until the lock has been removed.

network/transport/v2/handlers.go
woutslakhorst: Didn't know that, but yes that would be a very good optimization. Until then these will have to lock as well.

network/transport/v2/handlers.go
```diff
@@ -62,35 +64,69 @@ func (p protocol) Handle(peer transport.Peer, raw interface{}) error {
...
 	case *Envelope_State:
-		return p.handleState(peer, msg.State)
+		// no DB interaction so no lock required
+		return handleASync(peer, envelope, p.handleState)
 	case *Envelope_TransactionSet:
-		return p.handleTransactionSet(peer, msg)
+		// no DB interaction so no lock required
+		return handleASync(peer, envelope, p.handleTransactionSet)
 	case *Envelope_DiagnosticsBroadcast:
 		p.handleDiagnostics(peer.ID, msg.DiagnosticsBroadcast)
```
reinkrul, line 128: this can be handled async as well

network/transport/v2/transactionlist_handler.go
```diff
@@ -0,0 +1,125 @@
...
+	}()
+}
+
+func (p *protocol) handleTransactionList(peer transport.Peer, envelope *Envelope) error {
+	subEnvelope := envelope.Message.(*Envelope_TransactionList)
+	msg := envelope.GetTransactionList()
+	cid := conversationID(msg.ConversationID)
+	data := handlerData{}
+
+	log.Logger().Tracef("handling handleTransactionList from peer (peer=%s, conversationID=%s, message=%d/%d)", peer.ID.String(), cid.String(), msg.MessageNumber, msg.TotalMessages)
```
reinkrul, line 81:
```suggestion
	log.Logger().Tracef("Handling handleTransactionList from peer (peer=%s, conversationID=%s, message=%d/%d)", peer.ID, cid, msg.MessageNumber, msg.TotalMessages)
```

network/transport/v2/handlers.go
woutslakhorst: added #1055

PR #1049, opened 2022-04-29
VCR: Make VCR.SearchVCs() according to spec

reinkrul:


woutslakhorst (changes requested):
Update your oapi-codegen to v1.10.1

other than that 👍

woutslakhorst (approved):

PR #1052, opened 2022-04-29
Make boot logging (starting/stopping) more symmetric

reinkrul:
Makes it easier where the problem is when it freezes/crashes.


woutslakhorst (approved):

woutslakhorst:
This was actually by design. At info level you just want to know if a module started. At debug level you want to know when it starts, so you know which debug messages follow.

PR #1080, opened 2022-05-05
Fix didman organization search

stevenvegt:
The API endpoint `didman/v1/search/organizations?query=somename` did not return any results when it should. This PR fixes it by upgrading to a newer version of go-leia.


woutslakhorst (approved):

woutslakhorst:
Backport to V2

PR #1088, opened 2022-05-09
fixed incorrect observer for private TX

woutslakhorst:
private payloads also called the observer

also a small fix for the echo_integration_test that  kept showing up during race tests

### DIFF
--- a/network/dag/state.go
+++ b/network/dag/state.go
@@ -88,12 +88,10 @@ func NewState(dataDir string, verifiers ...Verifier) (State, error) {
 	newState.publisher = publisher
 
 	xorTree := newBBoltTreeStore(db, "xorBucket", tree.New(tree.NewXor(), PageSize))
-	newState.RegisterObserver(xorTree.dagObserver, true)
-	newState.xorTree = xorTree
-
 	ibltTree := newBBoltTreeStore(db, "ibltBucket", tree.New(tree.NewIblt(IbltNumBuckets), PageSize))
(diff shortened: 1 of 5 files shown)


network/network_integration_test.go
```diff
@@ -576,6 +576,46 @@ func TestNetworkIntegration_PrivateTransaction(t *testing.T) {
...
+
+		// Start 2 nodes: node1 and node2, node1 sends a private TX to node 2
+		node1 := startNode(t, "node1", testDirectory, func(cfg *Config) {
+			cfg.NodeDID = "did:nuts:node1"
+		})
+		node2 := startNode(t, "node2", testDirectory, func(cfg *Config) {
+			cfg.NodeDID = "did:nuts:node2"
+		})
+
+		// create 999 transactions on node 1
```
gerardsn, line 593: creates 9 tx?

network/dag/state.go
```diff
@@ -107,6 +105,21 @@ func (s *state) RegisterObserver(observer Observer, transactional bool) {
 
 }
 
+func (s *state) treeObserver(ctx context.Context, transaction Transaction, payload []byte) error {
```
gerardsn, line 108: should the be tested?

gerardsn (changes requested):

network/dag/state.go
woutslakhorst: added tests for non-happy flows. Happy flow is tested all over the place

gerardsn (approved):

PR #1089, opened 2022-05-09
Crypto: adjust ErrKeyNotFound to clearly indicate it concerns private keys

reinkrul:
Otherwise one could assume it's because a DID document is missing a verification method.


woutslakhorst (commented):
Is the correct error message passed to the API caller (guessing that is the use-case?)

reinkrul:
It is.

gerardsn (approved):

PR #1091, opened 2022-05-10
VDR: improve error messages returned when validating compound service references

reinkrul:


vdr/types/common.go
```diff
@@ -57,7 +58,19 @@ var ErrServiceNotFound = errors.New("service not found in DID Document")
...
+}
+
+// Error returns the error message.
+func (e ErrInvalidServiceQuery) Error() string {
+	return fmt.Sprintf("service query is invalid: %s", e.Cause)
+}
+
+// Is checks whether the other error is also a ErrInvalidServiceQuery
+func (e ErrInvalidServiceQuery) Is(other error) bool {
+	return fmt.Sprintf("%T", e) == fmt.Sprintf("%T", other)
```
woutslakhorst, line 72: So a wrapped ErrInvalidServiceQuery will not be equal

woutslakhorst (commented):
One comment on the errors.Is of the new error type

vdr/types/common.go
reinkrul: No, it isn't. Primary goal of this is to keep supporting the error to status code mapping in our APIs.

gerardsn (approved):

PR #1093, opened 2022-05-11
Network: Send State when peer is behind and there are no new transactions

gerardsn:
reflects latest protocol update, see https://github.com/nuts-foundation/nuts-specification/pull/189

If the Gossip does not result in the same XOR as its peer, the node now always makes an attempt at resolving their differences.

### DIFF
--- a/network/transport/v2/handlers.go
+++ b/network/transport/v2/handlers.go
@@ -248,21 +248,20 @@ func (p *protocol) handleGossip(peer transport.Peer, envelope *Envelope) error {
 	// TODO swap for trace logging
 	log.Logger().Infof("received %d new transaction references via Gossip", len(refs))
 
-	// send State if node is missing more refs than referenced in this Gossip
-	tempXor := xor.Xor(refs...)
-	if msg.LC >= clock && !tempXor.Equals(peerXor) {
-		// TODO swap for trace logging
-		log.Logger().Infof("xor is different from peer=%s and peers clock is equal or higher", peer.ID)
-		return p.sender.sendState(peer.ID, xor, clock)
-	}
-
 	// request missing refs
-	if len(refs) > 0 {
+	tempXor := xor.Xor(refs...)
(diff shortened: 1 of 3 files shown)


reinkrul:
Might be good to add an integration test for this? I suspect we could test this by starting one node which adds lots of TXs, then start a second node?

On the other hand, this really sounds like the "bootstrap/new node" scenario, which is probably already tested in the integration test?

network/transport/v2/handlers.go
```diff
@@ -248,21 +248,20 @@ func (p *protocol) handleGossip(peer transport.Peer, envelope *Envelope) error {
...
-	tempXor := xor.Xor(refs...)
-	if msg.LC >= clock && !tempXor.Equals(peerXor) {
-		// TODO swap for trace logging
-		log.Logger().Infof("xor is different from peer=%s and peers clock is equal or higher", peer.ID)
-		return p.sender.sendState(peer.ID, xor, clock)
-	}
-
 	// request missing refs
-	if len(refs) > 0 {
+	tempXor := xor.Xor(refs...)
```
reinkrul, line 252: Maybe some extra comments for understanding?
```suggestion
	// If our DAG is just missing the TXs from the gossip to get in sync with the peer's DAG, send TransactionListQuery.
	// Test this by XORing the TX refs from the gossip message with our DAG's XOR (should then equal peer DAG's XOR).
	tempXor := xor.Xor(refs...)
```

reinkrul (changes requested):

gerardsn:
> Might be good to add an integration test for this? I suspect we could test this by starting one node which adds lots of TXs, then start a second node?
> 
> On the other hand, this really sounds like the "bootstrap/new node" scenario, which is probably already tested in the integration test?

This specific change solves the case where a slow node is always behind and publishes a new transaction, but somehow the gossip didn't get through. In the old situation the peer would never be able to get its transaction out, but now it does. Added an integration test for this.

reinkrul (approved):

PR #1094, opened 2022-05-11
Network: Filter all gossiped Tx from peer's gossip queue

### LINKED ISSUE #1011: Filter gossip transactions when received from peer
If Alice adds A, B and C and gossips this to Bob, Bob adds those transactions to the gossip list for Alice. This is not needed since it's clear Alice already has them.

gerardsn:
fixes #1011

### DIFF
--- a/network/transport/v2/handlers.go
+++ b/network/transport/v2/handlers.go
@@ -222,11 +222,18 @@ func (p *protocol) handleGossip(peer transport.Peer, envelope *Envelope) error {
 		return nil
 	}
 
-	// check for unknown transactions
+	//
 	refs := make([]hash.SHA256Hash, len(msg.Transactions))
+	for i, bytes := range msg.Transactions {
+		refs[i] = hash.FromSlice(bytes)
+	}
+	if len(refs) > 0 {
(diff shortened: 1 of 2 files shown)


network/transport/v2/handlers.go
```diff
@@ -238,11 +245,8 @@ func (p *protocol) handleGossip(peer transport.Peer, envelope *Envelope) error {
 		}
 	}
 	refs = refs[:i]
-	if len(refs) > 0 {
-		// TODO swap for trace logging
-		log.Logger().Infof("received %d new transaction references via Gossip", len(refs))
-		p.gManager.GossipReceived(peer.ID, refs...)
-	}
+	// TODO swap for trace logging
+	log.Logger().Infof("received %d new transaction references via Gossip", len(refs))
```
reinkrul, line 249: fix TODO?

reinkrul (approved):

network/transport/v2/handlers.go
gerardsn: There are many handlers with the same TODO. We should all update them at the same time, but maybe wait until we are done tweaking the network protocol.

PR #1097, opened 2022-05-12
Remove v1 protocol

reinkrul:
Previously merged but then reverted because private TXs were broken (see integration test).


woutslakhorst:
What changed between then and now?

reinkrul:
> What changed between then and now?

Nothing. PEBCAP on my end, I think... Can't reproduce it any more.

woutslakhorst (approved):

PR #1098, opened 2022-05-12
Gossip: Reduce chance of test data race

gerardsn:
Several tests directly access a `peerQueue` without a lock. This can coincide with the Gossip manager trying to send Gossip messages and cause a data race.

Only 1 test actually depended on the gossip interval, so test duration is not increased.


reinkrul (approved):

reinkrul:
Wouldn't it be better if the test then also used the lock when acessing it?

Issue #1036, opened 2022-04-26
Storage: implement basic storage engine


reinkrul:
Define a storage engine that implements BBolt for the defined storage API.

Issue #1035, opened 2022-04-26
Storage: define interface for acquiring and using a KV store (simple version)


reinkrul:
First find out whether we build our own, or pick an off-the-shelf library that provides an API + adapters for different databases.

- Define interface for acquiring a KV store
- Define simple interface for reading and writing (no cursors/TXs yet).

To be used by the specific engines in the Nuts node.

PR #1099, opened 2022-05-12
Storage: define API and provide BBolt KV store

reinkrul:
Closes https://github.com/nuts-foundation/nuts-node/issues/1036
Closes https://github.com/nuts-foundation/nuts-node/issues/1035
Closes https://github.com/nuts-foundation/nuts-node/issues/1037

TODO:

- [x]  Fix copyrights after https://github.com/nuts-foundation/nuts-node/pull/1117 has been merged (rebase on master first)


echo/echo_integration_test.go
```diff
@@ -87,14 +87,16 @@ func startServer(t *testing.T) string {
 	go func() {
 		err := cmd.Execute(ctx, system)
 		if err != nil {
-			t.Fatal(err)
+			panic(err)
```
woutslakhorst, line 90: Do these changes fix the echo test failures?

storage/interface.go
```diff
@@ -0,0 +1,86 @@
+package storage
```
woutslakhorst, line 1: copyright

woutslakhorst (changes requested):
Buckets can't be translated 1-on-1 to all the key value stores we could support. Some KV stores would require a string to be prepended to the key to simulate a bucket. So maybe we have to change the name?

`Seek` on the cursor is only supported by bbolt (and similar KV stores). It won't be implementable by other implementations. (It could be, but very inefficient)

storage/engine.go
```diff
@@ -1,16 +1,61 @@
...
+	}
+	return nil
+}
+
+func (e *engine) Configure(config core.ServerConfig) error {
+	e.datadir = config.Datadir
+	return nil
+}
+
+func (e *engine) GetKVStore(engineKey string, storeName string) (KVStore, error) {
```
woutslakhorst, line 49: My guess is that this is not pluggable yet, but must be in the future. A TODO comment would give a heads up

vdr/integration_test.go
```diff
@@ -44,24 +43,28 @@ import (
...
+	var err error
 	// initialize logger, verbosity flag needs to be available
 	if lvl, err = log.ParseLevel(nutsConfig.Verbosity); err != nil {
 		return
 	}
 	log.SetLevel(lvl)
 	log.SetFormatter(&log.TextFormatter{ForceColors: true})
 
+	// Storage
+	warehouse := storage.New()
```
woutslakhorst, line 62: warehouse as var name is confusing..

storage/interface.go
```diff
@@ -0,0 +1,86 @@
...
+	// WriteBucket starts a writable transaction, retrieves the specified bucket and passes it to the given function.
+	// If the bucket does not exist, it will be created.
+	WriteBucket(bucketName string, fn func(BucketWriter) error) error
+	// ReadBucket starts a read-only transaction, retrieves the specified bucket and passes it to the given function.
+	// If the bucket does not exist, the function is not called.
+	ReadBucket(bucketName string, fn func(BucketReader) error) error
+}
+
+// WriteTransaction is used to write to a KVStore.
+type WriteTransaction interface {
```
woutslakhorst, line 43: `OnRollback` and `OnCommit` are required here

woutslakhorst (changes requested):

network/network.go
```diff
@@ -76,8 +75,9 @@ type Network struct {
 	decrypter              crypto.Decrypter
 	nodeDIDResolver        transport.NodeDIDResolver
 	didDocumentFinder      types.DocFinder
-	connectionsDB          *bbolt.DB
+	connectionStore        storage.KVStore
 	eventPublisher         events.Event
+	warehouse              storage.Warehouse
```
woutslakhorst, line 80: warehouse is a weird term, never encountered in IT before

reinkrul: Data Warehouse (not exactly the same though).

I thought: what holds/provides stores? -> Warehouses (to avoid having 1000 `*Manager` types)

echo/echo_integration_test.go
reinkrul: Doesn't fix it, but allows the error to be properly propagated to the test runner. Because members of `t` can't be called from gorotounes (but panic can)

storage/engine.go
reinkrul: It will become pluggable by providing multiple implementations and a specific store to be configured (in the storage engine) given `engineKey` and `storeName`. Or at least that's my vision for it.

I don't think we need a TODO here, it's obvious you'd be choosing the store implementation to return here, depending the engine's configuration.

storage/interface.go
reinkrul: I added them as AfterRollback and AfterCommit to make it clear it's really exactly that (and not before commit/rollback)

network/network.go
woutslakhorst: And a Factory doesn't creates stores. Builders do, but they are already reserved for a different kind of configuring....
`Provider` would be doable? `storage.Provider` would be the type then

storage/engine.go
woutslakhorst: Then `Engine` needs to be clarified a bit more. Is an engine something like Redis or based on the Nuts-go engine calling it?

network/network.go
reinkrul: I'll change it to `Provider`.

storage/engine.go
reinkrul: Latter one, so network module would pass in "network" and "connections". I'll rename it to something less specific and add some more godoc.

woutslakhorst:
Move to separate repo first?

storage/bbolt.go
```diff
@@ -101,11 +101,24 @@ func (b bboltStore) doTX(fn func(tx *bbolt.Tx) error, writable bool) error {
 		if err != nil {
```
woutslakhorst, line 101: this error should also trigger a afterRollback

storage/bbolt.go
```diff
@@ -101,11 +101,24 @@ func (b bboltStore) doTX(fn func(tx *bbolt.Tx) error, writable bool) error {
...
+			if ac, ok := opt.(*afterCommit); ok {
+				ac.fn()
+			}
+		}
 	} else {
 		log.Logger().Warnf("Rolling back transaction application due to error: %s", appError)
 		err := dbTX.Rollback()
 		if err != nil {
 			log.Logger().Errorf("Could not rollback BBolt transaction: %s", err)
+		} else {
```
woutslakhorst, line 115: should also be executed if rollback fails?

woutslakhorst (changes requested):

storage/bbolt.go
woutslakhorst: except when the error is `ErrTxClosed` or `ErrTxNotWritable`

reinkrul: why?

storage/bbolt.go
reinkrul: Well, question is what happens with the data when rollback fails. Since it's also not committed, the TX will be abandoned and DB state unchanged?

Feels like the `AfterRollback` should actually be `BeforeRollback`?

woutslakhorst: The hook is not for changing DB related data, but other data. Before or after rollback then doesn't really matter. In fact you can call it OnRollback :)

storage/bbolt.go
woutslakhorst: hmmm yes, it should always do a onRollback. The caller is expecting the TX to be committed, if this doesn't happen it should be notified.

storage/bbolt.go
woutslakhorst: For example, the in-memory IBLT/XOR tree have a change which needs to be rolled back when the write to disk fails. (By removing the tx or by reloading from disk)

network/network_integration_test.go
```diff
@@ -846,13 +847,17 @@ func startNode(t *testing.T, name string, testDirectory string, opts ...func(cfg
...
-	if err := eventPublisher.(core.Configurable).Configure(core.ServerConfig{Datadir: testDirectory}); err != nil {
+	serverConfig := core.ServerConfig{Datadir: path.Join(testDirectory, name)}
+	if err := eventPublisher.(core.Configurable).Configure(serverConfig); err != nil {
 		t.Fatal(err)
 	}
 	if err := eventPublisher.(core.Runnable).Start(); err != nil {
 		t.Fatal(err)
 	}
 
+	storeProvider := storage.New()
```
woutslakhorst, line 858: it might be wise to invest in a `storage.NewTestProvider(t *testing.T)`

storage/bbolt.go
reinkrul: Fixed in lib.

storage/bbolt.go
reinkrul: fixed in lib

woutslakhorst (commented):
~~Namespaces are redundant, we need something like storage `class`es that can be mapped to the interface. See #1169 ~~

storage/interface.go
```diff
@@ -0,0 +1,44 @@
...
+type Engine interface {
+	core.Engine
+	core.Configurable
+	core.Runnable
+	Provider
+}
+
+// Provider lets callers get access to stores.
+type Provider interface {
+	// GetIterableKVStore returns a key-value store like KVStore, but then iterable (supporting cursors).
```
woutslakhorst, line 36: We might have to assume not all implementations can iterate "in order".... So rewalking the DAG is actually reading values by bulk (since LC keys are calculated in advance).....

Is iterate still needed.

storage/engine.go
```diff
@@ -18,17 +18,83 @@
...
 }
+
+func (e engine) Start() error {
+	return nil
+}
+
+func (e engine) Shutdown() error {
+	failures := false
+	for storeName, store := range e.stores {
+		err := store.Close()
```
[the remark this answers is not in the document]
reinkrul: Added it to the library: https://github.com/nuts-foundation/go-stoabs/pull/1

network/network_integration_test.go
reinkrul: There actually is (`NewTestStorageEngine()`), I'll apply it

storage/test.go
```diff
@@ -0,0 +1,35 @@
...
+
+package storage
+
+import (
+	"github.com/nuts-foundation/go-stoabs"
+	"github.com/nuts-foundation/go-stoabs/bbolt"
+	"github.com/nuts-foundation/nuts-node/core"
+)
+
+func NewTestStorageEngine(testDirectory string) Engine {
```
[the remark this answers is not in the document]
reinkrul: `NewTestXYZ` now all take a testdirectory as param. I could change it here (and remove some shutdown code in other places), but the change would bubble up `NewTestNetworkEngine`, having to change all calls of both functions (and probably break a lot of tests in the meantime). Out of scope for now.

storage/interface.go
reinkrul: Waiting for https://github.com/nuts-foundation/go-stoabs/pull/3

storage/engine.go
```diff
@@ -18,17 +18,103 @@
...
+	store, err := p.getStore(p.moduleName, name, func(moduleName string, name string) (stoabs.Store, error) {
+		return bbolt.CreateBBoltStore(path.Join(p.engine.datadir, moduleName, name+".db"))
+	})
+	if store == nil {
+		return nil, err
+	}
+	return store.(stoabs.KVStore), err
+}
+
+func (p *provider) getStore(moduleName string, name string, creator func(namespace string, name string) (stoabs.Store, error)) (stoabs.Store, error) {
```
woutslakhorst, line 103: Probably not an issue when starting the node:
not thread safe, requesting the same store in parallel would cause one of the callers to lock

woutslakhorst (changes requested):
One minor thing, easy to fix

storage/engine.go
reinkrul: I considered this but came to the same conclusion. But indeed, an easy fix 👍

woutslakhorst (approved):

PR #1101, opened 2022-05-12
VCR: issuer tracks credentials it revoked and returns 409 on double revocation

### LINKED ISSUE #937: Revoking an already revoked VC does not return 409
It's specced in the V2 VCR API but there's no logic suggesting this is currently done.

gerardsn:
fixes #937

### DIFF
--- a/vcr/issuer/issuer.go
+++ b/vcr/issuer/issuer.go
@@ -22,14 +22,14 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"github.com/nuts-foundation/nuts-node/core"
-	"github.com/nuts-foundation/nuts-node/vcr/types"
 	"time"
 
 	"github.com/google/uuid"
+	
 	ssi "github.com/nuts-foundation/go-did"
 	"github.com/nuts-foundation/go-did/did"
 	"github.com/nuts-foundation/go-did/vc"
+	"github.com/nuts-foundation/nuts-node/core"
 	"github.com/nuts-foundation/nuts-node/crypto"
 	"github.com/nuts-foundation/nuts-node/jsonld"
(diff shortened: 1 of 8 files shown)


vcr/issuer/issuer.go
```diff
@@ -171,6 +181,11 @@ func (i issuer) Revoke(credentialID ssi.URI) (*credential.Revocation, error) {
 		return nil, fmt.Errorf("failed to publish revocation: %w", err)
 	}
 
+	// Store the revocation after it has been published
+	if err := i.store.StoreRevocation(*revocation); err != nil {
```
reinkrul, line 185: why isn't publishing it enough?

reinkrul (changes requested):

vcr/issuer/issuer.go
gerardsn: My understanding is that the issuer keeps track of the data (credentials and revocations) it has issued. A published revocation will end up in the verifier, but currently does not make its way to the issuer.

reinkrul: If you need to add it here, it was already broken?

But: the revocation should be stored when received through the network when published, for the local node as well. See `vcr/ambassador.go:92`

reinkrul (changes requested):

vcr/issuer/issuer.go
stevenvegt: Later, when we migrate to accumulator revocation, the information about which credential is revoked and when will be lost. So, since the issuer has it's own administration about issued credentials, I think we also should administer the relevant information about revocation, separate from the information received back from the network. Also, I don't like mixing the 3 roles, since we need a fourth place to integrate and we make them dependent on each other and more difficult to maintain and refactor.

reinkrul (approved):

PR #1109, opened 2022-05-16
How-to release Nuts Node

reinkrul:


docs/pages/development/2-releasing.rst
```diff
@@ -0,0 +1,37 @@
+.. _releasing-nuts-node:
+
+Releasing Nuts Node
+###################
+
+Nuts Node and auxiliary tools/applications follow a semantic versioning scheme (``<major>.<minor>.<patch>``).
+With every regular release the major version is incremented. The patch version is incremented for releases that contain bugfixes.
```
woutslakhorst, line 7: not true, we have a lot of major releases now because we are removing stuff. The minor release should be the primary choice

woutslakhorst: So we follow a normal semantic versioning scheme

reinkrul: > not true, we have a lot of major releases now because we are removing stuff. The minor release should be the primary choice

I thought we'd release a new major release every month now, which increments the major version. On second thought, that's true, because they generally contain breaking changes. So these months we'll probably keep releasing majors, but after v3 it'll be 3.1, 3.2, 3.3?

woutslakhorst: correct

docs/pages/development/2-releasing.rst
```diff
@@ -3,9 +3,14 @@
 Releasing Nuts Node
 ###################
 
-Nuts Node and auxiliary tools/applications follow a semantic versioning scheme (``<major>.<minor>.<patch>``).
-With every regular release the major version is incremented. The patch version is incremented for releases that contain bugfixes.
-The minor version generally isn't used.
+Nuts Node and auxiliary tools/applications follow a semantic versioning scheme (``<major>.<minor>.<patch>``):
+
+    Given a version number MAJOR.MINOR.PATCH, increment the:
+    1. MAJOR version when you make incompatible API changes,
```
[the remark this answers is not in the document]
reinkrul: Added a note.

PR #1112, opened 2022-05-16
send notification from DAG with AND without payload

### LINKED ISSUE #1116: Created private transactions are not added to xor/iblt
Because the observer didn't react to private TX with payload!

woutslakhorst:
fixes #1116 

it basically simulates 2 separate notifications: 1 for the TX and 1 for the payload. But when receiving the payload notification you'll want the TX as well.

Now the notifications are equal again for creating and receiving nodes of private TXs

backport to V2

### DIFF
--- a/network/dag/state.go
+++ b/network/dag/state.go
@@ -48,16 +48,18 @@ const (
 
 // State has references to the DAG and the payload store.
 type state struct {
-	db                        *bbolt.DB
-	graph                     *bboltDAG
-	payloadStore              PayloadStore
-	transactionalObservers    []Observer
-	nonTransactionalObservers []Observer
-	keyResolver               types.KeyResolver
-	publisher                 Publisher
-	txVerifiers               []Verifier
-	xorTree                   *bboltTree
-	ibltTree                  *bboltTree
+	db                               *bbolt.DB
+	graph                            *bboltDAG
+	payloadStore                     PayloadStore
+	transactionalObservers           []Observer
+	nonTransactionalObservers        []Observer
+	transactionalPayloadObservers    []PayloadObserver
+	nonTransactionalPayloadObservers []PayloadObserver
+	keyResolver                      types.KeyResolver
+	publisher                        Publisher
+	txVerifiers                      []Verifier
+	xorTree                          *bboltTree
+	ibltTree                         *bboltTree
 }
 
 // NewState returns a new State. The State is used as entry point, it's methods will start transactions and will notify observers from within those transactions.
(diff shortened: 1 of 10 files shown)


reinkrul:
End-to-end test not happy:

```
{\"error\":\"invalid_request\",\"error_description\":\"jwt bearer token validation failed: unable to find the DID document\"}\n","status":400,"title":"RequestAccessToken failed"}
```

network/dag/state_test.go
```diff
@@ -340,7 +342,7 @@ func TestState_treeObserver(t *testing.T) {
 	}
 	ctx := context.Background()
 
-	t.Run("no callback for public transaction without payload", func(t *testing.T) {
+	t.Run("callback for public transaction without payload", func(t *testing.T) {
```
reinkrul, line 345: Why is this allowed? Since v1 public TXs must always have payload?

network/dag/state.go
```diff
@@ -146,7 +144,17 @@ func (s *state) Add(ctx context.Context, transaction Transaction, payload []byte
 			return err
 		}
 
-		return s.notifyObservers(contextWithTX, transaction, payload)
+		// The double notification call makes sure that the node that created the TX gets the same amount of
+		// notifications as a node that received the TX from the network.
+		// Give a private transaction, both the receiver and creator get 2 notifications: 1 with payload and 1 without.
+		// The creator gets both from here while the receiver gets the notification with payload via WritePayload.
+		// For public transactions, all nodes get 2 notifications: 1 with and 1 without a payload.
+		if payload != nil {
```
reinkrul, line 152: given the comments above, the order is off? Because now it's first added with payload, and then without payload?

Notifying twice here is quite obscure because it depends on specific behavior of the observers. If we can't express what we want in the current observer setup, I'd rather refactor it instead. Seems like you'd actually want to emit the events here, rather than from the publisher?

reinkrul (changes requested):

network/dag/state_test.go
woutslakhorst: The behaviour changes so the test changed.

gerardsn (approved):
LGTM

network/dag/state.go
```diff
@@ -91,33 +93,32 @@ func NewState(dataDir string, verifiers ...Verifier) (State, error) {
...
-	newState.RegisterObserver(newState.treeObserver, true)
+	newState.RegisterTransactionObserver(newState.treeObserver, true)
 
 	return newState, nil
 }
 
-func (s *state) RegisterObserver(observer Observer, transactional bool) {
+func (s *state) RegisterTransactionObserver(observer Observer, transactional bool) {
 	if transactional {
 		s.transactionalObservers = append(s.transactionalObservers, observer)
```
reinkrul, line 103: should've been named transactionalTransactionObservers, but ok :-p

reinkrul (approved):

PR #1123, opened 2022-05-17
Deduplicate oapi ssi schema definitions

### LINKED ISSUE #970: Deduplicate OAS VC definition
is added to vcr and auth specs. Should be in a shard file.

gerardsn:
fixes #970


reinkrul:
Although I admire the effort, I think we shouldn't change the existing APIs at this moment because we promised them to be stable (unless we had to fix a bug).

gerardsn:
Some issues I encountered while trying to fix this issue:

This PR makes the schemas for VC and VP consistent with w3c specs, and also consistent between engines. This changes the API, but I think all changes to the schema properties make their types less strict than before.

Schemas are now defined with examples included for most properties. The examples seem engine specific or maybe even bolt specific. If we move to common definitions for schemas, the common examples probably need to be minimal and can be overloaded with context specific examples as long as these examples are defined according to the original schema.

https://nuts-node.readthedocs.io/en/iss-970-deduplicate-oapi-vc-definition/pages/api.html?urls.primaryName=Verifiable%20Credential%20Registry%20(v2)

woutslakhorst:
> Although I admire the effort, I think we shouldn't change the existing APIs at this moment because we promised them to be stable (unless we had to fix a bug).

The current situation could also be regarded as broken...

woutslakhorst:
I'll drop a question in Slack to assess the impact.

woutslakhorst:
Nobody confirmed they were using generated code for the V2 API.

woutslakhorst (approved):

PR #1131, opened 2022-05-19
added bbolt TX as param to some bboltDAG methods

woutslakhorst:
The `State` starts the transaction, bboltDAG reuses this transaction.

This is now done for:
- GetTransaction/Get
- IsPresent
- WritePayload

The goal is to remove the tx from the context. This is required for the DB abstraction layer

### DIFF
--- a/network/dag/bbolt_dag.go
+++ b/network/dag/bbolt_dag.go
@@ -128,14 +128,8 @@ func (dag *bboltDAG) Diagnostics() []core.DiagnosticResult {
 	return result
 }
 
-func (dag bboltDAG) Get(ctx context.Context, ref hash.SHA256Hash) (Transaction, error) {
-	var result Transaction
-	var err error
-	err = storage.BBoltTXView(ctx, dag.db, func(_ context.Context, tx *bbolt.Tx) error {
-		result, err = getTransaction(ref, tx)
-		return err
-	})
-	return result, err
+func (dag bboltDAG) Get(tx *bbolt.Tx, ref hash.SHA256Hash) (Transaction, error) {
+	return getTransaction(ref, tx)
 }
 
 func (dag bboltDAG) GetByPayloadHash(ctx context.Context, payloadHash hash.SHA256Hash) ([]Transaction, error) {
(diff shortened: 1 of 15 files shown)


gerardsn (approved):
Makes sense to me, but I can't oversee the impact of the changes. Relying a lot on the tests for the approval.

network/dag/bbolt_dag.go
```diff
@@ -114,31 +107,25 @@ func newBBoltDAG(db *bbolt.DB) *bboltDAG {
...
 
-func (dag bboltDAG) Get(ctx context.Context, ref hash.SHA256Hash) (Transaction, error) {
-	var result Transaction
-	var err error
-	err = storage.BBoltTXView(ctx, dag.db, func(_ context.Context, tx *bbolt.Tx) error {
-		result, err = getTransaction(ref, tx)
-		return err
-	})
-	return result, err
+func (dag bboltDAG) get(tx *bbolt.Tx, ref hash.SHA256Hash) (Transaction, error) {
```
reinkrul, line 124: can be removed

network/dag/bbolt_payloadstore.go
```diff
@@ -38,44 +35,27 @@ type bboltPayloadStore struct {
...
-		bucket := tx.Bucket([]byte(payloadsBucketName))
-		if bucket == nil {
-			return nil
-		}
-		data := bucket.Get(payloadHash.Slice())
-		result = len(data) > 0
-		return nil
-	})
-	return result, err
+func (store bboltPayloadStore) IsPayloadPresent(tx *bbolt.Tx, payloadHash hash.SHA256Hash) bool {
```
reinkrul, line 38: should probably be non-exported?

network/dag/state.go
```diff
@@ -284,32 +312,38 @@ func (s *state) Start() error {
...
+				return err
+			}
+		}
+		return nil
+	})
 }
 
 func (s *state) Walk(ctx context.Context, visitor Visitor, startAt hash.SHA256Hash) error {
-	return s.graph.Walk(ctx, visitor, startAt)
+	return storage.BBoltTXUpdate(ctx, s.db, func(contextWithTX context.Context, tx *bbolt.Tx) error {
```
reinkrul, line 342: This should be a read TX? Otherwise we always lock the db.

network/dag/state.go
```diff
@@ -139,60 +139,88 @@ func (s *state) Add(ctx context.Context, transaction Transaction, payload []byte
...
+func (s *state) writePayload(tx *bbolt.Tx, transaction Transaction, payloadHash hash.SHA256Hash, data []byte) error {
+	err := s.payloadStore.WritePayload(tx, payloadHash, data)
+	if err == nil {
+		// ctx passed with bbolt transaction
+		return s.notifyPayloadObservers(tx, transaction, data)
+	}
+	return err
+}
+
+func (s *state) ReadPayload(ctx context.Context, hash hash.SHA256Hash) (payload []byte, err error) {
```
reinkrul, line 218: you changed WritePayload, why not this one as well?

reinkrul (changes requested):
Not sure what to think of this at this state: feels like it's just halfway done. Can't we eliminate TX passing through context passing in one go?

(and then maybe test it by altering the implementation of BBoltTXView and BBoltTXUpdate so it crashes if there's already a TX in the context; at least for the testing)

woutslakhorst:
> Not sure what to think of this at this state: feels like it's just halfway done. Can't we eliminate TX passing through context passing in one go?

The PR would get considerably bigger due to the problematic dag.Publisher. I did an analysis on the problematic methods (called with and without transaction):

Screenshot 2022-05-23 at 09 07 36

The first step is to split/alter the methods in red. For some methods this can be done by splitting it in a public method and a private method where the public one starts the TX and the private one only reuses it.

network/dag/state.go
woutslakhorst: As can be seen in the diagram, ReadPayload always starts a transaction.

network/dag/state.go
woutslakhorst: Is called by migrate which does an update. I'll change migrate so it starts a transaction and calls the private walk.

network/dag/bbolt_payloadstore.go
woutslakhorst: Imade everything in the store private

network/dag/state.go
woutslakhorst: actually not needed, so changed to read tx

network/dag/bbolt_dag.go
```diff
@@ -114,31 +107,25 @@ func newBBoltDAG(db *bbolt.DB) *bboltDAG {
...
+	return getTransaction(ref, tx)
 }
```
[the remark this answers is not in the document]
woutslakhorst: get has been removed

reinkrul (approved):

PR #1136, opened 2022-05-23
Network: debug logging in case of authentication error

reinkrul:
Aids analysis.


woutslakhorst (approved):
Maybe even `trace` level

PR #1142, opened 2022-05-23
Network: add branch/tag to peer diagnostics

reinkrul:


network/network.go
```diff
@@ -571,7 +571,7 @@ func (n *Network) collectDiagnostics() transport.Diagnostics {
 	result := transport.Diagnostics{
 		Uptime:               time.Now().Sub(n.startTime.Load().(time.Time)),
 		NumberOfTransactions: uint32(n.state.Statistics(context.Background()).NumberOfTransactions),
-		SoftwareVersion:      core.GitCommit,
+		SoftwareVersion:      fmt.Sprintf("%s (%s)", core.GitBranch, core.GitCommit),
```
gerardsn, line 574: Should this be `core.GitVersion()`? This uses the branch if no version is available.

There seems to be some overlap with the output of `core.system.Diagnostics()`. Should these be made consistent?

reinkrul: > Should this be `core.GitVersion()`? This uses the branch if no version is available.

I added this because currently we only share the Git commit ID. Which is informative enough, but it's more pleasant to the eye to immediately see the branch when viewing peer information. Because for releases, it actually shows the tag (e.g. v2.1.0). Then you don't have to reverse-lookup the actual release from the commit ID.

> There seems to be some overlap with the output of `core.system.Diagnostics()`. Should these be made consistent?

Yes there is overlap, but the peer diagnostics is spec'd by the RFCs since it's a gRPC message. I think the diagnostics page should contain as much information in a parsable format (therefore the different fields for branch/git commit ID). The peer diagnostics gRPC message tries to contain "just enough" information, to avoid making it ambiguous or too complex for implementations.

reinkrul: I think the current solution is correct: the "version" property indicates the "lineage" of the current version, e.g. v1/v2/v3 (or more specific patch versions). If it is not tagged, and thus doesn't have a specific version, it's lineage is determined by the branch (e.g. some feature branch, or v2 branch, or master).

gerardsn: I assumed that this is what you were going for to give it a more readable name

gerardsn (approved):

PR #1180, opened 2022-06-08
change dbsize stats

woutslakhorst:
According to the docs when making a backup `tx.Size()` is the way to get the exact number of bytes.


reinkrul:
We should probably add this to `go-stoabs` as well, but maybe optionally supported (you probably don't want to support it for other data stores than local files due to computationally cost and/or permissions it would require).

reinkrul (approved):

Issue #1169, opened 2022-06-07
Storage config to DB selection


woutslakhorst:
In general, storage abstraction allows for organizations to run a Nuts node with their preferred storage. To enable this a config setup is needed that is somehow used when instantiating a kind of storage in code. We identified different types of storage:

- Volatile: storage that may be removed and nothing really bad happens. eg: db connection state
- Bulk: storage that is needed by a node but can be built from network state. You don't want to rebuild it since it will cost a lot of time. eg: DAG, public payloads
- Critical: storage that can't be rebuild and needs to be backed-up, eg: private payloads, issued credentials
- Index: storage that allows for searching, can be rebuild from Bulk. Data in this type of storage will typically reside in-memory (for faster searching). You don't want to rebuild it since it will cost a lot of time. eg: VCR, VDR.

The configuration file should be simple to use, eg: "this is my redis storage". It should feature sane defaults but still allow for a particular setup.

```yaml
storage:
  adapters:
    volatile:
      type: bbolt
      config:
        datadir: data
    default:
      type: redis
      config:
        key: value
        key2: value
    private:
      type: redis
      config:
        key: value
        key2: value
    index:
      type: elastic
      config:
        key: value
        key2: value

```

`default` will be used for the DAG (and payloads and private if not configured separately). No index means a default leia on bbolt store in the same as `volatile`. `volatile` by default will be bbolt on disk.

Engines in nuts-core know which type of storage they need, they can request it from the storage engine:

```
storage.GetVolatile(...)
storage.GetDefault(...)
// etc
```

So we have different storage `classes`. For each class multiple instances (identified by `name`) can exist. It's up to the adapter to translate the name to something else (buckets, tables, logical DBs, etc).

woutslakhorst:
For a future version:

The current indexing lib (go-leia) is hooked to bbolt on disk which doesn't allow for multiple readers. It would be best to create an interface for searching and provide multiple implementations. Some research is needed to see if goleia can be transformed so it uses a generic query language compatible with other indexing capable DBs

reinkrul:
[...]

PR #1189, opened 2022-06-10
Storage: basic storage configuration layer

reinkrul:
Required for implementing backups and Redis support.

Backup configuration for BBolt is there for demonstration purposes (of how to configure databases), will be actually implemented later on.

Fixes https://github.com/nuts-foundation/nuts-node/issues/1169

TODO:
- [x] Test coverage
- [x] Documentation (write it when we have alternative backup strategies and/or Redis support).


reinkrul:
Ready for first round of comments.

woutslakhorst (commented):
My initial thoughts on config and DB setup was that any specific config for a particular DB is handled by stoabs. The concerns for how a DB functions lie with stoabs, then the Nuts node only have to select which DB is used. If a new type of DB is supported by stoabs this should be available with a lib update and shouldn't require any code changes.

reinkrul:
> My initial thoughts on config and DB setup was that any specific config for a particular DB is handled by stoabs. The concerns for how a DB functions lie with stoabs, then the Nuts node only have to select which DB is used. If a new type of DB is supported by stoabs this should be available with a lib update and shouldn't require any code changes.

Not 100% sure which part you mean. Config or also operational features, such as backupping?

Depends, my initial thought was to keep `go-stoabs` as simple as possible, so just an API to provide access to multiple databases. Mapping database options (e.g. connection strings, file names, etc) to configuration (be it a map or an actual Go struct) _could_ be a concern of `go-stoabs`, but for the sake of simplicity of the library I left it in `nuts-node`. Otherwise part of the Nuts Node config would be injected into `go-stoabs` which is certainly possible, but would complicate the Nuts Node documentation (then we should refer to the `go-stoabs` documentation on how to configure it, making our docs more complicated).

On the same note, scheduling a backup for BBolt is a Nuts Node concern as well: it's not something you can abstract in `go-stoabs`'s API, since whether/how to backup is highly dependent on the underlying database.

reinkrul:
@woutslakhorst  Tried to move database-specific config to `go-stoabs`, but I think the added complexity is not worth it (especially at this moment in time).

woutslakhorst:
Too much to handle in a single PR. Let's make this part about supporting the example config (last example in issue). I think there will be many more PRs to get things the way we want it to be.

storage/engine.go
```diff
@@ -83,6 +94,40 @@ func (e engine) Shutdown() error {
...
+		if _, configured := dbConfig["path"]; !configured {
+			dbConfig["path"] = path.Join(config.Datadir, moduleName, storeName+".db")
+		}
+	}
+
+	// Register databases
+	for _, curr := range e.config.Databases {
+		if e.isDatabaseRegistered(curr.Type) {
+			// TODO: Will be supported in future
+			return fmt.Errorf("multiple databases configured of type '%s' (which is not supported)", curr.Type)
```
woutslakhorst, line 110: incorrect error?

woutslakhorst (changes requested):
And waiting for tests/successful build and a rebase

reinkrul:
Requires https://github.com/nuts-foundation/go-stoabs/pull/9 to be merged first (and then update library).

reinkrul:
@woutslakhorst I reworked the configuration. It's much simpler now, but more hardcoded. It reflects the configuration example as proposed by https://github.com/nuts-foundation/nuts-node/issues/1169 with one (important) difference: you can't configure a specific database (e.g. BBolt or Redis) multiple times. For the sake of simplicity I just assume data is not going to be in multiple Redis servers at this moment.

I don't think we need true "dynamic" support for any database supported by `go-stoabs` (in other words: when we update the library, the database is automagically supported in the Nuts Node): in the forseeable future we only support 2 databases, and having some config in the Nuts Node for each specific database makes configuration (for the end user) simpler imo (good auto-generated documentation, configuration in line with the rest of the node). Striving for that level of integration while being dynamic isn't worth the added complexity, I think.

And there will be database-specific features (e.g. backups) and metadata (storage class) that will still be database specific, implemented in the Nuts Node regardless of where they're configured.

woutslakhorst (commented):
1st a suggestion on the backup feature

storage/bbolt.go
```diff
@@ -0,0 +1,182 @@
...
+	// Backup specifies backup config for the database.
+	Backup BBoltBackupConfig `koanf:"backup"`
+}
+
+// BBoltBackupConfig specifies config for BBolt database backups.
+type BBoltBackupConfig struct {
+	// Directory specifies the directory in which the BBolt backup should be written.
+	Directory string `koanf:"directory"`
+	// Interval specifies the time between backups.
+	Interval time.Duration `koanf:"interval"`
```
woutslakhorst, line 56: How is koanf suppose to coerce "1h" to a time.Duration?

storage/bbolt.go
```diff
@@ -0,0 +1,182 @@
...
+	databasePath := path.Join(b.datadir, b.getRelativeStorePath(moduleName, storeName))
+	store, err := bbolt.CreateBBoltStore(databasePath)
+	if store != nil {
+		b.startBackup(moduleName, storeName, store)
+	}
+	return store, err
+}
+
+func (b bboltDatabase) getClass() Class {
+	return VolatileStorageClass
```
woutslakhorst, line 84: If it has a backup running, then it's not volatile anymore?

storage/bbolt.go
```diff
@@ -0,0 +1,182 @@
...
+	log.Logger().Debugf("Starting BBolt database backup (store: %s/%s), target: %s", moduleName, storeName, backupFilePath)
+	startTime := time.Now()
+	wipFilePath := backupFilePath + ".work"
+	previousFilePath := backupFilePath + ".previous"
+
+	// To avoid corrupting the backup in case of a crash, the backup procedure looks as follows:
+	// Write backup to "store.db.work"
+	// Rename existing backup ("store.db") to "store.db.previous"
+	// Rename "store.db.work" to "store.db"
+	// Remove work file
```
woutslakhorst, line 125: You just renamed it

storage/engine.go
```diff
@@ -83,6 +89,14 @@ func (e engine) Shutdown() error {
 
 func (e *engine) Configure(config core.ServerConfig) error {
 	e.datadir = config.Datadir
+
+	data, _ := json.MarshalIndent(e.config, "  ", "  ")
+	println(string(data))
```
woutslakhorst, line 94: println

woutslakhorst (changes requested):

storage/bbolt.go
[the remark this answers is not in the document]
reinkrul: Made an issue for this: https://github.com/nuts-foundation/nuts-node/issues/1223

reinkrul: At first I thought that too and implemented a custom type, but appears it is supported out of the box...

storage/bbolt.go
reinkrul: Probably? But if you configure a backup for BBolt, but configure a Redis server as well I'd say we'd give precedence to Redis for storing non-volatile data?

I don't feel like implementing it at this moment, because we don't support per-store database binding (e.g. DAG in Redis, network connections in BBolt) and backup is currently all or nothing (so having the separate classes won't change anything).

Change when we implemented Redis.

woutslakhorst (changes requested):
It looks like the storage engine now operates on a single BBolt DB instead of the backwards compatible files that used to back each store? This would require a migration for all current data!

reinkrul:
> It looks like the storage engine now operates on a single BBolt DB instead of the backwards compatible files that used to back each store? This would require a migration for all current data!

Nope, it should use the same files as before (`$datadir/$engine/$store.db`). See `bboltDatabase.createStore`.

woutslakhorst (approved):

Issue #1032, opened 2022-04-26
VCR: Assert that all fields in a VC are defined by the context


reinkrul:
When issuing a VC, we must validate that all fields are defined by the JSON-LD context(s). Otherwise the integrator could falsely assume the extra fields are protected by the signature (which they are not). Or when integrations contains bugs (incorrect field names) it could cause vulnerabilities (e.g. `usercontext` instead of `userContext` for a `NutsAuthorizationCredential`, leading to a user identity not being required when exchanging medical data).

reinkrul:
RFC014 specifies `legalBase`, but this isn't included in the context;

```
  "legalBase": {
    "consentType": "explicit",
    "evidence": {
      "path": "pdf/f2aeec97-fc0d-42bf-8ca7-0548192d4231",
      "type": "application/pdf"
    }
```

We can't add this to the context, because that will break existing VCs on the dev and stable network.

woutslakhorst:
A new context will be created including the legalBase

PR #1190, opened 2022-06-10
VCR: Validate that VCs only contain fields that are defined by JSON-LD context

reinkrul:
Fixes https://github.com/nuts-foundation/nuts-node/issues/1032

TODO:

- [x] Find out why proof is (almost) empty after compaction
- [x] Check whether we want VCs received through the network fail on these same rules (or should it only be a sanity check from our API). Answer: no; VCs should be stored, but only fields from context returned/indexed.

### DIFF
--- a/docs/pages/release_notes.rst
+++ b/docs/pages/release_notes.rst
@@ -13,8 +13,21 @@ Release date: 2022-10-14
 
 - HTTPS TLS offloading is now also possible at the Nuts node. Checkout the docs on TLS offloading for the details.
     By default this is turned off which corresponds to the current behaviour.
+- Issuing a Verifiable Credential will now fail when it includes a property not defined in its JSON-LD context(s).
+  The behavior was changed because undefined fields are not secured by the JSON-LD proof,
+  which allows an attacker to alter it while the developer assumes it is secured by the signature.
+  It also helps developers noticing they misspelled a property, which it previously accepted but may have caused issues at processing systems downstream.
 
 **Full Changelog**: https://github.com/nuts-foundation/nuts-node/compare/v4.0.0...v5.0.0
+
+========
+Breaking changes
+========
+
+While Nuts RFC014 (Authorization Credential) required ``legalBase`` to be present in all ``NutsAuthorizationCredential``s,
+this property was missing in the Nuts v1 JSON-LD context. When issuing Verifiable Credentials, now all fields must be defined in its context(s).
+This means, starting this version, the ``legalBase`` property can't used in new v1 ``NutsAuthorizationCredential``s.
+It has been added to the future v2 version of the credential, in which it can be used again.
 
 *****************
 Chestnut update (v4.2.4)
--- a/docs/pages/release_notes.rst
+++ b/docs/pages/release_notes.rst
@@ -27,7 +27,8 @@ Breaking changes
 While Nuts RFC014 (Authorization Credential) required ``legalBase`` to be present in all ``NutsAuthorizationCredential``s,
 this property was missing in the Nuts v1 JSON-LD context. When issuing Verifiable Credentials, now all fields must be defined in its context(s).
 This means, starting this version, the ``legalBase`` property can't used in new v1 ``NutsAuthorizationCredential``s.
-It has been added to the future v2 version of the credential, in which it can be used again.
+Until the next major version (v6.0.0) it should be removed.
+Then when v6.0.0 is released, upgrade to the v2 JSON-LD context version and re-add the ``legalBase``.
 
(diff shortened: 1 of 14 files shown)


vcr/credential/validator.go
```diff
@@ -78,28 +85,46 @@ func Validate(credential vc.VerifiableCredential) error {
...
-		return err
+// validateAllFieldsKnown verifies that all fields in the VC are specified by the JSON-LD context.
+func (d defaultCredentialValidator) validateAllFieldsKnown(input vc.VerifiableCredential) error {
+	// First expand, then compact and marshal to JSON, then compare
+	inputAsJSON, _ := input.MarshalJSON()
+	inputAsMap := make(map[string]interface{})
+	_ = json.Unmarshal(inputAsJSON, &inputAsMap)
+
+	processor := ld.NewJsonLdProcessor()
+	options := ld.NewJsonLdOptions("") // TODO: why?
```
woutslakhorst, line 99: doesn't matter for your own comparison. The given string acts as "root" context name.

woutslakhorst (changes requested):

reinkrul:
Merge for v5, add to release notes that integrations can't pass legalBase.

docs/pages/release_notes.rst
```diff
@@ -4,6 +4,28 @@ Release notes
 
 What has been changed, and how to update between versions.
 
+*****************
+Hazelnut (v5.0.0)
```
woutslakhorst, line 8: Coconut

docs/pages/release_notes.rst
```diff
@@ -4,6 +4,28 @@ Release notes
...
+  It also helps developers noticing they misspelled a property, which it previously accepted but may have caused issues at processing systems downstream.
+
+========
+Breaking changes
+========
+
+While Nuts RFC014 (Authorization Credential) required ``legalBase`` to be present in all ``NutsAuthorizationCredential``s,
+this property was missing in the Nuts v1 JSON-LD context. When issuing Verifiable Credentials, now all fields must be defined in its context(s).
+This means, starting this version, the ``legalBase`` property can't used in new v1 ``NutsAuthorizationCredential``s.
+It has been added to the future v2 version of the credential, in which it can be used again.
```
woutslakhorst, line 27: Advice would be to remove the prop until V6. Then use V2 context.

docs/pages/release_notes.rst
```diff
@@ -4,6 +4,28 @@ Release notes
...
+
 *****************
 Chestnut update (v4.2.2)
```
woutslakhorst, line 30: No 4.2.3 and 4.2.4? Not backported, rebase master

vcr/credential/validator_test.go
```diff
@@ -30,7 +35,9 @@ import (
 )
 
 func TestNutsOrganizationCredentialValidator_Validate(t *testing.T) {
```
woutslakhorst, line 37: I would have expected that those new funcs to have their own tests. Keeps the test small.

woutslakhorst (changes requested):

reinkrul:
Feature becomes more complicated; compaction causes map with only an "ID" field to be "unmapped", so given:

```
{ credentialSubject: { id: "1234" } }
```

After compaction, becomes:

```
{ credentialSubject: "1234" }
```

woutslakhorst:
what about using expanded form instead of compacted form?

reinkrul:
> what about using expanded form instead of compacted form?

Can you describe how it would work? When we take VCs as input or marshal them, they're in compacted form. So the validation now works as follows:

```
inputJSON = toJSON(inputVC)
compactedJSON = compact(inputJSON) // removes fields not in JSON-LD context
normalize(inputJSON)
normalize(compactedJSON)
assert inputJSON == compactedJSON
```

reinkrul:
Assessed impact; VCs are validated when:

- Issuing VC
- Creating JWT grant token (thus, access tokens can't be requested for invalid NutsAuthCred)
- When verifying VCs, which is when:
  - Searching VCs (invalid NutsAuthCreds can't be found through search anymore)
  - Resolving VCs, used when introspecting access tokens (already issued access tokens can't be used any more)
- When verifying VPs, which is when:
  - VPs from the Auth module are validated, e.g. checking an IRMA signature in an access token

Since parties have only issued NutsAuthorizationCredentials for testing purposes, I don't think the actual usage of now considered invalid VCs will be a problem. That they don't appear in the search, however, might cause confusion. I'll add a warning in the log when the VC search encounters invalid VCs. Which can then be removed later again, in a subsequent version.

PR #1197, opened 2022-06-16
Fix config merge

### LINKED ISSUE #1072: Fix config merge issues
When a config options is a list
And there are several (3) default values
And you want to override these with a new list with fewer items (2)
Then only the first few items (2) are overwritten, but the rest of the default values (the last one) are still there.

Example is the IRMA contractvalidators.

stevenvegt:
Fixes #1072 

For the ServerConfig:
Default values are defined in the pflags
Defaults are loaded from pflags
Followed by: 
* config file (default nuts.yaml)
* environment variables
* command line arguments
Server does not have PersistentFlags, so rewritten to regular Flags.

The config for the Client does not longer load the config file since it only uses the command line arguments and env vars.

Note that this PR changes behavior: For the server command, when a non-default value for `configfile` is provided and the file is not found, the node throws an error during startup.


reinkrul:
e1b6eec9bf66cd00b470935da98fb6672e5dbaa16fa88de36a03dc3a06bb7750

core/client_config.go
```diff
@@ -40,22 +40,24 @@ type ClientConfig struct {
...
 }
 
-// DefaultClientConfig creates a new CLI client config with default values set.
-func DefaultClientConfig() ClientConfig {
-	return ClientConfig{
-		Address:   defaultAddress,
-		Verbosity: defaultLogLevel,
-		Timeout:   defaultClientTimeout,
+// NewClientConfigForCommand loads the config map into a command
+func NewClientConfigForCommand(cmd *cobra.Command) (ClientConfig, error) {
```
woutslakhorst, line 44: What about a panic/fatal instead of the errors? Since the caller will always be a client command it can just exit right away when the config is incorrect. Saves a lot of test code as well.

woutslakhorst (commented):
one main question: use of a panic/fatal for the client config loading. Less lines of code.

core/config.go
```diff
@@ -49,14 +72,40 @@ func loadConfigIntoStruct(flags *pflag.FlagSet, target interface{}, configMap *k
...
+// loadFromFlagSet loads the config values set in the command line options into the configMap.
+func loadFromFlagSet(configMap *koanf.Koanf, flags *pflag.FlagSet) error {
+	return configMap.Load(posflag.Provider(flags, defaultDelimiter, configMap), nil)
+}
+
+// LoadConfigMap loads all the values for a given command into the provided configMap.
+// It loads the default values and then (when defined) overwrites them respectively with values from
+// the config file, environment variables and command line flags.
+func LoadConfigMap(configMap *koanf.Koanf, cmd *cobra.Command) error {
+	flags := cmd.Flags()
```
reinkrul, line 93: do you know what the difference is between persistent flags and non-persistent flags? (otherwise you could just supply the actual flags here, instead of the complete command)

reinkrul (changes requested):
Much more setup code now in the client commands now?

core/client_config.go
stevenvegt: I don't like the stacktraces which appear after a panic. It feels like not handling an error and just returning a 500 because is saves a lot of code.

core/config.go
stevenvegt: Yes, fixed it: server does not have any persistent flags, only the client commands share a common set. Since client commands do not load the config file, the loading code for server and client config is different now. So you no longer need to provide the command for loading the server config, only the flags.

reinkrul (approved):
run `make update-docs`, rest OK.

PR #1198, opened 2022-06-17
refactor state.Subscriber

woutslakhorst:
There's now one subscriber that can be configured to be transactional (at least 1 invocation of callback) or not (at most 1 callback). With DB support and other customizations.


woutslakhorst:
obsolete by other PRs

PR #1200, opened 2022-06-20
Docs: frequently asked errors

reinkrul:


woutslakhorst (changes requested):
Page is not added to index by default. Alter `index.rst`

woutslakhorst (approved):

PR #1204, opened 2022-06-20
added universal subscriber

woutslakhorst:
It allows for a subscriber to have its events be persisted in the DB (same as DAG). This makes sure the event reaches its destination.

The interface is splitted in a `Save` and `Notify`. Save is to be used within a go-stoabs transaction and notify on the AfterCommit.


network/dag/subscriber.go
```diff
@@ -0,0 +1,356 @@
...
+	// TransactionEventType is used as Type in an Event when a transaction is added to the DAG.
+	TransactionEventType = "transaction"
+	// PayloadEventType is used as Type in an Event when a payload is written to the DB.
+	PayloadEventType = "payload"
+)
+
+// Subscriber defines methods for a persistent retry mechanism.
+// Storing the event in the DB is separated from notifying the subscribers.
+// The event is sent to subscribers after the transaction is committed to prevent timing issues.
+type Subscriber interface {
```
reinkrul, line 46: Given the pub-sub pattern, Subscriber sounds a bit off; this is more like an EventPublisher? (since it does all the event-queue-retry stuff). I'd call only the receiver of the event the subscriber.

reinkrul: Shouldn't we actually use NATS for this?

network/dag/subscriber.go
```diff
@@ -0,0 +1,356 @@
...
+	// Save an event that needs to be retried even after a crash.
+	// It will not yet be sent, use Notify to notify subscribers.
+	// The event may be ignored due to configured filters.
+	// Returns nil if event already exists.
+	// An error is returned if there's a problem with the underlying storage.
+	Save(tx stoabs.WriteTx, event Event) error
```
reinkrul, line 52: Leaky abstraction: Save() apparently reuses some external TX for writing, Finished() starts a TX of its own.

reinkrul (changes requested):

network/dag/subscriber.go
woutslakhorst: > Shouldn't we actually use NATS for this?

We can't use Nats transactionally and Nats then becomes part of "MUST be backed up" family. Saving the extra backup requirement with a few hundred lines of code was worth the investment.

network/dag/subscriber.go
woutslakhorst: This is by design, the events are connected/hooked to the DAG. This guarantees the delivery and does not leave unwanted events on rollback.

network/dag/subscriber.go
woutslakhorst: Renamed to `(New)Notifier` with a `ReceiverFn`

reinkrul (approved):

PR #1205, opened 2022-06-20
Docs: simplify options generators

reinkrul:
Just map flags to engine name: this allows engines to specify options that don't necesarrily map to a Koanf struct entry.


woutslakhorst (changes requested):
Lots of duplicates in the generated list

reinkrul:
> Lots of duplicates in the generated list

Weird, didn't notice that at first...

PR #1206, opened 2022-06-20
Render graph on Lamport Clock ranges

### LINKED ISSUE #1199: Refactor render graph
It tries to render the entire graph, which is impossible. As a feature we could add a from/to LC param with a max size.
This allows for partial rendering.

gerardsn:
closes #1199 

Graphs can be created using open ended range requests.

not sure if this is the right api format


reinkrul (approved):

network/network.go
```diff
@@ -443,7 +436,12 @@ func (n *Network) GetTransactionPayload(transactionRef hash.SHA256Hash) ([]byte,
 
 // ListTransactions returns all transactions known to this Network instance.
 func (n *Network) ListTransactions() ([]dag.Transaction, error) {
```
woutslakhorst, line 438: Not used anymore, since it was removed from the interface?

woutslakhorst (changes requested):
Update monitoring.rst in docs

woutslakhorst (approved):

PR #1208, opened 2022-06-21
rewrite gossip callback

woutslakhorst:
The gossip callback is the first to remove the observer and use the universal notifier.
The others can't be done yet due to stoabs not being used yet in the DAG.
The new Notifier option on the state will replace all subscribers on the state eventually.


network/dag/state.go
```diff
@@ -279,6 +326,13 @@ func (s *state) Start() error {
 		return err
 	}
 
+	// resume all notifiers
+	for _, subscriber := range s.notifiers {
```
gerardsn, line 330: subscriber -> notifier

network/dag/interface.go
```diff
@@ -54,6 +54,14 @@ type State interface {
...
 	// IsPresent returns true if a transaction is present in the DAG
 	IsPresent(context.Context, hash.SHA256Hash) (bool, error)
+	// Notifier creates a new Notifier.
+	// It lets another part of the application receive events on new transactions. When a new transaction is received
+	// the `receiver` function is called. Notifiers can be persistent and will survive restarts.
+	// The name is used to keep different notifiers apart.
+	// Filters can be used to receive specific transactions. Filters are added via the WithSelectionFilter() option.
+	// A Notifier should only be created during `configuration` step since the `start` step will redeliver all events that have not been delivered yet.
+	// Returns an error when the Notifier already exists
+	Notifier(name string, receiver ReceiverFn, filters ...NotifierOption) (Notifier, error)
```
gerardsn, line 64: Probably doesn't need to return the `Notifier`?

gerardsn (changes requested):

network/dag/interface.go
woutslakhorst: Not related to earlier comment, but I'll take it:
see nuts-foundation/go-stoabs#10, when the DB is accessible we can compare the TX with the DB

gerardsn: Sounds like a good solution

woutslakhorst: @gerardsn implemented

gerardsn (approved):

PR #1212, opened 2022-06-22
Replace dag with go-stoabs

gerardsn:


gerardsn:
Integration tests are still a bit flaky since the tests sync everything to disk again. Is there a way to set `stoabs.Options` as defaults in the storage engine? We need a way to do this if there isn't

network/dag/state.go
```diff
@@ -246,48 +256,38 @@ func (s *state) IBLT(ctx context.Context, reqClock uint32) (tree.Iblt, uint32) {
...
-		return fmt.Errorf("failed to read ibltTree: %w", err)
-	}
+	s.graph.init()
+	s.payloadStore.(*payloadStore).init()
+	s.loadTrees()
 
 	if err := s.Verify(); err != nil {
 		return err
-	}
+	} // TODO: is this still needed?
```
woutslakhorst, line 275: Create an issue for that

network/dag/state_test.go
```diff
@@ -119,12 +115,13 @@ func TestState_Shutdown(t *testing.T) {
...
 		ctx := context.Background()
-		txState := createState(t, func(_ *bbolt.Tx, _ Transaction) error {
+		txState := createState(t, func(_ stoabs.ReadTx, _ Transaction) error {
 			return errors.New("failed")
 		})
 		tx := CreateTestTransactionWithJWK(0)
-
+		println(1)
 		err := txState.Add(ctx, tx, nil)
+		println(2)
```
woutslakhorst, line 124: and more

network/network.go
```diff
@@ -80,6 +80,12 @@ type Network struct {
 	storeProvider       storage.Provider
 }
 
+// Walk walks the DAG starting at the root, passing every transaction to `visitor`.
+func (n *Network) Walk(visitor dag.Visitor) error {
```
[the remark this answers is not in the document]
woutslakhorst: Unused I think

woutslakhorst (changes requested):
Small stuff and a lock

network/network.go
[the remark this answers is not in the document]
gerardsn: I as thinking the Reprocess command should use this, but removed for now

network/dag/state.go
```diff
@@ -137,11 +107,46 @@ func (s *state) Add(ctx context.Context, transaction Transaction, payload []byte
...
+
+func (s *state) updateTrees(tx stoabs.WriteTx, transaction Transaction) error {
+	if err := s.ibltTree.write(tx, transaction); err != nil {
+		return err
+	}
+	return s.xorTree.write(tx, transaction)
+}
+
+func (s *state) loadTrees() {
+	// get an exclusive lock
```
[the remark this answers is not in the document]
gerardsn: added lock

network/dag/bbolt_dag.go
```diff
@@ -19,224 +19,204 @@
...
-func (dag *bboltDAG) diagnostics() []core.DiagnosticResult {
-	result := make([]core.DiagnosticResult, 0)
-	ctx := context.Background()
-	stats := dag.statistics(ctx)
-	result = append(result, headsStatistic{heads: dag.heads(ctx)})
-	result = append(result, numberOfTransactionsStatistic{numberOfTransactions: stats.NumberOfTransactions})
-	result = append(result, dataSizeStatistic{sizeInBytes: stats.DataSize})
-	return result
+func (d *dag) init() error {
+	return d.db.Write(func(tx stoabs.WriteTx) error {
```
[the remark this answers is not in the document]
gerardsn: rewrote a bit so they are all initialized from State. (non of the stores actually needs to have access tot he db, passing a tx is enough)

network/dag/bbolt_tree.go
```diff
@@ -19,155 +19,110 @@
...
+// newTreeStore returns an instance of a BBolt based tree store. Buckets managed by this store are filled to treeBucketFillPercent
+func newTreeStore(db stoabs.KVStore, bucketName string, tree tree.Tree) *treeStore {
+	return &treeStore{
+		db:         db,
+		bucketName: bucketName,
+		tree:       tree,
 	}
 }
 
+func (store *treeStore) init() error {
```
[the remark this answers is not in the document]
gerardsn: added test on State.Start

network/network_integration_test.go
```diff
@@ -89,6 +89,8 @@ func TestNetworkIntegration_HappyFlow(t *testing.T) {
 	}
 	expectedDocLogSize++
 
+	time.Sleep(100 * time.Millisecond) // Need a small delay for root transaction to propagate within the nodes.
```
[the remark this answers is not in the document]
gerardsn: Adding the tx is already wrapped in a waitFor. The subscriber that is used in this wait appears to complete before state.Add does, allowing the waitFor to complete before the root tx is actually in the store. Not really sure what has changed though...

network/dag/state.go
```diff
@@ -301,33 +301,35 @@ func (s *state) Verify() error {
 	})
 }
 
+func (s *state) Walk(_ context.Context, visitor Visitor, startAt uint32) error {
```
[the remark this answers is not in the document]
gerardsn: added test

network/dag/bbolt_dag.go
```diff
@@ -252,145 +232,101 @@ func (dag *bboltDAG) addSingle(tx *bbolt.Tx, transaction Transaction) error {
...
 
-	log.Logger().Tracef("storing transaction logical clock, txRef: %s, clock: %d", ref.String(), clock)
+	log.Logger().Tracef("storing transaction logical clock, txRef: %s, clock: %d", ref.String(), clockKey)
 
 	return nil
 }
 
 // returns the highest clock for which a transaction is present in the DAG
-func (dag bboltDAG) getHighestClock(ctx context.Context) uint32 {
+func (d dag) getHighestClock() uint32 {
```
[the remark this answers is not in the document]
gerardsn: there is an issue (1055) to keep track of the LC in state instead of getting it from the db. We probably still need this to initialize the state?

network/dag/state.go
gerardsn: issue 1216

network/dag/bbolt_payloadstore.go
```diff
@@ -21,46 +21,42 @@ package dag
...
 		return nil
 	}
 	data, err := reader.Get(stoabs.NewHashKey(payloadHash))
 	return data
 }
 
 func (store payloadStore) writePayload(tx stoabs.WriteTx, payloadHash hash.SHA256Hash, data []byte) error {
-	writer, err := tx.GetShelfWriter(payloadsShelfName)
+	writer, err := tx.GetShelfWriter(payloadsShelf)
+	log.Logger().Errorf("failed to write payload (hash=%s): %s", payloadHash, err)
```
woutslakhorst, line 59: don't log here, in a higher level it'll be logged again.

woutslakhorst (changes requested):

network/dag/bbolt_tree.go
```diff
@@ -87,14 +78,19 @@ func (store *treeStore) read(tx stoabs.ReadTx) error {
 	// get data
 	rawData := map[uint32][]byte{}
 	err = reader.Iterate(func(k stoabs.Key, v []byte) error {
-		split := binary.LittleEndian.Uint32(k.Bytes())
+		split := binary.BigEndian.Uint32(k.Bytes()) // stoabs.Key uses BigEndian
```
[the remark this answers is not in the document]
woutslakhorst: Hmmm does this mean that our clocks shelf is broken if we deploy it?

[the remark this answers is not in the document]
gerardsn: changed to write using stoabs.BytesKey instead of Uint32Key to keep the keys in the same endianness

network/dag/bbolt_payloadstore.go
```diff
@@ -19,43 +19,45 @@
...
-	bucket := tx.Bucket([]byte(payloadsBucketName))
-	if bucket == nil {
+func (store payloadStore) isPayloadPresent(tx stoabs.ReadTx, payloadHash hash.SHA256Hash) bool {
+	reader, err := tx.GetShelfReader(payloadsShelf)
+	if err != nil {
+		log.Logger().Errorf("failed to verify payload existence (hash=%s): %s", payloadHash, err)
 		return false
 	}
-	data := bucket.Get(payloadHash.Slice())
+	data, err := reader.Get(stoabs.NewHashKey(payloadHash))
```
reinkrul, line 43: error is ignored

network/dag/bbolt_payloadstore.go
```diff
@@ -19,43 +19,45 @@
...
-	if bucket == nil {
+func (store payloadStore) readPayload(tx stoabs.ReadTx, payloadHash hash.SHA256Hash) []byte {
+	reader, err := tx.GetShelfReader(payloadsShelf)
+	if err != nil {
+		log.Logger().Errorf("failed to read payload (hash=%s): %s", payloadHash, err)
 		return nil
 	}
-	return bucket.Get(payloadHash.Slice())
+	data, err := reader.Get(stoabs.NewHashKey(payloadHash))
+	return data
```
reinkrul, line 54: error is ignored

network/dag/state.go
```diff
@@ -24,73 +24,65 @@ import (
...
-	newState.xorTree = xorTree
-	newState.ibltTree = ibltTree
-	newState.RegisterTransactionObserver(newState.treeObserver, true)
-
 	return newState, nil
 }
 
+func initShelfs(db stoabs.KVStore, shelfs ...string) error {
+	for _, shelf := range shelfs {
+		if err := db.WriteShelf(shelf, func(writer stoabs.Writer) error {
```
reinkrul, line 77: This will probably won't do anything when using Redis

network/dag/state.go
```diff
@@ -137,11 +122,49 @@ func (s *state) Add(ctx context.Context, transaction Transaction, payload []byte
...
+		}
+
+		if err := s.ibltTree.read(tx); err != nil {
+			return fmt.Errorf("failed to read ibltTree: %w", err)
+		}
+		return nil
+	}); err != nil {
+		log.Logger().Errorf("Failed to load the XOR and IBLT trees: %s", err)
+	}
+	log.Logger().Info("Loaded the XOR and IBLT trees")
```
reinkrul, line 164: not really INFO level information?

reinkrul (changes requested):
Aside from the TODOs found some stuff, rest is up to the tests to prove the implementation

network/dag/state.go
gerardsn: changed to TRACE

network/dag/bbolt_dag.go
```diff
@@ -19,224 +19,204 @@
...
+type dag struct {
+	db stoabs.KVStore
 }
 
 type headsStatistic struct {
 	// SHA256Hash is the last consistency hash.
 	heads []hash.SHA256Hash
 }
 
-func (d headsStatistic) Result() interface{} {
```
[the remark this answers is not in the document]
gerardsn: renaming `bboltDAG` to `dag` had some confusing side effects, but agree on the scope creep

network/dag/state.go
gerardsn: The `stoabs.ReadTx` interface for `GetShelfReader(...)` states that if a shelf does not exist, the returned `stoabs.Reader == nil`. 

For bbolt `initShelfs` means that we don't have to check if the reader is nil every time. Redis might still fail depending on its implementation. The existence of a shelf in Redis could be derived from a partial key match, but for certain shelfs (connections/payload_jobs) existence is not well defined/transient. This is rather confusing. I'm not sure what the right way of dealing with this is.

Any suggestions?

network/dag/bbolt_payloadstore.go
gerardsn: error returned (as described by the interface)

network/dag/bbolt_payloadstore.go
gerardsn: logged

network/dag/state.go
woutslakhorst: Mark current lines with a TODO.
Add an issue to go-stoabs for a feature request: support DB specific initialisation scripts.

woutslakhorst (approved):

network/dag/state.go
gerardsn: https://github.com/nuts-foundation/go-stoabs/issues/14

reinkrul: Maybe it would be simpler to have `GetShelfReader` return a reader, even when the shelf doesn't exist. But all read operations will return a value that indicates the key doesn't exist.

The check for Redis to determine "does shelf exist" involves doing a wildcard match for the shelf and then count the results, which doesn't seem very efficient to me.

If the existence check is required, we can add it as separate function (`store.ShelfExists()`)?

gerardsn: To me that would be a reasonable outcome of the issue mentioned above, but probably out of the scope of this PR?

woutslakhorst: The empty reader would be a good solution. https://github.com/nuts-foundation/go-stoabs/issues/14 may be left open for future use.

reinkrul (approved):

PR #1226, opened 2022-06-28
FAQ: 'JWT signing key not present on this node'

reinkrul:


docs/pages/faq-errors.rst
```diff
@@ -13,4 +13,17 @@ When connecting to a remote node, the following error can occur:
...
+*************************************************
+
+When inspecting an access token, the following error can occur:
+
+    Error while inspecting access token (error: JWT signing key not present on this node)
+
+This indicates the JWT token you send to the Nuts node can't be introspected by the Nuts node,
+because it can't find the private key to create the token. This can have 2 causes:
+
+# Calls are mixed up; token introspection must be done by the Nuts node (*authorizer*) that created it for the consuming node (*requestor*).
```
gerardsn, line 28: > Calls are mixed up;

Do you mean that the call went to the wrong node? (So nodes are mixed up?)

woutslakhorst: For JWT tokens the node creates its own keypair on startup, it's not DID based

PR #1238, opened 2022-07-04
VDR: replace bbolt storage with stoabs

gerardsn:


cmd/root.go
```diff
@@ -184,7 +184,7 @@ func CreateSystem() *core.System {
 	cryptoInstance := crypto.NewCryptoInstance()
 	jsonld := jsonld.NewJSONLDInstance()
 	storageInstance := storage.New()
-	didStore := store.NewBBoltStore()
+	didStore := store.NewStore(storageInstance.GetProvider(vdr.ModuleName))
```
woutslakhorst, line 187: moduleName is upper case. This is translated to a directory. This will work on your mac but will be not backwards compatible on linux.

vdr/store/bbolt.go
```diff
@@ -16,62 +16,56 @@
...
+	metadataShelf = "metadata"
+	// transactionIndexShelf has the transaction reference as key and metadata reference as value
+	transactionIndexShelf = "txRef"
+	// documentShelf has payload hash as key and document as value
+	documentShelf = "documents"
 )
 
-type bboltStore struct {
-	db *bbolt.DB
+type store struct {
```
woutslakhorst, line 42: filename may be changed as well.

vdr/store/bbolt.go
```diff
@@ -16,62 +16,56 @@
...
-func (store *bboltStore) Configure(config core.ServerConfig) error {
+func (s *store) Configure(_ core.ServerConfig) error {
 	var err error
-	filePath := path.Join(config.Datadir, "vdr", "didstore.db")
-	if err = os.MkdirAll(path.Join(config.Datadir, "vdr"), os.ModePerm); err != nil {
-		return err
-	}
-	store.db, err = bbolt.Open(filePath, 0600, bbolt.DefaultOptions)
-
+	s.db, err = s.storeProvider.GetKVStore("didstore", storage.PersistentStorageClass)
```
woutslakhorst, line 58: const

woutslakhorst (changes requested):

vdr/store/bbolt.go
gerardsn: I will change all files in a different PR. Github picks this up as a delete old + add new for some files instead of a rename. This is hard to review.

cmd/root.go
gerardsn: module name is convert to lowercase by storage engine

vdr/store/bbolt.go
gerardsn: added `didStoreName` const

woutslakhorst (approved):

PR #1241, opened 2022-07-05
Network: introduce 'analyze' command

reinkrul:


woutslakhorst:
This is very specific to a bbolt file. Therefore it's a bit weird it's part of the main executable?

reinkrul:
> This is very specific to a bbolt file. Therefore it's a bit weird it's part of the main executable?

Could argue these report/analytical features shouldn't be part of the nuts-node, but making it a separate repo/application is a bit much right now?

Could imporove it a bit when it just takes the server config, then any backing KVStore type is supported, rather than just BBolt.

woutslakhorst:
> > This is very specific to a bbolt file. Therefore it's a bit weird it's part of the main executable?
> 
> Could argue these report/analytical features shouldn't be part of the nuts-node, but making it a separate repo/application is a bit much right now?
> 
> Could imporove it a bit when it just takes the server config, then any backing KVStore type is supported, rather than just BBolt.

That would be better. Other remark is naming: we have diagnostics, metrics and now analyze. Shouldn't something like `signer summary` be a better name. Or make sub commands for different summaries?

reinkrul:
> > > This is very specific to a bbolt file. Therefore it's a bit weird it's part of the main executable?
> > 
> > 
> > Could argue these report/analytical features shouldn't be part of the nuts-node, but making it a separate repo/application is a bit much right now?
> > Could imporove it a bit when it just takes the server config, then any backing KVStore type is supported, rather than just BBolt.
> 
> That would be better. Other remark is naming: we have diagnostics, metrics and now analyze. Shouldn't something like `signer summary` be a better name. Or make sub commands for different summaries?

`nuts network analyze signers`

reinkrul:
This PR needs a redesign: it wants to use the stores as configured (e.g. `nuts.yaml`/environment variables). The easiest way would be to use `core.System` to configure and start the required modules. But then it would also start a gRPC listener and start connecting to other nodes.

So it wants to use the initialized storage, but does not want to start the services.

woutslakhorst:
Will be covered with other tooling.

PR #1242, opened 2022-07-05
elaborate on VC searching

### LINKED ISSUE #1133: Searching for a credential type should filter out other types
When searching for a credential, the type is used to expand the fields, but it's not used to filter out other types. So when a credential has matching fields but is from a different type, it still appears in the results. This is confusing.

woutslakhorst:
fixes #1133

### DIFF
--- a/docs/pages/integrating/vc.rst
+++ b/docs/pages/integrating/vc.rst
@@ -97,5 +97,41 @@ The example below searches for a `NutsOrganizationCredential` (note that the `qu
 
 Note the fields `@context` and `type`, these are required for making it a valid VC in JSON-LD.
 In the example above they also contain Nuts specific contexts and types (since we're searching for a Nuts VC).
+The fields `@context` and `type` are not used as query parameters for searching, they are required to determine the right context.
+The following query does not return all `NutsOrganizationCredential` but **all** credentials.
+
+.. code-block:: json
+
+    {
+        "query": {
+            "@context": [
+                "https://www.w3.org/2018/credentials/v1",
+                "https://nuts.nl/credentials/v1"
+            ],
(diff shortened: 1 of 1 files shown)


reinkrul (commented):
This might solve the issue for some who actually read and understand the documentation, but the API is still very counter-intuitive as a search API. Why not always filter out all other credentials which are not the specified type(s)?

woutslakhorst:
> Why not always filter out all other credentials which are not the specified type(s) (maybe unless a flag is specified)?
because that's not how JSON-LD works. We must not encourage implementations to rely on specific type identifiers

reinkrul:
> > Why not always filter out all other credentials which are not the specified type(s) (maybe unless a flag is specified)?
> > because that's not how JSON-LD works. We must not encourage implementations to rely on specific type identifiers

Then why do we have to specify specific type identifiers, if we musn't encourage implementations to rely on them? (don't answer; rhetorical question).

I understand that it makes sense from a JSON-LD point of view, but that doesn't make the search API any better. Imo it's naive to expect developers who'll be using the search API to understand the pecularities of JSON-LD and how its used when searching for VCs.

Nonetheless, given we don't change the API I think we need the docs as proposed by this PR.

But that doesn't make the search API any better. That said, the Nuts node might not be the right application to provide good DX (developer experience) for non-JSON LD-experts. Just like we introduced DIDMan (and maybe `Auth`), a domain-friendly search API might take the form of a `(Nuts) Registry` service deployed in front of a Nuts node.

reinkrul (approved):
See comment.

woutslakhorst:
I believe DID SIOP has introduced a query language that is also used in OpenID-Connect4VP. That would be a solid future improvement.

Issue #1243, opened 2022-07-05
Add lc to tx crit headers again


woutslakhorst:
and change version to 2.

v4 accepts both {1, 2}

PR #1244, opened 2022-07-05
revert tx version back to v1, remove LC from crit headers

### LINKED ISSUE #1240: Transaction version not backwards compatible
Master already creates version 2 transactions (LC in crit). But v3 only supports version 1 transactions.
V4 should support both and V5 should emit them.

woutslakhorst:
fixes #1240 

v5 issue: #1243

### DIFF
--- a/network/dag/parser_test.go
+++ b/network/dag/parser_test.go
@@ -68,6 +68,39 @@ func TestParseTransaction(t *testing.T) {
 		assert.NotNil(t, transaction.Data())
 		assert.False(t, transaction.Ref().Empty())
 	})
+	t.Run("ok v2", func(t *testing.T) {
+		headers := makeJWSHeaders(key, "123", true)
+		_ = headers.Set("pal", []string{base64.StdEncoding.EncodeToString([]byte{5, 6, 7})})
+		_ = headers.Set(versionHeader, 2)
+		_ = headers.Set(jws.CriticalKey, []string{signingTimeHeader, versionHeader, previousHeader, lamportClockHeader})
+		signature, _ := jws.Sign(payloadAsBytes, headers.Algorithm(), key, jws.WithHeaders(headers))
+
+		transaction, err := ParseTransaction(signature)
+		if !assert.NoError(t, err) {
+			return
+		}
+
+		var actualKey ecdsa.PublicKey
+		err = transaction.SigningKey().Raw(&actualKey)
+		if !assert.NoError(t, err) {
+			return
+		}
+
+		assert.NotNil(t, transaction)
+		if !assert.NoError(t, err) {
+			return
+		}
+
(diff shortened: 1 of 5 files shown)


reinkrul (approved):

network/dag/parser_test.go
```diff
@@ -245,7 +278,8 @@ func makeJWSHeaders(key crypto.Signer, kid string, embedKey bool) jws.Headers {
 	headerMap := map[string]interface{}{
 		jws.AlgorithmKey:   jwa.ES256,
 		jws.ContentTypeKey: "foo/bar",
-		jws.CriticalKey:    []string{signingTimeHeader, versionHeader, previousHeader, lamportClockHeader},
+		// TODO add lamportClockHeader after v4.0.0 tag
+		jws.CriticalKey:    []string{signingTimeHeader, versionHeader, previousHeader},
 		lamportClockHeader: 0,
```
gerardsn, line 283: lamportClockHeader is always set making the test for v1 incorrect

gerardsn (changes requested):
This still does not support or comply to the specs for transactions v1.
v1 does not contain an lc (RFC004), and according to v2 (RFC016) a virtual lc must be calculated to overcome this. Currently parser.go#parseLamportClock returns an error if the header does not contain an lc instead of calculating one.

woutslakhorst:
> This still does not support or comply to the specs for transactions v1. v1 does not contain an lc (RFC004), and according to v2 (RFC016) a virtual lc must be calculated to overcome this. Currently parser.go#parseLamportClock returns an error if the header does not contain an lc instead of calculating one.

Yes, we made a mistake in the transition. Either altering the spec or changing the code has 0 influence on the outcome for V5. It'll be more work and could introduce bugs.

It's not totally as it should be, but since there are no other implementations we can get away with it.

network/dag/parser_test.go
woutslakhorst: we're not going to introduce calculated LV values at this point, not worth the trouble. Let's agree V3 and V4 are not totally according to spec.

gerardsn:
So nuts-node v3 actually publishes v1 transactions with an extra and header and parses them as a v1-v2 hybrid (tx labeled as v1 and no `lc` in `crit`, but assumes that it does contain the `lc` header)

Since there exists transactions labeled v1 on the dag, we will forever have to accept v1 transactions. Right?
Currently there exists no other implementation, but if someone decides to build one that conforms to the spec, our node will not accept correct v1 transactions. 

I don't think this wil be fixed in node v5 or after unless we fully implement the spec for both versions. Other options could be a big-bang that deprecates v1 transactions, or add the lc to RFC004 to match wat node v3 does, but that might violate JWS RFC7515.

gerardsn:
Moved this discussion to #1249

gerardsn (approved):

PR #1247, opened 2022-07-06
Docs: generate CLI command reference

reinkrul:
See https://nuts-node.readthedocs.io/en/docs-cli-command-reference/pages/deployment/cli/index.html


docs/pages/deployment/cli/commands/nuts_network.rst
```diff
@@ -0,0 +1,34 @@
+.. _nuts_network:
```
gerardsn, line 1:
```
nuts network -h
```
returns
```
network commands

Usage:
  nuts network [command]

Available Commands:
  get         Gets a transaction from the network
  list        Lists the transactions on the network
  payload     Retrieves the payload of a transaction from the network
  peers       Get diagnostic information of the node's peers
  reprocess   Reprocess all transactions with the give contentType (ex: application/did+json)

Flags:
      --address string     Address of the remote node. Must contain at least host and port, URL scheme may be omitted. In that case it 'http://' is prepended. (default "localhost:1323")
  -h, --help               help for network
      --timeout duration   Client time-out when performing remote operations, such as '500ms' or '10s'. Refer to Golang's 'time.Duration' syntax for a more elaborate description of the syntax. (default 10s)
      --verbosity string   Log level (trace, debug, info, warn, error) (default "info")
```

The `Available Commands` section from the help are in the docs replaced by the `SEE ALSO` section and moved to the end. To me it was not clear that this section contains the available commands and looks more like a suggested reading section that I can ignore. Can this be changed to look a bit more like the help?

woutslakhorst:
Screenshot 2022-07-08 at 11 16 38

The menu options are a bit overkill. I think that each command can do without the submenu

woutslakhorst (changes requested):
see above comment

docs/pages/deployment/cli/commands/nuts_network.rst
```diff
@@ -0,0 +1,34 @@
...
+nuts network
+------------
+
+network commands
+
+Synopsis
+~~~~~~~~
+
+
+network commands
```
gerardsn, line 12: shouldn't this be `network [command]`

docs/pages/deployment/cli/commands/nuts_network_get.rst
```diff
@@ -0,0 +1,39 @@
...
+  nuts network get [ref] [flags]
+
+Options
+~~~~~~~
+
+::
+
+  -h, --help   help for get
+
+Options inherited from parent commands
```
gerardsn, line 25: - Inconsistent use of `[flags]` and `options` is confusing.
- Not all options/flags/inputs are listed, but this is probably more a documentation thing.

gerardsn (dismissed):
Great addition to the docs.

I think the way the information presented by the cli help is clearer than the docs though.

I've made some comments on the network command, but I think they apply everywhere.

docs/pages/deployment/cli/commands/nuts_network_get.rst
reinkrul: out of scope

reinkrul:
Removed See Also section, joined flags/options because "inherited from parent" didn't add anything. Changed the subsections to bolded text, so they don't show up in the TOC. Added horizontal lines between commands.

woutslakhorst (approved):

PR #1255, opened 2022-07-12
Network: store number of TXs

### LINKED ISSUE #1055: Keep track of highest LC value in dag.State
currently it uses a bbolt cursor to find the latest element. This is not needed, it can be updated with each Add and use the cursor on startup.

### LINKED ISSUE #1253: Network: persist number of TXs in store (instead of calculating it on database)
The network diagnostics shows the number of transactions. This is calculating by requesting the number of entries in a storage shelf. Since not every database supports this (Redis), this solution doesn't work.

Solve this by persisting the number of transactions in the database. We need to introduce a new store for state metadata (more will follow, e.g. current highest LC).

reinkrul:
Closes #1253 
Closes #1055


woutslakhorst:
Add a prometheus metric for #TXs.
Add highest LC value.

reinkrul:
> * [ ]  Add a prometheus metric for #TXs.

out of scope I'd say

woutslakhorst (approved):

Issue #1216, opened 2022-06-24
DAG integrity check at startup


gerardsn:
State checks the integrity of the DAG at startup by verifying all transactions. This can become problematic once the dag contains millions of transactions.

Is this still needed?

woutslakhorst:
nope, but do we want a command to run such a check?

gerardsn:
I think it would be nice to have a way to trigger this check manually. Maybe move if up to Network and let it use state.Walk?

woutslakhorst:
continued in #1506

PR #1259, opened 2022-07-13
disable DAG.Verify at startup

woutslakhorst:
part of #1216 

Also removed `walk`, not used anymore.

### DIFF
--- a/network/dag/bbolt_dag.go
+++ b/network/dag/bbolt_dag.go
@@ -22,7 +22,6 @@ import (
 	"encoding/binary"
 	"errors"
 	"fmt"
-	"math"
 	"sort"
 
 	"github.com/nuts-foundation/go-stoabs"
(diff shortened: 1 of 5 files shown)


reinkrul (commented):
Remove in itself does look OK, but now we have dead code around. Is it useful to have as a CLI command? If we don't think so at this moment: it's so little code, i'd say just remove it (can be easily rewritten later).

reinkrul (changes requested):
See comment

woutslakhorst:
> Remove in itself does look OK, but now we have dead code around. Is it useful to have as a CLI command? If we don't think so at this moment: it's so little code, i'd say just remove it (can be easily rewritten later).

As you can see, this is part of #1216 not the entire solution for that issue. Removing the verify has a higher priority as the command. The command still has to be added. TODOs have been added to the correct places in code.

reinkrul (approved):

PR #1268, opened 2022-07-17
Auth: log module name with IRMA log entries

reinkrul:
Suggested change in irmago: https://github.com/privacybydesign/irmago/issues/232


woutslakhorst (commented):
But if we configure our logger to do JSON, then this logger will still do text?

reinkrul:
> But if we configure our logger to do JSON, then this logger will still do text?

No it will output as JSON: the `JSONFormatter` is set on the `StandardLogger` before the auth module is initialized, and this code copies the StandardLogger and wraps the `Formatter` (text or JSON) with a decorator that adds a field.

woutslakhorst (approved):

PR #1273, opened 2022-07-18
Drop bbolt from store names

gerardsn:
renamed files and removed several references to bbolt


network/dag/dag_test.go
```diff
@@ -109,7 +109,7 @@ func TestBBoltDAG_Get(t *testing.T) {
 			return nil
 		})
 	})
-	t.Run("bbolt byte slice is copied", func(t *testing.T) {
+	t.Run("bbolt byte slice is copied", func(t *testing.T) { // TODO: can this be removed?
```
woutslakhorst, line 112: Yes, This is not part of go-stoabs

woutslakhorst (approved):
See comment on test removal

PR #1274, opened 2022-07-19
Network: clean up bucket init code

reinkrul:
Depending on the backing store, this might not do anything.

### DIFF
--- a/network/dag/test.go
+++ b/network/dag/test.go
@@ -113,7 +113,6 @@ func createBBoltDB(testDirectory string) stoabs.KVStore {
 func CreateDAG(t *testing.T) *dag {
 	testDirectory := io.TestDirectory(t)
 	d := newDAG(createBBoltDB(testDirectory))
-	d.init()
 	return d
 }
 
--- a/network/dag/payloadstore_test.go
+++ b/network/dag/payloadstore_test.go
@@ -28,7 +28,6 @@ import (
 
 	"github.com/nuts-foundation/go-stoabs"
 	"github.com/nuts-foundation/nuts-node/crypto/hash"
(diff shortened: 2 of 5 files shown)


network/dag/test.go
```diff
@@ -113,7 +113,6 @@ func createBBoltDB(testDirectory string) stoabs.KVStore {
 func CreateDAG(t *testing.T) *dag {
 	testDirectory := io.TestDirectory(t)
 	d := newDAG(createBBoltDB(testDirectory))
-	d.init()
```
gerardsn, line 116: bbolt buckets are no longer initialized without this. (Currently happens in `State`.) Tests do not fail because all happen to add a tx first (write) that initializes the buckets. This is the most probable scenario, so not sure if its worth doing anything about it...

reinkrul: `go-stoabs` return a reader that always return `nil` when it operates on a bucket that doesn't exist. In what scenario will this cause a problem?

gerardsn: The actual behavior in `go-stoabs` depends on the underlying store. Redis will never be nil for example. Also see [go-stoabs #14](https://github.com/nuts-foundation/go-stoabs/issues/14). Since this function hides the underlying store, having to initialize buckets/shelfs first is unexpected behavior to me. (Also, stoabs doesn't have a way to initialize shelfs other than writing to it.)

gerardsn: stoabs has had some changes that I was unaware of.
storage/engine.go#InitializeShelfs and all of its used can also be removed.

reinkrul: removed it as well

PR #1295, opened 2022-07-25
Add page for custom credentials to docs

stevenvegt:


woutslakhorst (commented):
Maybe add a note on indexing. Custom creds are not indexed on any custom fields. So large scale usage requires a new feature: custom indices?

woutslakhorst (approved):

PR #1296, opened 2022-07-25
Fix options table layout

### LINKED ISSUE #1282: Fix options table layout
The default values for the JSON-LD context is so large it completely ruins the table of server options. Both in README and on readthedocs

Large values could be replaced with placeholders which are then explained below the table?

stevenvegt:
Fix #1282 
And fix several docs creation warnings.


woutslakhorst (changes requested):
If I generate the docs, the table still looks like this:
Screenshot 2022-07-26 at 11 25 59

woutslakhorst (approved):

PR #1306, opened 2022-07-27
add transaction count metric to prometheus output

### LINKED ISSUE #1287: Add number of transactions processed to metrics
basically the same number in the diagnostics but then as metric.

woutslakhorst:
closes #1287 

adds metric with name: `nuts_dag_transaction_count`
Count is incremented on state.Add()

### DIFF
--- a/network/dag/state.go
+++ b/network/dag/state.go
@@ -108,6 +108,7 @@ func (s *state) Add(ctx context.Context, transaction Transaction, payload []byte
 		Transaction: transaction,
 		Payload:     payload,
 	}
+	txAdded := false
 	emitPayloadEvent := false
 
 	return s.db.Write(ctx, func(tx stoabs.WriteTx) error {
@@ -115,6 +116,8 @@ func (s *state) Add(ctx context.Context, transaction Transaction, payload []byte
 		if present {
 			return nil
 		}
+		// control the afterCommit hooks
+		txAdded = true
 
 		if err := s.verifyTX(tx, transaction); err != nil {
 			return err
@@ -145,12 +148,16 @@ func (s *state) Add(ctx context.Context, transaction Transaction, payload []byte
 		log.Logger().Warn("Reloading the XOR and IBLT trees due to a DB transaction Rollback")
 		s.loadTrees(ctx)
 	}), stoabs.AfterCommit(func() {
-		s.notify(txEvent)
-		if emitPayloadEvent {
-			s.notify(payloadEvent)
+		if txAdded {
+			s.notify(txEvent)
+			if emitPayloadEvent {
+				s.notify(payloadEvent)
+			}
 		}
 	}), stoabs.AfterCommit(func() {
-		s.transactionCount.Inc()
+		if txAdded {
+			s.transactionCount.Inc()
+		}
 	}), stoabs.WithWriteLock())
 }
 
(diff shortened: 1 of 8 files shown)


network/dag/state.go
```diff
@@ -43,17 +44,14 @@ const (
...
-	xorTree                          *treeStore
-	ibltTree                         *treeStore
+	db               stoabs.KVStore
+	graph            *dag
+	payloadStore     PayloadStore
+	txVerifiers      []Verifier
+	notifiers        map[string]Notifier
+	xorTree          *treeStore
+	ibltTree         *treeStore
+	transactionCount prometheus.Counter
```
reinkrul, line 54: we now have 3 types of diagnostical data, diagnostics, `State`'s `func (s *state) Statistics(ctx context.Context) Statistics` and internal prometheus counters.

Maybe Statistics should be merged with Diagnostics?

network/dag/state.go
```diff
@@ -65,19 +63,36 @@ func NewState(db stoabs.KVStore, verifiers ...Verifier) (State, error) {
 	graph := newDAG(db)
 
 	payloadStore := NewPayloadStore()
+	transactionCount := transactionCountCollector()
+	err := prometheus.Register(transactionCount)
+	if err != nil && err.Error() != (prometheus.AlreadyRegisteredError{}).Error() { // No unwrap on prometheus.AlreadyRegisteredError
```
reinkrul, line 68: is this going to work when you start multiple instances in a test?

reinkrul (changes requested):

network/dag/state_test.go
```diff
@@ -364,6 +365,38 @@ func TestState_IBLT(t *testing.T) {
...
 }
 
+func TestState_InitialTransactionCountMetric(t *testing.T) {
+	// create state
+	ctx := context.Background()
+	txState := createState(t).(*state)
+	payload := []byte("payload")
+	tx, _, _ := CreateTestTransactionEx(1, hash.SHA256Sum(payload), nil)
+	dagClock := 3 * PageSize / 2
+	tx.(*transaction).lamportClock = dagClock
```
gerardsn, line 375: Are these needed for this test?

gerardsn (changes requested):

network/dag/state_test.go
woutslakhorst: nope

network/dag/state.go
woutslakhorst: no and it will never will with the way the prometheus client library works. It has one global registry.
The tests for seeing if the counter works are never executed in parallel so will work.

Other tests could have a wrong metric counter, which isn't a big deal

network/dag/state.go
woutslakhorst: Statistics removed

reinkrul (approved):

network/dag/state.go
```diff
@@ -134,6 +149,8 @@ func (s *state) Add(ctx context.Context, transaction Transaction, payload []byte
 		if emitPayloadEvent {
 			s.notify(payloadEvent)
 		}
+	}), stoabs.AfterCommit(func() {
+		s.transactionCount.Inc()
```
gerardsn, line 153: duplicate checking is done within the context of this transaction (L114). If the transaction already exists it exits, but still increments the counter

gerardsn (changes requested):

network/dag/state.go
woutslakhorst: good catch

gerardsn (approved):

Issue #1298, opened 2022-07-25
DB locks without timeout can cause node to lockup


woutslakhorst:
The bbolt write lock has no timeout feature. This causes API calls to create go routines which must finish no matter what. It's better to cancel a call and return the appropriate http code when the node is busy.

For example a 1 sec timeout on acquiring a lock increases performance and makes the node responsive under high load.

PR #1316, opened 2022-07-29
Storage: update go-stoabs, so lock acquisition expires in 1s

reinkrul:
Closes https://github.com/nuts-foundation/nuts-node/issues/1298

@woutslakhorst can you repeat the test with this change?


woutslakhorst:
> @woutslakhorst can you repeat the test with this change?

also requires expanded send buffer

woutslakhorst (approved):

PR #1323, opened 2022-08-01
reset conversation timeout on processed message

### LINKED ISSUE #1322: conversation validity check on receive
check conversation validity on receiving the message not processing it. 
Also reset the conversation timeout after each processed message. (This means the node is alive and processing, no need to stop the conversation)

woutslakhorst:
closes #1322 

also fixed transactionList processing channel size.


network/transport/v2/conversation.go
```diff
@@ -56,8 +56,9 @@ func (cid conversationID) String() string {
 }
 
 type conversation struct {
-	conversationID   conversationID
-	createdAt        time.Time
+	conversationID conversationID
+	// lastValidAt is the time the conversation was created or when the last message for the conversation was successfully processed.
+	lastValidAt      time.Time
```
reinkrul, line 61: `lastValidAt` sounds a bit quirky, would it make more sense to call it `expiry`? Then you only need to do calculations with `Add()` on it in `resetTimeout`.

network/transport/v2/transactionlist_handler.go
```diff
@@ -108,6 +110,8 @@ func (p *protocol) handleTransactionList(peer transport.Peer, envelope *Envelope
 
 	if msg.MessageNumber >= msg.TotalMessages {
 		p.cMan.done(cid)
+	} else {
+		p.cMan.resetTimeout(cid)
```
reinkrul, line 114: According to the linked issue it should be reset upon receival, not after processing?

reinkrul (changes requested):

network/transport/v2/transactionlist_handler.go
woutslakhorst: From issue:

> Also reset the conversation timeout after each processed message.

reinkrul (approved):

Issue #1303, opened 2022-07-27
Improve TLS support for HTTP connections


reinkrul:
We now support TLS offloading for gRPC with features and documentation. We don't have that same support for TLS on HTTP connections.

The least we should do is document it with the same level of detail as we did for TLS on gRPC, and maybe even support TLS on Echo (keep in mind we have 3 HTTP flavours: no TLS, server-certificate TLS (`/public`) and mutual TLS (`/n2n`).

PR #1334, opened 2022-08-04
Echo: support TLS for HTTP endpoints

reinkrul:
Closes https://github.com/nuts-foundation/nuts-node/issues/1303

This first iteration just adds support for TLS on Echo interfaces. Interfaces can be configured to have no TLS, one-way TLS (server cert) and mutual TLS (server and client cert).

TODO:

- [x] Let Network module use core `tls` properties when not set for the Network module (but only if `network.enabletls` is `true`).
- [x] Document TLS HTTP configuration
- [ ] Test example TLS proxy configuration

Considerations for further improvements:

- Remove TLS properties from `network` module (e.g. `network.certfile`) and use the ones from `core`.
- Apply `tls.offload` and `tls.enable` to all (gRPC and HTTP) endpoints.
  - Pro: makes TLS configuration easier to understand (and simpler to document)
  - Con: less flexible (but we can be opinionated; why would you offload TLS on gRPC and not on HTTP, or vice versa?)
- If no offloading, register sane defaults for HTTP interfaces. This means registering multiple HTTP endpoints instead of one:
  - `/n2n` (e.g. on `:1323`) with mutual TLS
  - `/public` (e.g. on `:1322`) with one-way TLS
  - other endpoints (`/internal`, `/status`, etc) don't need TLS, so can stay on `:1323`


core/server_config.go
```diff
@@ -197,10 +247,14 @@ func FlagSet() *pflag.FlagSet {
...
 	flagSet.Bool("internalratelimiter", true, "When set, expensive internal calls are rate-limited to protect the network. Always enabled in strict mode.")
 	flagSet.String("datadir", "./data", "Directory where the node stores its files.")
 	flagSet.StringSlice("http.default.cors.origin", nil, "When set, enables CORS from the specified origins for the on default HTTP interface.")
+	flagSet.String("tls.certfile", "", "PEM file containing the certificate for the server (also used as client certificate). Required when `network.enabletls` is `true`.")
+	flagSet.String("tls.certkeyfile", "", "PEM file containing the private key of the server certificate. Required when `tls.enable` is `true`.")
+	flagSet.String("tls.truststorefile", "truststore.pem", "PEM file containing the trusted CA certificates for authenticating remote servers.")
 	flagSet.String("tls.offload", "", "Whether to enable TLS offloading for incoming connections. If enabled `tls.certheader` must be configured as well.")
 	flagSet.String("tls.certheader", "", "Name of the HTTP header that will contain the client certificate when TLS is offloaded.")
```
gerardsn, line 259: defaults that are left blank `""` seem to be replaced by the first thing that appears in back ticks in the usage string when generating `cli-reference.rst`

``` 
    --tls.certfile network.enabletls                PEM file containing the certificate for the server (also used as client certificate). Required when network.enabletls is `true`.
    --tls.certheader string                         Name of the HTTP header that will contain the client certificate when TLS is offloaded.
    --tls.certkeyfile tls.enable                    PEM file containing the private key of the server certificate. Required when tls.enable is `true`.
    --tls.offload tls.certheader                    Whether to enable TLS offloading for incoming connections. If enabled tls.certheader must be configured as well.
    --tls.truststorefile string                     PEM file containing the trusted CA certificates for authenticating remote servers. (default "truststore.pem")
```

core/server_config.go
```diff
@@ -197,10 +247,14 @@ func FlagSet() *pflag.FlagSet {
...
 	flagSet.String("loggerformat", "text", "Log format (text, json)")
 	flagSet.String("http.default.address", ":1323", "Address and port the server will be listening to")
+	flagSet.String("http.default.tls", string(DisabledHTTPTLSMode), fmt.Sprintf("Whether to enable TLS for the default interface (options are `%s`, `%s`, `%s`).", DisabledHTTPTLSMode, ServerCertTLSMode, MutualTLSMode))
 	flagSet.Bool("strictmode", false, "When set, insecure settings are forbidden.")
 	flagSet.Bool("internalratelimiter", true, "When set, expensive internal calls are rate-limited to protect the network. Always enabled in strict mode.")
 	flagSet.String("datadir", "./data", "Directory where the node stores its files.")
 	flagSet.StringSlice("http.default.cors.origin", nil, "When set, enables CORS from the specified origins for the on default HTTP interface.")
+	flagSet.String("tls.certfile", "", "PEM file containing the certificate for the server (also used as client certificate). Required when `network.enabletls` is `true`.")
+	flagSet.String("tls.certkeyfile", "", "PEM file containing the private key of the server certificate. Required when `tls.enable` is `true`.")
```
gerardsn, line 256: `tls.enable` does not exist, but probably makes sense to move `network.enabletls` to `tls.enabled`

core/server_config.go
```diff
@@ -90,8 +126,22 @@ type HTTPConfig struct {
...
+// HTTPTLSMode defines the values for TLS modes
+type HTTPTLSMode string
+
+const (
+	// DisabledHTTPTLSMode specifies that TLS is not enabled for this interface.
+	DisabledHTTPTLSMode HTTPTLSMode = "disabled"
+	// ServerCertTLSMode specifies that TLS is enabled for this interface, but no client certificate is required.
+	ServerCertTLSMode = "server-cert"
+	// MutualTLSMode specifies that TLS is enabled for this interface, and that it will require a client certificate.
+	MutualTLSMode = "server-and-client-cert"
```
gerardsn, line 142: these are TLS options, so to me it makes more sense to describe the options in these terms. (disabled / tls / mutual-tls)

docs/pages/deployment/cli-reference.rst
```diff
@@ -129,8 +134,11 @@ Imports private keys from filesystem based storage into Vault. The given directo
       --storage.redis.password string                 Redis database password. If set, it overrides the username in the connection URL.
       --storage.redis.username string                 Redis database username. If set, it overrides the username in the connection URL.
       --strictmode                                    When set, insecure settings are forbidden.
+      --tls.certfile network.enabletls                PEM file containing the certificate for the server (also used as client certificate). Required when network.enabletls is `true`.
       --tls.certheader string                         Name of the HTTP header that will contain the client certificate when TLS is offloaded.
+      --tls.certkeyfile tls.enable                    PEM file containing the private key of the server certificate. Required when tls.enable is `true`.
       --tls.offload tls.certheader                    Whether to enable TLS offloading for incoming connections. If enabled tls.certheader must be configured as well.
+      --tls.truststorefile string                     PEM file containing the trusted CA certificates for authenticating remote servers. (default "truststore.pem")
```
gerardsn, line 141: incorrect flag types

core/server_config.go
gerardsn, line 142:
```suggestion
	// DisabledHTTPTLSMode specifies that TLS is not enabled for this interface.
	DisabledHTTPTLSMode HTTPTLSMode = "disabled"
	// ServerCertTLSMode specifies that TLS is enabled for this interface, but no client certificate is required.
	ServerCertTLSMode HTTPTLSMode = "server-cert"
	// MutualTLSMode specifies that TLS is enabled for this interface, and that it will require a client certificate.
	MutualTLSMode HTTPTLSMode = "server-and-client-cert"
```

core/server_config.go
```diff
@@ -197,10 +247,14 @@ func FlagSet() *pflag.FlagSet {
...
 	flagSet.Bool("strictmode", false, "When set, insecure settings are forbidden.")
 	flagSet.Bool("internalratelimiter", true, "When set, expensive internal calls are rate-limited to protect the network. Always enabled in strict mode.")
 	flagSet.String("datadir", "./data", "Directory where the node stores its files.")
 	flagSet.StringSlice("http.default.cors.origin", nil, "When set, enables CORS from the specified origins for the on default HTTP interface.")
+	flagSet.String("tls.certfile", "", "PEM file containing the certificate for the server (also used as client certificate). Required when `network.enabletls` is `true`.")
+	flagSet.String("tls.certkeyfile", "", "PEM file containing the private key of the server certificate. Required when `tls.enable` is `true`.")
+	flagSet.String("tls.truststorefile", "truststore.pem", "PEM file containing the trusted CA certificates for authenticating remote servers.")
 	flagSet.String("tls.offload", "", "Whether to enable TLS offloading for incoming connections. If enabled `tls.certheader` must be configured as well.")
```
gerardsn, line 258: unrelated to this PR, but `tls.offload` needs a default as well, or at least made clear that it is a bool

gerardsn (changes requested):
Looks good to me. some minor things left

network/network.go
```diff
@@ -264,12 +264,32 @@ func (n *Network) emitEvents(event dag.Event) (bool, error) {
...
-func loadCertificateAndTrustStore(moduleConfig Config) (tls.Certificate, *core.TrustStore, error) {
-	clientCertificate, err := tls.LoadX509KeyPair(moduleConfig.CertFile, moduleConfig.CertKeyFile)
+func loadCertificateAndTrustStore(moduleConfig Config, serverConfig core.ServerConfig) (tls.Certificate, *core.TrustStore, error) {
+	var certFile, certKeyFile, trustStoreFile string
+	if len(moduleConfig.CertFile) > 0 {
+		log.Logger().Warn("Deprecated: use `tls.certfile` instead of `network.certfile`")
+		certFile = moduleConfig.CertFile
+	} else {
+		certFile = serverConfig.TLS.CertFile
+	}
```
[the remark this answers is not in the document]
reinkrul: Why would it be better?

docs/pages/deployment/tls-configuration.rst
```diff
@@ -3,31 +3,66 @@
...
 .. raw:: html
     :file: ../../_static/images/network_layouts_directwan.svg
 
-No additional configuration is required.
+For this setup you need to configure TLS and set up the HTTP interfaces so the endpoints are properly secured.
+The example below shows how to:
+
+* configure TLS for HTTP and gRPC connections,
+* enable TLS (with required client certificate) for node-to-node (``/n2n``) HTTPS connections on port ``8443``,
+* enable TLS (server certificate only) for ``/public`` HTTPS connections on port ``443``.
```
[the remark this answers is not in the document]
reinkrul: No, since the public (aka the IRMA client app) doesn't have a client certificate

docs/pages/deployment/tls-configuration.rst
```diff
@@ -3,31 +3,66 @@
...
+* Connections from the "outside world" (HTTP), e.g. mobile devices, are secured using TLS with only a server certificate.
+
+Your TLS configuration depends mostly on where you `terminate the TLS connection <https://en.wikipedia.org/wiki/TLS_termination_proxy>`_.
+This page describes the different layouts for TLS and how to configure them.
 
-    HTTP connections between nodes (all calls to ``/n2n``) must be secured using TLS which is not handled by the Nuts node.
-    You need to have a reverse proxy in front of the Nuts node for terminating the (node-to-node) HTTPS traffic and forwarding it to the Nuts node.
-    Refer to :ref:`Interfaces/Endpoints <nuts-node-recommended-deployment>` for the requirements on this HTTP endpoint (and others).
+All layouts your node's certificate must be issued by a Certificate Authority which is trusted by the other nodes in the network.
+Each layout requires ``tls.certfile``, ``tls.certkeyfile`` and ``tls.truststorefile`` to be configured.
```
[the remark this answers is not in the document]
reinkrul: That's right, but I don't want people to stimulate disabling TLS. I'll change it to

> Any case using TLS requires ``tls.certfile``, ``tls.certkeyfile`` and ``tls.truststorefile`` to be configured.

core/server_config.go
reinkrul: That's weird, I noticed something earlier but thought it was just me.

core/server_config.go
reinkrul: I just removed the last sentences about TLS being enabled

core/server_config.go
reinkrul: I think "mTLS" is a bad term, since it implicates otherwise TLS is only used one way, correct term would be "TLS with mutual authentication", but that's probably too long. Will settle for your suggestion.

docs/pages/deployment/cli-reference.rst
reinkrul: ?

gerardsn: this is related to the other comment in core/server_config.go

core/server_config.go
gerardsn: It does look a bit weird, but I can live with this. 
Is authentication a better word for the first tls? so `authentication: disabled | tls | mutual-tls`.

network/network.go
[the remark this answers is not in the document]
gerardsn: Looking at it again it is probably fine. It should be obvious that these options should never be declared more than once.

gerardsn (approved):

core/server_config.go
reinkrul: An upcoming change adds authentication, so then we'd get an `auth` field already. I'm not a big fan of my proposal either. Maybe just; `disabled`, `server`, `server-client`?

gerardsn (approved):

PR #1336, opened 2022-08-04
Storage: basic TLS support for Redis

reinkrul:
Allowing a different CA or certificate to be configured is future work. Because at this point, not sure if it's required.


woutslakhorst (changes requested):
What about bypassing config and do an autodetect based on the URL scheme: redis vs rediss:

https://www.iana.org/assignments/uri-schemes/prov/rediss

woutslakhorst (approved):

PR #1339, opened 2022-08-04
expose notifier dlq to metrics/diagnostics/API

### LINKED ISSUE #1260: Network: improve visibility/analysability for unretrievable private TXs
There are be several reasons a node can't retrieve a private TX, including:

- It can't connect to the other node(s) for retrieving the private TX
- It can connect to the other node, but the node won't/can't send the TX
- etc

If a private TX is received but the payload can't be retrieved, an operator would have to find out about it and analyse the issue using the log. But this should be a metric (for alerting) and should be present on the node's diagnostics page. Maybe including the reason why it failed.

In fact; sort of dead letter queue?

woutslakhorst:
closes #1260

Doesn't currently add a Delete API.

### DIFF
--- a/docs/_static/network/v1.yaml
+++ b/docs/_static/network/v1.yaml
@@ -171,8 +171,70 @@ paths:
       responses:
         "202":
           description: "The request was accepted, process was started"
+  /internal/network/v1/events:
+    get:
+      summary: "Lists the state of the internal events"
+      description: >
+        Internal events are used to update network state, retrieve private transactions and update the VDR and VCR.
+        New transactions will cause events to be omitted. In normal operation no events should remain.
+        The diagnostics page displays the number of failed events. If there are any, this API can be used to get the cause.
+        Events are listed by subscriber.
+        
+        error returns:
+        * 500 - internal server error
+      operationId: "ListEvents"
+      tags:
+        - diagnostics
+      responses:
+        "200":
+          description: "Successfully listed the events"
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: '#/components/schemas/EventSubscriber'
+        default:
+          $ref: '../common/error_response.yaml'
 components:
   schemas:
+    Event:
+      type: object
+      description: Non-completed event. An event represents a transaction that is of interest to a specific part of the Nuts node.
+      required:
+        - hash
+        - retries
+        - transaction
+      properties:
+        'type':
+          description: "'transaction' or 'payload'"
+          type: string
+        hash:
+          description: Hash is the ID of the Event, usually the same as the transaction reference.
+          type: string
+        retries:
+          description: Number of times the event has been retried.
+          type: integer
+        transaction:
+          description: The transaction reference
+          type: string
+        error:
+          description: Lists the error if the event processing failed due to an error.
+          type: string
+    EventSubscriber:
+      type: object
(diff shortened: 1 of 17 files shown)


docs/_static/network/v1.yaml
```diff
@@ -171,8 +171,70 @@ paths:
...
+          description: Hash is the ID of the Event, usually the same as the transaction reference.
+          type: string
+        retries:
+          description: Number of times the event has been retried.
+          type: integer
+        transaction:
+          description: The transaction reference
+          type: string
+        error:
+          description: Lists the error if the event processing failed due to an error.
```
reinkrul, line 222:
```suggestion
          description: Lists the last error if the event processing failed due to an error.
```

network/api/v1/api.go
```diff
@@ -124,6 +124,32 @@ func (a Wrapper) RenderGraph(ctx echo.Context, params RenderGraphParams) error {
...
+	for _, notifier := range a.Service.Subscribers() {
+		eventSubscriber := EventSubscriber{
+			Name: notifier.Name(),
+		}
+		events, err := notifier.GetFailedEvents()
+		if err != nil {
+			return err
+		}
+		for _, event := range events {
+			eventError := event.Error
```
reinkrul, line 138: `len(event.Error) == 0` should be passed as `nil` error. Why is eventType optional?

docs/_static/network/v1.yaml
```diff
@@ -171,8 +171,70 @@ paths:
...
+        'type':
+          description: "'transaction' or 'payload'"
+          type: string
+        hash:
+          description: Hash is the ID of the Event, usually the same as the transaction reference.
+          type: string
+        retries:
+          description: Number of times the event has been retried.
+          type: integer
+        transaction:
```
reinkrul, line 218: why both `hash` and `transaction`, since they will be equal? (if `Event` were generic, `transaction` shouldn't be here).

network/dag/state.go
```diff
@@ -93,6 +93,40 @@ func transactionCountCollector() prometheus.Counter {
 	)
 }
 
+func (s *state) initPrometheusCounters() error {
+	s.transactionCount = transactionCountCollector()
+	err := prometheus.Register(s.transactionCount)
+	if err != nil && err.Error() != (prometheus.AlreadyRegisteredError{}).Error() { // No unwrap on prometheus.AlreadyRegisteredError
+		return err
+	}
+	s.eventsNotifyCount = prometheus.NewCounter(
```
reinkrul, line 102: wouldn't it make sense to split this into successfully published events and failed ones?

reinkrul (changes requested):

network/dag/state.go
woutslakhorst: Som events won't error, they just won't finish. The private payloads for example, they won't error, but will retry until finished.

docs/_static/network/v1.yaml
woutslakhorst: It's a DAG event, so transactions are always there. It could be hazardous to set the hash to always be the transaction ref.

network/api/v1/api.go
woutslakhorst: event type already was optional, so left it.

reinkrul (approved):

Issue #1337, opened 2022-08-04
Support TLS offloading via HAProxy


woutslakhorst:
HAProxy can send the client certificate as base64 DER certificate not PEM:

```
http-request set-header X-SSL-Client-Cert          %{+Q}[ssl_c_der,base64]
```

The node currently only supports PEM.

reinkrul:
Test: https://github.com/nuts-foundation/nuts-go-e2e-test/pull/33

PR #1340, opened 2022-08-06
Network: support TLS offloading with DER encoded client certs

reinkrul:
Closes https://github.com/nuts-foundation/nuts-node/issues/1337

### DIFF
--- a/network/transport/grpc/tls_offloading.go
+++ b/network/transport/grpc/tls_offloading.go
@@ -22,6 +22,7 @@ import (
 	"context"
 	"crypto/tls"
 	"crypto/x509"
+	"encoding/base64"
 	"errors"
 	"fmt"
 	"github.com/nuts-foundation/nuts-node/core"
@@ -33,6 +34,7 @@ import (
 	"google.golang.org/grpc/peer"
 	"google.golang.org/grpc/status"
 	"net/url"
+	"strings"
 )
 
 type tlsOffloadingAuthenticator struct {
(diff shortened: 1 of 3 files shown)


gerardsn (approved):

network/transport/grpc/tls_offloading.go
```diff
@@ -78,13 +80,39 @@ func (t *tlsOffloadingAuthenticator) authenticate(serverStream grpc.ServerStream
...
 		return nil, fmt.Errorf("expected exactly 1 client certificate in header, found %d", len(certificates))
 	}
+	return certificates, err
+}
+
+func (t *tlsOffloadingAuthenticator) parseDERCert(data string) ([]*x509.Certificate, error) {
+	bytes, err := base64.StdEncoding.DecodeString(data)
+	if err != nil {
+		return nil, fmt.Errorf("unable to base64 decode client cert header: %w", err)
+	}
```
gerardsn, line 103: the (DER + )base64 encoding comes from the `[ssl_c_der,base64]` in the HAProxy config:

```
backend grpc_servers
         mode http
         http-request set-header X-SSL-CERT %{+Q}[ssl_c_der,base64]
         server node1 nuts_node:5555 check proto h2
```

It is possible to specify other encodings in haproxy.
not sure if this should be more generic or that we can just specify that we expect DER+base64

gerardsn (changes requested):

network/transport/grpc/tls_offloading.go
reinkrul: That's right, I see you could also use hex encoding. I don't think we need to be more generic; base64 is probably the most common occurrence. I'll add it to the documentation.

reinkrul:
> certificate parsing is not specific too grpc right? maybe move logic to `core.ParseCertificate()`

Parsing base64-encoded DER format is (currently) only used in gRPC TLS offloading, so it is specific. When it's used somewhere else as well, then maybe we should move it to `ParseCertificate`. But not at this time.

I'm also a bit weary against too smart/magicy functions (ParseCertificate that tries just whatever format), because it can lead to buggy/slow/wieldy code.

gerardsn (approved):

PR #1351, opened 2022-08-11
Fix link to production deployment docs on landingpage

AikedeJongste:
The link to the production deployment documentation on the landingpage doesn't work. This should fix that.


reinkrul (approved):
Thanks!

Issue #665, opened 2021-12-08
Update JWT creation using builder pattern


woutslakhorst:
Latest update of our JWT library has introduced a builder pattern (https://github.com/nuts-foundation/nuts-node/pull/663).
This might clean up some of our code.

jeroendee:
I will have a look @woutslakhorst / @reinkrul

jeroendee:
Issue can be closed after evaluating the proposed PR.

PR #1360, opened 2022-08-18
refactor(crypto): use jwt.NewBuilder() instead of jwt.New()

jeroendee:
Proposal in order to evaluate if this (builder) way of generating a
JWT is cleaner.

Initial impression is that this way of generating a JWT does not yield
a big benefit. It possibly introduces inconsistent way's of generating
a JWT. Actual builder like method chaining is limited given the claims
map it has to parse.

There are similar calls in oauth_test.go that can be refactored. These
have been ignored given the outcome of this proposal.

Refs: #665


reinkrul:
I agree there's not much benefit, given we set the claims as map.

jeroendee:
As per @reinkrul comment, closing this PR.

PR #1365, opened 2022-08-19
Add monitoring api specs

gerardsn:
This adds the remaining endpoints to the API docs and makes the overview complete

for a preview see [nuts-node.readthedocs.io/en/api-monitoring](https://nuts-node.readthedocs.io/en/api-monitoring/pages/integrating/api.html?urls.primaryName=Monitoring)


docs/_static/monitoring/v1.yaml
```diff
@@ -0,0 +1,65 @@
...
+  /status:
+    get:
+      summary: "healthcheck to confirm node is up and running"
+      description: |
+        Check if the node is up and running.
+      operationId: getStatus
+      tags:
+        - status
+      responses:
+        200:
```
reinkrul, line 20: and otherwise?

docs/_static/monitoring/v1.yaml
```diff
@@ -0,0 +1,65 @@
...
+        * DAG state
+        * node build info
+      operationId: getDiagnostics
+      tags:
+        - status
+      responses:
+        200:
+          description: Basic nuts-node diagnostics
+          content:
+            application/json:
```
reinkrul, line 41: It returns it in YAML, unless `application/json` is specified in the `Accept` header

reinkrul (changes requested):

docs/_static/monitoring/v1.yaml
gerardsn: We only have a 200 set for this (and all other monitoring) endpoints. Not sure what else you would like to see here?

docs/_static/monitoring/v1.yaml
gerardsn: The default `Accept` header only contains `text/plain`. Should this be changed to reflect that it is yaml as well?

reinkrul: - Maybe: respond with `Content-Type: application/yaml` when returned in YAML (instead of text/plain)
- Return YAML content-type when requested with Accept

reinkrul (approved):

PR #1368, opened 2022-08-22
Core: let engines shutdown in reverse order from which they're started

reinkrul:
There's dependencies between engines, so order for `Start()` matters. The same then applies to `Shutdown()`, but then in reverse.

### DIFF
--- a/core/engine.go
+++ b/core/engine.go
@@ -115,15 +115,21 @@ func (system *System) Start() error {
 
 // Shutdown shuts down all engines in the system.
 func (system *System) Shutdown() error {
-	var err error
-	return system.VisitEnginesE(func(engine Engine) error {
+	var engines []Runnable
+	system.VisitEngines(func(engine Engine) {
 		if m, ok := engine.(Runnable); ok {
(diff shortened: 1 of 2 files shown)


core/engine.go
```diff
@@ -115,15 +115,21 @@ func (system *System) Start() error {
...
-			err = m.Shutdown()
-			coreLogger.Infof("Stopped %s", getEngineName(engine))
+			engines = append(engines, m)
 		}
-		return err
 	})
+	for i := len(engines) - 1; i >= 0; i-- {
+		curr := engines[i]
+		coreLogger.Infof("Stopping %s...", getEngineName(curr))
+		if err := curr.Shutdown(); err != nil {
```
gerardsn, line 127: should it ignore this error and shutdown the remaining engines?

reinkrul: Not sure. There's something to say for either approach.

You could say you try to shut it down again after X time, then force shutdown after a timeout. But that makes it probably more complex than required.

When running in a Docker container, if the system does not shut down before the timeout, it's killed anyways.

gerardsn (approved):

PR #1375, opened 2022-08-24
Fix datarace on network.state by not nil-ing it on shutdown

reinkrul:
Peer diagnostics collector is called periodically by v2 which reads it, while there might be a Shutdown call in-flight which sets the field to `nil`, causing a nil deref. Now it can still fail because the databases are closed (which is OK, because the node is shutting down anyways), but at least it's not a nil deref due to a data race.

Most correct fix would wait for the protocols  (v2) and its child goroutines (diagnostics collector) to finish, but that would require a lot more code, increasing complexity.

### DIFF
--- a/network/network.go
+++ b/network/network.go
@@ -562,12 +562,9 @@ func (n *Network) Shutdown() error {
 	n.connectionManager.Stop()
 
 	// Close State and underlying DBs
-	if n.state != nil {
-		err := n.state.Shutdown()
-		if err != nil {
-			return err
-		}
-		n.state = nil
+	err := n.state.Shutdown()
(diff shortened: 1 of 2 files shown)


gerardsn (changes requested):
The comment above needs to be updated as well

network/network.go
```diff
@@ -562,12 +562,9 @@ func (n *Network) Shutdown() error {
 	n.connectionManager.Stop()
 
 	// Close State and underlying DBs
```
reinkrul, line 564:
```suggestion
```

gerardsn (approved):

Issue #1376, opened 2022-08-24
Add support for custom CA for TLS (Redis)


olafz:
From https://github.com/nuts-foundation/nuts-node/pull/1336#issue-1328311110:

> Allowing a different CA or certificate to be configured is future work. Because at this point, not sure if it's required.

It is. Many providers will have an internal Certificate Authority or a non-publicly trusted CA-chain. Currently, it is not possible to have that CA-chain trusted by the Nuts node. This currently requires ugly workarounds or custom builds of the Nuts docker container.

olafz:
I can confirm that the mentioned PR properly adds support for custom CAs:

```
time="2022-08-26T06:59:40Z" level=debug msg="Configured Crypto" module=core
time="2022-08-26T06:59:40Z" level=debug msg="Configuring VCR" module=core
time="2022-08-26T06:59:40Z" level=debug msg="Creating Redis store" module=Storage store=vcr/backup-issued-credentials
time="2022-08-26T06:59:40Z" level=debug msg="Checking connection to Redis database (attempt=1/5, address=a.b.c.d:1234)..."
time="2022-08-26T06:59:40Z" level=debug msg="Connection check successful"
time="2022-08-26T06:59:40Z" level=debug msg="Configured VCR" module=core
time="2022-08-26T06:59:40Z" level=debug msg="Configuring Network" module=core
time="2022-08-26T06:59:40Z" level=debug msg="Creating Redis store" module=Storage store=network/data
time="2022-08-26T06:59:40Z" level=debug msg="Checking connection to Redis database (attempt=1/5, address=a.b.c.d:1234)..."
time="2022-08-26T06:59:40Z" level=debug msg="Connection check successful"
```

PR #1378, opened 2022-08-25
Storage: support configuring a truststore when connecting to a Redis server over TLS

reinkrul:
Closes https://github.com/nuts-foundation/nuts-node/issues/1376


gerardsn (approved):
Seems to work. Maybe confirm with @olafz that this solves the issue

PR #1380, opened 2022-08-26
Security: validate key IDs before passing them to storage

reinkrul:
Prevents attackers to read/write outside of intended key storage (other Vault engines or directories on disk). And prevents accidents with key naming.

Fixes https://github.com/nuts-foundation/nuts-node/security/code-scanning/19

### DIFF
--- a/crypto/crypto.go
+++ b/crypto/crypto.go
@@ -29,12 +29,15 @@ import (
 	"github.com/nuts-foundation/nuts-node/crypto/log"
 	"github.com/nuts-foundation/nuts-node/crypto/storage"
 	"path"
+	"regexp"
 )
 
 const (
 	// ModuleName contains the name of this module
 	ModuleName = "Crypto"
 )
+
+var kidPattern = regexp.MustCompile(`^[\da-zA-Z_\- :#.]+$`)
 
 // Config holds the values for the crypto engine
 type Config struct {
@@ -119,6 +122,9 @@ func (client *Crypto) New(namingFunc KIDNamingFunc) (Key, error) {
 	if err != nil {
 		return nil, err
 	}
+	if err = validateKID(kid); err != nil {
+		return nil, err
+	}
 	if err = client.Storage.SavePrivateKey(kid, keyPair); err != nil {
 		return nil, fmt.Errorf("could not create new keypair: could not save private key: %w", err)
 	}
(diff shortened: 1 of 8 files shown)


crypto/crypto_test.go
```diff
@@ -156,3 +176,18 @@ func createCrypto(t *testing.T) *Crypto {
...
+		assert.NoError(t, validateKID("did:nuts:3dGjPPeEuHsyNMgJwHkGX3HuJkEEnZ8H19qBqTaqLDbt#JwIR4Vct-EELNKeeB0BZ8Uff_rCZIrOhoiyp5LDFl68"))
+		assert.NoError(t, validateKID("did:nuts:BC5MtUzAncmfuGejPFGEgM2k8UfrKZVbbGyFeoG9JEEn#l2swLI0wus8gnzbI3sQaaiE7Yvv2qOUioaIZ8y_JZXs"))
+	})
+	t.Run("bad KIDs", func(t *testing.T) {
+		assert.Error(t, validateKID("../server-certificate"))
+		assert.Error(t, validateKID("\\"))
+		assert.Error(t, validateKID(""))
+		assert.Error(t, validateKID("\t"))
+	})
+}
```
gerardsn, line 193: don't need to change this, but perfect usecase for table driven tests

crypto/crypto.go
```diff
@@ -29,13 +29,16 @@ import (
...
 	"path"
+	"regexp"
 )
 
 const (
 	// ModuleName contains the name of this module
 	ModuleName = "Crypto"
 )
 
+var kidPattern = regexp.MustCompile(`^[\da-zA-Z_\- :#.]+$`)
```
gerardsn, line 40: package already depends on `nuts-node/crypto/storage`. Export the pattern there and use it here? (Or export from somewhere else to prevent double declaration.)

crypto/storage/fs.go
```diff
@@ -139,8 +152,11 @@ func (fsc fileSystemBackend) readEntry(kid string, entryType entryType) ([]byte,
 	return data, nil
 }
 
-func (fsc fileSystemBackend) getEntryPath(key string, entryType entryType) string {
-	return filepath.Join(fsc.fspath, getEntryFileName(key, entryType))
+func (fsc fileSystemBackend) getEntryPath(key string, entryType entryType) (string, error) {
```
gerardsn, line 155: error case is untested

gerardsn (changes requested):

crypto/crypto_test.go
reinkrul: I think Go's data-driven test setup is quite unreadable and verbose, but that might just be me.

crypto/crypto.go
reinkrul: I first implemented in the filesystem backend, then I realized it applies to Vault as well, so I copied it to there. Forgot to remove it here.

crypto/crypto_test.go
gerardsn: I agree. This is the first time I spot a test where I think it makes sense

crypto/crypto.go
gerardsn: This still needs to happen right?

gerardsn (approved):

Issue #1377, opened 2022-08-25
CLI Reference lists too many options for cmds that use server config


reinkrul:
E.g. `crypto fs2vault` or `http gen-token`

PR #1381, opened 2022-08-26
Docs: don't repeat server flags

reinkrul:
Closes https://github.com/nuts-foundation/nuts-node/issues/1377


docs/pages/deployment/cli-reference.rst
```diff
@@ -1,29 +1,18 @@
 .. _nuts-cli-reference:
 
 CLI Command Reference
-*********************
+#####################
 
+There are 2 types of commands: server command and client commands. Server commands (e.g. ``nuts server``) can only be run on the system where the node is (or will be) running, because they require the node's config. Client commands are used to remotely administer a Nuts node and require the node's API address.
```
gerardsn, line 6: Most of these flags don't do anything for `nuts crypt fs2vault` and the `nuts http gen-token` right? To me it is still not clear what flags are required to successfully use these commands.
This does improve readability though.

reinkrul: The current commands of the engines,  that use the server config, don't have additional flags, other than those here. If we introduce a command that does, it will not show up in these docs. That's an issue, but to be solved when we need it.

gerardsn (approved):
I ment to approve this the first time

reinkrul:
> I ment to approve this the first time

No hard feelings ;-)

Issue #1357, opened 2022-08-17
ldutils_test.go reads (HTTP) external JSON-LD schemas during test


reinkrul:
Tests should not use external resources because it makes test flaky. E.g. when the external resources goes away/relocates, is slow, or the local network has a hickup.

Replace with a local webserver.

jeroendee:
Will have a look.

reinkrul:
> Will have a look.

The test is actually `ldutils_test.go` on line 146.

jeroendee:
h3. What
In the unittest _"it resolves an external doc when allowingExternalCalls is true"_ in `ldutils_test.go` (on line 146) don't depend on an external call being made. 

h3. Why
During unit-testing depending on external calls being made increases the chance of tests becoming flaky.  

h3. How
Implement a test-double, in this case a stub, through the usage of the httptest package.

PR #1383, opened 2022-08-26
Feature/1357 use test double serving json ld

jeroendee:
> By introducing a server stub the test is not dependent on an external
resource being available (= out of our control). By intercepting the
(real) request being made, we take control over the response outcome.
The underlying test behavior/goal is unchanged.

> Side effect, the execution time has improved by a factor of 40.
400 ms => 10 ms.

Fixes: #1357


jsonld/ldutils_test.go
```diff
@@ -146,9 +149,39 @@ func TestNewContextLoader(t *testing.T) {
...
+
+		// Arrange the stubSrv; stubSrv logic is included inline,
+		// it makes the test story telling more clear.
+		//
+		// Stub behavior is based on the code behavior when processing:
+		// loader.LoadDocument("http://schema.org")
+		mux := http.NewServeMux()
+
+		// Stage 1 of the stub
+		f1 := func(w http.ResponseWriter, r *http.Request) {
```
reinkrul, line 161: `f1` and `f2` could use a more descriptive name, something like `rootHandler` and `schemaHandler` (since they're handler functions)

reinkrul (changes requested):
Test seems OK, some naming improvements

jsonld/ldutils_test.go
jeroendee: @reinkrul removed the usage of those variables all together.

reinkrul (approved):

reinkrul:
🚀

PR #1411, opened 2022-09-02
gRPC sets ip based on X-Forwarded-For header if available

### LINKED ISSUE #1389: Network: gRPC should respect X-Forwarded-For header (or similar)
When logging HTTP calls Echo checks the `X-Forwarded-For` header in case traffic is forwarded by a proxy. The should by applied to gRPC traffic.

gerardsn:
closes #1389

Behavior matches that of the echo handler
Will still produce IP of the proxy when configured for level 4 forwarding

### DIFF
--- a/network/transport/grpc/ip_interceptor.go
+++ b/network/transport/grpc/ip_interceptor.go
@@ -2,6 +2,7 @@ package grpc
 
 import (
 	"net"
+	"net/netip"
 	"strings"
 
 	"google.golang.org/grpc"
@@ -65,11 +66,14 @@ func addrFrom(ip string) net.Addr {
 	if ip == "" {
 		return nil
 	}
-	addr, err := net.ResolveIPAddr("ip", ip)
+	ipAddr, err := netip.ParseAddr(ip)
 	if err != nil {
 		return nil
 	}
-	return addr
+	return &net.IPAddr{
+		IP:   ipAddr.AsSlice(),
+		Zone: ipAddr.Zone(),
+	}
 }
 
 func isInternal(ip net.IP) bool {
--- a/network/transport/grpc/ip_interceptor.go
+++ b/network/transport/grpc/ip_interceptor.go
@@ -1,0 +1,65 @@
+package grpc
+
+import (
+	"net"
+	"strings"
+
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/metadata"
+	"google.golang.org/grpc/peer"
+)
+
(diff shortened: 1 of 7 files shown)


network/transport/grpc/ip_interceptor.go
```diff
@@ -0,0 +1,77 @@
...
+	"google.golang.org/grpc/metadata"
+	"google.golang.org/grpc/peer"
+)
+
+const headerXForwardedFor = "X-Forwarded-For"
+
+// ipInterceptor tries to extract the IP from the X-Forwarded-For header and sets this as the peers address.
+// No address is set if the header is unavailable.
+func ipInterceptor(srv interface{}, serverStream grpc.ServerStream, _ *grpc.StreamServerInfo, handler grpc.StreamHandler) error {
+	addr := addrFrom(extractIPFromXFFHeader(serverStream))
```
reinkrul, line 17: I think this should only be done when TLS is offloaded. Because then, the reverse proxy is expected to set this header. Otherwise, an attacker could set it on the HTTP request, spoofing its IP address.

network/transport/grpc/ip_interceptor.go
```diff
@@ -0,0 +1,77 @@
...
+	// All IPs are trusted; return first element because it is furthest from server (best effort strategy).
+	return strings.TrimSpace(ips[0])
+}
+
+// addrFrom returns nil if ip is not a valid IP
+func addrFrom(ip string) net.Addr {
+	if ip == "" {
+		return nil
+	}
+	addr, err := net.ResolveIPAddr("ip", ip)
```
reinkrul, line 68: I'd not resolve it, since it could involve (slow) DNS queries when passed a hostname. I'd say just use

```
netip.ParseAddr()
```

When you want a `net.Addr` from a IP address as string.

Update: `extractIPFromXFFHeader` already parses the IP and then returns it as string, then you parse it again. Probably just return it there as parsed IP.

reinkrul:
Maybe an integration test that asserts the IP returned from diagnostics (or `Peers()`).

reinkrul (changes requested):

network/transport/grpc/ip_interceptor.go
gerardsn: I agree, but this matches the current behavior on the http endpoints. We should change both at the same time.

gerardsn: created issue #1416 for this

reinkrul (approved):

PR #1412, opened 2022-09-05
Network: validate discovered NutsComm address

### LINKED ISSUE #1385: Network: do not connect to invalid/weird discovered addresses
On the stable network, someone registered a localhost address as `NutsComm` address. This causes all nodes to connect to their own node, which will never succeed and leads to odd logging.

Do not connect to discovered addresses that are:

- localhost
- not hostnames (e.g. IP addresses)
- reversed TLDs (https://www.ietf.org/archive/id/draft-chapin-rfc2606bis-00.html)

Should be able to be disabled, but not in strict mode.

Though these should still be allowed in bootstrap config, since they aid cluster-like setup (e.g. connecting to an internal bootstrap node).

gerardsn:
closes #1385

Service discovery no longer accepts NutsComm endpoints that consist of reserved TLD's / addresses, or IP addresses.
Behavior of bootstrap nodes is unchanged.

- (How) Should this be documented?

### DIFF
--- a/network/transport/types.go
+++ b/network/transport/types.go
@@ -21,12 +21,14 @@ package transport
 import (
 	"errors"
 	"fmt"
-	"github.com/nuts-foundation/nuts-node/core"
-	"github.com/sirupsen/logrus"
+	"net"
 	"net/url"
+	"strings"
 	"time"
 
 	"github.com/nuts-foundation/go-did/did"
+	"github.com/nuts-foundation/nuts-node/core"
+	"github.com/sirupsen/logrus"
 )
 
 // PeerID defines a peer's unique identifier.
@@ -35,18 +37,6 @@ type PeerID string
 // String returns the PeerID as string.
 func (p PeerID) String() string {
 	return string(p)
-}
-
-// ParseAddress parses the given input string to a gRPC target address. The input must include the protocol scheme (e.g. grpc://).
-func ParseAddress(input string) (string, error) {
-	parsed, err := url.Parse(input)
-	if err != nil {
-		return "", err
-	}
-	if parsed.Scheme != "grpc" {
-		return "", errors.New("invalid URL scheme")
-	}
-	return parsed.Host, nil
 }
 
 // Peer holds the properties of a remote node we're connected to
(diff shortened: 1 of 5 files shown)


network/transport/types.go
```diff
@@ -106,3 +96,59 @@ type ConnectorStats struct {
...
+	if isReserved(parsed) {
+		return "", errors.New("hostname is reserved")
+	}
+	return parsed.Host, nil
+}
+
+// isReserved returns true if URL uses any of the reserved TLDs or addresses
+func isReserved(URL *url.URL) bool {
+	parts := strings.Split(strings.ToLower(URL.Hostname()), ".")
+	if len(parts) < 2 {
```
reinkrul, line 124: this could break docker-compose setups. Maybe only disallow in strict mode?

reinkrul (changes requested):

network/transport/types.go
gerardsn: This is only used in NutsComm service discovery, bootstrap nodes do not check for this.
- Reserved addresses will generally not resolve to anything, and when they do (like localhost), the nodeDIDs will not match and probably cause some weird hard to debug error messages.
- Internal endpoints should never be published publicly?

gerardsn:
We currently recommend to register care organization NutsComm endpoint as
```
POST <internal-node-address>/internal/didman/v1/did/<care-organization-did>/endpoint
{
    "type": "NutsComm",
    "endpoint": "<vendor-did>/serviceEndpoint?type=NutsComm"
}
```
Should we resolve the vendor and verify that we are connected to it? Can we do anything meaningful it not?

woutslakhorst:
> We currently recommend to register care organization NutsComm endpoint as
> 
> ```
> POST <internal-node-address>/internal/didman/v1/did/<care-organization-did>/endpoint
> {
>     "type": "NutsComm",
>     "endpoint": "<vendor-did>/serviceEndpoint?type=NutsComm"
> }
> ```
> 
> Should we resolve the vendor and verify that we are connected to it? Can we do anything meaningful it not?

Out of scope for this PR I guess

woutslakhorst:
> * (How) Should this be documented?

Add a **note** here: https://nuts-node.readthedocs.io/en/stable/pages/getting-started/3-configure-your-node.html#node-identity

woutslakhorst (approved):

network/transport/types.go
reinkrul: The private-transactions end-to-end test registers a Docker service+port (`nodeA:5555`) as NutsComm address in `prepare.sh`

We could always rename the node in the test, or add a network link so we can use something like `nodeA.docker` which isn't disallowed.

gerardsn: E2E tests currently don't fail because the node is bootstrapped. 
For Node Discovery to work there needs to be an intermediary node that connects nodes that are not directly connected, so we need a minimum of 3 nodes. We can extend this e2e test with a third node, or remove the part that publishes the endpoint.

reinkrul (approved):

PR #1417, opened 2022-09-06
Docs: Updated HAProxy pass-through config

gerardsn:
A single proxy config should only contain `listen`, or a combination of `frontend` and `backend`

### DIFF
--- a/docs/pages/deployment/tls-configuration.rst
+++ b/docs/pages/deployment/tls-configuration.rst
@@ -78,19 +78,19 @@ Configuration for `HAProxy <https://www.haproxy.com/>`_ could look like this (gi
 
 .. code-block::
 
-    default
-        mode tcp
-
     listen grpc
         bind *:5555
+        mode tcp
         server node1 nodeA-backend:5555 check
 
     listen public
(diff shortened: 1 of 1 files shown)


docs/pages/deployment/tls-configuration.rst
```diff
@@ -72,37 +72,25 @@ When using a (level 4) load balancer that does not inspect or alter requests, TL
...
-        mode tcp
-        use_backend nuts_node_public_https
-
-    backend nuts_node_public_https
-        mode tcp
         server node1 nodeA-backend:443 check
 
-    listen n2n_https
+    listen n2n
         bind *:8443
```
reinkrul, line 93: mode tcp isn't required?

reinkrul (approved):

docs/pages/deployment/tls-configuration.rst
gerardsn:
```
    default
         mode tcp
```
sets everything to tcp. But this is not any shorter than just adding `mode tcp` to all listeners, so I will change it back.

PR #1426, opened 2022-09-08
Docs: update nginx tls offload config

### LINKED ISSUE #1392: Document nginx settings te disable buffering and max message length
max message length to something very large or 0
buffering off

gerardsn:
closes #1392?

I could not reproduce issues with max message size, so I have only added the supposed solution as a note.

The current config didn't work for me, so I changed some things to make it a copy-paste solution to get started with NGINX

### DIFF
--- a/docs/pages/deployment/tls-configuration.rst
+++ b/docs/pages/deployment/tls-configuration.rst
@@ -161,8 +161,6 @@ For `NGINX <https://www.nginx.com/>`_ the proxy configuration could look as foll
 
           location / {
             grpc_pass grpc://nuts-node:5555;
-            proxy_buffering off; # immediately forward all traffic to node
-            grpc_buffer_size 128M; # prevents node syncing issues
             grpc_set_header X-SSL-CERT $ssl_client_escaped_cert;        # add peer's SSL cert
             grpc_set_header X-Forwarded-For $proxy_add_x_forwarded_for; # for correct IP logging
           }
--- a/docs/pages/deployment/tls-configuration.rst
+++ b/docs/pages/deployment/tls-configuration.rst
@@ -145,10 +145,6 @@ For `NGINX <https://www.nginx.com/>`_ the proxy configuration could look as foll
 
 .. code-block::
 
-    events {
-        worker_connections  1024;
-    }
-
     http {
         server {
           server_name nuts-grpc;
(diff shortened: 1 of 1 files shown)


docs/pages/deployment/tls-configuration.rst
```diff
@@ -145,58 +145,60 @@ For `NGINX <https://www.nginx.com/>`_ the proxy configuration could look as foll
...
-      ssl_verify_client         on;
-      ssl_verify_depth          1;
-
-      location / {
-        grpc_pass grpc://nuts-node-grpc;
-        grpc_set_header X-SSL-CERT $ssl_client_escaped_cert;
-        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for; // for correct IP logging
-      }
+    events {
+        worker_connections  1024;
```
reinkrul, line 149: not sure if we should include this. I think it's default? But there's more defaults, like log files. We should probably only show the stuff here that's relevant for TLS configuration? Otherwise it can get quite elaborate.

docs/pages/deployment/tls-configuration.rst
```diff
@@ -145,58 +145,60 @@ For `NGINX <https://www.nginx.com/>`_ the proxy configuration could look as foll
...
-      ssl_certificate_key       /etc/nginx/ssl/key.pem;
+.. note::
 
-      location / {
-        proxy_pass http://nuts-node-http-public
-        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for; // for correct IP logging
-      }
-    }
+    During synchronization of a new Nuts node it is possible that the gRPC stream contains messages larger than NGINX is willing to accept.
+    To limit these issues add ``proxy_buffering off`` to the grpc config section, and increase the buffer size to any reasonably large number by adding ``grpc_buffer_size 128M``.
```
reinkrul, line 201: I think it has to be much, much larger than 128mb? At least from what @woutslakhorst read.

reinkrul (changes requested):

docs/pages/deployment/tls-configuration.rst
gerardsn: I got an error message that the events section was missing when leaving this out.

Probably fine to leave it out though.

reinkrul (approved):

PR #1433, opened 2022-09-13
Docs: correct NGINX config flag

gerardsn:
Added the wrong flag in previous PR since I was not able to reproduce the issue.


reinkrul:
`grpc_buffer_size` also exists, but is the buffer size of the server response (local Nuts node). Should this be enlarged as well?

reinkrul (commented):
`grpc_buffer_size` also exists, but is the buffer size of the server response (local Nuts node). Should this be enlarged as well?

gerardsn:
`client_max_body_size` is the one mentioned in the original issue and verified to fix the problem (at least partially). I prefer not to add settings without being able to verify their effects. We can try to reproduce the issue, or request help from users to create a more robust config.

reinkrul (approved):

Issue #1432, opened 2022-09-12
Add VCR document count to diagnostics


woutslakhorst:
Use go-leia v3.2.0

This will allow us to test/debug the VCR recreate command.

PR #1438, opened 2022-09-13
VCR: Add stats on credentials in store

reinkrul:
Closes https://github.com/nuts-foundation/nuts-node/issues/1432

E.g.:

```yaml
vcr:
    credential_count: 7
    issuer:
        issued_credentials_count: 0
        revoked_credentials_count: 0
    verifier:
        revocations_count: 18
```

### DIFF
--- a/docs/pages/deployment/monitoring.rst
+++ b/docs/pages/deployment/monitoring.rst
@@ -46,6 +46,13 @@ Returns the status of the various services in ``yaml`` format:
             dag_xor: 6aada4464e380db16d0316e597956fcdaeada0e8f6023be82eeb9c798e1815c6
             stored_database_size_bytes: 106496005
             transaction_count: 9001
+    vcr:
+        credential_count: 7
+        issuer:
+            issued_credentials_count: 0
+            revoked_credentials_count: 0
+        verifier:
+            revocations_count: 18
     status:
         git_commit: d36837bae48b780bfb76134e85b506472fc207a6
         os_arch: linux/amd64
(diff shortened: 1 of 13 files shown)


vcr/vcr.go
```diff
@@ -404,3 +404,29 @@ func (c *vcr) Untrusted(credentialType ssi.URI) ([]ssi.URI, error) {
...
+			Title: "issuer",
+			Items: c.issuerStore.Diagnostics(),
+		},
+		core.DiagnosticResultMap{
+			Title: "verifier",
+			Items: c.verifierStore.Diagnostics(),
+		},
+		core.GenericDiagnosticResult{
+			Title:   "credential_count",
+			Outcome: credentialCount,
```
gerardsn, line 429: What exactly are these?

Is the last one the 'holder'?

gerardsn (commented):
I don't think I know what all the fields are based on these names. 
Can they be split based on 'my/node credentials' and 'network/all credentials'?

Should any of these also be prometheus metrics?

vcr/vcr.go
reinkrul: @woutslakhorst ?

woutslakhorst:
Descriptions can be added to docs.

vcr/vcr.go
woutslakhorst: `credential_count` is global + holder
`verifier` holds globally revoked

woutslakhorst (changes requested):
Some info in the docs: monitoring and maybe backup/restore?

woutslakhorst:
metrics are needed when you want to turn the number into a graph... Can't think of theause-case for that.

reinkrul:
> metrics are needed when you want to turn the number into a graph... Can't think of a use-case for that.

Not required for metrics I think

docs/pages/deployment/monitoring.rst
```diff
@@ -54,6 +61,13 @@ Returns the status of the various services in ``yaml`` format:
 
 If you supply ``application/json`` for the ``Accept`` HTTP header it will return the diagnostics in JSON format.
 
+Explanation of ambiguous/complex entries in the diagnostics:
```
gerardsn, line 64: Should value of -1 be explained?

gerardsn (approved):
I'd prefer some more self-explanatory naming like: (if correct of course)
```
    vcr:
         credential_count: 7
         revocation_count: 18
         issued:
             credential_count: 0
             revocation_count: 0
```
but that doesn't change the content.

docs/pages/deployment/monitoring.rst
reinkrul: Not really I think, there will be an error in the log

woutslakhorst (approved):

PR #1442, opened 2022-09-13
Docs: Mark network flags deprecated

gerardsn:
Marking the flags as deprecated removes them from generated documentation, and produces the following warning on the command line:
```
/ # nuts config --network.maxcrlvaliditydays 10
Flag --network.maxcrlvaliditydays has been deprecated, use 'tls.crl.maxvaliditydays' instead
```


reinkrul (approved):
Nice, didn't know about that

PR #1453, opened 2022-09-15
added V2 context

### LINKED ISSUE #1452: Release new Nuts JSON-LD context (v2) that specifies legalBase

woutslakhorst:
Closes #1452 

Example expanded form: https://tinyurl.com/2gwplpnx

backport to V4 (regenerate readme)

set as default context for V5


vcr/assets/assets/contexts/nuts-v2.ldjson
```diff
@@ -0,0 +1,97 @@
...
+            "userContext": "nuts:userContext"
+          }
+        },
+        "localParameters": {"@id": "nuts:localParameters", "@container": "@graph"},
+        "legalBase": {
+          "@id":"nuts:legalBase",
+          "@context": {
+            "@version": 1.1,
+            "@protected": true,
+            "consentType": "nuts:consentType",
```
reinkrul, line 83: JSON-LD doesn't support enumerations?

reinkrul (approved):
I'm wondering whether this is the way to work with schema updates. Just some thoughts: another approach would be
-  to have a separate "legal base" context, then you'd use both when creating a Nuts Auth credential
- have v2 extend v1?

vcr/assets/assets/contexts/nuts-v2.ldjson
woutslakhorst: That would require a json schema I think

woutslakhorst:
> I'm wondering whether this is the way to work with schema updates. Just some thoughts: another approach would be
> 
> * to have a separate "legal base" context, then you'd use both when creating a Nuts Auth credential
This would create a "global" definition for legal base instead of a nested only context. I don't think it's our role to define the world wide meaning of legal base? I couldn't find anything matching our case on schema.org 
> * have v2 extend v1?
No longer possible due to version 1 having `@protected: true`. We could make this version not protected, but that would enable people creating VCs with a nested context where they redefine certain definitions. Feature or something to avoid?

woutslakhorst:
@stevenvegt added test

stevenvegt (approved):

PR #1457, opened 2022-09-16
change time in tests to 1970 to avoid weird timezone in 1900

woutslakhorst:
apparently, dutch timezone in 1900 has changed from +00:19:32 to +00:17:xx


stevenvegt:
Under what conditions did these fail?

reinkrul:
Good find

reinkrul (approved):

Issue #1467, opened 2022-09-19
VCR: it's not possible to create a VC with a different context


woutslakhorst:
The defaultBuilder set it to the VC and Nuts context.

woutslakhorst:
fixed

PR #1468, opened 2022-09-19
Allow VCs with less than 1 type

### LINKED ISSUE #1448: VCR: Allow received VCs with only VerifiableCredential as type
Stable network contains VCs that contain only "VerifiableCredential" as type, missing e.g. "NutsAuthCredential". This currently causes an error, which keeps popping up every time the VC is retried.

The RFC states that the VC must contain "VerifiableCredential" and _can_ have 1 additional type. So; the Nuts node MUST accept VCs that only specify "VerifiableCredential" as type.

Prerequisite is that the expanded version of the NutsAuthorizationCredential is indexed/stored?

woutslakhorst:
Closes #1448

This still not allows for other VCs with different context: #1467


vcr/credential/resolver.go
```diff
@@ -39,7 +39,7 @@ func FindValidatorAndBuilder(credential vc.VerifiableCredential) (Validator, Bui
 			}
 		}
 	}
-	return defaultCredentialValidator{}, nil
+	return defaultCredentialValidator{}, defaultBuilder{vcType: vc.VerifiableCredentialType}
```
stevenvegt, line 42: Fix the function description:
```// When no additional type is provided, it returns the default validator and a nil builder```

stevenvegt: Looks like the defaultBuilder is completely unused except for tests. All the return values of the `Builder` type of the FindValidatorAndBuilder method are ignored as well.

stevenvegt (changes requested):

stevenvegt:
I saw the Builder interface was not used anymore so I created a PR on this branch which removes it: https://github.com/nuts-foundation/nuts-node/pull/1471

stevenvegt (approved):

PR #1471, opened 2022-09-20
Remove VC Builder interface

stevenvegt:
The Builder interface and defaultBuilder are not used except in their test. I think they can be safely removed.


woutslakhorst (approved):
I'm shocked

PR #1465, opened 2022-09-19
fix overriding VAULT_TOKEN with empty default

woutslakhorst:
fixes #1464

PR #1474, opened 2022-09-21
fix overriding VAULT_TOKEN with empty default (#1465)

woutslakhorst:
backport of #1465 to V4


gerardsn (approved):
E2E are all good when run manually

PR #1482, opened 2022-09-26
only add default VC context when missing

woutslakhorst:
This is a small fix to prevent duplicate VC context when set within the POST body


gerardsn (approved):
Should a missing context also log a warning to let users know that they should specify the context? (could make the transition to v2 a bit easier)

woutslakhorst:
> Should a missing context also log a warning to let users know that they should specify the context? (could make the transition to v2 a bit easier)

Maybe in a later version (V5.1 or V6), when we started migrating the API to a "no defaults" API

PR #1484, opened 2022-09-26
Feature/879 disallow updating deactivated doc

### LINKED ISSUE #879: Mark resurrected DID Documents as such and show them in diagnostics
Consider the following:

- Given A and B are updates of the same DID Document in parallel
- A is a deactivation and B is a regular update. Due to the parallelism, the deactivation must not block the update. It'll be a conflict but manageable.
- C is a TX that deactivates the controller of a DID Document.
- D is a DID Document update that uses the previous of C as controller/signing key.
- D would be a legit transaction because source TX will point to a valid TX, just not the latest one....
- D may not be blocked by C since that would introduce another parallel attack case.

D does not currently point to the previous transaction of the same DID Document when another controller is used. It's therefore not possible to check if the previous TX of the DID Document was a deactivation. It's therefore a possibility to resurrect deactivated DID Documents without notice.

woutslakhorst:
closes #879 

conflict resolution has added the requirement a DID Document update MUST always point to all its previous versions. This makes detection of reactivation of a deactivated document easier. The conflicted count now also covers resurrection.

### DIFF
--- a/vdr/types/common.go
+++ b/vdr/types/common.go
@@ -39,7 +39,7 @@ var ErrDIDNotManagedByThisNode = errors.New("DID document not managed by this no
 // ErrNotFound The DID resolver was unable to find the DID document resulting from this resolution request.
 var ErrNotFound = errors.New("unable to find the DID document")
 
-// ErrDeactivated The DID document has been deactivated.
+// ErrDeactivated signals rejection due to document deactivation.
 var ErrDeactivated = deactivatedError{msg: "the DID document has been deactivated"}
 
 // ErrNoActiveController The DID supplied to the DID resolution does not have any active controllers.


gerardsn (commented):
Doesn't this allow a timing attack?

Eve sends
- Deactivation to Alice
- DID (key) update + some other transaction referencing this update to Bob

Alice can never accept the transaction Bob has accepted

woutslakhorst:
> Doesn't this allow a timing attack?
> 
> Eve sends
> 
> * Deactivation to Alice
> * DID (key) update + some other transaction referencing this update to Bob
> 
> Alice can never accept the transaction Bob has accepted

This would create a conflict. That's why the check also looks at the length of the sourceTransactions

vdr/ambassador.go
```diff
@@ -320,6 +320,11 @@ func (n *ambassador) handleUpdateDIDDocument(transaction dag.Transaction, propos
 	// Stable order for metadata.SourceTransactions (derived from unordered maps): makes it easier to analyse and test.
 	sortHashes(sourceTransactions)
 
+	if len(sourceTransactions) == 1 && currentDIDMeta.Deactivated {
```
reinkrul, line 323: Why the check for len == 1?

woutslakhorst: See previous discussion. If length > 1 then there's a conflict which is "allowed"

gerardsn (approved):

reinkrul (approved):

PR #1485, opened 2022-09-27
fix staticcheck issues

gerardsn:
This is the first step towards adding [staticcheck](https://staticcheck.io) linter to the CI pipeline

### DIFF
--- a/auth/services/x509/x509_validator_test_utils_test.go
+++ b/auth/services/x509/x509_validator_test_utils_test.go
@@ -19,7 +19,6 @@
 package x509
 
 import (
-	"crypto"
 	"crypto/rand"
 	"crypto/rsa"
 	"crypto/x509"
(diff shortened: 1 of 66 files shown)


network/transport/v2/senders.go
```diff
@@ -91,7 +91,6 @@ func (p *protocol) sendTransactionListQuery(id transport.PeerID, refs []hash.SHA
 			Debug("Did not request a TransactionList while another conversation is in progress")
 		return nil
 	}
-	conversation.set("refs", refs)
```
woutslakhorst, line 94: We do not check these anymore?

network/transport/grpc/outbound_connector.go
```diff
@@ -150,7 +150,7 @@ func (c *outboundConnector) tryConnect() (*grpcLib.ClientConn, error) {
 	if c.tlsConfig != nil {
 		dialOptions = append(dialOptions, grpcLib.WithTransportCredentials(credentials.NewTLS(c.tlsConfig))) // TLS authentication
 	} else {
-		dialOptions = append(dialOptions, grpcLib.WithInsecure()) // No TLS, requires 'insecure' flag
+		dialOptions = append(dialOptions, grpcLib.WithTransportCredentials(insecure.NewCredentials())) // No TLS, requires 'insecure' flag
```
woutslakhorst, line 153: What is the reason for this change? Deprecation?

woutslakhorst (commented):
Are there any instances of the 'append' fix that could benefit from a `copy`? Making sure the original slice is not returned?

network/transport/grpc/outbound_connector.go
gerardsn: correct, this is the recommended replacement

network/transport/v2/senders.go
gerardsn: We do, but directly on the refs in the TransactionListQuery message that created conversation. The additional properties were unused/unchecked.

auth/services/x509/x509_validator_test_utils_test.go
```diff
@@ -124,28 +123,3 @@ func createLeafCert(parent *x509.Certificate, caKey *rsa.PrivateKey) (*x509.Cert
 	cert, err := createTestCert(parent, template, &priv.PublicKey, caKey)
 	return cert, priv, err
 }
-
-func createCrl(currentCrl *x509.RevocationList, issuer *x509.Certificate, priv crypto.Signer, certsToRevoke []*x509.Certificate) ([]byte, error) {
```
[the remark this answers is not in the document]
gerardsn: Good question. It looks like Crls are now handled in the `crl` package and this is package is used in the `x509_validator_test.go`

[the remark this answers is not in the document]
gerardsn: I'm not familiar enough with the topic to determine the quality of testing

gerardsn:
> Are there any instances of the 'append' fix that could benefit from a `copy`? Making sure the original slice is not returned?

made some adjustments to make sure new slices are returned where needed.

`b := make([]newSlice, len(a)); copy(b, a)` and `c := append([]newSlice, a...)` are both new slices that only differ in the capacity of the slice. https://go.dev/play/p/C2nPQFflsM2

woutslakhorst (approved):

Issue #1449, opened 2022-09-15
VCR: Do not retry VCs that are semantically/syntactically incorrect


reinkrul:
When VCs can never be parsed, because e.g., the format is invalid (not JSON, or invalid JSON-LD), the scheduler should not retry it as it currently does. And probably log a warning.

Follow up on https://github.com/nuts-foundation/nuts-node/issues/1423

woutslakhorst:
This requires a specific (wrapping) error that is recognised by the scheduler

PR #1499, opened 2022-09-30
Network: cancel notifier on receiving invalid data

gerardsn:
part of #1449 
- [ ] testing
- [ ] network.emitEvents -> what is needed for this?
- [ ] VCR currently only retries context errors. Since The VCR only contains indices in a local bbolt store, the impact of DB errors should be minimal. go-leia probably needs better errors to distinguish between recoverable and unrecoverable errors


network/dag/notifier.go
```diff
@@ -291,7 +314,15 @@ func (p *notifier) Notify(event Event) {
 	}
 
 	if err := p.notifyNow(event); err != nil {
```
pascaldekloe, line 316: 🚨 The error is silently discarded now. Should at least do a `log.Print("dag: Notify retry on: ", err)`.

network/dag/notifier.go
```diff
@@ -291,7 +314,15 @@ func (p *notifier) Notify(event Event) {
...
-		p.retry(event)
+		if isUnrecoverable(err) {
+			log.Logger().
+				WithError(err).
+				WithField(core.LogFieldTransactionRef, event.Hash.String()).
+				WithField(core.LogFieldEventSubscriber, p.name).
+				Errorf("Notify failed")
+		} else {
+			p.retry(event)
```
pascaldekloe, line 324: So a hardcoded single retry and the error ignored again?

network/dag/notifier.go
```diff
@@ -291,7 +314,15 @@ func (p *notifier) Notify(event Event) {
 	}
 
 	if err := p.notifyNow(event); err != nil {
-		p.retry(event)
+		if isUnrecoverable(err) {
+			log.Logger().
+				WithError(err).
+				WithField(core.LogFieldTransactionRef, event.Hash.String()).
+				WithField(core.LogFieldEventSubscriber, p.name).
+				Errorf("Notify failed")
```
pascaldekloe, line 322: Errors already imply "failed". It is better to log the decision instead.

```go
log.Printf("DAG event %#x dropped: %s", event.Hash, err)
```

network/transport/v2/protocol.go
```diff
@@ -240,6 +240,9 @@ func (p *protocol) handlePrivateTxRetry(event dag.Event) (bool, error) {
 	// Sanity check: if we have the payload, mark this job as finished
 	payload, err := p.state.ReadPayload(context.Background(), event.Transaction.PayloadHash())
 	if err != nil {
+		if !errors.Is(err, stoabs.ErrDatabase{}) {
```
pascaldekloe, line 243: Misuse of errors.Is here. Use `errors.As(err, new(stoabs.ErrDatabase))` instead.

network/transport/v2/protocol.go
```diff
@@ -255,6 +258,9 @@ func (p *protocol) handlePrivateTxRetry(event dag.Event) (bool, error) {
 
 	pal, err := p.decryptPAL(epal)
 	if err != nil {
+		if !errors.Is(err, stoabs.ErrDatabase{}) {
```
pascaldekloe, line 261: Dito.

vdr/ambassador.go
```diff
@@ -133,6 +134,10 @@ func (n *ambassador) handleReprocessEvent(msg *nats.Msg) {
 
 func (n *ambassador) handleNetworkEvent(event dag.Event) (bool, error) {
 	if err := n.callback(event.Transaction, event.Payload); err != nil {
+		if !errors.Is(err, stoabs.ErrDatabase{}) {
```
pascaldekloe, line 137: Dito.

network/dag/notifier.go
```diff
@@ -42,6 +42,28 @@ const (
...
+	error
+}
+
+func (e ErrUnrecoverableEvent) Unwrap() error {
+	return e.error
+}
+
+func (e ErrUnrecoverableEvent) Is(other error) bool {
+	_, ok := other.(ErrUnrecoverableEvent)
+	return ok
```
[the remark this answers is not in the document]
gerardsn: Excellent point. A quick search of the code results in 0 usages of `errors.As`, and plenty of implementations of `func (...) Is(`. Our errors + checking can probably do with a makeover.

gerardsn:
This is still a draft, but some excellent points on the errors.

A quick search of the code results in 0 usages of `errors.As`, and plenty of implementations of `func (...) Is(`. Our errors + checking can probably do with a makeover.

network/dag/notifier.go
gerardsn: retry() starts the retry mechanism and does not return anything

network/dag/notifier.go
gerardsn: It was only discarded when it triggered the retry mechanism. Changed to log all errors.

network/transport/v2/protocol.go
gerardsn: I'm proposing to let `stoabs.DatabaseError` return `stoabs.ErrDatabase` so we can use `errors.As(err, new(stoabs.ErrDatabase))`, see https://github.com/nuts-foundation/go-stoabs/pull/29

network/dag/notifier.go
```diff
@@ -42,6 +42,28 @@ const (
 	PayloadEventType = "payload"
 )
 
+func UnrecoverableEvent(err error) error {
```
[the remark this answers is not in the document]
gerardsn: changed to NewEventFatal

network/dag/notifier.go
```diff
@@ -42,6 +42,28 @@ const (
...
+	return &ErrUnrecoverableEvent{err}
+}
+
+// ErrUnrecoverableEvent signals that Event notification failed and should not be retried.
+type ErrUnrecoverableEvent struct {
```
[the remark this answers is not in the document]
gerardsn: changed to EventFatal

network/dag/notifier.go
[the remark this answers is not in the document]
gerardsn: removed

network/dag/notifier.go
gerardsn: this is consistent with the other error loggin. made some modifications.

network/transport/v2/protocol.go
pascaldekloe: The pattern is good. Just the type name s/stoabs.ErrDatabase/stoabs.DatabaseError/ again.

network/dag/notifier.go
```diff
@@ -42,6 +42,25 @@ const (
 	PayloadEventType = "payload"
 )
 
+// NewEventFatal is used by a subscriber to signal that it cannot process an Event,
+// and notification of the Event should not be retried.
+// Returns nil when err == nil.
+func NewEventFatal(err error) error {
+	if err == nil {
+		return nil
```
pascaldekloe, line 50: I'd be very surprised to find myself with a nil value from a NewXyz function. Accidents waiting to happen. The literal looks good. We don't need a constructor. 😀 Code should be boring—not a surprise around every corner.

reinkrul (commented):
Why use `errors.As` over `Is`?

network/dag/notifier.go
```diff
@@ -291,7 +311,16 @@ func (p *notifier) Notify(event Event) {
 	}
 
 	if err := p.notifyNow(event); err != nil {
-		p.retry(event)
+		retryErrMsg := "Notify event dropped"
+		if !errors.As(err, new(EventFatal)) {
```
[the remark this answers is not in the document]
gerardsn: See first 3 comments from pascal above. Might be a good idea to revise our current error handling.

network/dag/notifier.go
gerardsn: That would require exporting the error on EventFatal, and implementing the error interface. I'm fine either way, but I think we need to have a larger discussion about what the pattern for our custom errors should be.

network/dag/notifier.go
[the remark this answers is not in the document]
pascaldekloe: As is a classification, and Is matches a specific case @reinkrul. In this case we want to match anything classified as fatal, regardless of the case [cause].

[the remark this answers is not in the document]
reinkrul: In my own words, `As` matches on type, while `Is` matches on equalness? (which can also be used, but only if you implement `func Is(err) bool`). So we should've been using `As`?

Thing to keep in mind is `As` alters the passed argument (tricky behavior).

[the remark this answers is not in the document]
pascaldekloe: Yes, correct @reinkrul. Generics would have been nice for As. Now it needs this allocation just to pass a type to query for. I expect something better to show up within a year.

network/dag/notifier.go
```diff
@@ -291,7 +311,16 @@ func (p *notifier) Notify(event Event) {
 	}
 
 	if err := p.notifyNow(event); err != nil {
-		p.retry(event)
+		retryErrMsg := "Notify event dropped"
```
woutslakhorst, line 314: the variable name no longer makes sense.

network/dag/notifier.go
```diff
@@ -351,16 +383,20 @@ func (p *notifier) notifyNow(event Event) error {
...
 
-	if finished, err := p.receiver(*dbEvent); err != nil {
-		log.Logger().
-			WithError(err).
-			WithField(core.LogFieldTransactionRef, dbEvent.Hash.String()).
-			WithField(core.LogFieldEventSubscriber, p.name).
-			Errorf("Retry failed")
+	finished, err := p.receiver(*dbEvent)
+	if err != nil {
+		if errors.As(err, new(EventFatal)) {
```
woutslakhorst, line 388: The previous if can be removed. This call also checks for nil

woutslakhorst (changes requested):
There are two checks, one is in `Notify` the other in `NotifyNow`. isn't one enough (`Notify` calls `NotifyNow`)? Could be complicated though

network/dag/notifier.go
gerardsn: This would mark an event with a non-fatal error and finished == true as completed. This changes the current logic of the method

network/dag/notifier.go
gerardsn: changed to notifyErrMsg

network/dag/notifier.go
```diff
@@ -351,16 +383,20 @@ func (p *notifier) notifyNow(event Event) error {
...
+			// mark as failed event
+			dbEvent.Retries = maxRetries
+			err = retry.Unrecoverable(err)
+		}
 
 		dbEvent.Error = err.Error()
 	} else if finished {
 		return p.Finished(dbEvent.Hash)
+	} else {
```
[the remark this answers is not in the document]
gerardsn: We may want to use a different log level based on what error is returned by `notifyNow`. Previously every error returned by `p.receiver` was logged on error level in `p.notifyNow` (this has been dropped), and all errors returned by `p.notifyNow` were/are logged on debug level in `p.retry`.

network/transport/v2/protocol.go
gerardsn: different library / larger discussion

gerardsn:
> There are two checks, one is in `Notify` the other in `NotifyNow`. isn't one enough (`Notify` calls `NotifyNow`)? Could be complicated though

The obvious option would be to always call `retry` from `Notify`. This would create a lot of unnecessary go routines though, since the vast majority of notifications will complete or return a fatal error on the first attempt.

network/dag/notifier.go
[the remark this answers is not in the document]
gerardsn: https://github.com/nuts-foundation/nuts-node/pull/1499/commits/e7dc1b52aa4bbb2a5603f25142b411e53b9a544a demonstrates the idea

woutslakhorst (changes requested):
One thing left: The `NewEventFatal()` may never return nil since `New` funcs shouldn't do this. Suggestion is to change to `EventFatal{err}` where needed. (The Go way)

woutslakhorst (approved):

Issue #1502, opened 2022-10-03
VCR: NutsAuthorizationCredential validator must respect v1/v2


reinkrul:
Recently, Nuts JSON-LD context v2 was introduced, which added `legalBase` for `NutsAuthorizationCredential`s. In upcoming PR https://github.com/nuts-foundation/nuts-node/pull/1190 validation for VC is added, from which points VCs can't contain fields that aren't defined in the JSON-LD context. However, the `NutsAuthorizationCredential` validator requires such VC to have a populated `legalBase`, which is spec'd for v2 but not for v1.

The validator must only check `legalBase` when it concerns a v2 credential.

PR #1503, opened 2022-10-03
VCR: Only check LegalBase for v2 VCs

reinkrul:
Fixes https://github.com/nuts-foundation/nuts-node/issues/1502

### DIFF
--- a/vcr/credential/validator.go
+++ b/vcr/credential/validator.go
@@ -176,16 +176,19 @@ func (d nutsAuthorizationCredentialValidator) Validate(credential vc.VerifiableC
 	if len(strings.TrimSpace(cs.PurposeOfUse)) == 0 {
 		return failure("'credentialSubject.PurposeOfUse' is nil")
 	}
-	switch cs.LegalBase.ConsentType {
-	case "implied":
-		// no additional requirements
-		break
-	case "explicit":
-		if cs.Subject == nil || len(strings.TrimSpace(*cs.Subject)) == 0 {
-			return failure("'credentialSubject.Subject' is required when consentType is 'explicit'")
+
+	if credential.ContainsContext(NutsV2ContextURI) {
+		switch cs.LegalBase.ConsentType {
+		case "implied":
+			// no additional requirements
+			break
+		case "explicit":
+			if cs.Subject == nil || len(strings.TrimSpace(*cs.Subject)) == 0 {
+				return failure("'credentialSubject.Subject' is required when consentType is 'explicit'")
+			}
+		default:
+			return failure("'credentialSubject.LegalBase.ConsentType' must be 'implied' or 'explicit'")
 		}
-	default:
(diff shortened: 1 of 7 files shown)


vcr/credential/validator.go
```diff
@@ -176,16 +176,19 @@ func (d nutsAuthorizationCredentialValidator) Validate(credential vc.VerifiableC
...
-		if cs.Subject == nil || len(strings.TrimSpace(*cs.Subject)) == 0 {
-			return failure("'credentialSubject.Subject' is required when consentType is 'explicit'")
+
+	if credential.ContainsContext(NutsV2ContextURI) {
+		switch cs.LegalBase.ConsentType {
+		case "implied":
+			// no additional requirements
+			break
+		case "explicit":
+			if cs.Subject == nil || len(strings.TrimSpace(*cs.Subject)) == 0 {
```
pascaldekloe, line 186: Instead of `len(…) == 0` a `strings.TrimSpace(*cs.Subject) != ""` would read nicely here. The Go compiler is optimised for this scenario.

pascaldekloe (approved):
LGTM

vcr/credential/validator.go
reinkrul: Assuming _The Internet_ is right; I had to do some digging for proof, but it indeed seems to be faster (or at least lead to less operations); https://stackoverflow.com/a/71202228/1651690

pascaldekloe: 😆 👍  We just have to know.

Reading the output from `go build -gcflags=-S` on a regular basis comes highly recommend b.t.w.. The compiler almost never does what I expect it to do.

vcr/credential/validator.go
```diff
@@ -176,16 +176,19 @@ func (d nutsAuthorizationCredentialValidator) Validate(credential vc.VerifiableC
...
 	}
-	switch cs.LegalBase.ConsentType {
-	case "implied":
-		// no additional requirements
-		break
-	case "explicit":
-		if cs.Subject == nil || len(strings.TrimSpace(*cs.Subject)) == 0 {
-			return failure("'credentialSubject.Subject' is required when consentType is 'explicit'")
+
+	if credential.ContainsContext(NutsV2ContextURI) {
```
woutslakhorst, line 180: This will break as soon there's a V3 context that also implies....
Either make it fool proof or create an issue that we have to get rid of these validators.

woutslakhorst (changes requested):

vcr/credential/validator.go
reinkrul: Question for V3 is when, and what. Long as we have these validators, introducing a new JSON-LD context will include altering these validators. I'm all for getting rid of these validators, but there should be something in return to keep people from making mistakes when issuing credentials (these are all too common). I'd prefer pluggable validation rules, like JSON-LD contexts are. Maybe in the form of JSON schemas.

woutslakhorst (approved):

PR #1504, opened 2022-10-04
Storage: check presence of required Redis Sentinel params

reinkrul:


woutslakhorst (changes requested):
Given the complexity of the URL parsing, can't we just switch to config params?

reinkrul:
Sentinel config will be reworked to actual config parameters.

PR #1513, opened 2022-10-05
Speedup integration tests

### LINKED ISSUE #1219: Speedup network integration test
Network integration tests are slow because the underlying bbolt store is syncing everything to disk, see #1218.

Turn off synching to disk when the storage provider supports this and reduce gossip intervals that are raised in #1212

gerardsn:
closes #1219 

network/vdr integration test use bbolt store without syncing

test speedup:
network 26s -> 9s
vdr 5s -> 0.7s

### DIFF
--- a/network/network.go
+++ b/network/network.go
@@ -107,12 +107,10 @@ func NewNetworkInstance(
 	}
 }
 
-var dagStorageClass storage.Class = storage.PersistentStorageClass
-
 // Configure configures the Network subsystem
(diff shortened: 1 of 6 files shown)


woutslakhorst (changes requested):
Creating off the TestBBoltStore can be moved to `storage/test.go` where a func can have `testing.T` as argument for the cleanup.

reinkrul (approved):

woutslakhorst (approved):

PR #1514, opened 2022-10-05
fix panic when VCs do not have a custom type

### LINKED ISSUE #1512: Change in validation allows for VCs without type which causes a panic
`in vcr/store.go#71` there's a line that says: "// validation has made sure there's exactly one!".

I think validation has changed and therefore the code now reaches this point and causes a panic!

woutslakhorst:
fixes #1512

less strict validation on VC type introduced a panic. Extra painful is that it was only used for logging.

### DIFF
--- a/vcr/store.go
+++ b/vcr/store.go
@@ -65,8 +65,11 @@ func (c *vcr) StoreCredential(credential vc.VerifiableCredential, validAt *time.
 }
 
 func (c *vcr) writeCredential(subject vc.VerifiableCredential) error {
-	// validation has made sure there's exactly one!
-	vcType := credential.ExtractTypes(subject)[0]
+	vcType := "VerifiableCredential"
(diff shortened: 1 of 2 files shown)


vcr/store.go
```diff
@@ -65,8 +65,11 @@ func (c *vcr) StoreCredential(credential vc.VerifiableCredential, validAt *time.
 }
 
 func (c *vcr) writeCredential(subject vc.VerifiableCredential) error {
-	// validation has made sure there's exactly one!
-	vcType := credential.ExtractTypes(subject)[0]
+	vcType := "VerifiableCredential"
+	customTypes := credential.ExtractTypes(subject)
```
gerardsn, line 69: subject.Type is guaranteed to contain "VerifiableCredential", and at most 1 other type.

gerardsn (approved):

PR #1522, opened 2022-10-07
Network: add highest LC to cached state

gerardsn:
XOR and IBLT are cached in memory, but every request for an IBLT/XOR still required an DB call for the highest LC.

### DIFF
--- a/network/dag/state.go
+++ b/network/dag/state.go
@@ -23,14 +23,13 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"github.com/prometheus/client_golang/prometheus"
-	"sync"
-
 	"github.com/nuts-foundation/go-stoabs"
 	"github.com/nuts-foundation/nuts-node/core"
 	"github.com/nuts-foundation/nuts-node/crypto/hash"
 	"github.com/nuts-foundation/nuts-node/network/dag/tree"
 	"github.com/nuts-foundation/nuts-node/network/log"
+	"github.com/prometheus/client_golang/prometheus"
+	"sync"
 )
 
 const (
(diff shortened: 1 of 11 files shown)


network/dag/state.go
```diff
@@ -483,3 +481,21 @@ func (s *state) Diagnostics() []core.DiagnosticResult {
 	diag = append(diag, &core.GenericDiagnosticResult{Title: "failed_events", Outcome: len(s.getFailedEvents())})
 	return diag
 }
+
+type highestLC struct {
```
woutslakhorst, line 485: Why not an atomic value?

network/dag/state.go
```diff
@@ -217,15 +222,19 @@ func (s *state) Add(ctx context.Context, transaction Transaction, payload []byte
 	}), stoabs.WithWriteLock())
 }
 
-func (s *state) updateTrees(tx stoabs.WriteTx, transaction Transaction) error {
+func (s *state) updateState(tx stoabs.WriteTx, transaction Transaction) error {
+	if s.highestLC.get() < transaction.Clock() {
```
woutslakhorst, line 226: This is the point where it'll break in the future. This works because the write tx has an exclusive lock on the DB. As soon as parallel write TXs are allowed (not anytime soon) you'll need a `setIfBigger` operation on the `highestLC`.

woutslakhorst (changes requested):

network/dag/state.go
gerardsn: because it needs a `setIfBigger` ;)

network/dag/state.go
gerardsn: added + tests

woutslakhorst (approved):

PR #1524, opened 2022-10-10
Bump go-stoabs to v1.3.0

gerardsn:
this updates the go-redis library that should fix an error in redis sentinel


gerardsn:
same as #1532

PR #1509, opened 2022-10-05
Add benchmark for tree.Load

gerardsn:
Some benchmarks in preparation for changes to state trees

PR #1528, opened 2022-10-10
State: build tree bottom up

gerardsn:
Changed the way how trees are built after loading from disk. See #1509 for some benchmarks.

### DIFF
--- a/network/dag/tree/tree_test.go
+++ b/network/dag/tree/tree_test.go
@@ -360,12 +360,13 @@ func filledTestTree(data Data, leafSize uint32) (*tree, treeData) {
 	return tr, td
 }
 
-/* treeData for the following structure
-       r
-	   / \
-    p0   p1
-   / \   / \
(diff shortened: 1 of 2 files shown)


woutslakhorst (changes requested):
Tests still work so seems legit...

network/dag/tree/tree_test.go
```diff
@@ -360,12 +360,13 @@ func filledTestTree(data Data, leafSize uint32) (*tree, treeData) {
...
 }
 
-/* treeData for the following structure
-       r
-	   / \
-    p0   p1
-   / \   / \
-  c0 c1 c2 c3
+/*
+	 treeData for the following structure
```
woutslakhorst, line 364: spaces vs tabs in this comment

gerardsn: IDE view is different from github

Issue #1517, opened 2022-10-06
Docs: document requirements/advice regarding key rotation


reinkrul:
Parties probably don't realize they should be rotating their DID documents' keys on a regular basis. This should probably some "going to production" documentation.

Maybe summarized in a checklist that people can use to make sure they did everything they need to do to make it secure, like;

- Require (correct) client certificate on /n2n, gRPC and FHIR endpoints
- Have a key rotation procedure
- Have a backup procedure
- Protect access to /internal
- etc

woutslakhorst:
- What about an API that does this
- What about a node policy to do this? Totally automated (clustering issue maybe)

woutslakhorst:
+ backport

reinkrul:
First approach which uses existing operations is to use `vdr addvm` and `vdr delvm`. The following changes are still required:

- Make `addvm` also add the keys as vm relationships (`keyAgreement` and friends), so it can actually be used.
- Make `delvm` remove the key from all vm relationships.
- Document procedure

PR #1537, opened 2022-10-11
VDR: Support & document key rotation operations

reinkrul:
Closes https://github.com/nuts-foundation/nuts-node/issues/1517

TODO:

- [x] End-to-end test: https://github.com/nuts-foundation/nuts-go-e2e-test/pull/51
- [x] Actually apply key usage in AddNewVerificationMethod API
- [x] Documentation

### DIFF
--- a/docs/pages/deployment/key-rotation.rst
+++ b/docs/pages/deployment/key-rotation.rst
@@ -62,8 +62,7 @@ Corresponding REST API operation: ``GET /internal/vdr/v1/did/{did}``
 2. Register a new key
 ==================
 
-Then, you add a new key which will be used to sign future transactions.
-This generates a new key pair in your crypto storage and adds it to the DID document.
+Then, you add a new key which generates a new key pair in your crypto storage and adds it to the DID document:
 
 .. code-block:: shell
 
--- a/docs/pages/deployment/key-rotation.rst
+++ b/docs/pages/deployment/key-rotation.rst
@@ -1,0 +1,100 @@
+.. _key-rotation:
+
+Key rotation procedure
+######################
+
+To minimize the impact of stolen/leaked keys, private keys should be rotated at a regular, scheduled interval.
+This applies to any (vendor, care organization, or any other) private key used for a longer period of time.
+The node aids this procedure by supporting operations to add and remove keys from DID documents.
+
+Procedure
+*********
+
+The procedure to rotate a key is as follows:
+
+1. find out the ID of the key to rotate,
+2. register a new key (verification method in the DID document)
+3. remove the previous key from the DID document
+
+To do this you need the following:
+
+- DID of the DID document which contains the key
+- the allowed usage of the key (verification method relationships)
+
+The examples below use the CLI to perform the key rotation, but the same process can be performed using the REST API.
+Each step below notes the corresponding REST API operation (refer to its documentation for the exact usage).
+
+1. Determine which key to rotate
+=============================
+
+First, you need to determine which key you'll rotate. You either have a system to administer to schedule key rotations (recommended),
+or you have to resolve the DID document to lookup the ID of the key, e.g. (given the DID ``<DID>``):
+
+.. code-block:: shell
+
+    nuts vdr resolve <DID>
+
+Use the result to lookup the ID of key (verification method) to rotate, e.g.:
+
+.. code-block:: json
+    {
+      ...
+      "verificationMethod": [
+        {
+          "controller": "did:nuts:Gjkkn5PgY3hEbKBJe5xot2mGcpD7MVN9Cs4j2XU7wZZp",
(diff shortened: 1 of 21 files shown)


docs/pages/deployment/key-rotation.rst
```diff
@@ -0,0 +1,100 @@
...
+
+    did:nuts:Gjkkn5PgY3hEbKBJe5xot2mGcpD7MVN9Cs4j2XU7wZZp#wPB7g3ipPKastvXiLJOITdCqLFDn4nJRFsCNVxaI1us
+
+
+Corresponding REST API operation: ``GET /internal/vdr/v1/did/{did}``
+
+2. Register a new key
+==================
+
+Then, you add a new key which will be used to sign future transactions.
```
woutslakhorst, line 65: or other actions (encrypt, sign JWT, etc)

vdr/types/common.go
```diff
@@ -136,31 +136,37 @@ type ResolveMetadata struct {
...
 
+// KeyUsage is a bitmask used for specifying for what purposes a key in a DID document can be used (a.k.a. Verification Method relationships).
+type KeyUsage int
+
+// Is returns whether the specified KeyUsage is enabled.
+func (k KeyUsage) Is(other KeyUsage) bool {
+	return k&other == other
+}
+
+const (
```
woutslakhorst, line 147: you could have used the left shift operator, C style

woutslakhorst (changes requested):

vdr/types/common.go
reinkrul: Nice

vdr/types/common.go
```diff
@@ -136,31 +136,37 @@ type ResolveMetadata struct {
 // CompoundService is a service type that can be used as target for github.com/nuts-foundation/go-did/did/document.go#UnmarshalServiceEndpoint
 type CompoundService map[string]string
 
+// KeyUsage is a bitmask used for specifying for what purposes a key in a DID document can be used (a.k.a. Verification Method relationships).
```
pascaldekloe, line 139: KeyUsage is not a mask, but a flag set. When filtered with a logical AND the filter becomes the mask. Flags are unsigned by nature. I would propose `type DocKeyFlags uint` instead.

vdr/api/v1/types.go
```diff
@@ -25,3 +25,63 @@ type DIDDocument = did.Document
...
+	result = setKeyUsage(result, r.AssertionMethod, types.AssertionMethodUsage)
+	result = setKeyUsage(result, r.CapabilityDelegation, types.CapabilityDelegationUsage)
+	result = setKeyUsage(result, r.CapabilityInvocation, types.CapabilityInvocationUsage)
+	result = setKeyUsage(result, r.KeyAgreement, types.KeyAgreementUsage)
+	return result
+}
+
+func setKeyUsage(current types.KeyUsage, value *bool, keyUsageToSet types.KeyUsage) types.KeyUsage {
+	// If not set, do nothing (keep existing value)
+	// If set (may be true or false), disable or enable
```
pascaldekloe, line 76: These comments belong at go doc level. A "set…" function would require a pointer receiver. Use "with…" when returning a (possibly modified) copy.

```go
// WithFlagTo returns reg with value applied as flag.
func withFlagTo(reg, flag types.DocKeyFlags, value *bool) types.DocKeyFlags {
        switch {
        case value == nil: // no setting
                return reg
        case *value: // true
                return reg | flag
        default: // false
                return reg &^ flag
        }
}
```

pascaldekloe (changes requested):
Kudos for the readme. :-)

vdr/types/common.go
reinkrul: I like the "flags" portion, because `KeyUsage` sounded a bit too much like X509 while it isn't. Renamed to `DIDKeyFlags`.

woutslakhorst (approved):

vdr/api/v1/types.go
```diff
@@ -25,3 +25,63 @@ type DIDDocument = did.Document
...
+	SelfControl *bool `json:"selfControl,omitempty"`
+}
+
+// VerificationMethodRelationship defines model for VerificationMethodRelationship.
+type VerificationMethodRelationship struct {
+	// indicates if the generated key pair can be used for assertions.
+	AssertionMethod *bool `json:"assertionMethod,omitempty"`
+
+	// indicates if the generated key pair can be used for authentication.
+	Authentication *bool `json:"authentication,omitempty"`
```
pascaldekloe, line 47: I think it is better not to use pointers for absent (configuration) values where possible. If AssertMethod was just a `bool`, then it would document the default of `false` on the fly. Authentication would have to be renamed to NoAuthentication as such.

No pointers also prevents the ugly `&varFalse` construction above, plus it eliminates the need for a separate withKeyFlag function—just inline the if/else in  ToFlags. ✨

pascaldekloe (approved):

vdr/api/v1/types.go
reinkrul: There because of ugly OpenAPI code generator, but fixed ugly/dangerous bool ptr reusage.

PR #1541, opened 2022-10-11
refactor types.ErrInvalidServiceQuery to doc.DIDServiceQueryError

pascaldekloe:
Error type names should end with Error. The error strings each should classify the full context of what is already explicit in code. In this case, describe the error is about "DID Subjects" and "endpoint URIs".

Errors should live together with the code that encounters any of the respective cases. The package name becomes a descriptive part of the scenario. Package names like `types` or `core` must be killed with fire.

https://go.dev/doc/effective_go#names

Do not introduce functions for cases handled well in plain code, especially if such function has an odd structure like core.ResolveStatusCode does. The new `switch` is more readable, it requires no knowledge of the Nuts packages, and it provides a sane default [HTTP 500].

### DIFF
--- a/vdr/doc/util.go
+++ b/vdr/doc/util.go
@@ -79,8 +79,9 @@ func ValidateServiceReference(endpointURI ssi.URI) error {
 		return DIDServiceQueryError{errors.New("endpoint URI with multiple " + serviceTypeQueryParameter + " query parameters")}
 	}
 
+	// “Other query parameters, paths or fragments SHALL NOT be used.”
+	// — RFC006, subsection 4.2
 	if len(q) > 1 {
-		// weird requirement; needs explaination why
 		return DIDServiceQueryError{errors.New("endpoint URI with query parameter other than " + serviceTypeQueryParameter)}
 	}
 
--- a/vdr/doc/util.go
+++ b/vdr/doc/util.go
@@ -19,20 +19,20 @@
 package doc
 
 import (
+	"errors"
 	"fmt"
 	"github.com/nuts-foundation/go-did"
 	"github.com/nuts-foundation/go-did/did"
-	"github.com/nuts-foundation/nuts-node/vdr/types"
 	"strings"
 )
 
 const serviceTypeQueryParameter = "type"
-const serviceEndpointPath = "serviceEndpoint"
+const serviceEndpointPath = "/serviceEndpoint"
 
 // MakeServiceReference creates a service reference, which can be used as query when looking up services.
 func MakeServiceReference(subjectDID did.DID, serviceType string) ssi.URI {
 	ref := subjectDID.URI()
-	ref.Opaque += "/" + serviceEndpointPath
+	ref.Opaque += serviceEndpointPath
 	ref.Fragment = ""
 	ref.RawQuery = fmt.Sprintf("%s=%s", serviceTypeQueryParameter, serviceType)
 	return ref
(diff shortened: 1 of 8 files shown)


vdr/doc/util.go
```diff
@@ -43,31 +43,46 @@ func IsServiceReference(endpoint string) bool {
...
 }
 
+// DIDServiceQueryError denies the query based on validation constraints.
+type DIDServiceQueryError struct {
+	Err error // cause
+}
+
+// Error implements the error interface.
+func (e DIDServiceQueryError) Error() string {
+	return "DID service query invalid: " + e.Err.Error()
```
reinkrul, line 53: isn't there a safer way which doesn't cause a panic when the inner error is nil?

vdr/doc/util.go
```diff
@@ -43,31 +43,46 @@ func IsServiceReference(endpoint string) bool {
...
+		return DIDServiceQueryError{errors.New("endpoint URI without " + serviceTypeQueryParameter + " query parameter")}
+	default:
+		return DIDServiceQueryError{errors.New("endpoint URI with multiple " + serviceTypeQueryParameter + " query parameters")}
 	}
-	if len(endpointURI.Query()) > 1 {
-		// Service reference contains more than just `type` query parameter
-		return typeQueryParameterError
+
+	if len(q) > 1 {
+		// weird requirement; needs explaination why
```
reinkrul, line 83: We spec'd what service references look like. If there are other query parameters, it's a user error/bug so we kindly inform them. Maybe in future we allow custom/flexible queries, we can always relax the requirement.

vdr/doc/util.go
pascaldekloe: In Go pointer may not be nil, unless specified explicitly. For these rare cases it is recommended to name the field accordingly, such as `ErrOrNil`. In here, the embedded error is significant. Without it the structure/instance makes no sense.

vdr/doc/util.go
pascaldekloe: You are absolutely right. RFC006 states "Other query parameters, paths or fragments SHALL NOT be used.".

vdr/doc/util.go
```diff
@@ -43,31 +43,47 @@ func IsServiceReference(endpoint string) bool {
...
+	case 0:
+		return DIDServiceQueryError{errors.New("endpoint URI without " + serviceTypeQueryParameter + " query parameter")}
+	default:
+		return DIDServiceQueryError{errors.New("endpoint URI with multiple " + serviceTypeQueryParameter + " query parameters")}
 	}
-	if len(endpointURI.Query()) > 1 {
-		// Service reference contains more than just `type` query parameter
-		return typeQueryParameterError
+
+	// “Other query parameters, paths or fragments SHALL NOT be used.”
```
reinkrul, line 82: good way of referencing

reinkrul (approved):

vdr/doc/util.go
woutslakhorst: But nothing is preventing someone (uninitiated dev) to create such an error without the underlying error: `DIDServiceQueryError{}.Error()`

pascaldekloe: The developer is responsible for setting all fields when creating a struct, unless the names or documentation state otherwise. Creating nil checks everywhere down to road is no solution. For example, an io.LimitReader panics when .R is nil, and that is no mistake.

If we think that our developers can not bare this responsibility, then we should use constructors. Linters notice direct struct construction when a matching New function exists.

reinkrul: Agreed, and it would (should) be found by a unit test/review if it happens

PR #1557, opened 2022-10-14
VCR: Tolerate undefined fields for v1 NutsAuthCred

reinkrul:
To maintain backwards compatibility with v4. Remove for v6.

Backport to v5


woutslakhorst:
Hard to value without a broader analysis of where verification happens and how this works with the combination of v4/v5 data and v4/v5 nodes.

reinkrul:
We decided differently: https://github.com/nuts-foundation/nuts-node/issues/1580#issuecomment-1288880589

PR #1560, opened 2022-10-17
Network: add highest LC to cached state (#1522)

gerardsn:
back port of #1522

### DIFF
--- a/network/dag/state.go
+++ b/network/dag/state.go
@@ -23,14 +23,13 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"github.com/prometheus/client_golang/prometheus"
-	"sync"
-
 	"github.com/nuts-foundation/go-stoabs"
 	"github.com/nuts-foundation/nuts-node/core"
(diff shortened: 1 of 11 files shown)


network/dag/state.go
```diff
@@ -51,6 +50,7 @@ type state struct {
 	notifiers           map[string]Notifier
 	xorTree             *treeStore
 	ibltTree            *treeStore
+	highestLC           highestLC
```
pascaldekloe, line 53: Mutex not ideal for atomic access. The classic solution was `atomic.Value`. Now with generics, I`d recommend `atomic.StorePointer` and  `atomic.LoadPointer` on this field—without the highestLC type— in combination with an explicit field name, such as atomicHighestLC.

reinkrul (approved):

network/dag/state.go
gerardsn: The current PR is a back port to the V5 branch. No point in changing it here, but sounds like a good change for the master branch

pascaldekloe: Well there you go: #1565.

PR #1565, opened 2022-10-17
Eliminate mutex use in dag package

pascaldekloe:
* locks held while invoking listener methods
* missing atomic access on Prometheus mock
* dedicaded atomic type elimination

### DIFF
--- a/network/dag/state.go
+++ b/network/dag/state.go
@@ -333,7 +333,7 @@ func (s *state) Notifier(name string, receiver ReceiverFn, options ...NotifierOp
 
 	_, loaded := s.notifiers.LoadOrStore(name, n)
 	if loaded {
-		return nil, fmt.Errorf("event receiver %q registarion denied on duplicate name", name)
+		return nil, fmt.Errorf("Nuts event receiver %q registration denied on duplicate name", name)
 	}
 	return n, nil
 }
--- a/network/dag/state.go
+++ b/network/dag/state.go
@@ -30,6 +30,7 @@ import (
 	"github.com/nuts-foundation/nuts-node/network/log"
 	"github.com/prometheus/client_golang/prometheus"
 	"sync"
+	"sync/atomic"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


network/dag/state.go
```diff
@@ -324,36 +325,33 @@ func (s *state) Head(ctx context.Context) (hash.SHA256Hash, error) {
...
 	options = append(options, withCounters(s.eventsNotifyCount, s.eventsFinishedCount))
 
-	notifier := NewNotifier(name, receiver, options...)
-	s.notifiers[name] = notifier
+	n := NewNotifier(name, receiver, options...)
 
-	return notifier, nil
+	_, loaded := s.notifiers.LoadOrStore(name, n)
+	if loaded {
+		return nil, fmt.Errorf("event receiver %q registarion denied on duplicate name", name)
```
reinkrul, line 336: Why not keep the message?
```suggestion
		return nil, fmt.Errorf("notifier already exists (name=%s)", name)
```

network/dag/state.go
```diff
@@ -435,72 +429,37 @@ func (s *state) Verify(ctx context.Context) error {
...
-	for name, notifier := range s.notifiers {
-		events, err := notifier.GetFailedEvents()
+func (s *state) failedEventCount() int {
+	var n int
+	s.notifiers.Range(func(key, value any) bool {
+		events, err := value.(Notifier).GetFailedEvents()
 		if err != nil {
-			log.Logger().Errorf("Failed to retrieve failed events: %v (notifier=%s)", err, name)
-			continue
+			log.Logger().Errorf("failed events lookup %q: %s", key, err)
```
reinkrul, line 452: our log messages start with a capital, and might be useful to keep the notifier name to aid future analysis. We moved to structured logging recently (for fields worth indexing), so this should probably be something like:
```suggestion
			log.Logger().WithError(err).Errorf("Unable to retrieve failed events for %q", key)
```

reinkrul (changes requested):

network/dag/state.go
pascaldekloe: * fails to tell which notifier this is about
* fails to tell what the result/decission is
* the (key=value) syntax adds no information
* use quoted strings to detect whitespace and such

network/dag/state.go
pascaldekloe: Even more log bells and whistles. 😢 The core log.Printf is such a beauty.
Any way, "unable to" and "failed to" are kind of useless prefixes in an error message.

network/dag/state.go
reinkrul: Ok, but then fix the typo in "registarion" (typos trigger me 🤣 )

network/dag/state.go
reinkrul: I can live without the prefix, as long as we keep the field logs (they help log analysis by providing indexable fields). Form (consistency in casing) is probably more of a personal quest; I get annoyed and thus distracted when message formatting/casing isn't consistent. But that might be my own shortcoming..

reinkrul (approved):

Issue #1561, opened 2022-10-17
VCR: Validating VCs fields breaks backwards compatiblity


reinkrul:
In v5 we added validation of VC fields: all fields must be defined by the JSON-LD context. This has broken NutsAuthorizationCredentials, because it required LegalBase to be present which was missing from the JSON-LD context.

Validation is performed when:
- Issuing a VC (self-issued)
- Retrieving a VC (resolving or searching) (self-issued or by someone else)
- Creating an access token (self-issued, or issued by a trusted 3rd party in future)

**Given**
- Contents of VCs are only validated ("are all fields defined in JSON-LD context?") when issuing it. It exists to help vendors find bugs in their integration.
- Nuts v1 JSON-LD context OR RFC014 will have to be changed: JSON-LD context is missing LegalBase, while RFC014 states it MUST be present. Both are considered "released".

**Possible solutions**
- v2 Nuts JSON-LD context becomes default in v5 and v4.3
  - 👍  Pros:
     - Quick adoption of v2 JSON-LD context
     - Technically "pure" solutions: no shortcuts, e.g., altering released specifications
     - No changes required for existing integrations from vendors
     - No/minimal changes required in RFCs
   - 👎 Cons:
     - Nuts Node v4 users must upgrade to v4.3 of v5: nodes running <= v4.3 can't process the new credentials
- Current situation: don't allow LegalBase in v5, allow it starting v6 (through Nuts v2 context)
  - 👍  Pros:
     - ...
   - 👎 Cons:
     - Nuts Spec v1 will have to be altered, which is already released.
     - Existing integrations from vendors will have to be changed
     - Adoption of Nuts v2 is deferred until v6 is released (which is unscheduled)
- Lenience (variant on above): keep current situation, but accept LegalBase in API, remove it from VC before issuing it
  - 👍  Pros:
     - No changes required for existing integrations from vendors. We keep our promise "you may run the previous version, and it will keep working".
   - 👎 Cons:
     - Odd API behavior from developer perspective, which might lead to incorrect conclusions (LegalBase expected to be there, but it isn't)
     - Nuts Spec v1 will have to be altered, which is already released.
- Add LegalBase to Nuts v1 JSON-LD context in v5; remove it from VC before validating signature for backwards compatibility (based on issuance date, or verify signature twice; with legalBase and without)
   Can be made less awkward if the VC are sanitized (legalBase removed if there 'illegally') before storing, complemented with a migration to do the same to existing VCs in storage.
[...]

PR #1570, opened 2022-10-18
v4: Forward compatibility for NutsAuthenticationCredentials in v5 release 

reinkrul:
Upcoming v5 release will start issuing NutsAuthorizationCredential@v2, because NutsAuthorizationCredential@v1 was missing definition for `LegalBase`. But at the same time, the API requires it to be present.

The proposed solution is as follows (taken from https://github.com/nuts-foundation/nuts-node/issues/1561):
- New v4 minor release that:
  - Adds support for NutsAuthorizationCredential@v2, such that when issued by v5 are understood
  - Relaxes validation of NutsAuthorizationCredential@v1 credentials so `LegalBase` does not need to be present, so correctly issued  NutsAuthorizationCredential@v1 (without LegalBase being present) are valid according to v4.
- After some time-to-upgrade, v5 will be released which by default issues NutsAuthorizationCredential@v2

This PR cherry-picks commits already present on master.


reinkrul:
Tested against end-to-end test v4 branch

woutslakhorst (changes requested):
Relaxation of validation: yes.
New V2 context: no

reinkrul:
Closed in favor of a new PR

PR #1571, opened 2022-10-18
Release notes: make clear v5 is not yet released

reinkrul:


gerardsn (approved):
These nutes should only be visible after deliberately opening the latest branch of the docs anyway

PR #1572, opened 2022-10-18
VCR: Issue NutsAuthorizationCredential@v2 by default

reinkrul:
Instead of v1.


reinkrul:
We decided differently: https://github.com/nuts-foundation/nuts-node/issues/1580#issuecomment-1288880589

PR #1582, opened 2022-10-21
Network: Add Rebuild

gerardsn:
Proposal for Rebuild state from DAG, code is untested.

`data.db` has a bunch of shelves. The `transactionsShelf` contains the actual data and should be considered correct/leading. We can add another command to verify the data in here (`state.Verify` still exists), but that is a different feature.

All other shelfs are derived from the `transactionsShelf` and may need fixing/rebuilding. The main difficulty in doing this on a running node is dealing with new transactions (both network and api) while rebuilding. The following shelfs are derived from the `transactionsShelf` (and updated transactionally): 
- `clockShelf`: maps lamport clock to tx references
- `metadataShelf`: head, highest LC, # tx on dag
- `ibltShelf` & `xorShelf`: state trees
- `payload`: not considered (yet) as its behavior is a bit different

The `clockShelf` is the only one that must be rebuild from the `transactionsShelf`, the rest can be rebuild from either of the two. The only strategy I could come up with to rebuild the `clockShelf` is locking the entire `data.db` and iterating over the entire thing. This locks the DB for a very long time (all other requests for write locks are guaranteed to expire), parses all transactions (resource heavy), and keeps track of a potentially very large map between clock values and transaction references.

All other shelfs (with the exception of the `numberOfTransactions` value on the `metadataShelf`) can be rebuild in batches where the DB is locked while a small range of LC values is evaluated. However, this actually rebuilds all other shelfs to match the data on the `clockShelf` and is thus possible IFF the `clockShelf` is correct.

Questions/Options:
- Benchmark takes 4s to complete for a single `state.Rebuild` (for ~50k txs on DAG) when there is nothing to fix. The write/reloads when there actually is an inconsistency will add a significant amount of time to this. Should this be done on a live node in the first place?
- Just rebuilding state in batches seems possible. Should we provide 2 commands. One to rebuild `clockShelf` and one to rebuild the rest? (not entirely sure where the `metadataShelf` belongs) 
- If processing in batches, how should a write lock request timeout be dealt with?
- There is a good chance the logs will provide a hint to what LC value/range might have issues. Adding a range to the Rebuild command could make the procedure more manageable.

### DIFF
--- a/crypto/hash/sha256.go
+++ b/crypto/hash/sha256.go
@@ -74,7 +74,7 @@ func (h SHA256Hash) Slice() []byte {
 
 // Equals determines whether the given Hash is exactly the same (bytes match).
 func (h SHA256Hash) Equals(other SHA256Hash) bool {
-	return h.Compare(other) == 0
+	return bytes.Equal(h[:], other[:])
 }
 
 // Compare compares this Hash to another Hash using bytes.Compare.
--- a/network/dag/dag.go
+++ b/network/dag/dag.go
@@ -37,8 +37,8 @@ const metadataShelf = "metadata"
 // numberOfTransactionsKey is the key of the metadata property that holds the number of transactions on the DAG
 const numberOfTransactionsKey = "tx_num"
 
-// highestClockValue is the key of the metadata property that holds the highest lamport clock value of all transactions on the DAG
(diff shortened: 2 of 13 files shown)


gerardsn:
See linked issue for reason of closure

PR #1593, opened 2022-10-26
Synchronous Reprocess Call

pascaldekloe:
Bug fix, code simplification, and a more future proof API.

### DIFF
--- a/network/network.go
+++ b/network/network.go
@@ -636,17 +636,20 @@ func (n *Network) PeerDiagnostics() map[transport.PeerID]transport.Diagnostics {
 }
 
 func (n *Network) Reprocess(contentType string) {
-	go n.ReprocessContentType(context.Background(), contentType)
+	go func() {
+		err := n.ReprocessContentType(context.Background(), contentType)
+		if err != nil {
+			log.Logger().Error(err)
+		}
+	}()
 }
 
-func (n *Network) ReprocessContentType(ctx context.Context, contentType string) {
+func (n *Network) ReprocessContentType(ctx context.Context, contentType string) error {
 	log.Logger().Infof("Starting reprocess of %s", contentType)
 
 	_, js, err := n.eventPublisher.Pool().Acquire(ctx)
 	if err != nil {
-		log.Logger().
-			WithError(err).
-			Error("Failed to start reprocessing transactions")
+		return fmt.Errorf("reprocess abort on message client: %w", err)
 	}
 
 	// The Lamport's clock stamps count from 0, with a step size of 1.
@@ -654,15 +657,11 @@ func (n *Network) ReprocessContentType(ctx context.Context, contentType string) 
 	for offset := 0; ; offset += clockSteps {
 		end := offset + clockSteps
 		if end >= 1<<32 {
-			log.Logger().Error("reprocess abort on Lamport clock uint32 overflow")
-			return
+			return errors.New("reprocess abort on Lamport clock uint32 overflow")
 		}
 		txs, err := n.state.FindBetweenLC(ctx, uint32(offset), uint32(end))
 		if err != nil {
-			log.Logger().
-				WithError(err).
-				Errorf("reprocess abort on transaction lookup in clock range [%d, %d)", offset, end)
-			return
+			return fmt.Errorf("reprocess abort on transaction lookup, clock range [%d, %d): %w", offset, end, err)
 		}
 
 		for _, tx := range txs {
@@ -674,12 +673,7 @@ func (n *Network) ReprocessContentType(ctx context.Context, contentType string) 
 			subject := fmt.Sprintf("%s.%s", events.ReprocessStream, contentType)
 			payload, err := n.state.ReadPayload(ctx, tx.PayloadHash())
 			if err != nil {
-				log.Logger().
-					WithError(err).
-					WithField(core.LogFieldTransactionRef, tx.Ref()).
-					WithField(core.LogFieldEventSubject, subject).
-					Error("Failed to publish transaction")
-				return
+				return fmt.Errorf("reprocess abort on transaction %#x payload %#x: %w", tx.Ref(), tx.PayloadHash(), err)
 			}
 			twp := events.TransactionWithPayload{
 				Transaction: tx,
@@ -692,12 +686,7 @@ func (n *Network) ReprocessContentType(ctx context.Context, contentType string) 
 				Trace("Publishing transaction")
 			_, err = js.PublishAsync(subject, data)
 			if err != nil {
-				log.Logger().
-					WithError(err).
-					WithField(core.LogFieldTransactionRef, tx.Ref()).
-					WithField(core.LogFieldEventSubject, subject).
-					Error("Failed to publish transaction")
-				return
+				return fmt.Errorf("reprocess abort on transaction %#x publish: %w", tx.Ref(), err)
 			}
 		}
 
@@ -708,6 +697,8 @@ func (n *Network) ReprocessContentType(ctx context.Context, contentType string) 
 		// give some time for Update transactions that require all read transactions to be closed
 		time.Sleep(time.Second)
 	}
+
+	return nil
 }
 
 func (n *Network) collectDiagnosticsForPeers() transport.Diagnostics {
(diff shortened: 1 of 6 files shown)


network/network.go
```diff
@@ -611,70 +611,86 @@ func (n *Network) PeerDiagnostics() map[transport.PeerID]transport.Diagnostics {
...
+		log.Logger().Infof("reprocess %q complete", contentType)
+	}()
+}
+
+// ReprocessReport describes the reprocess exection.
+type ReprocessReport struct {
+	// reserved for future use
+}
 
+func (n *Network) ReprocessContentType(ctx context.Context, contentType string) (*ReprocessReport, error) {
```
reinkrul, line 631: shouldn't be exported?

woutslakhorst (changes requested):
Since there's now a synchronous call, the tests can be simplified as well where a wait loop is used now, this could be replaced by a context with timeout.

network/network.go
pascaldekloe: This should be used instead of the current Reprocess. Ideally just drop and replace, if that is something we can do?

woutslakhorst: It's called from an API which is returning `202 accepted`. The API can't wait for the process since it'll timeout on the client side.

network/network_test.go
```diff
@@ -742,91 +739,122 @@ func TestNetwork_Reprocess(t *testing.T) {
...
+		if len(messages) == 0 {
+			t.Error("subscription received no messages")
+		}
 	})
 
-	t.Run("ignores other transactions", func(t *testing.T) {
-		ctx, eventManager := setup(t)
-		ctx.state.EXPECT().FindBetweenLC(gomock.Any(), uint32(0), uint32(1000)).Return([]dag.Transaction{tx}, nil)
-		ctx.state.EXPECT().ReadPayload(gomock.Any(), tx.PayloadHash()).Return([]byte("payload"), nil)
+	t.Run("mismatch", func(t *testing.T) {
```
woutslakhorst, line 789: not really a descriptive test name

network/network_test.go
```diff
@@ -742,91 +739,122 @@ func TestNetwork_Reprocess(t *testing.T) {
...
-		ctx, eventManager := setup(t)
-		ctx.state.EXPECT().FindBetweenLC(gomock.Any(), uint32(0), uint32(1000)).Return(nil, errors.New("b00m!"))
+	t.Run("error", func(t *testing.T) {
+		t.Run("query", func(t *testing.T) {
+			ctx, cancel := context.WithTimeout(context.Background(), time.Second)
+			defer cancel()
 
-		var found int
-		subscribe(&found, eventManager)
+			testErr := errors.New("test error")
```
woutslakhorst, line 814: may be put one level higher?

network/network_test.go
```diff
@@ -742,91 +739,122 @@ func TestNetwork_Reprocess(t *testing.T) {
...
-			foundMutex.Lock()
-			defer foundMutex.Unlock()
-			return found > 0, nil
-		}, 100*time.Millisecond)
+			messages := subscribe(ctx, t, eventManager)
 
-		assert.Equal(t, 0, found)
-	})
+			_, err := setup.network.ReprocessContentType(ctx, "application/did+vc")
+			switch {
```
woutslakhorst, line 822: This switch would have been a single assert when using testify

network/network_test.go
```diff
@@ -742,91 +739,122 @@ func TestNetwork_Reprocess(t *testing.T) {
...
-		}, 100*time.Millisecond)
+			setup, eventManager := newSetup(t)
+			setup.state.EXPECT().FindBetweenLC(gomock.Any(), uint32(0), uint32(1000)).Return([]dag.Transaction{tx}, nil)
+			setup.state.EXPECT().ReadPayload(gomock.Any(), tx.PayloadHash()).Return(nil, testErr)
 
-		assert.Equal(t, 0, found)
+			messages := subscribe(ctx, t, eventManager)
+
+			_, err := setup.network.ReprocessContentType(ctx, "application/did+json")
+			switch {
```
woutslakhorst, line 847: same here

network/network.go
```diff
@@ -611,70 +611,87 @@ func (n *Network) PeerDiagnostics() map[transport.PeerID]transport.Diagnostics {
...
+		// transactions.
+		time.Sleep(time.Second)
+	}
+
+	// flush publish queue
+	select {
+	case <-js.PublishAsyncComplete():
+		break
+	case <-ctx.Done():
+		return nil, fmt.Errorf("reprocess terminate during final flush: %w", ctx.Err())
```
woutslakhorst, line 691: This error will be logged and viewed by a system operator. What does this mean for him? Failed, should he/she do it again?

woutslakhorst (changes requested):

network/network.go
pascaldekloe: Sure. The HTTP 202 coupling makes this launch, so that's where the `go` fork should be—not somewhere hidden in a subpackage.

network/network_test.go
```diff
@@ -742,91 +739,122 @@ func TestNetwork_Reprocess(t *testing.T) {
...
+			case err == nil:
+				t.Error("reprocess gave no error")
+			case !errors.Is(err, testErr):
+				t.Errorf("reprocess error did not include error %q: %s", testErr, err)
+			}
+
+			if n := len(messages); n != 0 {
```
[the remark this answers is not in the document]
pascaldekloe: That assert thing prints a ton with little info. When something fails you have to dive in the test code to see what happens. This code is fully descriptive. No need to know what assert considers to be equal. The language is all we need.

network/network_test.go
pascaldekloe: The test name is `ReprocessWithContext/mismatch`. What would be a better description?

network/network_test.go
pascaldekloe: Better a little copy than a little dependency.

network/network.go
pascaldekloe: Exactly the point of the message. 😅 It is impossible to tell. Maybe it worked, or maybe it didn't.

network/network_test.go
pascaldekloe: True. The message printed on either case is terrible though.

network/network.go
```diff
@@ -611,70 +611,87 @@ func (n *Network) PeerDiagnostics() map[transport.PeerID]transport.Diagnostics {
...
+				Trace("Publishing transaction")
+			_, err = js.PublishAsync(subject, data)
+			if err != nil {
+				return nil, fmt.Errorf("reprocess abort on transaction %#x publish: %w", tx.Ref(), err)
 			}
+		}
 
-			// give some time for Update transactions that require all read transactions to be closed
-			time.Sleep(time.Second)
+		if len(txs) == 0 || int(uint(txs[len(txs)-1].Clock())) < end-1 {
```
stevenvegt, line 677: Can you put the output of `int(uint(txs[len(txs)-1].Clock())) < end-1` into a descriptive variable? This is painful to read.

network/network.go
stevenvegt: The `final flush` is an internal concept which does not help a system operator. I would suggest a different wording.

network/network_test.go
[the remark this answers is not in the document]
stevenvegt: Or `assert.Len()`

network/network_test.go
```diff
@@ -742,91 +739,122 @@ func TestNetwork_Reprocess(t *testing.T) {
...
+				t.Error("reprocess gave no error")
+			case !errors.Is(err, testErr):
+				t.Errorf("reprocess error did not include error %q: %s", testErr, err)
+			}
 
-	t.Run("stops on error 2", func(t *testing.T) {
-		ctx, eventManager := setup(t)
-		ctx.state.EXPECT().FindBetweenLC(gomock.Any(), uint32(0), uint32(1000)).Return([]dag.Transaction{tx}, nil)
-		ctx.state.EXPECT().ReadPayload(gomock.Any(), tx.PayloadHash()).Return(nil, errors.New("b00m!"))
+			if n := len(messages); n != 0 {
```
stevenvegt, line 829: `if assert.Len(t, messages, 0) {return}`

network/network.go
pascaldekloe: Yeah, bit weird now thinking about it. 😆 Perhaps "reprocess terminate while awaiting last enqueded"?

woutslakhorst (approved):

PR #1598, opened 2022-10-27
Feature/1530 did context

### LINKED ISSUE #1530: Missing @context's for the DID document
The key material is of `@type`: `JsonWebKey2020` and needs a `https://w3c-ccg.github.io/lds-jws2020/contexts/lds-jws2020-v1.json context`.
The service of `@type` `node-contact-info` needs its own context to support the `email`, `name`, `phone` and `website` fields.

woutslakhorst:
closes #1530 

it adds the nuts and JWS context URLS to did Docs when updating and creating. A nuts context is available in the assets.
No checking or JSON-LD magic is currently applied to did docs.

### DIFF
--- a/vdr/vdr.go
+++ b/vdr/vdr.go
@@ -32,6 +32,7 @@ import (
 	"github.com/nuts-foundation/nuts-node/vdr/doc"
 	"github.com/nuts-foundation/nuts-node/vdr/store"
 	"github.com/sirupsen/logrus"
+	"strings"
 
 	"github.com/nuts-foundation/nuts-node/crypto"
 	"github.com/nuts-foundation/nuts-node/network"
@@ -176,6 +177,9 @@ func (r VDR) Update(id did.DID, current hash.SHA256Hash, next did.Document, _ *t
 		return err
 	}
 
+	// #1530: add nuts and JWS context if not present
+	next = patchContext(next)
+
 	payload, err := json.Marshal(next)
 	if err != nil {
 		return err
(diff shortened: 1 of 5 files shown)


vdr/vdr.go
```diff
@@ -243,3 +247,25 @@ func (r VDR) resolveControllerWithKey(doc did.Document) (did.Document, crypto.Ke
 
 	return did.Document{}, nil, fmt.Errorf("could not find capabilityInvocation key for updating the DID document: %w", err)
 }
+
+func patchContext(next did.Document) did.Document {
```
reinkrul, line 251: will this be here forever? Or is temporary? When/how will it be removed?

vdr/vdr.go
```diff
@@ -243,3 +247,25 @@ func (r VDR) resolveControllerWithKey(doc did.Document) (did.Document, crypto.Ke
...
+	NutsContextPresent := false
```
reinkrul, line 252: Ugly code imo, better to have a function `addContext` which adds it if not present and call it for both contexts?

reinkrul (changes requested):

vdr/vdr.go
```diff
@@ -243,3 +247,25 @@ func (r VDR) resolveControllerWithKey(doc did.Document) (did.Document, crypto.Ke
...
+	JWSContextPresent := false
+
+	for _, c := range next.Context {
+		if strings.EqualFold(c.String(), doc.NutsDIDContextV1) {
```
pascaldekloe, line 256: Namespace URIs, including those for JSON-LD are case sensitive.

vdr/vdr.go
pascaldekloe: A `s/patchContext/withLDContext/` makes it explicit that you return a copy with changes, as opposed to a (pointer) setter.

pascaldekloe (dismissed):
I believe the namespaces should be added if, and only if the respective elements are actually in the document.

pascaldekloe:
To determine the namespaces in use, one needs to iterate over each "extra" field in the document (root) and each verification method. Something like the following perhaps?

```go
// NamespaceResolver returns the namespace URI for a property on a DID Document, with
// zero for unknown.
type NamespaceResolver(property string, content json.RawMessage) string

// VerificationMethodNamespaceResolver returns the namespace URI for a property on a
// VerificationMethod, with zero for unknown.
type VerificationMethodNamespaceResolver(m *VerificationMethod, property string, content json.RawMessage) string

// ResetContext clears the JSON-LC context with a single "https://www.w3.org/ns/did/v1"
// entry, and then resolves each additional property on the Document with r, and each
// addition property on a VerificationMethod with m.
func (d *Document) ResetContext(r NamespaceResolver, m VerifacionMethodNamespaceResolver) error {
```

vdr/vdr.go
woutslakhorst: > will this be here forever? Or is temporary? When/how will it be removed?

Unknown, if in the future users are going to be responsible then the CRUD API should also do checks to prevent non-context members (like in VCs).

woutslakhorst:
> I believe the namespaces should be added if, and only if the respective elements are actually in the document.

Currently it's a certainty that they are required. 

We currently don't treat DID Docs as JSON-LD but decided on adding the correct context so they DID Docs may be treated as JSON-LD by third parties.

woutslakhorst:
How to properly support JSON-LD is part of a larger issue that is still to be researched

reinkrul (approved):

PR #1606, opened 2022-10-30
Network/Auth: send custom UserAgent on HTTP calls

reinkrul:
HTTP header for gRPC calls:

```
nuts-node-refimpl/v4.3.0 grpc-go/1.50.1
```
(`grpc-go` user agent is always added by the library)

### DIFF
--- a/core/build_vars.go
+++ b/core/build_vars.go
@@ -64,3 +64,9 @@ func BuildInfo() string {
 
 	return b.String()
 }
+
+// UserAgent returns a string that can be used as HTTP user agent, containing the version of the node (e.g. nuts-node-refimpl/5.0.0)
+func UserAgent() string {
+	// Remove leading "v" from version to keep it standard for user agent strings.
+	return "nuts-node-refimpl/" + strings.TrimPrefix(Version(), "v")
+}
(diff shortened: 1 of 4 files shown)


core/build_vars.go
```diff
@@ -64,3 +64,9 @@ func BuildInfo() string {
 
 	return b.String()
 }
+
+// UserAgent returns a string that can be used as HTTP user agent, containing the version of the node (e.g. nuts-node-refimpl/5.0.0)
+func UserAgent() string {
+	// Remove leading "v" from version to keep it standard for user agent strings.
+	return "nuts-node-refimpl/" + strings.TrimPrefix(Version(), "v")
```
gerardsn, line 71: If the node is not running a tagged version, `Version()` returns the branch name. Is this okay?

gerardsn: if so, what about trimming of branch names starting with `vdr/...` or `vcr/...`

reinkrul: Just return 0 if it isn't a version (so it must start with `v` and following a digit)?

gerardsn: Or use `GitVersion`? This makes it easy to differentiate between official releases and custom builds.

core/build_vars.go
```diff
@@ -67,6 +67,9 @@ func BuildInfo() string {
 
 // UserAgent returns a string that can be used as HTTP user agent, containing the version of the node (e.g. nuts-node-refimpl/5.0.0)
 func UserAgent() string {
-	// Remove leading "v" from version to keep it standard for user agent strings.
-	return "nuts-node-refimpl/" + GitVersion
+	version := GitVersion
+	if version == "" {
+		version = "unknown"
```
gerardsn, line 72: Should the default for `GitVersion` just be set to `unknown` instead of `""`?

gerardsn (approved):
LGTM

PR #1609, opened 2022-11-01
NATS: Fix tests giving NATS timeouts

reinkrul:
Unit tests often gave NATS-related time-out issues when running unit tests on my dev machine. CircleCI didn't have this problem. This is due to the NATS port (`4222`) didn't get released soon enough by the OS, which caused subsequent tests (when speedy enough) to cause error (time-outs) due to the port not being available. Fixed by making the NATS port random in tests.

Also made the error returned when pool acquisition fails less ambiguous vs the subscription error.

### DIFF
--- a/vcr/ambassador.go
+++ b/vcr/ambassador.go
@@ -90,7 +90,7 @@ func (n ambassador) Start() error {
 		network.MaxReprocessBufferSize)
 	conn, _, err := n.eventManager.Pool().Acquire(context.Background())
 	if err != nil {
-		return fmt.Errorf("time-out acquiring NATS connection from pool: %w", err)
+		return fmt.Errorf("failed to subscribe to REPROCESS event stream: %w", err)
 	}
 
 	err = stream.Subscribe(conn, "VCR", fmt.Sprintf("%s.*", events.ReprocessStream), n.handleReprocessEvent)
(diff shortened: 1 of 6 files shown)


vcr/ambassador.go
```diff
@@ -90,7 +90,7 @@ func (n ambassador) Start() error {
 		network.MaxReprocessBufferSize)
 	conn, _, err := n.eventManager.Pool().Acquire(context.Background())
 	if err != nil {
-		return fmt.Errorf("failed to subscribe to REPROCESS event stream: %v", err)
+		return fmt.Errorf("time-out acquiring NATS connection from pool: %w", err)
```
woutslakhorst, line 93: This is weird:

- is a timeout the only error you can get?
- why a timeout if the given context is Background?

vdr/ambassador.go
```diff
@@ -95,7 +95,7 @@ func (n *ambassador) Start() error {
 		network.MaxReprocessBufferSize)
 	conn, _, err := n.eventManager.Pool().Acquire(context.Background())
 	if err != nil {
-		return fmt.Errorf("failed to subscribe to REPROCESS event stream: %v", err)
+		return fmt.Errorf("time-out acquiring NATS connection from pool: %w", err)
```
woutslakhorst, line 98: same comment as the other ambassador

woutslakhorst (changes requested):

vcr/ambassador.go
reinkrul: Time-out was the error returned during test, but with this "REPROCESS" error log. But looking at the code, I can't really find out how it is triggered, due to the flow of `Acquire`. I think it's more complex than it should be, will make an issue for refactoring it. I'll revert these error messages for now.

pascaldekloe (approved):

woutslakhorst (approved):

PR #1614, opened 2022-11-02
Update deepmap/oapi-codegen to 1.12.2

gerardsn:
This PR contains
- no functional changes (or shouldn't)
- dependency update
- make gen-api
- some changes to make the above work:
   - moved some schemas from vdr/v1.yaml file to common/ssi_types.yaml
   - deleted vdr_v1_test as codegen was failing on this and moved tests to /api/ssi_types
- fixed typos

And many changes in generated code


reinkrul:
> moved some schemas from vdr/v1.yaml file to common/ssi_types.yaml

I had a discussion with @woutslakhorst whether we allow (backwards compatible) changes OAS between major versions, which we didn't conclude yet.

gerardsn:
`common/ssi_types.yaml` is just a collection of schemas that can be referenced by the other APIs. If I understand it correctly, this does not change the specification.

Having a single schema collection adds to the consistency of our OAS. I started this because we had conflicting definitions of VCs, but it also makes testing a bit easier.

gerardsn:
https://nuts-node.readthedocs.io/en/deps-bump-oapi-codegen/

network/api/v1/api_test.go
```diff
@@ -43,18 +43,15 @@ var payload = []byte("Hello, World!")
...
-		}{
-			AdditionalProperties: map[string]PeerDiagnostics{"peer": {
-				Uptime:               10,
-				Peers:                nil,
-				NumberOfTransactions: 5,
-				SoftwareVersion:      "",
-				SoftwareID:           "aaa",
-			}},
-		},
+		JSON200: &map[string]PeerDiagnostics{"peer": {
```
woutslakhorst, line 46: These first 10 lines (upto the println!!) shouldn't be here at all?

woutslakhorst (changes requested):
ssi_types should also be added to the downloads on the API documentation page.

woutslakhorst (approved):

Issue #1619, opened 2022-11-04
VCR: Replace prefix search with explicit wildcard


reinkrul:
The current VCR search API always uses the prefix matcher, meaning any value that starts with the provided value matches, e.g.:

```json
{
    "path": "/fhir/Task/1"
}
```

matches:
```
/fhir/Task/1
/fhir/Task/10
/fhir/Task/1/etc/etc
/fhir/Task/10/etc/etc
```

This causes issues because most developers assume it matches exactly, causing many questions/bug reports.

**Background**
Prefix matching was originally introduced to support cases where partial search is required, the only current case a user searching for an organization's partial name/location.

**Suggestion**
Make default matching exact, requiring applications to provide the exact value (`/fhir/Task/1` in the example above) to have it returned.

Introduce the wildcard character (`*`) to enable prefix searching for the particular field, with the following constraints:
- It is only used when the search query ends with `*`
- When appearing anywhere else in the string it will be part of the actual matching, e.g. `*/Task/*` can only match values starting with `*`.
- It is only supported on string values (makes little sense on numeric fields or others)

Out of scope:
- Escaping `*` for cases where used in the actual value ("Ziekenhuis de Vijf Sterren *****")

**Impact**
Developers most probably use fully specified search queries for technical fields (e.g. URIs like `NutsAuthorizationCredential.credentialSubject.resource.path`), which will now only return more relevant matches.
Prefix searching for organization name/location will have to change: an asterisk will have to be added.

(e.g. Demo EHR will also have to be fixed)

Also; we should make it clear to developers they need to check the VCs they get in return, to make sure they pick the right one (e.g. filter on `/fhir/Task/1`). The more reason because fuzzy/phonetic indexes may have been applied on the particular field. Just taking whatever `SearchVCs` returns is like using Google's "I'm Feeling Lucky", but then in a security-sensitive context...

PR #1620, opened 2022-11-04
VCR: SearchVCs match exact by default, allow wildcard for notnil/exact match

reinkrul:
Closes https://github.com/nuts-foundation/nuts-node/issues/1619

### DIFF
--- a/docs/pages/integrating/vc.rst
+++ b/docs/pages/integrating/vc.rst
@@ -122,6 +122,8 @@ By default, matching is exact: it only returns the result when the given value e
 There are 2 other matchers for strings:
 - ``"*"`` to match credential fields that contain the field (non-empty)
 - ``*`` as postfix to match credential fields that start with the given string, e.g. ``Hospital Amst*``
+When ``*`` is used anywhere else in the string it won't be interpreted as wildcard and matched as-is.
+Wildcards are not supported for other types than strings.
 
 The following example showcases both matchers:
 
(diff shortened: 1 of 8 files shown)


docs/pages/integrating/vc.rst
```diff
@@ -118,9 +118,12 @@ The following query does not return all `NutsOrganizationCredential` but **all**
     }
 
 To find certain credentials, you'll need to add fields that are required to exist in the desired credential.
-The following query searches for credentials that have a organization name that starts with an empty string.
-Any credential that does not have an organization name will be ignored.
-By default, field selection is done by matching the given value as prefix.
+By default, matching is exact: it only returns the result when the given value exactly matches.
+There are 2 other matchers for strings:
+- ``"*"`` to match credential fields that contain the field (non-empty)
+- ``*`` as postfix to match credential fields that start with the given string, e.g. ``Hospital Amst*``
```
woutslakhorst, line 124: Also add that prefix wildcards are not supported, only postfix

woutslakhorst: Also numbers and booleans don't support it. They are always exact match

woutslakhorst (changes requested):

docs/pages/integrating/vc.rst
reinkrul: I'll make it explicit.

woutslakhorst:
Also: breaking changes in release notes

woutslakhorst (approved):

Issue #1628, opened 2022-11-08
VCR: Revoked VCs spam log using SearchVCs


reinkrul:
When searching for VCs and encountering revoked VCs, the Nuts Node logs an info message for every revoked VC that would otherwise be returned:

```
 level=info msg="Encountered invalid VC, omitting from search results." credentialID="did:nuts:Fhn6hdQQpe29SXPpcASpttJLqZzFyHTtfyRmGkWpzVp3#d9f56379-38ae-4469-814e-f7747da49171" error="credential is revoked" module=VCR
```

This should be a trace log at best (when err = revoked)

woutslakhorst:
Added bug label because logs need to be clean on production

PR #1640, opened 2022-11-11
VCR: SearchVCs now logs invalid VCs on debug to avoid log clutter

reinkrul:
Fixes https://github.com/nuts-foundation/nuts-node/issues/1628


gerardsn (approved):
Should there be an `allowInvalid` similar to the `allowUntrusted` for debugging purposes?

reinkrul:
> Should there be an `allowInvalid` similar to the `allowUntrusted` for debugging purposes?

Not sure, invalid could mean anything, so could be dangerous?

PR #1653, opened 2022-11-15
self-authenticate on startup

gerardsn:
Verifies that the configuration is valid for authenticated connections. This should simplify dealing with errors on the users side.

still needs:
- documentation
- double check on logged errors


gerardsn:
replaced by #1695

PR #1657, opened 2022-11-16
Fix crl validator tests

stevenvegt:
The CRL testdata expired on 15 november 2022.
This PR adds
* a `nowFunc` to `crl/validator` which can be overridden during testing to set it to a valid date.
* an extra test to check if a failing crl is actually resulting in an error.

### DIFF
--- a/crl/validator_test.go
+++ b/crl/validator_test.go
@@ -124,9 +124,6 @@ func TestValidator_IsRevoked(t *testing.T) {
 	nowFunc = func() time.Time {
 		return time.Date(2021, 12, 1, 0, 0, 0, 0, time.UTC)
 	}
-	t.Cleanup(func() {
-		nowFunc = time.Now
-	})
 
 	data, err := os.ReadFile(pkiOverheidCRL)
 	if !assert.NoError(t, err) {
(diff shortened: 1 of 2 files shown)


woutslakhorst (approved):

crl/validator_test.go
```diff
@@ -142,12 +146,27 @@ func TestValidator_IsRevoked(t *testing.T) {
 		assert.True(t, crlValidator.IsSynced(0))
 	})
 
-	t.Run("should return false if the certificate was not revoked even though the bit was set", func(t *testing.T) {
-		t.Parallel()
+	t.Run("should return false if the crl is expired", func(t *testing.T) {
+		oldNowFunc := nowFunc
+		nowFunc = func() time.Time {
```
reinkrul, line 151: I always wonder what happens at the end of the function, should you restore `nowFunc` in a defer? Or are package-level variables re-initialized for every test?

stevenvegt: For subtests you need to restore it, for complete different test package, I don't know.

crl/validator_test.go
```diff
@@ -142,12 +146,27 @@ func TestValidator_IsRevoked(t *testing.T) {
 		assert.True(t, crlValidator.IsSynced(0))
 	})
 
-	t.Run("should return false if the certificate was not revoked even though the bit was set", func(t *testing.T) {
-		t.Parallel()
+	t.Run("should return false if the crl is expired", func(t *testing.T) {
+		oldNowFunc := nowFunc
```
reinkrul, line 150: I think you're referring to the inline now func within the function now, due to shadowing. So give it an alternative name.

And restore the `nowFunc` in a `defer` or `t.Cleanup()` to make sure it's executed if the function exits prematurely.

PR #1663, opened 2022-11-18
Docs: where to find trust anchors

reinkrul:


stevenvegt (changes requested):
Shouldn't this info come in the paragraph below, which introduces the concept of TLS certificates?

stevenvegt (approved):

PR #1669, opened 2022-11-18
Increment client backoff when server returned unauthenticated

### LINKED ISSUE #1265: Network: backoff when connection fails on protocol level
The outbound connector backoff works when the connection fails on TCP/TLS level (e.g. time-out, DNS error, etc). But then, when a connection can be made (TLS handshake succeeds) but an application protocol level check fails (e.g. node DID authentication), the backoff isn't increased. The node immediately tries to reconnect, while it is often a non-temporary error (e.g. node DID misconfigured). In this case (and maybe other cases), the backoff should be increased as well.

woutslakhorst:
fixes #1265 

store grpc error before closing connection, can then be used to increment backoff
currently backoff only increased for Code==16 (Unauthenticated)

### DIFF
--- a/network/transport/grpc/connection.go
+++ b/network/transport/grpc/connection.go
@@ -365,9 +365,6 @@ func (mc *conn) IsProtocolConnected(protocol Protocol) bool {
 }
 
 func (mc *conn) CloseError() *status.Status {
-	mc.mux.RLock()
-	defer mc.mux.RUnlock()
-
 	return mc.status.Load()
 }
 
--- a/network/transport/grpc/connection.go
+++ b/network/transport/grpc/connection.go
@@ -91,11 +91,8 @@ type Connection interface {
 	// IsProtocolConnected returns whether the given protocol is active on the connection.
 	IsProtocolConnected(protocol Protocol) bool
 
-	// ErrorStatus returns the status when the connection closed with an error or nil otherwise
-	ErrorStatus() *status.Status
-
-	// SetErrorStatus sets the status when the connection closed with an error
-	SetErrorStatus(status *status.Status)
+	// CloseError returns the status when the connection closed with an error or nil otherwise
+	CloseError() *status.Status
 }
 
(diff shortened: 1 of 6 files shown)


network/transport/grpc/connection.go
```diff
@@ -107,6 +113,7 @@ type conn struct {
 	peer             atomic.Value
 	ctx              context.Context
 	cancelCtx        func()
+	errStatus        *status.Status
```
reinkrul, line 116: you could use a `atomic.Pointer` (with generics), to avoid having to use the mutex;

```go
status atomic.Pointer[status.Status]
```

network/transport/grpc/connection.go
```diff
@@ -90,6 +90,12 @@ type Connection interface {
 
 	// IsProtocolConnected returns whether the given protocol is active on the connection.
 	IsProtocolConnected(protocol Protocol) bool
+
+	// ErrorStatus returns the status when the connection closed with an error or nil otherwise
+	ErrorStatus() *status.Status
+
+	// SetErrorStatus sets the status when the connection closed with an error
+	SetErrorStatus(status *status.Status)
```
reinkrul, line 98: I don't think the interface should allow access to this func. Replace it with `atomic.Pointer` and it can go anyways

reinkrul (changes requested):

network/transport/grpc/connection.go
```diff
@@ -359,6 +364,13 @@ func (mc *conn) IsProtocolConnected(protocol Protocol) bool {
 	return mc.ctx != nil && mc.streams[protocol.MethodName()] != nil
 }
 
+func (mc *conn) CloseError() *status.Status {
+	mc.mux.RLock()
```
reinkrul, line 368: you don't need the mutex now?

reinkrul (changes requested):

network/transport/grpc/connection.go
woutslakhorst: true

reinkrul (approved):

PR #1677, opened 2022-11-21
VCR: Fix validator allowing localParameters

reinkrul:
Latest version of `piprate/json-gold` introduces `SafeMode` which causes `Expand()` to fail when it encounters properties not defined in the VCs context.

Nuts JSON-LD context also wasn't entirely correct using `@vocab`, we need to determine whether we can safely fix it.


stevenvegt (changes requested):
A new context should also be published on the nuts.nl website.
Also, for each context change, create a test credential with proof and make sure all previous versions are still compatible and still have a valid signature.

reinkrul:
The plot thickens:
- Current Nuts v1 context: canonicalized form includes the `localParameters` property, but not the child properties under it. Meaning the `localParameters` properties can be changed without invalidating the signature. Although JSON-LD playground includes them, but weird enough, not completely correctly.
- The suggested context change (adding `@vocab`) fixes this, but this invalidates current signatures (since now canonicalized form does contain the child properties).

Apparently, there is a `@json` type which does exactly what we want (https://www.w3.org/TR/json-ld11/#json-literals):
> At times, it is useful to include JSON within JSON-LD that is not interpreted as JSON-LD. Generally, a JSON-LD processor will ignore properties which don't map to [IRIs](https://tools.ietf.org/html/rfc3987#section-2), but this causes them to be excluded when performing various algorithmic transformations. But, when the data that is being described is, itself, JSON, it's important that it survives algorithmic transformations.

Tested this; canonicalized form does contain the JSON as-is (but normalized).

vcr/signature/proof/jsonld.go
```diff
@@ -86,6 +86,7 @@ func (p LDProof) Verify(document Document, suite signature.Suite, key crypto.Pub
 	if err != nil {
 		return err
 	}
+	println(string(canonicalDocument))
```
stevenvegt, line 89: You missed a spot over here.

stevenvegt (changes requested):
almost done!

reinkrul:
> +update context on website

https://github.com/nuts-foundation/nuts-webpresence/pull/135

stevenvegt (approved):

PR #1678, opened 2022-11-21
Serialise DIDMan calls per DID

### LINKED ISSUE #1676: DIDMan: prevent parallel updates on DID Documents
Currently creates conflicting documents.

stevenvegt:
Fixes #1676.

For each DID, create a mutex in a sync.Map.
Before each `docResolver.Resolve` combined with a `vdr.Update` a Lock is needed to prevent accidental parallel updates.

Note: The mutexes in the map are reused and not deleted.
A `sync.Mutex` is only 2x32 bit ints large, so this should not be a problem.

### DIFF
--- a/didman/didman.go
+++ b/didman/didman.go
@@ -59,14 +59,14 @@ type keyedMutex struct {
 func (m *keyedMutex) Lock(key string) func() {
 	value, _ := m.mutexes.LoadOrStore(key, &sync.Mutex{})
 	mtx := value.(*sync.Mutex)
-	log.Logger().Debugf("aquiring lock for: %s", key)
+	log.Logger().Tracef("aquiring lock for: %s", key)
 	mtx.Lock()
-	log.Logger().Debugf("lock aquired for: %s", key)
+	log.Logger().Tracef("lock aquired for: %s", key)
 
 	// use this to unlock the mutex
 	return func() {
 		mtx.Unlock()
-		log.Logger().Debugf("unlocked for: %s", key)
+		log.Logger().Tracef("unlocked for: %s", key)
 	}
 }
 


reinkrul:
Don't forget the map key (50-ish bytes) and the map entry. Not the nicest of solutions, but will realistically only noticable memory  usage during stress tests (when you alter 100.000s of DID documents)

didman/didman.go
```diff
@@ -46,6 +47,29 @@ const ModuleName = "Didman"
...
+type keyedMutex struct {
+	mutexes sync.Map
+}
+
+// Lock tries to lock a mutex for a certain key.
+// It returns a function which can be used to unlock mutex for the key
+func (m *keyedMutex) Lock(key string) func() {
+	value, _ := m.mutexes.LoadOrStore(key, &sync.Mutex{})
+	mtx := value.(*sync.Mutex)
+	log.Logger().Debugf("aquiring lock for: %s", key)
```
reinkrul, line 62: trace might be more applicable

didman/didman.go
```diff
@@ -87,6 +114,9 @@ func (d *didman) Name() string {
 }
 
 func (d *didman) AddEndpoint(id did.DID, serviceType string, u url.URL) (*did.Service, error) {
+	unlockFn := d.callSerializer.Lock(id.String())
```
reinkrul, line 117: I think you can even make this a one-liner

stevenvegt: Yes, but that will be unreadable:
`defer d.callSerializer.Lock(id.String())()`

reinkrul: I can read that 🤣

reinkrul (approved):

Issue #1607, opened 2022-10-31
OpenAPI: Switch to strict server interfaces


reinkrul:
Starting v1.12 `deepmap/oapi-codegen` has a "strict" flag, which instructs the generator to produce "strict" server interfaces, which only support the exact request/response bodies as defined in the OAS: https://github.com/deepmap/oapi-codegen#strict-server-generation

This should help validating request bodies and avoid returning invalid (according to the OAS) response bodies.

woutslakhorst:
How will response bodies be validated? Currently our API impls return data directly via echo.

reinkrul:
You then return the response body, instead of an (Echo) error. So you can't return anything else than spec'd.

woutslakhorst:
What is the impact on the error return handler? Currently errors are automatically converted to the correct body. Would this then be custom?

PR #1680, opened 2022-11-21
Network: strict OpenAPI interface for API

reinkrul:
PoC of https://github.com/nuts-foundation/nuts-node/issues/1607, so we can decide whether we think this is better.

I personally like the RPC-style API functions more (because they are safer and more explicit), but the naming is bad (long request/response object names) leading to code that doesn't read very nice.

But it still prevents bugs where the API implementation would handle parameters/response codes/types differently than spec'd.

Test code is more or less the same, some ugly casting required but no more weird HTTP request/response recorders.

I'd 4-8 hours to rewrite all APIs and tests. When subtracting generated code from added LoCs, the LoCs with and without strict server interfaces (including tests) are more or less equal.

Toughts?


gerardsn:
I like the strict definitions and structure this creates for a API call.

Some extra consideration related to this change
- CLI commands need to be updated accordingly
- New VC / DID are currently specced to use application/vc+json / application/did+json as content type in transactions, in oapi we defined something else, and implementation returns them as application/json on creation. (There seems to be no consensus on the mediatypes yet.) Using strict interfaces probably requires fixing this issue first.
- `/api/*` currently exists purely for comparing oapi generated structs with our internal struct. Would we be able to drop these tests when adopting the strict interfaces?

woutslakhorst:
It seems that our code looks cleaner (at least less) plus it adds testability for the actual content-type. The echo.Context is gone, also a plus.

What is the default behaviour when param/body parsing results in an error? Is our middleware still returning a nice RFC style Problem?

network/api/v1/api_test.go
```diff
@@ -328,52 +239,40 @@ func TestApiWrapper_Reprocess(t *testing.T) {
...
 func TestWrapper_Preprocess(t *testing.T) {
-	ctrl := gomock.NewController(t)
-	w := &Wrapper{}
-	ctx := mock.NewMockContext(ctrl)
-	ctx.EXPECT().Set(core.OperationIDContextKey, "foo")
-	ctx.EXPECT().Set(core.ModuleNameContextKey, "Network")
-
-	w.Preprocess("foo", ctx)
+	t.SkipNow()
+	// TODO: Test strict middleware
```
gerardsn, line 268: is still needed now the Preprocessed is no longer generated?

test/http/api.go
```diff
@@ -0,0 +1,27 @@
+package http
```
gerardsn, line 1: license + some comments on the methods

gerardsn (changes requested):

test/http/api.go
reinkrul: I don't really care about comments on test funcs?

gerardsn: I'm still slow at reading code / processing new concepts. Oneliners telling me what to expect or when to use it help a lot. Having to go to a test function to figure out what it does requires moving through code and distracts from what you are trying to test.

Although these functions are probably simple enough to understand

gerardsn (approved):

PR #1682, opened 2022-11-22
HTTP: improve test stability by for HTTP server to start

reinkrul:


http/engine_test.go
```diff
@@ -440,6 +451,21 @@ func TestDecodeURIPath(t *testing.T) {
 	})
 }
 
+func assertServerStarted(t *testing.T, address string) {
```
woutslakhorst, line 454: require instead of assert?

woutslakhorst (approved):

http/engine_test.go
reinkrul: We use `assertXYZ` all over the place, so kinda trivial in this case I think

Issue #1674, opened 2022-11-21
Non-finding from pentest


woutslakhorst:
Stuff they mentioned but are not part of the report:

<img width="1080" alt="Screenshot 2022-11-21 at 09 09 53" src="https://user-images.githubusercontent.com/703298/202997747-8672de85-e503-41ce-8082-c768186913cf.png">

PR #1689, opened 2022-11-24
Crypto: cleanup FS backend initialization (NTS-006)

reinkrul:
Closes https://github.com/nuts-foundation/nuts-node/issues/1674


gerardsn (approved):
Should probably include #1707

PR #1695, opened 2022-11-28
Network: self-authenticate on startup

gerardsn:
check if NutsComm is listed on TLS certificate.
Node cannot start in strictmode if there is a misconfiguration in the nodeDID, NutsComm, and TLS cert triangle.


network/network.go
```diff
@@ -411,10 +412,28 @@ func (n *Network) validateNodeDID(nodeDID did.DID) error {
...
+	if n.certificate.Leaf == nil { // n.config.DisableNodeAuthentication is for incoming connections.
+		return nil
+	}
+
+	// Check if the NutsComm endpoint matches a SAN on the certificate
+	var nutsCommURLStr string
+	if err = nutsCommService.UnmarshalServiceEndpoint(&nutsCommURLStr); err != nil {
+		return fmt.Errorf("invalid %s service endpoint: %w", transport.NutsCommServiceType, err)
+	}
+	nutsCommURL, err := url.Parse(nutsCommURLStr)
```
reinkrul, line 430: You can use `transport.ParseNutsCommAddress(nutsCommStr)`

network/network.go
```diff
@@ -411,10 +412,28 @@ func (n *Network) validateNodeDID(nodeDID did.DID) error {
...
 	serviceResolver := doc.NewServiceResolver(n.didDocumentResolver)
 	serviceRef := doc.MakeServiceReference(nodeDID, transport.NutsCommServiceType)
-	_, err = serviceResolver.Resolve(serviceRef, doc.DefaultMaxServiceReferenceDepth)
+	nutsCommService, err := serviceResolver.Resolve(serviceRef, doc.DefaultMaxServiceReferenceDepth)
 	if err != nil {
 		return fmt.Errorf("unable to resolve %s service endpoint, register it on the DID document (did=%s): %v", transport.NutsCommServiceType, nodeDID, err)
 	}
+
+	// Self-authentication requires a certificate. Running non-TLS is evaluated in Configure
+	if n.certificate.Leaf == nil { // n.config.DisableNodeAuthentication is for incoming connections.
```
reinkrul, line 421: shouldn't NutsComm still be checked (presence and validated, but not vs certificate)?

docs/pages/integrating/faq-errors.rst
```diff
@@ -44,4 +44,18 @@ It's probably the latter problem. This typically happens when you create transac
...
+The second situation is fixed by configuring the correct bootstrap node.
+
+Error: `invalid nodeDID configuration`
+*******************************************
+
+When starting your node, the following error can occur:
+
+    invalid nodeDID configuration: <error details>
+
+Correctly configured nodes have a ``NutsComm`` service endpoint on the did document of the configured ``network.nodeDID``,
```
reinkrul, line 56:
```suggestion
Correctly configured nodes have a ``NutsComm`` service endpoint on the DID document of the configured ``network.nodeDID``,
```

docs/pages/integrating/faq-errors.rst
```diff
@@ -44,4 +44,18 @@ It's probably the latter problem. This typically happens when you create transac
...
+and this address must be listed in the Subject Alternative Name section of the configured ``TLS certificate``.
+A misconfiguration in any of these three items means that peers cannot authenticate the connection to your node,
+and in strictmode this will prevent the node from starting. The <error details> will point you to the exact issue.
```
reinkrul, line 59:
```suggestion
and in strictmode this will prevent the node from starting. The ``<error details>`` will point you to the exact issue.
```

docs/pages/integrating/faq-errors.rst
```diff
@@ -44,4 +44,18 @@ It's probably the latter problem. This typically happens when you create transac
...
+Remove the nodeDID from the configuration in strictmode to view or change the ``NutsComm`` address.
```
reinkrul, line 60:
```suggestion
When in strict mode, remove the nodeDID from the configuration to view or change the ``NutsComm`` address.
```

docs/pages/integrating/faq-errors.rst
```diff
@@ -44,4 +44,18 @@ It's probably the latter problem. This typically happens when you create transac
...
+In non-strictmode, the misconfiguration is logged as an error, but is not fatal.
```
reinkrul, line 61:
```suggestion
In non-strictmode, the misconfiguration is logged as an error but it doesn't prevent the node from starting.
```

reinkrul (changes requested):

network/network.go
gerardsn: This should probably return an error, not nil. Strict mode won't get here, and in non-strict mode it doesn't do any harm to have an error logged for missing TLS certificate.

network/network.go
reinkrul: Debugging setups probably don't use strict mode?

Now you're also missing the validation of `grpc://` (which `ParseNutsCommAddress` does)

gerardsn: Reserved addresses and IPs are not prohibited in the RFCs. Shouldn't they be added there first? (I'm assuming you can't register these addresses on a PKIo certificate anyway?)

network/network.go
gerardsn: moved this further down the checklist and now returns an error

network/network.go
gerardsn: Our testing makes use of reserved addresses (localhost, local, etc.). I've fixed everything for this repo, but e2e still needs to be done

reinkrul (approved):

woutslakhorst:
e2e fail, they are probably incorrectly configured?

Issue #1196, opened 2022-06-16
Turn of "examples" in swagger UI


woutslakhorst:
They are auto-filled with garbage. We should not show them on readthedocs

woutslakhorst:
Not entirely fixed by #1697

PR #1697, opened 2022-11-28
Docs: improve OpenAPI examples to reflect actual data

reinkrul:
During past hackatons people keep getting confused about incorrect examples. Mostly about service ID which appears to be the service type in some examples, but which is actually the hash of the service.

This PR fixes examples containing DIDs (DIDs, verification method IDs and service IDs).

Partially fixes https://github.com/nuts-foundation/nuts-node/issues/1196

### DIFF
--- a/docs/_static/didman/v1.yaml
+++ b/docs/_static/didman/v1.yaml
@@ -330,7 +330,7 @@ components:
         serviceEndpoint:
           description: An endpoint URL or a reference to another service.
           type: string
-          example:
+          examples:
             - 'did:nuts:EwVMYK2ugaMvRHUbGFBhuyF423JuNQbtpes35eHhkQic/serviceEndpoint?type=eOverdracht-fhir'
             - 'http://example.com/some/service/endpoint'
     EndpointProperties:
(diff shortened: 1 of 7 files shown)


docs/_static/didman/v1.yaml
```diff
@@ -330,7 +330,9 @@ components:
         serviceEndpoint:
           description: An endpoint URL or a reference to another service.
           type: string
-          example: 'did:nuts:123/serviceEndpoint?type=eOverdracht-fhir'
+          example:
```
stevenvegt, line 333: If you want to show multiple examples, I think you need to use `examples:`

docs/_static/didman/v1.yaml
```diff
@@ -344,7 +346,9 @@ components:
         endpoint:
           description: An endpoint URL or a reference to another service.
           type: string
-          example: 'did:nuts:123/serviceEndpoint?type=eOverdracht-fhir'
+          example:
```
stevenvegt, line 349: `examples`

stevenvegt (changes requested):
Change `example` to `examples` when you want to show multiple values.

reinkrul:
`examples` is not allowed, but I found a map of examples is the most idiomatic way (although not perfect since it is presented as the actual example).

woutslakhorst (approved):

PR #1700, opened 2022-11-29
Docs: tell people not to use key rotation

reinkrul:


docs/pages/deployment/key-rotation.rst
```diff
@@ -3,6 +3,10 @@
 Key rotation procedure
 ######################
 
+.. warning::
+    Rotating keys currently prevents private transactions (e.g. ``NutsAuthorizationCredential``) from being received properly:
```
woutslakhorst, line 7: It's not that black and white. Only happens if keys rotate and credentials are in transit at the same time.

reinkrul: Which you don't know, so it can't be used safely in any circumstance?

woutslakhorst (approved):

PR #1705, opened 2022-11-29
Fix crypto fs storage

### LINKED ISSUE #1704: NTS-008: Crypto FS storage file creation permissions
Files are created with `666`, should be `600`.
Files can overwrite files with same name (very small change since the filename is the hash of the public key).

stevenvegt:
Fixes #1704 
File permissions to 0600 and don't create when already exists.

### DIFF
--- a/crypto/storage/fs.go
+++ b/crypto/storage/fs.go
@@ -95,7 +95,7 @@ func (fsc *fileSystemBackend) GetPrivateKey(kid string) (crypto.Signer, error) {
 // SavePrivateKey saves the private key for the given key to disk. Files are postfixed with '_private.pem'. Keys are stored in pem format.
 func (fsc *fileSystemBackend) SavePrivateKey(kid string, key crypto.PrivateKey) error {
 	filenamePath := fsc.getEntryPath(kid, privateKeyEntry)
-	outFile, err := os.Create(filenamePath)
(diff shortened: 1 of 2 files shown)


reinkrul (commented):
What permission should the directory (createdirs) have?

stevenvegt:
> What permission should the directory (createdirs) have?
ROS advises 0700. That is part of NTS-006.

crypto/storage/fs.go
```diff
@@ -95,7 +95,7 @@ func (fsc *fileSystemBackend) GetPrivateKey(kid string) (crypto.Signer, error) {
 // SavePrivateKey saves the private key for the given key to disk. Files are postfixed with '_private.pem'. Keys are stored in pem format.
 func (fsc *fileSystemBackend) SavePrivateKey(kid string, key crypto.PrivateKey) error {
 	filenamePath := fsc.getEntryPath(kid, privateKeyEntry)
-	outFile, err := os.Create(filenamePath)
+	outFile, err := os.OpenFile(filenamePath, os.O_WRONLY|os.O_CREATE|os.O_EXCL, os.FileMode(0600))
```
reinkrul, line 98: this function could be simpler using `os.WriteFile()` which doesn't involve handling file handles

reinkrul (approved):

crypto/storage/fs.go
stevenvegt: Almost, but that one allows overwriting existing files. Here we include the `os.O_EXCL` flag to prevent that.

PR #1718, opened 2022-12-01
Auth: clearly distinguish errors returned by remote nodes

### LINKED ISSUE #1527: Auth: sanitize errors returned by CreateAccessToken
In development it should help the caller (remote Nuts node) identify the problem, but on production the errors should be sanitized to avoid leaking internals, which could point a hacker to a vulnerability.

in production (strictmode=true) it should have a generic error message.

reinkrul:
Partially fixes #1527 

- More clearly indicates that the error is returned by the remote Nuts node
- Stop returning the received response code as the local API call's response code, a downstream error should have a specific status code, not influenced by the downstream system being called. Otherwise, you never know which status code is going to be returned, which can even be a vulnerability.

### DIFF
--- a/docs/_static/auth/v1.yaml
+++ b/docs/_static/auth/v1.yaml
@@ -191,7 +191,7 @@ paths:
 
         error returns:
         * 400 - one of the parameters has the wrong format
-        * 502 - the authorizer could not be reached or returned an error
+        * 503 - the authorizer could not be reached or returned an error
       tags:
         - auth
       requestBody:
--- a/docs/_static/auth/v1.yaml
+++ b/docs/_static/auth/v1.yaml
@@ -191,6 +191,7 @@ paths:
 
         error returns:
         * 400 - one of the parameters has the wrong format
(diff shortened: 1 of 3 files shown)


woutslakhorst:
be careful with the "partial fixes" comments. Github still reads "fixes X" and thus auto closes the issue.

docs/_static/auth/v1.yaml
```diff
@@ -191,6 +191,7 @@ paths:
 
         error returns:
         * 400 - one of the parameters has the wrong format
+        * 502 - the authorizer could not be reached or returned an error
```
woutslakhorst, line 194: This should be part of the generic errors (502) which might suggest that 502 is a bad idea to return when the remote server isn't responding. 503 might be better. 502's are returned by reverse proxies when the internal server is offline. Not really part of this PR, may be a new issue.

woutslakhorst (changes requested):

woutslakhorst (approved):

PR #1719, opened 2022-12-02
Core: remove server response body HTTP client errors

reinkrul:
Since the server may return anything (e.g., binary response, or a very large blob), it's unsafe to include it in the error message: it gets logged and sometimes returned to the client.

Better to record it in the error struct, but not to include it in the error message.

Since it's still useful to print when a CLI command fails (due to non-OK server response), introduced a wrapper for all client commands.

### DIFF
--- a/auth/api/v1/client.go
+++ b/auth/api/v1/client.go
@@ -20,13 +20,13 @@ package v1
 
 import (
 	"context"
-	"github.com/nuts-foundation/nuts-node/auth/log"
 	"net/http"
 	"net/url"
 	"strings"
 	"time"
 
 	"github.com/nuts-foundation/nuts-node/auth"
+	"github.com/nuts-foundation/nuts-node/auth/log"
 	"github.com/nuts-foundation/nuts-node/core"
 )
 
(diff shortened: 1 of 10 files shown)


woutslakhorst (commented):
Just to check: response bodies are no longer logged server side when an error occurs? (Could be added to trace logging then?)

reinkrul:
Added debug logging for response body of remote Nuts Node's CreateAccessToken response.

Clients now just pass the response in the returned error, with a centralized error handling wrapper in root command registerer, that logs the body if the returned error is a RemoteServerError.

auth/api/v1/client.go
```diff
@@ -20,6 +20,7 @@ package v1
 
 import (
 	"context"
+	"github.com/nuts-foundation/nuts-node/auth/log"
```
woutslakhorst, line 23: sorting

core/http_client.go
```diff
@@ -26,11 +26,23 @@ import (
 	"net/http"
 )
 
+// RemoteServerError describes an error returned when invoking a remote server.
+type RemoteServerError struct {
```
woutslakhorst, line 30: I thought it was a custom error with specific contents. But it's actually a HttpError (>= 400)

woutslakhorst (changes requested):
Just the error name is misleading

woutslakhorst (approved):

PR #1727, opened 2022-12-02
Auth: CreateAccessToken must only return error details for validation errors

reinkrul:
This PR sanitizes errors created by `CreateAccessToken`, which is called by remote Nuts nodes. It should not return any internal error details, since they could point an attacker to possible vulnerabilities. Validation errors should be fine to return. 

This behavior is only applied in strict mode: in non-strict mode, it just returns the error details, as it did before (to aid debugging in non-production environments).

In order to achieve this, the following things changed:
- Always log the returned error as warning
- Move OAuth service interface to `oauth` package
- Rename `OAuthClient` to client
- removed validation checks from`buildAccessToken` that are already performed by `validateAccessTokenRequest`

Fixes https://github.com/nuts-foundation/nuts-node/issues/1527


woutslakhorst (changes requested):
I don't see a test that checks if the description is empty when securemode is true

reinkrul:
Note: it will only fail with a generic error when signing fails, because it can't find a signing key or the signing itself fails.

woutslakhorst (changes requested):

woutslakhorst (approved):

PR #1728, opened 2022-12-03
Mark ssi_types.yaml as OpenAPI file

reinkrul:
Without it, OpenAPI Generator Tools for Java doesn't recognize it as a YAML OpenAPI file.

### DIFF
--- a/docs/_static/common/ssi_types.yaml
+++ b/docs/_static/common/ssi_types.yaml
@@ -1,3 +1,4 @@
+openapi: "3.0.0"
 components:
   schemas:
 # VC


woutslakhorst (commented):
This doesn't conflict when included by the other files? It should not be used as root file to generate code from but be included from other files.

gerardsn:
OAPI files downloaded from readthedocs are missing the required folder structure for the references to work. It's a different issue, but could this be something other generators get stuck on?

woutslakhorst:
> OAPI files downloaded from readthedocs are missing the required folder structure for the references to work. It's a different issue, but could this be something other generators get stuck on?

Better to offer a zip/tarball?

reinkrul:
We use it to generate a common model, otherwise you have that model duplicated. This is natively supported by OpenAPI 3.1.0, which makes `paths` optional (support is coming "soon" in OpenAPI Generator Tools).

OpenAPI Generator Tools (and I think our Golang deepmap/codegen as well) don't support specifiying an alternative package for specific types to fix this without separate generation for the common model (ssi_types.yaml), as far as I know.

This approach works in both OpenAPI Generator Tools and deepmap/codegen.

gerardsn (approved):
Everything in the common folder (`error_response.yaml` and `ssi_types.yaml`) should not be used in generators on its own. The components should only be referenced from other files. 

We currently do have a build target for `ssi_types.yaml`, but that only exists for testing so we can make sure the generated components are consistent with the structs defined elsewhere in our code.

but adding this doesn't break anything either, right?

gerardsn:
> > OAPI files downloaded from readthedocs are missing the required folder structure for the references to work. It's a different issue, but could this be something other generators get stuck on?
> 
> Better to offer a zip/tarball?

creates #1739 for this

PR #1741, opened 2022-12-06
remove VDR memstore

woutslakhorst:
It was only used in test and had custom logic in it. Switching to the BBolt one (also in memory) actually made some tests fail!

Part of #1709


vdr/store/store.go
```diff
@@ -288,6 +294,10 @@ func (s *store) Resolve(id did.DID, metadata *vdr.ResolveMetadata) (returnDocume
 				return err
 			}
 
+			if metadataRecord.Deactivated && metadata == nil {
```
reinkrul, line 297: this belongs in a different PR

vdr/store/store.go
```diff
@@ -47,20 +47,26 @@ const (
...
-	return &store{storeProvider: storeProvider}
+func NewStore(storeProvider storage.Provider, inMemory bool) vdr.Store {
+	return &store{storeProvider: storeProvider, inMemory: inMemory}
 }
 
 func (s *store) Configure(_ core.ServerConfig) error {
 	var err error
-	s.db, err = s.storeProvider.GetKVStore(didStoreName, storage.PersistentStorageClass)
+	if s.inMemory {
+		s.db, err = s.storeProvider.GetKVStore(didStoreName, storage.VolatileStorageClass)
```
reinkrul, line 66: Volatile doesn't mean in-memory per se. Just that the storage may be lost without losing data (e.g. connection backoffs).

Other tests use `StaticKVProvider` with a precreated test-BBolt store (`CreateTestBBoltStore`, `storage/test.go:38`).

reinkrul (changes requested):

vdr/store/store.go
woutslakhorst: Not really, since the in memory store did it correctly. With the change to the other store, tests fail.

reinkrul (approved):

PR #1743, opened 2022-12-06
VCR: Retry VC on certain JSON-LD errors

reinkrul:


gerardsn (approved):

woutslakhorst (changes requested):
This changes the default behaviour from recoverable to non-recoverable. So any unknowns will be marked unrecoverable and can thus no longer be found in the failed events.

ContextURLNotAllowed should be marked unrecoverable and the rest as recoverable.

reinkrul:
> This changes the default behaviour from recoverable to non-recoverable. So any unknowns will be marked unrecoverable and can thus no longer be found in the failed events.
> 
> ContextURLNotAllowed should be marked unrecoverable and the rest as recoverable.

This is still how it works, only added thing here is `ContextURLNotAllowedErr` is not recoverable?

woutslakhorst:
Hmm true

woutslakhorst (approved):

reinkrul:
To fix :

- non-recoverable errors should only be logged, not retried (out of scope)
- migrate current failed events, to remove unrecoverable cases this PR fixes

woutslakhorst:
@reinkrul

PR #1744, opened 2022-12-06
move and rename doc package for clarity

woutslakhorst:
part of #1709 

the `doc` and `store` package are too generic for what they do. Placed the doc package under `diddocuments` and renamed it to `dochelper` to clarify they only contain helper functionality and not the actual data.

The restructuring is also required to prevent circular dependencies.


vdr/diddocuments/dochelper/creator.go
```diff
@@ -14,10 +13,9 @@
  *
  * You should have received a copy of the GNU General Public License
  * along with this program.  If not, see <https://www.gnu.org/licenses/>.
- *
  */
 
-package doc
+package dochelper
```
gerardsn, line 18: I keep reading this as dochel-per

gerardsn (commented):
Aside from the empty `diddocuments` folder (that I'm assuming will be filled by another PR), this PR looks alright. Without further details on the redesign of the VDR it is hard to give a proper review of this

reinkrul:
Not sure about the rename; without knowing what will come after this (empty diddocuments package), it's hard to judge.

In itself, I find it an uglier setup than what it was; `diddocuments` can just be omitted, because VDR is only about DID documents (it could be `documents`).

`dochelpers` is now a sort of util package, since `helpers` is just a synonym for `util`. I get the feeling that everything in it might belong in the root of `vdr`, or split it up in `vdr/resolve` and (failed to come up with a good name..).

Why is `store` too generic?

woutslakhorst:
@gerardsn @reinkrul checkout #1709 for the end result. The problem lies with the fact these packages are used outside of the VDR. Using store as package outside the VDR can mean anything. The root cmd would say something like `store.New`. Which store? VDR, VCR, DAG, Connections, etc. Same goes for the helpers, they are also used outside the VDR and read as `doc.Resolver`. `doc` does not exclusively mean DID Document.

woutslakhorst:
I could rename `dochelper` to `didservice`?

reinkrul:
> The root cmd would say something like store.New. Which store? VDR, VCR, DAG, Connections, etc.

These imports are then generally aliased to `vdrStore`, `vcrStore` (same is happening with `cmd` packages).

reinkrul (approved):

PR #1747, opened 2022-12-07
implement ranged ListTransactions (?start=N&end=M)

beardedfoo:
A ranged operation for ListTransactions was always intended and is implemented via this PR. The query string now accepts an optional ?start=N&end=M pattern. Following convention of the surrounding code the start parameter is inclusive and the end parameter is exclusive. These parameters restrict the lamport clock values of the returned transactions. This feature is needed to implement a scalable data viewer client.

### DIFF
--- a/docs/_static/network/v1.yaml
+++ b/docs/_static/network/v1.yaml
@@ -30,7 +30,6 @@ paths:
       summary: "Lists transactions on the DAG"
       description: >
         Lists transactions on the DAG. Since this call returns all transactions on the DAG, care should be taken when there are many of them. A range (start,end) are recommended at all times to avoid overloading the endpoint.
-        TODO: We may need a more elaborate querying interface (pagination, filtering, etc).
 
         error returns:
         * 500 - internal server error
--- a/docs/_static/network/v1.yaml
+++ b/docs/_static/network/v1.yaml
@@ -9,6 +9,23 @@ servers:
   - url: http://localhost:1323
 paths:
   /internal/network/v1/transaction:
+    parameters:
+      - name: start
+        description: "Inclusive start of range (in lamport clock)"
+        example: 1
+        in: query
+        required: false
+        schema:
+          type: integer
+          minimum: 0
+      - name: end
+        description: "Exclusive stop of range (in lamport clock)"
+        example: 1000
+        in: query
+        required: false
+        schema:
+          type: integer
+          minimum: 0
     get:
       summary: "Lists the transactions on the DAG"
       description: >
--- a/docs/_static/network/v1.yaml
+++ b/docs/_static/network/v1.yaml
@@ -11,7 +11,7 @@ paths:
   /internal/network/v1/transaction:
     parameters:
       - name: start
-        description: "Inclusive start of range (in lamport clock)"
+        description: "Inclusive start of range (in lamport clock); default=0"
         example: 1
         in: query
         required: false
@@ -19,7 +19,7 @@ paths:
           type: integer
           minimum: 0
       - name: end
-        description: "Exclusive stop of range (in lamport clock)"
+        description: "Exclusive stop of range (in lamport clock); default=∞"
         example: 1000
         in: query
         required: false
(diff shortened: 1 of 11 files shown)


network/cmd/cmd.go
```diff
@@ -133,12 +134,39 @@ const sortFlagType = "type"
 
 func listCommand() *cobra.Command {
 	var sortFlag string
+	var rangeStart string
+	var rangeEnd string
```
gerardsn, line 138: missing some tests for the range part of the command

gerardsn:
The PR adds a range query to the `nuts network list` command. Should we have a separate `nuts network range` command where the range arguments are mandatory?

This simplifies parsing and makes deprecation of the list command in the future a bit easier

docs/_static/network/v1.yaml
```diff
@@ -9,6 +9,23 @@ servers:
   - url: http://localhost:1323
 paths:
   /internal/network/v1/transaction:
+    parameters:
```
reinkrul, line 12: The TODO in the description can go now.

       > Lists all transactions on the DAG. Since this call returns all transactions on the DAG, care should be taken when there
        are many of them.
        TODO: By then we'd need a more elaborate querying interface (ranging over timestamps/hashes, pagination, filtering, etc).


Also adjust the "all transactions on the DAG" description

network/cmd/cmd.go
```diff
@@ -151,7 +179,11 @@ func listCommand() *cobra.Command {
 			return nil
 		},
 	}
+
 	cmd.Flags().StringVar(&sortFlag, "sort", sortFlagTime, "sort the results on either time or type")
+	cmd.Flags().StringVar(&rangeStart, "start", "", "inclusive start of lamport clock range")
```
reinkrul, line 184: isn't there an `IntVar`?

reinkrul (changes requested):

beardedfoo:
I think we'll keep the list command but implement pagination for it. We don't need to stop users from doing something against every transaction (there are legitimate scripting use cases for this) but rather stop this from happening at the API level. In this case I think it's best to add these as parameters rather than a separate command.

> The PR adds a range query to the `nuts network list` command. Should we have a separate `nuts network range` command where the range arguments are mandatory?
> 
> 
> 
> This simplifies parsing and makes deprecation of the list command in the future a bit easier

network/cmd/cmd.go
beardedfoo: Yes but I don't think the behavior for an unspecified value is very nice. StringVar is nice in that you get an empty string if the option was not specified.

I'll take a second look and make sure IntVar  doesn't work for our use case.

gerardsn (approved):

docs/_static/network/v1.yaml
```diff
@@ -9,12 +9,28 @@ servers:
...
+          minimum: 0
     get:
-      summary: "Lists the transactions on the DAG"
+      summary: "Lists transactions on the DAG"
       description: >
-        Lists all transactions on the DAG. Since this call returns all transactions on the DAG, care should be taken when there
-        are many of them.
-        TODO: By then we'd need a more elaborate querying interface (ranging over timestamps/hashes, pagination, filtering, etc).
+        Lists transactions on the DAG. Since this call returns all transactions on the DAG, care should be taken when there are many of them. A range (start,end) are recommended at all times to avoid overloading the endpoint.
+        TODO: We may need a more elaborate querying interface (pagination, filtering, etc).
```
reinkrul, line 33:
```suggestion
```

reinkrul (changes requested):

docs/_static/network/v1.yaml
beardedfoo: I had left this in (since many things were mentioned other than just ranges), but I'm happy to take it out and say that if we want more functionality we can do it with some kind of /search endpoint instead of the list endpoint.

reinkrul: I'd say just remove it as we don't have any specific plans for now.

PR #1748, opened 2022-12-07
Notifier: add timestamp of last notification attempt

### LINKED ISSUE #1731: Network failed event list should return moment last retried
Helpful for debugging.

```
      {
        "error": "",
        "hash": "a7c62cef74e0c0891ae0ac8afbf7cddf82c148345caa991d8c48418252cd915d",
        "retries": 20,
        "transaction": "a7c62cef74e0c0891ae0ac8afbf7cddf82c148345caa991d8c48418252cd915d",
        "type": "payload"
      },
```

gerardsn:
closes #1731


docs/_static/network/v1.yaml
```diff
@@ -215,6 +215,9 @@ components:
         retries:
           description: Number of times the event has been retried.
           type: integer
+        latest:
```
reinkrul, line 218: `latest` doesn't say the latest of what, ("latest event"?)

Maybe `last_publish`, `last_attempt`, `last_delivery_attempt`, `last_retry`? (I find "retry" not the best name, since the first attempt isn't a retry, I think?)

(not 100% sure about last/latest)

network/dag/notifier.go
```diff
@@ -99,12 +102,14 @@ type NotifierOption func(notifier *notifier)
...
 type Event struct {
-	// Type of an event, can be used to filter
+	// Type of event, can be used to filter
 	Type string `json:"type,omitempty"`
 	// Hash is the ID of the Event, usually the same as the dag.Transaction.Ref()
 	Hash hash.SHA256Hash `json:"Hash"`
 	// Retries is the current number of retries
 	Retries int `json:"retries"`
+	// Latest records the timestamp of the last notification attempt. It is not used in the backoff calculation.
+	Latest *time.Time `json:"latest,omitempty"`
```
reinkrul, line 112: Why can it be nil? An event is always attempted to be delivered at least once (initial publish)?

Makes for a harder interface

network/dag/notifier.go
```diff
@@ -118,6 +123,7 @@ func (j *Event) UnmarshalJSON(bytes []byte) error {
 		Type        string          `json:"type,omitempty"`
 		Hash        hash.SHA256Hash `json:"Hash"`
 		Retries     int             `json:"retries"`
+		Latest      *time.Time      `json:"latest,omitempty"`
```
reinkrul, line 126: why is this unmarshal function here 🤨

reinkrul (changes requested):

docs/_static/network/v1.yaml
gerardsn: Will change to `latest_notification_attempt` as it is probably the best fit considering the namespace this lives in, but it is very length . Anything containing `last` could be misinterpreted as the last time it will be attempted

network/dag/notifier.go
gerardsn: `Save` method only saves the event without notification, so there is no latest attempt time yet

woutslakhorst: also existing entries do not have it

PR #1749, opened 2022-12-08
update to stoabs v1.5.1

woutslakhorst:
New version of stoabs no longer returns an error on GetShelfWriter. This eliminates a lot of error checks


reinkrul (approved):
Could use `go mod tidy`

Issue #1722, opened 2022-12-02
NTS-005: No audit log in Crypto FS backend


stevenvegt:
**Description:**
The Crypto private key FS backend has no audit log, limiting detection of leaks and abuse.

**Technical description:**
The FS backend is storing sensitive key material, any access to it should be logged to a append-only service, so that abuses and leaks can be detected and traced.

**Impact:**
An adversary can access sensitive key-material undetected.

**Recommendation:**
Implement append-only logging for the fs backend.

PR #1759, opened 2022-12-09
Crypto: add audit logging on key operations

reinkrul:
Things to figure out:

- [x] Do we want to log the actor? This is problematic, since Go doesn't have thread-local variables. The only way is to pass around a context with the values in it, but that would mean context passing on (almost) every layer, on probably most of the functions.
- [x] Which operations do we want to audit-log? `Exists`, `List` and `Resolve` don't seem very useful (same info can be derived from DID documents)
- [ ] Which additional data do we want to log? E.g. source IP? Or maybe add the subject (e.g. key ID) as `audit.subject` field?

TODO:
- [x] Add assertions for log to tests
- [x] Documentation
- [x] Add to release notes
- [x] Are there any error cases that need to be logged?
- [x] Test panics
- [x] To get it compatible with current HTTP error logs:
  - [x] Rename `operation` to `event`
  - [x] Include `operationID` which is set in the HTTP layer

Requires https://github.com/nuts-foundation/nuts-node/pull/1782 to avoid key usage all over the place.

Fixes https://github.com/nuts-foundation/nuts-node/issues/1722
Uses https://github.com/nuts-foundation/nuts-node/pull/1758 and https://github.com/nuts-foundation/nuts-node/pull/1782


reinkrul:
Audit logging in service layer, then context passing doesn't need to go too deep.

reinkrul:
Auditable events for crypto are key creation and usage i.m.o.

reinkrul:
Ideally you'd want to log the size of the message being signed (or encrypted), but this isn't (practically) possible, because hashing generally happens outside the crypto module code (e.g. JWX library).

audit/audit.go
```diff
@@ -0,0 +1,72 @@
...
+
+// NewInfo formats audit info
+func NewInfo(actor string, operationModuleName string, operation string) Info {
+	return Info{
+		Actor:     actor,
+		Operation: operationModuleName + "." + operation,
+	}
+}
+
+var auditInfoContextKey = struct{}{}
```
woutslakhorst, line 35: An empty struct could easily collide with types created by other packages/libs?

audit/audit.go
```diff
@@ -0,0 +1,72 @@
...
+
+// Context returns a child context of the given parent context, enriched with the auditable actor and performed operation.
+func Context(parent context.Context, provider func() Info) context.Context {
+	return context.WithValue(parent, auditInfoContextKey, provider)
+}
+
+// InfoFromContext extracts the audit info from the given context.
+func InfoFromContext(ctx context.Context) *Info {
+	actor, ok := ctx.Value(auditInfoContextKey).(func() Info)
```
woutslakhorst, line 44: using actor as var name is confusing.

http/audit.go
```diff
@@ -0,0 +1,37 @@
+package http
```
woutslakhorst, line 1: copyright

docs/pages/deployment/audit-logging.rst
```diff
@@ -0,0 +1,19 @@
...
+    This feature is under development, not all relevant operations might be included in the audit log.
+
+Important events are logged as audit events. Examples are creation of a new cryptographic key pair or its usage when signing or decrypting.
+
+Audit events are logged to application log and can be recognized by the ``log`` field being set to ``audit``.
+In addition, the events contain the following fields:
+
+- ``operation`` which contains name action that was performed
+- ``actor`` which contains the name of the user/system that performed the action.
+  Actors from inside the application (e.g. triggered by a scheduled event) are prefixed with ``app-``.
```
woutslakhorst, line 16: I don't know how valuable these are?

woutslakhorst (changes requested):

audit/audit.go
reinkrul: That's right, I accidentally did this the wrong way. The correct way is;

```
type auditInfoContextKey struct{}

context.WithValue(ctx, auditInfoContextKey{}, value)
```

audit/audit.go
```diff
@@ -0,0 +1,72 @@
...
+	return Info{
+		Actor:     actor,
+		Operation: operationModuleName + "." + operation,
+	}
+}
+
+var auditInfoContextKey = struct{}{}
+
+// Context returns a child context of the given parent context, enriched with the auditable actor and performed operation.
+func Context(parent context.Context, provider func() Info) context.Context {
```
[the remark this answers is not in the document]
reinkrul: Because the audit middleware gets invoked before the API's preprocessors/middleware, which (the latter) sets the module and operation name, which I want to be part of the audit log. But it is not set yet when the audit middleware is invoked.

I realized this can be solved by invoking the audit middleware from the API layer, then it can be just `Info` again instead of a func.

network/dag/notifier.go
```diff
@@ -207,6 +207,11 @@ func NewNotifier(name string, receiverFn ReceiverFn, options ...NotifierOption)
 		option(subscriber)
 	}
 
+	// enrich context with audit info
+	subscriber.ctx = audit.Context(subscriber.ctx, func() audit.Info {
```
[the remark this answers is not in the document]
reinkrul: There is a couple of things to consider here;
- I'm dogmatic when it comes to writing similar events in the audit log; we decided to that private key usage (signing and decryption) should be in the audit log (pentest reported that key _access_ should be audit logged), so not having decryption audit logged is dangerous; if we add an interface in future that decrypts data at user request, we're missing an event in the audit log.
- On the other hand, we're going to remove decryption altogether in the near future (practicality argument)
- But, to address the finding from the pentest we should (i.m.o.) audit decryption as well (compliancy argument)
- Then again, it's already there, removing it costs more than just leaving it there (but more code = more cost to maintain)

So I'm 50/50 on removing v.s. keeping it.

docs/pages/deployment/audit-logging.rst
reinkrul: I'd say we need to log _some_ actor indicating it's from an internal app process. But maybe that should be indicated by an `interface` field? (e.g. `HTTP` or `App`)?

network/network.go
```diff
@@ -263,7 +263,7 @@ func (n *Network) Configure(config core.ServerConfig) error {
 }
 
 // emitEvents is called when a payload is added.
-func (n *Network) emitEvents(event dag.Event) (bool, error) {
+func (n *Network) emitEvents(_ context.Context, event dag.Event) (bool, error) {
```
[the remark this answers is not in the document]
reinkrul: Not required, since there's nothing to audit here. Although you could say the audit info should travel with the event to the subscriber (then, a retried event retains the original audit info fields, e.g. the original actor)?

I'll still add the context, since we can replace `context.Background()` with it.

network/dag/notifier.go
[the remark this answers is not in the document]
woutslakhorst: The argument is mainly about decrypting from the network layer, which is not a user but a machine. The context injections into the network layer go very deep.

If keeping, mark the additional context passing with some comments why it's there so we can remove it in the future.

network/network.go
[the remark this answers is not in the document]
woutslakhorst: > I'll still add the context, since we can replace context.Background() with it.

But the added contexts are not used for cancellation, that would give a bad hint for anyone using it?

audit/http_test.go
```diff
@@ -0,0 +1,72 @@
...
+		next        func(ctx echo.Context, args interface{}) (interface{}, error)
+		moduleName  string
+		operationID string
+	}
+	tests := []struct {
+		name string
+		args args
+		want func(ctx echo.Context, args interface{}) (interface{}, error)
+	}{
+		// TODO: Add test cases.
```
woutslakhorst, line 65: There's a todo here

didman/api/v1/api.go
```diff
@@ -81,6 +82,7 @@ func (w *Wrapper) Preprocess(operationID string, context echo.Context) {
 	context.Set(core.StatusCodeResolverContextKey, w)
 	context.Set(core.OperationIDContextKey, operationID)
 	context.Set(core.ModuleNameContextKey, didman.ModuleName)
+	audit.Middleware(context, didman.ModuleName, operationID)
```
woutslakhorst, line 85: Given the fact that controllers need to call the audit processing manually, it should no longer be called Middleware? It's no longer registered as middleware.

woutslakhorst (changes requested):
Also (maybe not for now), why do half of the APIs define a `Preprocess` and the other half set the middleware in `RegisterHandlers`.

reinkrul:
> Also (maybe not for now), why do half of the APIs define a `Preprocess` and the other half set the middleware in `RegisterHandlers`.

The modules which already use strict server interfaces don't need the `Preprocess` function anymore, instead can use middleware in `RegisterHandlers`.

network/dag/notifier.go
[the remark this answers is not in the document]
reinkrul: > The context injections into the network layer go very deep.

Let's remove it then.

didman/api/v1/api.go
reinkrul: renamed it to InitEchoContext

woutslakhorst (approved):

PR #1760, opened 2022-12-11
OpenAPI: order response types to support OAPI Java Generator

reinkrul:
OpenAPI Generator Tools Java generates an API for the first content type it sees, but deserializes it as JSON. So if the first is `text/plain` it generates an API returning a String, but returns the JSON as-is when the returned content-type is `application/json`.

Putting JSON on top fixes this.


woutslakhorst (approved):
wait wut? Java.....

reinkrul:
> wait wut? Java.....

OpenAPI Generators... 🤷

PR #1765, opened 2022-12-12
Allow for multiple organization credentials to exist

### LINKED ISSUE #987: Auth: support organizations having multiple, non-matching credentials
Currently, only the first organization credential found for an organization is used when creating or validating contracts. This can create issues where the organization name/city in the contract doesn't match the name/city when validating the contract.

Use case might be when an organization changes its name or location(?). And it's bound to happen during development/testing.

Solution is probably to:
- Accept an organization VC in `Notary.DrawUpContract` instead of providing organization DID
- Check the signed contract against all resolved organization VCs in `Notary.validateIssuer`

woutslakhorst:
Fixes #987

Name/City from signed login contract may now match any OrganizationCredential from requesting party. 
The API now supports an optional VC as argument. Since it's optional it's backwards compatible.

minor issue: non backport


auth/services/oauth/oauth.go
```diff
@@ -244,8 +248,14 @@ func (s *service) CreateAccessToken(request services.CreateAccessTokenRequest) (
...
-	if actualName != context.requesterName || actualCity != context.requesterCity {
-		log.Logger().Warnf("Token request validation failed, legal entity mismatch (expected name=%s, actual name=%s, expected city=%s, actual city=%s)", context.requesterName, actualName, context.requesterCity, actualCity)
+	found := false
+	for _, identity := range context.requesterOrganizationIdentities {
+		if actualName == identity.name && actualCity == identity.city {
+			found = true
+		}
+	}
+	if !found {
+		log.Logger().Warn("Token request validation failed, requester does not have any credentials that match the organization name and city in the contract.")
```
reinkrul, line 258:
```suggestion
		log.Logger().Warn("Token request validation failed, requester does not have any credential that match the organization name and city in the contract.")
```

auth/services/oauth/oauth.go
```diff
@@ -244,8 +248,14 @@ func (s *service) CreateAccessToken(request services.CreateAccessTokenRequest) (
 func (s *service) validateRequester(context *validationContext) error {
 	actualName := context.contractVerificationResult.ContractAttribute(contract.LegalEntityAttr)
 	actualCity := context.contractVerificationResult.ContractAttribute(contract.LegalEntityCityAttr)
-	if actualName != context.requesterName || actualCity != context.requesterCity {
-		log.Logger().Warnf("Token request validation failed, legal entity mismatch (expected name=%s, actual name=%s, expected city=%s, actual city=%s)", context.requesterName, actualName, context.requesterCity, actualCity)
+	found := false
+	for _, identity := range context.requesterOrganizationIdentities {
+		if actualName == identity.name && actualCity == identity.city {
+			found = true
```
reinkrul, line 254:
```suggestion
			found = true
			break
```

reinkrul (changes requested):
`notary.findVC` refers to the issue this fixes:

```go
	// Having multiple VCs with non-matching credentialSubjects for this care organization is not supported.
	// Will be supported when https://github.com/nuts-foundation/nuts-node/issues/987 is implemented.
	if len(result) > 1 {
		var credentialSubject interface{}
		for _, current := range result {
			if credentialSubject != nil && !reflect.DeepEqual(credentialSubject, current.CredentialSubject) {
				return "", "", errors.New("found multiple non-matching VCs, which is not supported")
			}
			credentialSubject = current.CredentialSubject
		}
	}
```

gerardsn (commented):
`make install-tools && make gen-api`

reinkrul (approved):

PR #1766, opened 2022-12-13
Auth: add missing mock Controller.Finish() calls

reinkrul:
Otherwise, mock assertions aren't performed.


gerardsn (commented):
would a `t.Cleanup(ctrl.Finish)` in createContext not be easier?

gerardsn:
No longer needed to call Finish on gomock controller

Issue #1508, opened 2022-10-05
Validate service endpoint before adding to DID document


reinkrul:
When the receiving care organization's NutsComm endpoint is incorrect (in this case, it wasn't a valid reference to its vendor's NutsComm service), the eOverdracht sender can issue the `NutsAuthorizationCredential`, but the private transaction's payload never comes available at the receiving party.

The NutsComm endpoint in question was literally: `<vendor>?serviceType=eOverdracht-receiver` (so `<vendor>` instead of the DID).

A blocking error would be nice, but there should at least be an error in the logging.

Also consider tweaking the endpoint validating, so that we capture these kind of errors, since a `NutsComm` endpoint is either a gRPC endpoint or a reference.

gerardsn:
It should not be possible to issue a private VC for a DID without a resolvable and valid NutsComm endpoint.

woutslakhorst:
1 - yes for a syntax check on DID services
2 - Checking for validity of a VC at runtime is extremely dependant on the VC type. Which turns into a discussion on how. much of validation should be hardcoded.

gerardsn:
I cannot reproduce the issue.

Even if I give both nodes an invalid NutsComm address:
```
"service": [
            {
                "id": "did:nuts:123abc",
                "serviceEndpoint": "%3Cvendor-did%3E/serviceEndpoint?type=NutsComm",
                "type": "NutsComm"
            }
        ],
```
`NutsAuthorizationCredentials` can be issued by, and resolved on both nodes. (I've tried master and v4.2.4 since this was latest tagged release at the time.) In the tests, private VC retrieval always works if 
1. there is a direct connection between involved nodes, 
2. the DID document contains a NutsComm service, doesn't matter if it is valid, and 
3. the node is restarted after creating a node DID to start the private VC scheduler.

If I remember the event correctly, restarting the node fixed some synchronization issues (node was ~10 tx behind and could no longer sync) and updating the NutsComm address allowed newly issued private VCs to be received, but private VCs issued prior to this correction were never received. 
- Condition 1: Not just the new, but all private VCs issued that week should have arrived if it was simply a connection issue.
- Condition 2: Does not cause any issues. 
- Condition 3: Most likely scenario by deduction. The node was not restarted -> private VC scheduler was not started -> scheduler never subscribed to private VC notifications -> after a restart the node did not know it had missed a private VCs...
[...]

PR #1769, opened 2022-12-14
VDR: Add ManagedDocumentValidator and managedServiceValidator

gerardsn:
part of #1508 

This splits did doc validators in `NetworkDocumentValidator` that validates documents received through the network, and `ManagedDocumentValidator` that provides stricter validation on did documents managed by a node before they are published. 

The latter has an additional validator: `managedServiceValidator` provides addition validation on services before publication. All services are validated on every did document update. _Could this be an issue with something like automated key rotation if a service reference is no longer valid?_

remaining after this PR is finished
- go through APIs and didman to validate the new did doc before it is updated/published
- remove checks in the above that are now redundant
- move ServiceResolver interface to vdr/types with the rest

### DIFF
--- a/vdr/validators.go
+++ b/vdr/validators.go
@@ -142,7 +142,8 @@ func (m managedServiceValidator) Validate(document did.Document) error {
 	// make sure that it resolves when if it's a reference
 	var resolvedEndpoint any
 	// Cache resolved DID documents because most of the time all (compound) services will refer to the same DID document in all service references.
-	cache := make(map[string]*did.Document, 0)
+	cache := make(map[string]*did.Document, 1)
+	cache[document.ID.String()] = &document // add the updated document to the cache in case it contains self-references
 	for _, service := range document.Service {
 		switch se := service.ServiceEndpoint.(type) {
 		case map[string]interface{}:
@@ -173,7 +174,12 @@ func (m managedServiceValidator) Validate(document did.Document) error {
 		}
 
 		// specific service.Type need additional validation
-		if err = serviceTypeValidation(service.Type, resolvedEndpoint); err != nil {
+		resolvedService := did.Service{
+			ID:              service.ID,
+			Type:            service.Type,
+			ServiceEndpoint: resolvedEndpoint,
+		}
+		if err = serviceTypeValidation(resolvedService); err != nil {
 			return invalidServiceError{fmt.Errorf("%s: %w", service.Type, err)}
 		}
 	}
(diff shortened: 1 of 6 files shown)


vdr/validators.go
```diff
@@ -74,40 +87,191 @@ func (v verificationMethodValidator) verifyThumbprint(method *did.VerificationMe
...
+	}
+	if URL, err := url.Parse(endpointStr); err != nil {
+		return err
+	} else if URL.Scheme != "grpc" {
+		return errors.New("scheme must be grpc")
+	}
+	return nil
+}
+
+func validateNodeContactInfo(endpoint any) error {
```
reinkrul, line 218: This is more complicated than necessarry, RFC006 only states that `email` must be present, and others (name, website, etc) _may_ be present. Doesn't state other keys can't be present. Nothing should break when other stuff is there as well.

I'd say just assert `email` is present, and just take the rest for granted.

reinkrul (changes requested):
This PR rescopes the original issue: "prevent mistakes by local sysadmin/app, adding invalid endpoints to their own DID documents" to "prevent DID documents  (from anyone) with invalid endpoints from being added to the VDR".

Can you elaborate why this is necessary? It feels like we're making the same mistake here we made with VC validation in VCR (hardcoding validating specifiek VC types in VCR, which should've been extensible and configurable).

It also creates a problem for VDR consistency on the network; new nodes don't accept DID documents with invalid endpoints, but which are already accepted in existing nodes. Means existing nodes would have to drop their VDR storage and reprocess.

gerardsn:
This PR only changes validation of DID documents before they are published on the network. Validation of published documents is unchanged. This should improve quality of data in the network without breaking anything.

vdr/validators.go
```diff
@@ -38,14 +40,24 @@ import (
...
 	}}
 }
 
+// ManagedDocumentValidator extends NetworkDocumentValidator with extra safety checks to be performed on DID documents managed by this node before they are published on the network.
+func ManagedDocumentValidator(serviceResolver didservice.ServiceResolver) did.Validator {
+	validator := NetworkDocumentValidator().(*did.MultiValidator)
+	validator.Validators = append(validator.Validators, managedServiceValidator{
+		serviceResolver: serviceResolver,
+	})
+	return validator
```
reinkrul, line 58: you could just wrap it in another multivalidator, then you don't need casting (dont mind the formatting)?

```suggestion
	return &did.MultiValidator{Validators: []did.Validator{NetworkDocumentValidator(), managedServiceValidator{
		serviceResolver: serviceResolver,
	}}
```

vdr/vdr.go
```diff
@@ -172,7 +172,7 @@ func (r VDR) Update(id did.DID, current hash.SHA256Hash, next did.Document, _ *t
 		return types.ErrDeactivated
 	}
 
-	if err = CreateDocumentValidator().Validate(next); err != nil {
+	if err = NetworkDocumentValidator().Validate(next); err != nil {
```
reinkrul, line 175: shouldn't this be the `ManagedDocumentValidator`?

reinkrul (changes requested):

vdr/validators.go
gerardsn: The idea is that ManagedDocumentValidator does everything the NetworkDocumentValidator + some more. Extending the NetworkDocumentValidator guarantees that this is the case.

vdr/vdr.go
gerardsn: This PR introduses the validator, the next PR applies it + removes checks in Didman and elsewhere that are now redundant. It is split to minimise conflicts with VDR redesign.

reinkrul (changes requested):
Some things

vdr/vdr.go
gerardsn: I wanted to split approval of the concept and implementing it, but I've added it where appropriate.

gerardsn: changed the design a bit in [2faee38](https://github.com/nuts-foundation/nuts-node/pull/1769/commits/2faee38d15038389d35bed2baaa47ff9f536a5a0) to deal with the different formats of ServiceEndpoints

gerardsn:
> Some things

review request was by accident

reinkrul (approved):

PR #1771, opened 2022-12-14
VDR: move ServiceResolver interface to vdr/types

gerardsn:
the only vdr interface that was not in interfaces.go


gerardsn:
not relevant after #1709

Issue #1761, opened 2022-12-12
CRL: revocation checking fails when unrelated CRL can't be updated


reinkrul:
`crl/validator.go:256`:

```go
// If the CRL validator is outdated kill the connection
if !v.IsSynced(maxValidityDays) {
    return errors.New("CRL database is outdated, certificate revocation can't be checked")
}
```

If _any_ CRL is outdated, it blocks the token from being verified. But it should only check the relevant root CAs and intermediate CAs (the ones used by the peer).

woutslakhorst:
Why the reopen?

reinkrul:
Found an edge case where it could go wrong; https://github.com/nuts-foundation/nuts-node/issues/1883

Issue #1762, opened 2022-12-12
CRL: must not try to update CRLs of expired CA certificates


reinkrul:
When the truststore contains an expired certificate, the CRL validator will always report it is out-of-sync, because it won't publish new CRLs. And the CRL validator doesn't need to update it, since any certificate issued by it will be invalid (because the CA certificate is expired).

PR #1773, opened 2022-12-15
CRL: Ignore expired certificates

reinkrul:
Fixes https://github.com/nuts-foundation/nuts-node/issues/1761 and https://github.com/nuts-foundation/nuts-node/issues/1762

### DIFF
--- a/crl/validator.go
+++ b/crl/validator.go
@@ -117,13 +117,18 @@ func (v *validator) IsRevoked(issuer string, serialNumber *big.Int) bool {
 	return false
 }
 
-func (v *validator) updateCRL(endpoint string) error {
+func (v *validator) updateCRL(endpoint string, issuer *x509.Certificate) error {
+	if nowFunc().After(issuer.NotAfter) {
+		// Certificate has expired, no need to update CRL since it can't sign any new CRLs.
+		return nil
+	}
+
 	v.listsLock.RLock()
 	crl, ok := v.lists[endpoint]
 	v.listsLock.RUnlock()
 
-	if !ok || crl.HasExpired(time.Now()) {
-		return v.downloadCRL(endpoint)
+	if !ok || crl.HasExpired(nowFunc()) {
+		return v.downloadCRL(endpoint, issuer)
 	}
 
 	return nil
@@ -143,7 +148,7 @@ func (v *validator) verifyCRL(crl *pkix.CertificateList) error {
 	return certificate.CheckCRLSignature(crl)
 }
 
-func (v *validator) downloadCRL(endpoint string) error {
+func (v *validator) downloadCRL(endpoint string, issuer *x509.Certificate) error {
 	response, err := v.httpClient.Get(endpoint)
 	if err != nil {
 		return err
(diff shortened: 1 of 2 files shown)


dmeijboom:
I see that a review was requested from me so I shall respond, whether it was by accident or not :P

In the `IsSynced` method for loop is changed to this:

```go
endpoints := v.parseCRLEndpoints()

for endpoint, issuer := range endpoints {
```

I'm not sure if that's correct as `parseCRLEndpoints` returns a map of an CRL endpoint with the certificate as it's value. The certificate doesn't have to be the issuer of the CRL, it could be any certificate in the chain that has the CRL endpoint configured.

> // Certificate has expired, no need to update CRL since it can't sign any new CRLs.

The CRL is signed by the CA or an intermediate CA. If you grab any certificate and download it's CRL it should be signed by the issuer of that certificate. That's also what has been implemented in the `verifyCRL` function:

```go
	certificate, ok := v.certificates[issuerName]
	if !ok {
		return errors.New("CRL signature could not be validated against known certificates")
	}

	return certificate.CheckCRLSignature(crl)
```

So I doubt this is correct as the certificate doesn't have to be the certificate that signed the CRL.

crl/validator.go
```diff
@@ -143,7 +148,7 @@ func (v *validator) verifyCRL(crl *pkix.CertificateList) error {
 	return certificate.CheckCRLSignature(crl)
```
dmeijboom, line 148: Not sure if Nuts is already using Go 1.19 but `Certificate.CheckCRLSignature` is deprecated:

> Deprecated: Use RevocationList.CheckSignatureFrom instead.

reinkrul: Yeah, I noticed. More work :'-)

woutslakhorst:
+backport to V5

woutslakhorst:
> In the IsSynced method for loop is changed to this:
>
> endpoints := v.parseCRLEndpoints()
> 
> for endpoint, issuer := range endpoints {
> I'm not sure if that's correct as parseCRLEndpoints returns a map of an CRL endpoint with the certificate as it's value. The certificate doesn't have to be the issuer of the CRL, it could be any certificate in the chain that has the CRL endpoint configured.

It might just be the `issuer` term which refers to the CA and not the issuer of the CRL?

dmeijboom:
> > In the IsSynced method for loop is changed to this:
> > endpoints := v.parseCRLEndpoints()
> > for endpoint, issuer := range endpoints {
> > I'm not sure if that's correct as parseCRLEndpoints returns a map of an CRL endpoint with the certificate as it's value. The certificate doesn't have to be the issuer of the CRL, it could be any certificate in the chain that has the CRL endpoint configured.
> 
> It might just be the `issuer` term which refers to the CA and not the issuer of the CRL?

Iirc the certificates stored in the CRL validator could be any certificate. Whether that's a CA, intermediate CA or a normal certificate. There is (as far as I'm aware) no direct relationship to the issuer. Otherwise the code in `verifyCRL` would have been really straightforward (it wouldn't have to look up the issuer).

reinkrul:
The signer of the CRL can be, indeed, any (trusted) CA:

>  A CRL is issued by a CRL issuer, which is typically the CA which also issued the corresponding certificates, but could alternatively be some other trusted authority.

(https://en.wikipedia.org/wiki/Certificate_revocation_list)

Fixing this actually made the code a bit simpler and more correct; now it takes into account multiple CA certificates using the same CRL endpoint. If there are no (more) active CA certificates using the CRL endpoint (in other words, all expired), it skips the CRL endpoint.

woutslakhorst (approved):

PR #1776, opened 2022-12-19
some cleanup (goimports)

woutslakhorst:
In #1709 lots of files have changed, some due to goimports and changes to some test code.
I bundled those (unimportant) changes into this PR.


reinkrul:
If you `goimports` the code, maybe we can add it to the workflow. I recently saw `golangci-lint` being used (https://github.com/piprate/json-gold/blob/master/.github/workflows/ci.yml), which also checks `goimports` compliance.

reinkrul (approved):

PR #1781, opened 2022-12-20
Crypto: strict OpenAPI-generated interface

reinkrul:


gerardsn (approved):

crypto/api/v1/api.go
```diff
@@ -80,35 +83,22 @@ func (signRequest SignJwsRequest) validate() error {
...
-func (w *Wrapper) SignJwt(ctx echo.Context) error {
-	var signRequest = &SignJwtRequest{}
-	err := ctx.Bind(signRequest)
-	if err != nil {
-		return err
-	}
-
-	if err := signRequest.validate(); err != nil {
-		return core.InvalidInputError("invalid sign request: %w", err)
+func (w *Wrapper) SignJwt(_ context.Context, signRequest SignJwtRequestObject) (SignJwtResponseObject, error) {
```
gerardsn, line 86: I believe we set a default timeout for echo that can cancel these contexts. Shouldn't canceling of the context also result in cancellation of the action on the node? (The question is not specific to this engine as it is probably more relevant in the Network engine than it is here)

reinkrul: As long as the code actually uses the context, yes. But `Crypto.SignJWT` doesn't take a context, so it's not cancellable.

Could be a good improvement, as operations involving private keys could invoke external resources (Vault), so cancellation is applicable here. Out of scope for the PR though, since behavior isn't changed.

PR #1782, opened 2022-12-20
Crypto: keep private keys inside crypto module

reinkrul:
`Resolve()` should not return a struct from which the private key itself can be extracted (e.g., through reflection), because:
- Parties can export the private key, which is unintended
- Compatible issues with future changes, in which the private key might not be in application memory at all (but in an HSM, for instance).
- Audit logging of key usage becomes near impossible, since you don't know when/how/where the key is actually used

The bulk of the changes are caused by not calling `SignJWS/SignJWT` directly, but they now how to go through `Crypto.SignJWS/SignJWT`.


crypto/interface.go
```diff
@@ -64,20 +64,28 @@ type Decrypter interface {
...
 	// KID returns the unique ID for this key.
 	KID() string
-	// Public returns the public key. This is a short-hand for Signer().Public()
+	// Public returns the public key.
 	Public() crypto.PublicKey
 }
+
+// keyContainer is a Key that contains the private key itself and thus is exportable.
+// Should only be used for select purposes (e.g. ephemeral keys).
+type keyContainer interface {
```
woutslakhorst, line 87: What about aliasing PrivateKey and add missing methods?

crypto/jwx.go
```diff
@@ -271,6 +240,29 @@ func signJWS(payload []byte, protectedHeaders map[string]interface{}, privateKey
 	return string(data), nil
 }
 
+func (client *Crypto) getPrivateKey(key interface{}) (crypto.Signer, string, error) {
```
woutslakhorst, line 243: The `interface{}` type in the interface and this method are a bit smelly. But a nice solution doesn't come to mind.

woutslakhorst (commented):
2 Questions

crypto/interface.go
reinkrul: Because it needs to return the key ID as well. I changed the name to `exportableKey` to make it more clear what's it about

crypto/jwx.go
reinkrul: I agree, but I couldn't find another solution either...

woutslakhorst (approved):

PR #1791, opened 2023-01-02
Cleanup mocks

gerardsn:
removed some unused mocks (only if entire file is unused, there are still plenty of unused mocked methods)

added mocking of echo.Context, and moved the existing `mock/mock_echo.go` to `echo/mock_echo.go`


stevenvegt:
Can you add the ones you have removed, and rename the `echo/echo_mock.go` -> `echo/mock.go`?

gerardsn:
removed:
- auth/mock/go
- auth/services/mock.go#CompoundServiceClient -> previously moved to didman
- jsonld/mock.go
- network/transport/v2/protocol_grpc_mock.pb.go
- vdr/mock.go

renamed:
- mock/mock_echo.go -> echo/mock.go

makefile
```diff
@@ -12,40 +12,38 @@ gen-readme:
 	./generate_readme.sh
 
 gen-mocks:
```
gerardsn, line 14: sorted alphabetically, no other changes than mentioned in the conversation

stevenvegt:
I like to discuss if we let unused mocks available for later usages. Testing is good, should be encouraged, less work lowers the threshold for writing that extra test. Mocks are cheap, are not part of the binary. I like it when I'm writing a test and my editor can find the mock. I would argue for the following rule: 

> If you write an interface, make sure you add a mock generation line to the makefile.

http/requestlogger_test.go
```diff
@@ -127,7 +127,7 @@ func Test_bodyLoggerMiddleware(t *testing.T) {
 		responseRecorder := httptest.NewRecorder()
 		response := echo.NewResponse(responseRecorder, e)
 		response.Header().Set("Content-Type", "application/json")
-		echoMock := mock.NewMockContext(ctrl)
+		echoMock := echo2.NewMockContext(ctrl)
```
reinkrul, line 130: bad import name imo

echo/echo_alias.go
```diff
@@ -0,0 +1,8 @@
+package echo
+
+import "github.com/labstack/echo/v4"
+
+// Context is an alias/composition of echo.Context that allows generating a mock of echo.Context
+type Context interface {
```
reinkrul, line 6: maybe move it to `test/echo`?

reinkrul (changes requested):

reinkrul:
> I like to discuss if we let unused mocks available for later usages. Testing is good, should be encouraged, less work lowers the threshold for writing that extra test. Mocks are cheap, are not part of the binary. I like it when I'm writing a test and my editor can find the mock. I would argue for the following rule:
> 
> > If you write an interface, make sure you add a mock generation line to the makefile.

I disagree, any unused code just adds clutter to the Makefile, codebase and makes generating mocks take longer (which already takes too long imo, I usually copy paste the line of a specific file to regenerate the mocks).

If a mock isn't used when its interface is introduced, it's probably not needed any time soon, since you write interfaces to use them. And where you use them, you need a unit test to assert the behavior. Thus, you need the mock (or you use the concrete type when it's easier than using the mock).

echo/echo_alias.go
gerardsn: the `echo/` package only contains tests, maybe just rename that to `echo_test/`.

gerardsn: The strict open-API interfaces seem to remove the need for mocking of echo.Context completely.

gerardsn:
conclusion of discussion: Usage of mocks should be minimized. Prefer concrete implementations where possible, and only use a mock if needed. Small interfaces should facilitate writing of test implementations. --> Remove all unused mock interfaces.

echo/echo_alias.go
woutslakhorst, line 6: Why do we need this? It wasn't required before?

gerardsn: This mocks an external interface. This mock was probably created externally and then imported into our codebase, but this change allows regenerating it easily. The echo.Context interface has not changed, so the mock has not stopped working. 

However, this mock file can go as soon as all APIs are are converted to use strict OAPI interfaces.

woutslakhorst: Add a comment to the type?

gerardsn: rebased and removed changes to mock/ since it was starting to conflict with the strict openAPI interface PRs. Made #1827 to remove this later.

woutslakhorst (approved):

reinkrul (approved):

PR #1792, opened 2023-01-02
Storage: handle ErrKeyNotFound returned by stoabs.Reader.Get

gerardsn:
Implements changes in https://github.com/nuts-foundation/go-stoabs/pull/41

Prevents some potential nil dereferencing in cases where `data, err := staobs.Reader.Get()` returns double nil

### DIFF
--- a/vdr/store/store.go
+++ b/vdr/store/store.go
@@ -136,7 +136,8 @@ func (s *store) Update(id did.DID, current hash.SHA256Hash, next did.Document, m
 
 		latestBytes, err := metadataWriter.Get(stoabs.BytesKey(latestRef))
 		if errors.Is(err, stoabs.ErrKeyNotFound) {
-			return vdr.ErrNotFound // TODO: is this correct? existence of a latestRef ~10 lines up means that this error cannot occur
+			// Existence of a latestRef means that this error should never occur. Could indicate corrupted DB.
+			err = fmt.Errorf("latest document metadata missing for %s (metadataShelfRef=%s)", didString, latestRef)
 		}
 		if err != nil {
 			return err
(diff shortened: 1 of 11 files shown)


woutslakhorst (changes requested):
Wait for official version of stoabs

vdr/store/store.go
```diff
@@ -126,16 +126,18 @@ func (s *store) Update(id did.DID, current hash.SHA256Hash, next did.Document, m
...
+		if errors.Is(err, stoabs.ErrKeyNotFound) {
 			return vdr.ErrNotFound
 		}
+		if err != nil {
+			return err
+		}
 
 		latestBytes, err := metadataWriter.Get(stoabs.BytesKey(latestRef))
+		if errors.Is(err, stoabs.ErrKeyNotFound) {
+			return vdr.ErrNotFound // TODO: is this correct? existence of a latestRef ~10 lines up means that this error cannot occur
```
reinkrul, line 139: Is from a different shelf, so it would indicate corrupt storage

reinkrul (approved):

vdr/store/store.go
gerardsn: so it should probably not return vdr.ErrNotFound since that indicates the DID does not exist and is handled different than the original error?

reinkrul: Yes, probably just return a specific error?

gerardsn: now returns 
```
fmt.Errorf("latest document metadata missing for %s (metadataShelfRef=%s)", didString, latestRef)
```

woutslakhorst (approved):

PR #1799, opened 2023-01-04
Crypto: update storage documentation

gerardsn:
Minor changes to crypto.storage config description. Changing the default storage is a breaking change and is delayed to v6.0.0

*** below this line is the old description ***
Code contains the following for crypto storage:
```
switch client.config.Storage {
case "fs":
	return client.setupFSBackend(config)
case "vaultkv":
	return client.setupVaultBackend(config)
case "":
	if config.Strictmode {
		return errors.New("backend must be explicitly set in strict mode")
	}
...
```
However, the default value is set to `fs`.

If the proposed change is correct I will add tests. Note that this also adds another required config field when changing to strictmode=true.


reinkrul (changes requested):
Also add to release notes (for v6?), that `storage` is now required to be set in strict mode.

gerardsn:
Can this be added in v5.1? (considering strictmode is also changing its default value in v5.1)

reinkrul:
> Can this be added in v5.1? (considering strictmode is also changing its default value in v5.1)

No, is not backwards compatible, so v6

gerardsn:
This PR now only contains some minor textual changes and created #1811 to remove the default value in v6.0

reinkrul (approved):

Issue #1379, opened 2022-08-25
OAPI specs use invalid media types for DID documents


gerardsn:
In OAPI we specify `application/json+did-document` and the node returns `application/json`.
DID spec is trying to register `application/did+json` and `application/did+ld+json` as official media types, but these are not accepted yet.

We should either
- update to `application/did+json` and `application/did+ld+json`, or
- use `application/json` and `application/ld+json`

and set the Content-Type header accordingly

gerardsn:
RFC006 uses `application/did+json` for the JWS content type

gerardsn:
yes. All responses already were in json, now the OAS match this.

PR #1807, opened 2023-01-06
VDR: Strict OpenAPI interfaces

reinkrul:
This also changes the response content-type of `application/json+did-document` to `application/json`. The code generator doesn't play nice with specialized JSON types (it will treat it as a binary/stream type), and I think the same will applies to other JSON unmarshallers as well (needing extra mapping for the content type). So why not make everyone's life easier and just return `application/json`. It's just JSON, after all.

### DIFF
--- a/docs/pages/release_notes.rst
+++ b/docs/pages/release_notes.rst
@@ -3,7 +3,15 @@
 Release notes
 #############
 
-What has been changed, and how to update between versions.
+*************************
+(To be released)
+*************************
+
+Release date: ?
+
+- Some VDR OpenAPI operations specified ``application/json+did-document`` as Content-Type, while they actually returned ``application/json``.
+  This inconsistency is fixed by changing the OpenAPI specification to ``application/json``.
+
 
 *************************
 Hazelnut release (v5.1.0)
--- a/vdr/api/v1/api.go
+++ b/vdr/api/v1/api.go
@@ -20,8 +20,9 @@
 package v1
 
 import (
+	"context"
 	"fmt"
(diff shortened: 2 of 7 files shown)


docs/pages/release_notes.rst
```diff
@@ -5,6 +5,18 @@ Release notes
 
 What has been changed, and how to update between versions.
 
+*****************
+Hazelnut (to be released) (v6.0.0)
```
woutslakhorst, line 9: Current master branch is elected to become v5.1. This will create a conflict.

woutslakhorst (changes requested):
blocked because of versioning

gerardsn:
also closes #1379

docs/pages/release_notes.rst
reinkrul: Will wait with merging after v5.1 got released

vdr/api/v1/api.go
```diff
@@ -56,132 +58,131 @@ func (a *Wrapper) ResolveStatusCode(err error) int {
...
-		return err
+		return nil, err
 	}
-	return ctx.JSON(http.StatusOK, *vm)
+	return AddNewVerificationMethod200JSONResponse(*vm), nil
 }
 
 func (a *Wrapper) Routes(router core.EchoRouter) {
-	RegisterHandlers(router, a)
+	RegisterHandlers(router, NewStrictHandler(a, []StrictMiddlewareFunc{
```
gerardsn, line 101: move this one to the top and add audit routing?

reinkrul: Fixed (had to rebase due to conflicts from this)

Issue #1721, opened 2022-12-02
NTS-004: Configuration defaults to the insecure fs backend


stevenvegt:
**Description:**
In case there is no configuration and `strict-mode` is not mandated by the server, the default private key storage backend for the crypto engine is the fs backend, which is less secure than the Vault backend.

**Technical description:**
The function func (client *Crypto) Configure(config core.ServerConfig) error in crypto/ crypto.go has this case for setting the storage backend:

```
case "":
        if config.Strictmode {
            return errors.New("backend must be explicitly set in strict mode")
        }
        // default to file system and run this setup again
        return client.setupFSBackend(config)
```

This is a "fail open" style default, which for security applications is to be avoided. For a discussion on "fail open" vs "fail closed" see this blog post: https://blogs.keysight.com/blogs/tech/nwvs.entry.html/2020/05/20/fail_closed_failop-ZYAt.html

**Impact:**
Sensitive key material might leak due to the less comprehensive defenses provided by the fs backend. 

**Recommendation:**
Make the default the vault backend, and have the users consciously degrade to fs if they need to do so.

reinkrul:
We're going to keep filesystem storage for cryptographic keys, but it may not be the default when in strict mode. Since this is a breaking change, we can't fix it in v5.1, so it is moved to v6.

Note that we're, at the same time, working on a system to allow pluggable externalized cryptographic key storage (https://github.com/nuts-foundation/nuts-node/issues/1820).

PR #1816, opened 2023-01-09
Crypto: remove default storage backend

### LINKED ISSUE #1811: Crypto: storage backend should be set explicitly in strict mode
The codebase is already setup to enforce this behavior, but the config still contains crypto.storage='kv' as default. Also see #1799.

Should be added in v6.0.0 since it is a breaking change.

gerardsn:
closes #1811

should be included in v6.0 release.

TODO: add to release notes


gerardsn:
Part of NTS-004 #1721

PR #1826, opened 2023-01-11
change startup order for VDR and network

### LINKED ISSUE #1823: Migrating DID store conflicts with loading NodeDID
The network module is started before the VDR. The VDR migrates at startup to the new layout, but before startup it is empty. NodeDID loading then fails.

woutslakhorst:
fixes #1823 

the network engine was started before the VDR engine. With the new DID store, the migration is run at startup which was too late for the network engine.

now the VDR engine starts before the network engine. VDR ambassador subscription on the network is now started in Start and not in Configure.

### DIFF
--- a/docs/pages/release_notes.rst
+++ b/docs/pages/release_notes.rst
@@ -12,6 +12,7 @@ Coconut update (v5.1.0)
 Release date: **DRAFT**
 
 - Default value of strictmode changed to true
+- Internal storage of VDR has changed. A migration will run at startup. If the node is stopped during this process, DID Documents will have to be reprocessed manually (restore functionality)
 
 **Full Changelog**: https://github.com/nuts-foundation/nuts-node/compare/v5.0.0...v5.1.0
 
--- a/vdr/ambassador.go
+++ b/vdr/ambassador.go
@@ -78,16 +78,20 @@ func NewAmbassador(networkClient network.Transactions, didStore didstore.Store, 
 	}
 }
 
-// Configure instructs the ambassador to start receiving DID Documents from the network.
 func (n *ambassador) Configure() error {
-	return n.networkClient.Subscribe("vdr", n.handleNetworkEvent,
+	return nil
+}
+
+func (n *ambassador) Start() error {
+	err := n.networkClient.Subscribe("vdr", n.handleNetworkEvent,
 		n.networkClient.WithPersistency(),
 		network.WithSelectionFilter(func(event dag.Event) bool {
 			return event.Type == dag.PayloadEventType && event.Transaction.PayloadType() == didDocumentType
 		}))
-}
+	if err != nil {
+		return err
+	}
 
-func (n *ambassador) Start() error {
 	stream := events.NewDisposableStream(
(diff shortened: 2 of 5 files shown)


vdr/ambassador.go
```diff
@@ -78,16 +78,20 @@ func NewAmbassador(networkClient network.Transactions, didStore didstore.Store,
 	}
 }
 
-// Configure instructs the ambassador to start receiving DID Documents from the network.
```
gerardsn, line 81: I found the symmetry this had with the VCR easier to understand. Should this be changed in the VCR too?

gerardsn (commented):
LGTM, just one question

vdr/ambassador.go
gerardsn: Or add a comment explaining why subscribing happens in Start instead of Configure

docs/pages/release_notes.rst
```diff
@@ -12,6 +12,7 @@ Coconut update (v5.1.0)
 Release date: **DRAFT**
 
 - Default value of strictmode changed to true
+- Internal storage of VDR has changed. A migration will run at startup. If the node is stopped during this process, DID Documents will have to be reprocessed manually (restore functionality)
```
reinkrul, line 15: If `Migrate()` calls are not atomic (1 db TX), we probably shouldn't allow the node to be stopped during migration (but it probably won't, since signal handlers are registered after `Start()` has finished, I think).

vdr/ambassador.go
reinkrul: I'd prefer changing it in VCR as well

reinkrul (changes requested):

docs/pages/release_notes.rst
woutslakhorst: `kill -9` is unstoppable

vdr/ambassador.go
woutslakhorst: The symmetry gives you an incorrect 'feeling' that both engines are the same. They aren't, the network is dependant on DIDs, not on VCs.

woutslakhorst: Added comment

docs/pages/release_notes.rst
woutslakhorst: That would be a different issue, this PR only fixes a required restart after updating.

woutslakhorst: I'm a bit weary about atomic operations and Redis:

```
// Pay attention, that Pipeline is not a transaction, so you can get unexpected
// results in case of big pipelines and small read/write timeouts.
// Redis client has retransmission logic in case of timeouts, pipeline
// can be retransmitted and commands can be executed more then once.
// To avoid this: it is good idea to use reasonable bigger read/write timeouts
// depends of your batch size and/or use TxPipeline.
```

vdr/ambassador.go
reinkrul: We probably don't need `Configure` in ambassadors at all, after this change. But for the sake of progress let's say out of scope for now and fix later.

reinkrul (approved):

gerardsn (approved):

Issue #1729, opened 2022-12-05
VDR: DIDUpdateRequest.currentHash from API


reinkrul:
Deprecated in v5.1 (https://github.com/nuts-foundation/nuts-node/issues/1710), to be removed in v6

gerardsn:
vdr/v1 spec is already deprecated; no reason to start breaking things

PR #1831, opened 2023-01-11
remove current hash from VDR Update interface

### LINKED ISSUE #1710: Deprecate Hash in document metadata
it was meant to prevent updates to documents that were already updated. But the main source of unintended updates come from race conditions. It's pretty useless currently.

It can be removed backwards compatible by just removing the check and remove the attribute from the API in the next major.

woutslakhorst:
closes #1710 

Mostly an interface simplification

### DIFF
--- a/docs/_static/vdr/v1.yaml
+++ b/docs/_static/vdr/v1.yaml
@@ -256,14 +256,9 @@ components:
     DIDUpdateRequest:
       required:
         - document
-        - currentHash
       properties:
         document:
           $ref: '#/components/schemas/DIDDocument'
-        currentHash:
-          type: string
-          description: The hash of the document in hex format. No longer used, contents ignored.
-          deprecated: true
     DIDCreateRequest:
       properties:
         controllers:
--- a/docs/_static/vdr/v1.yaml
+++ b/docs/_static/vdr/v1.yaml
@@ -262,7 +262,8 @@ components:
           $ref: '#/components/schemas/DIDDocument'
         currentHash:
           type: string
-          description: The hash of the document in hex format.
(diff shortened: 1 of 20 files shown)


gerardsn (approved):

gerardsn:
Not sure what is remaining for #1729. Removal of the hash from the API would make require a VDR V2 api?

docs/_static/vdr/v1.yaml
```diff
@@ -262,7 +262,8 @@ components:
           $ref: '#/components/schemas/DIDDocument'
         currentHash:
           type: string
-          description: The hash of the document in hex format.
+          description: The hash of the document in hex format. No longer used, contents ignored.
```
stevenvegt, line 265: Why is it still a required field?

stevenvegt (changes requested):
`currentHash` is still a required field on the Update Document request. I think this should be removed so people can already remove it from their request since it is deprecated.

docs/_static/vdr/v1.yaml
woutslakhorst: Because is some languages the generated code will change. We could also just remove it, then already generated code will still work.

reinkrul: I'd say just remove it for v6. Or maybe merge this PR as is, then remove the field from API for v6?

Regenerated code will cause integrations to break, but it's just the removal of 1 field.

reinkrul (approved):

woutslakhorst:
I reverted latest commit so the hash is still in there due to failing e2e tests on the command usage. Therefore changes are too big for v5.1. Will have to be completed in v6

stevenvegt (approved):

Issue #1820, opened 2023-01-09
Create a crypto storage API specification


stevenvegt:
Create a API specification:

Based on https://developer.hashicorp.com/vault/api-docs/secret/kv/kv-v1
* list secrets (might not be needed)
GET /secrets/keys/
* store / update secret
POST /secrets/keys/:id
* read secret
GET /secrets/keys/:id
* delete secret
DELETE /secrets/keys/:id

reinkrul:
Removed from v5.1 milestone since it doesn't necessarily need to be in v5

stevenvegt:
Closed by #1828

Issue #1810, opened 2023-01-09
Crypto: Create storage abstraction


stevenvegt:
This is a epic tracking issue.
In order to address NTS-004 #1721 and support multiple crypto storage back-ends #1361, we are going to build a storage abstraction layer which allows developers creating simple storage engines or proxies in their language of choice for their key store of choice.

For this we need multiple things:
* [x] API Specification, published in the docs #1820 
* [x] Build Nuts node client implementation (v5.1) #1819 
* [ ] Vault implementation (beta in v5.1)
* [ ] Remove the current Hashicorp Vault implementation from the nuts-node (v6) #1818 
* [ ] Document the new configuration, including examples for a side care setup
* [ ] Present this during a developers meeting

woutslakhorst:
I doubt if the new Vault implementation fits in the v5.1 timeframe.

reinkrul:
I don't see how this addresses NTS-004 https://github.com/nuts-foundation/nuts-node/issues/1721

Which crypto backend is used by default has nothing to do with where the Vault client lives (in the Nuts Node or in a separate process).

Removing from v5.1 milestone.

PR #1832, opened 2023-01-11
Add APIClient implementation of storage interface

### LINKED ISSUE #1819: Build storage back-end based on storage API
Create new storage backend which implement the /crypto/storage Storage interface
Implement it using a generated HTTP client based on the API specification #1820 .

stevenvegt:
Part of #1810. Closes #1819.
Adds an implementation of `crypto.Storage` interface.
Connects to an external secure storage server using a generated client from the API spec from #1820 .


crypto/crypto.go
```diff
@@ -87,6 +88,16 @@ func (client *Crypto) setupFSBackend(config core.ServerConfig) error {
 	return nil
 }
 
+func (client *Crypto) setupStorageAPIBackend() error {
+	log.Logger().Info("Setting up StorageAPI backend for storage of private key material.")
```
reinkrul, line 92: not very useful INFO comment (downgrade to debug)

crypto/crypto.go
```diff
@@ -111,6 +122,8 @@ func (client *Crypto) Configure(config core.ServerConfig) error {
 		return client.setupFSBackend(config)
 	case "vaultkv":
 		return client.setupVaultBackend(config)
+	case "storageapi":
```
reinkrul: could it just be a URL instead, and check if it starts with `<something>://`? I find `storageapi` a weird backend to configure.

Or if we want to have a config struct for future extensibility (although that could still be the case), maybe rename it to `secure-store` (also extra indicates the others aren't considered secure) or `external-store` / `external`.

crypto/storage/client.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+func NewAPIClient(url string) (Storage, error) {
+	client, _ := httpclient.NewClientWithResponses(url)
+	response, err := client.HealthCheckWithResponse(context.Background())
+	if err != nil {
+		return nil, fmt.Errorf("unable to connect to storage server: %w", err)
+	}
+	if response.StatusCode() != http.StatusOK {
+		return nil, fmt.Errorf("unable to connect to storage server: unexpected status code: %d", response.StatusCode())
```
reinkrul, line 33: not sure whether this is a good idea, in cloud environment containers should be able to start even if it depends on something that's not (yet) running (might still be starting).

Better to have the health check of Nuts Node execute this health check (although be wary of slow responses, so configure time-outs tightly), and have the environment deal with it (monitoring/restarting/etc)?

crypto/storage/client.go
```diff
@@ -0,0 +1,109 @@
...
+
+	return &APIClient{httpClient: client}, nil
+}
+
+type backendError struct {
+	error httpclient.ErrorResponse
+}
+
+func (r backendError) Error() string {
+	return fmt.Sprintf("remote error: the backend %s returned an error with status %d, title=%s details=%s", r.error.Backend, r.error.Status, r.error.Title, r.error.Detail)
```
reinkrul, line 44: We should take care this is not returned to clients, since it potentially leaks sensitive information. Maybe just log the backend/status/details and not return it in an error the calling HTTP client.

Might be a good point to document in the implementation guide (what to return, and what specifically not)

crypto/storage/client.go
```diff
@@ -0,0 +1,109 @@
...
+
+func (c APIClient) GetPrivateKey(kid string) (crypto.Signer, error) {
+	response, err := c.httpClient.LookupSecretWithResponse(context.Background(), kid)
+	if err != nil {
+		return nil, fmt.Errorf("unable to get private-key: %w", err)
+	}
+	switch response.StatusCode() {
+	case http.StatusOK:
+		if contentType := response.HTTPResponse.Header.Get("Content-Type"); contentType != "application/json" {
+			return nil, fmt.Errorf("unable to get private-key: unexpected content-type: %s", contentType)
```
reinkrul, line 55: could be a problem content-type, now you're losing the problem response

crypto/storage/client.go
```diff
@@ -0,0 +1,109 @@
...
+type APIClient struct {
+	httpClient *httpclient.ClientWithResponses
+}
+
+type APIClientConfig struct {
+	URL string `koanf:"url"`
+}
+
+func NewAPIClient(url string) (Storage, error) {
+	client, _ := httpclient.NewClientWithResponses(url)
```
reinkrul, line 27: Tight time-outs are important here, we shouldn't block for seconds when just signing a transaction. Implementations should respond really, really quick (I'd say <100ms or even less), since network message handling is synchronously calling this, blocking when reading a private key (when it receives any private TX, it tries to decrypt the PAL header with all of its private keys, to find out whether its meant for the local node).

crypto/storage/client.go
reinkrul, line 55:
```suggestion
			return nil, fmt.Errorf("unable to get private key: unexpected content-type: %s", contentType)
```

crypto/storage/client.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+	switch response.StatusCode() {
+	case http.StatusOK:
+		if contentType := response.HTTPResponse.Header.Get("Content-Type"); contentType != "application/json" {
+			return nil, fmt.Errorf("unable to get private-key: unexpected content-type: %s", contentType)
+		}
+
+		privateKey, err := util.PemToPrivateKey([]byte(response.JSON200.Data))
+		if err != nil {
+			return nil, fmt.Errorf("unable to parse private key as pem: %w", err)
```
reinkrul, line 60:
```suggestion
			return nil, fmt.Errorf("unable to parse private key as PEM: %w", err)
```

crypto/storage/client.go
```diff
@@ -0,0 +1,109 @@
...
+		}
+		return privateKey, nil
+	case http.StatusNotFound:
+		return nil, errKeyNotFound
+	case http.StatusBadRequest:
+		return nil, backendError{error: *response.JSON400}
+	default:
+		return nil, fmt.Errorf("unable to get private-key: unexpected status code from storage server: %d", response.StatusCode())
```
reinkrul, line 68:
```suggestion
		return nil, fmt.Errorf("unable to get private key: unexpected status code from storage server: %d", response.StatusCode())
```

crypto/storage/client.go
```diff
@@ -0,0 +1,109 @@
...
+	"fmt"
+	"github.com/nuts-foundation/nuts-node/crypto/storage/httpclient"
+	"github.com/nuts-foundation/nuts-node/crypto/util"
+	"net/http"
+)
+
+// APIClient implements the Storage interface. It uses a simple HTTP protocol to connect to an external storage server.
+// This server can either be a secret store itself, or proxy the request to a key store such as Hashicorp Vault or Azure Key Vault.
+// It allows us to keep the codebase clean and allow other parties to write their own adaptor.
+type APIClient struct {
```
reinkrul, line 17: general comment: allow `context.Context` to be passed to allow cancellation (usage of `context.Background` indicates missing context passing).

crypto/storage/client.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+func (c APIClient) SavePrivateKey(kid string, key crypto.PrivateKey) error {
+	pem, err := util.PrivateKeyToPem(key)
+	if err != nil {
+		return fmt.Errorf("unable to convert private key to pem format: %w", err)
+	}
+	response, err := c.httpClient.StoreSecretWithResponse(context.Background(), kid, httpclient.StoreSecretJSONRequestBody{Data: pem})
+	if err != nil {
+		return fmt.Errorf("unable to save private-key: %w", err)
```
reinkrul, line 87:
```suggestion
		return fmt.Errorf("unable to save private key: %w", err)
```

crypto/storage/client.go
```diff
@@ -0,0 +1,109 @@
...
+	if err != nil {
+		return false
+	}
+	return response.StatusCode() == http.StatusOK
+}
+
+func (c APIClient) SavePrivateKey(kid string, key crypto.PrivateKey) error {
+	pem, err := util.PrivateKeyToPem(key)
+	if err != nil {
+		return fmt.Errorf("unable to convert private key to pem format: %w", err)
```
reinkrul, line 83:
```suggestion
		return fmt.Errorf("unable to convert private key to PEM format: %w", err)
```

crypto/storage/client.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+	response, err := c.httpClient.StoreSecretWithResponse(context.Background(), kid, httpclient.StoreSecretJSONRequestBody{Data: pem})
+	if err != nil {
+		return fmt.Errorf("unable to save private-key: %w", err)
+	}
+	switch response.StatusCode() {
+	case http.StatusOK:
+		return nil
+	case http.StatusBadRequest:
+		return backendError{error: *response.JSON400}
```
reinkrul, line 93: nil deref?

crypto/storage/client.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+func (c APIClient) ListPrivateKeys() []string {
+	response, err := c.httpClient.ListKeysWithResponse(context.Background())
+	if err != nil {
+		return nil
+	}
+	switch response.StatusCode() {
+	case http.StatusOK:
+		return *response.JSON200
```
reinkrul, line 108: nil deref?

crypto/storage/storage.go
```diff
@@ -31,6 +31,8 @@ import (
 
 // ErrNotFound indicates that the specified crypto storage entry couldn't be found.
 var ErrNotFound = errors.New("entry not found")
+var errKeyNotFound = errors.New("key not found")
```
reinkrul, line 34: why not `ErrNotFound`?

reinkrul (changes requested):

crypto/crypto.go
stevenvegt: I don't think we should use a url there, but I agree with the storageapi name. I opted to `external-store`.

crypto/storage/client.go
stevenvegt: I've implemented the `HealthCheckable` interface, but this currently only results in the `/health` endpoint in reporting to be down. The node still continues to run. Keys will get lost if a LB doesn't catch this.

crypto/storage/client.go
stevenvegt: I Agree, but that changes the whole crypto.Storage interface. Perhaps part of a separate PR?

stevenvegt: The ListKeys is also a bit weird since it doesn't return errors.

crypto/cmd/cmd.go
```diff
@@ -48,11 +48,55 @@ func ServerCmd() *cobra.Command {
...
 	cmd.AddCommand(fs2VaultCommand())
+	cmd.AddCommand(fs2ExternalStore())
 	return cmd
 }
 
+func fs2ExternalStore() *cobra.Command {
+	return &cobra.Command{
+		Use:   "fs2external [directory]",
+		Short: "Imports private keys from filesystem based storage (located at the given directory) into the storage server.",
+		Long: "Imports private keys from filesystem based storage into the secret store server. The given directory must contain the private key files." +
```
reinkrul, line 59:
```suggestion
		Long: "Imports private keys from filesystem based storage into the secret store server. The given directory must contain the private key files. " +
```

makefile
```diff
@@ -57,6 +57,7 @@ gen-api:
 	oapi-codegen --config codegen/configs/vcr_v2.yaml docs/_static/vcr/v2.yaml | gofmt > vcr/api/v2/generated.go
 	oapi-codegen --config codegen/configs/auth_v1.yaml docs/_static/auth/v1.yaml | gofmt > auth/api/v1/generated.go
 	oapi-codegen --config codegen/configs/didman_v1.yaml docs/_static/didman/v1.yaml | gofmt > didman/api/v1/generated.go
+	oapi-codegen --config codegen/configs/crypto_store_client.yaml docs/_static/crypto/nuts-storage-api-v1.yaml | gofmt > crypto/storage/httpclient/generated.go
```
reinkrul, line 60: technically this doesn't belong under `gen-api`, but OK

reinkrul, line 60: Why the `httpclient` package isntead of putting it just into `storage`?

crypto/storage/vault.go
```diff
@@ -67,6 +67,20 @@ type vaultKVStorage struct {
 	client logicaler
 }
 
+func (v vaultKVStorage) Name() string {
+	return VaultConfigKey
+}
+
+func (v vaultKVStorage) CheckHealth() map[string]core.Health {
+	health := make(map[string]core.Health)
+	if err := v.checkConnection(); err != nil {
```
reinkrul, line 76: what is the time-out of `checkConnection`, in other words, how long can it block calls to the health check endpoint?

crypto/storage/client.go
```diff
@@ -0,0 +1,146 @@
...
+			return nil, fmt.Errorf("unable to parse private key as PEM: %w", err)
+		}
+		return privateKey, nil
+	case http.StatusNotFound:
+		return nil, ErrNotFound
+	case http.StatusBadRequest:
+		if response.JSON400 != nil {
+			return nil, backendError{error: *response.JSON400}
+		}
+		return nil, fmt.Errorf("unable to get private key: server responded with bad-request: %s", string(response.Body))
```
reinkrul, line 94: probably not the best idea to include the response in the error, since it might be returned to the end-user client (or it might be binary). Maybe just log it?

And I'd say error 400 is also unexpected, so you can just fallback to the default case (and maybe try to parse it as problem response, so you can log it).

crypto/storage/client.go
```diff
@@ -0,0 +1,146 @@
...
+		return fmt.Errorf("unable to save private key: %w", err)
+	}
+	switch response.StatusCode() {
+	case http.StatusOK:
+		return nil
+	case http.StatusBadRequest:
+		if response.JSON400 != nil {
+			return backendError{error: *response.JSON400}
+		}
+		return fmt.Errorf("unable to save private key: server responded with bad-request: %s", string(response.Body))
```
reinkrul, line 124: same here

crypto/crypto.go
```diff
@@ -41,8 +41,9 @@ var kidPattern = regexp.MustCompile(`^[\da-zA-Z_\- :#.]+$`)
 
 // Config holds the values for the crypto engine
 type Config struct {
-	Storage string              `koanf:"storage"`
-	Vault   storage.VaultConfig `koanf:"vault"`
+	Storage       string                  `koanf:"storage"`
+	Vault         storage.VaultConfig     `koanf:"vault"`
+	StorageClient storage.APIClientConfig `koanf:"connection"`
```
reinkrul, line 46: wouldn't `storage` be nicer?

`crypto.storage.whatever`

crypto/storage/client.go
```diff
@@ -0,0 +1,146 @@
...
+// It allows us to keep the codebase clean and allow other parties to write their own adaptor.
+type APIClient struct {
+	httpClient *httpclient.ClientWithResponses
+}
+
+func (c APIClient) Name() string {
+	return "Crypto"
+}
+
+func (c APIClient) CheckHealth() map[string]core.Health {
```
reinkrul, line 29: Since `CheckHealth()` performs calls to external resources, it might be nice to have it pass a context which canb e used for time-outs and cancellation.

crypto/crypto.go
```diff
@@ -59,6 +60,17 @@ type Crypto struct {
 	config  Config
 }
 
+func (client *Crypto) CheckHealth() map[string]core.Health {
+	response := make(map[string]core.Health)
+	if client.storage == nil {
+		response[ModuleName] = core.Health{Status: core.HealthStatusDown, Details: "no storage"}
```
reinkrul, line 66: is this also the case if it isn't used? That would be a bit weird

reinkrul (changes requested):

makefile
stevenvegt: Yes the name `gen-api` name does not compass everything since it also generate types and other api-clients.

stevenvegt: All generated clients are called `generated.go`, these files also get ignored during code scannen etc. I wanted to follow this convention. But than you get `/crypto/storage/generated.go` and that seemed weird. So I gave it its own package to live in.

crypto/storage/vault.go
stevenvegt: It uses `VaultConfig.Timeout` which is initialized in `vault.DefaultVaultConfig` and set to 5 seconds.

crypto/storage/client.go
stevenvegt: it implements a Healthcheckable interface. I consider it outside the scope of this PR to fix that. Perhaps open an issue for it?

crypto/crypto.go
stevenvegt: There should always be a form of storage: fs, vault or external. So this is a case of "This should never happen ™️ ".

crypto/crypto.go
stevenvegt: `storage` is already in use.

crypto/crypto.go
stevenvegt: It implies that this is possible, it is not, so remove the nil-check.

crypto/storage/client.go
```diff
@@ -0,0 +1,154 @@
...
+	"github.com/nuts-foundation/nuts-node/crypto/log"
+	"github.com/nuts-foundation/nuts-node/crypto/storage/httpclient"
+	"github.com/nuts-foundation/nuts-node/crypto/util"
+	"net/http"
+	"net/url"
+	"time"
+)
+
+const StorageAPIConfigKey = "external"
+const httpClientTimeout = 100 * time.Millisecond
```
stevenvegt, line 17: Make it configurable

makefile
stevenvegt: Decided: move it to a better name: `external`

reinkrul (approved):

PR #1833, opened 2023-01-12
VCR: Remove unrecoverable RemoteContextNotAllowedErr from failed events list

reinkrul:
Fixes https://github.com/nuts-foundation/nuts-node/issues/1662

Refactored to reusable collector for https://github.com/nuts-foundation/nuts-node/issues/1754

PR #1834, opened 2023-01-12
added removal of jsonld errors from failed event storage

### LINKED ISSUE #1662: Allow removal of failed events with migration
Failed events from the dag to its subscribers may cause entries in the events list. Those events can have a legitimate reason to be there. It can also be the result of a bug or incorrect configuration. This is always cause for an investigation. When fixed (or ignored) a migration should remove the entries.

API for getting the list GET /internal/network/v1/events

Examples of current causes:

- Unknown VC context: should not remain in the list of events.

woutslakhorst:
fixes #1662 

missing cleanup logic from #1743

### DIFF
--- a/network/interface.go
+++ b/network/interface.go
@@ -36,6 +36,8 @@ type Transactions interface {
 	Subscribe(name string, receiver dag.ReceiverFn, filters ...SubscriberOption) error
 	// Subscribers returns the list of notifiers on the DAG that emit events to subscribers.
 	Subscribers() []dag.Notifier
+	// CleanupSubscriberEvents removes events that errored but should be removed due to a bugfix.
+	CleanupSubscriberEvents(subcriberName, errorPrefix string) error
 	// GetTransactionPayload retrieves the transaction Payload for the given transaction.
 	// If the transaction or Payload is not found, dag.ErrPayloadNotFound is returned.
 	GetTransactionPayload(transactionRef hash.SHA256Hash) ([]byte, error)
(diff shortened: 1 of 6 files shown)


woutslakhorst:
Conflicts with #1833 

This impl has 100 lines less code ;)

vcr/ambassador.go
```diff
@@ -98,7 +98,10 @@ func (n ambassador) Start() error {
 	if err != nil {
 		return fmt.Errorf("failed to subscribe to REPROCESS event stream: %v", err)
 	}
-	return nil
+
+	// removing failed events required for #1743
+	// remove after v6 release
+	return n.networkClient.CleanupSubscriberEvents("vcr_vcs", "canonicalization failed: unable to normalize the json-ld document: loading remote context failed")
```
reinkrul, line 104: isn't  this too broad; e.g. remote server temporary unavailable?

reinkrul (commented):
I liked my garbage collector better but hey, that's probably my not-invented-here syndrome..

network/interface.go
```diff
@@ -36,6 +36,8 @@ type Transactions interface {
 	Subscribe(name string, receiver dag.ReceiverFn, filters ...SubscriberOption) error
 	// Subscribers returns the list of notifiers on the DAG that emit events to subscribers.
 	Subscribers() []dag.Notifier
+	// CleanupSubscriberEvents removes events that errored but should be removed due to a bugfix.
```
reinkrul, line 39:
```suggestion
	// CleanupSubscriberEvents removes events. Example use is cleaning up events that errored but should be removed due to a bugfix.
```

reinkrul (approved):
Trivial suggestion left

PR #1840, opened 2023-01-19
Network: connection manager cleanup

gerardsn:
This should contain no functional changes
removes unused struct values / interfaces / tests
small refactor in start/stop to improve readability

### DIFF
--- a/network/transport/grpc/connection_manager.go
+++ b/network/transport/grpc/connection_manager.go
@@ -36,7 +36,6 @@ import (
 	grpcPeer "google.golang.org/grpc/peer"
 	"google.golang.org/grpc/status"
 	"net"
-	"sync"
 )
 
 const defaultMaxMessageSizeInBytes = 1024 * 512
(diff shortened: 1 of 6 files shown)


network/transport/grpc/connection_manager.go
```diff
@@ -179,13 +154,31 @@ func (s *grpcConnectionManager) Start() error {
 	serverOpts = append(serverOpts, grpc.ChainStreamInterceptor(serverInterceptors...))
 
 	// Create gRPC server for inbound connectionList and associate it with the protocols
-	s.grpcServer = grpc.NewServer(serverOpts...)
-	for _, prot := range s.protocols {
-		func(protocol Protocol) {
-			prot.Register(s, func(stream grpc.ServerStream) error {
-				return s.handleInboundStream(protocol, stream)
-			}, s.connections, s)
-		}(prot)
```
gerardsn, line 188: does wrapping this in a function have any side effects?

woutslakhorst (changes requested):
With removing the `Stop()` there's no longer an option to stop the grpcServer gracefully. The connectionManager creates a background context and has to stop it on its own.

gerardsn:
Not sure what you mean. `Stop()` is not removed, biggest change there is that it no longer sets everything to nil after closing.

woutslakhorst:
> Not sure what you mean. `Stop()` is not removed, biggest change there is that it no longer sets everything to nil after closing.

Hmmm don't know what I was smoking...

woutslakhorst (approved):

PR #1846, opened 2023-01-20
migrate irma auth config to current format

### LINKED ISSUE #1845: Irma config not injected properly
It still uses the "dot" format instead of a nested Config struct.

Backport to V5.0

woutslakhorst:
fixes #1845

config was always set to default


stevenvegt:
How come no tests were failing? Perhaps add one which parses an actual config?

woutslakhorst:
> How come no tests were failing? Perhaps add one which parses an actual config?

added test.
Most config injection is untested.

reinkrul (approved):

PR #1850, opened 2023-01-24
Network: add authenticated plag to peer

### LINKED ISSUE #1664: Bootstrapnodes can send private VCs over an unauthenticated connection
Connections to bootstrapnodes are accepted as authenticated, even if authentication fails (failed authentication is logged). This allows private VC exchange over unauthenticated connections, even in strict mode. (possibly allows DIDComm in the future too)

Unauthenticated connections solve a chicken-egg problem: need to a connection with the network to sync to get the details to authenticate the connection. 

We can't authenticate the connection as soon as the information becomes available since this could kill the synchronization process: 
- If the bootstrap node has changed its NutsComm/certificate since inception, authentication fails as soon as its first NutsComm address is received and the connection will be dropped. 
- If all other discovered NutsComm endpoints up to that point in the DAG are offline, no peers are remaining to complete synchronization.
- Makes changing NutsComm impossible in a multi-graph scenario

Proposal:
1. boostrap nodes should be removed from the configuration after initial synchronization
2. marking these connections as unauthenticated prevents the exchange of private VCs and enforces step 1

Does this impact HA setups?

gerardsn:
This adds Peer.Authenticated to signal if a stream is authenticated or not.
closes #1664, and is part of a ongoing refactor on the connection manager (peer is not the right place for this flag)

`/status/diagnostics` currently has a `acceptunauthenticated: true/false` should this be replaced with the actual `Authenticated` value?

### DIFF
--- a/network/transport/grpc/authenticator.go
+++ b/network/transport/grpc/authenticator.go
@@ -55,6 +55,7 @@ func (t tlsAuthenticator) Authenticate(nodeDID did.DID, grpcPeer grpcPeer.Peer, 
 				WithError(err).
 				Warn("Connection manually authenticated with authentication error")
 			peer.NodeDID = nodeDID
+			peer.Authenticated = false
 			return peer, nil
 		}
 		return peer, err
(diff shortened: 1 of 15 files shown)


network/transport/grpc/authenticator.go
```diff
@@ -55,6 +55,7 @@ func (t tlsAuthenticator) Authenticate(nodeDID did.DID, grpcPeer grpcPeer.Peer,
 				WithError(err).
 				Warn("Connection manually authenticated with authentication error")
 			peer.NodeDID = nodeDID
+			peer.Authenticated = false
```
woutslakhorst, line 58: Default would be false already. Is there a way to go from true to false in this flow?

woutslakhorst (approved):
This will work. But is not the complete/clean solution. Next step would be to remove `AcceptUnauthenticated` which will always be true? And list the authenticated status at the connection list.

network/transport/grpc/authenticator.go
gerardsn: Yes, see last commit. However, I just added this for explicitness/readability.

PR #1865, opened 2023-01-31
resolve previous document based on tx prev hash in VDR update

### LINKED ISSUE #1754: VDR: Failed event on controller
On the stable network the following error is in my events list:

```{"error":"network document not signed by one of its controllers","hash":"fdec9e572b0ab2955c22c0de4c62654a0b46f180150fb2f5b068b1a0fd2f8fb8","retries":101,"transaction":"fdec9e572b0ab2955c22c0de4c62654a0b46f180150fb2f5b068b1a0fd2f8fb8","type":"payload"}```

 This is an error returned by the VDR which checks the controller hierarchy. Note that the TX has been correctly signed.

Question is if this is recoverable? Could mishaps happen due to timing/parallel updates?

### PR DESCRIPTION
**Pull request by woutslakhorst**
fixes #1754 

This correctly resolves the latest DID Document. This was already done for the controllers...

+backport to V5.0

### DIFF
--- a/vdr/ambassador.go
+++ b/vdr/ambassador.go
@@ -232,13 +232,29 @@ func (n *ambassador) handleUpdateDIDDocument(transaction dag.Transaction, propos
 		WithField(core.LogFieldTransactionRef, transaction.Ref()).
 		WithField(core.LogFieldDID, proposedDIDDocument.ID).
 		Debug("Handling DID document update")
-	// Resolve latest version of DID Document
-	currentDIDDocument, _, err := n.didStore.Resolve(proposedDIDDocument.ID, &types.ResolveMetadata{AllowDeactivated: true})
-	if err != nil {
-		return fmt.Errorf("unable to update DID document: %w", err)
+
+	// Resolve version of DID Document referred to by transaction
+	var currentDIDDocument *did.Document
+	var err error
+	for _, ref := range transaction.Previous() {
+		currentDIDDocument, _, err = n.didStore.Resolve(proposedDIDDocument.ID, &types.ResolveMetadata{AllowDeactivated: true, SourceTransaction: &ref})
+		if err != nil && !errors.Is(err, types.ErrNotFound) {
+			return fmt.Errorf("unable to update DID document: %w", err)
+		}
+		if currentDIDDocument != nil {
+			break
+		}
+	}
+	// fallback
+	if currentDIDDocument == nil {
+		currentDIDDocument, _, err = n.didStore.Resolve(proposedDIDDocument.ID, &types.ResolveMetadata{AllowDeactivated: true})
+		if err != nil {
+			return fmt.Errorf("unable to update DID document: %w", err)
+		}
+		log.Logger().Errorf("Failed to resolve DID Document by ref. Using latest version. (DID=%s)", proposedDIDDocument.ID)
 	}
 
-	// Resolve controllers of current version (could be the same document)
+	// Resolve controllers of previous version (could be the same document)
 	didControllers, err := n.resolveControllers(*currentDIDDocument, transaction)
 	if err != nil {
 		return fmt.Errorf("unable to resolve DID document's controllers: %w", err)
--- a/vdr/ambassador_test.go
+++ b/vdr/ambassador_test.go
@@ -436,6 +436,8 @@ func TestAmbassador_handleUpdateDIDDocument(t *testing.T) {
 		tx := newTX()
 		tx.signingKeyID = currentDoc.CapabilityInvocation[0].ID.String()
 		tx.payloadHash = payloadHash
+		prev := hash.RandomHash()
+		tx.prevs = []hash.SHA256Hash{prev}
 
 		// This is the metadata of the current version of the document which will be returned by the resolver
 		currentMetadata := &types.DocumentMetadata{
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: vdr/ambassador.go [hunk 1]
```diff
@@ -232,13 +232,29 @@ func (n *ambassador) handleUpdateDIDDocument(transaction dag.Transaction, propos
...
-	currentDIDDocument, _, err := n.didStore.Resolve(proposedDIDDocument.ID, &types.ResolveMetadata{AllowDeactivated: true})
-	if err != nil {
-		return fmt.Errorf("unable to update DID document: %w", err)
+
+	// Resolve version of DID Document referred to by transaction
+	var currentDIDDocument *did.Document
+	var err error
+	for _, ref := range transaction.Previous() {
+		currentDIDDocument, _, err = n.didStore.Resolve(proposedDIDDocument.ID, &types.ResolveMetadata{AllowDeactivated: true, SourceTransaction: &ref})
+		if err != nil && !errors.Is(err, types.ErrNotFound) {
```
gerardsn, line 241: Due to out of order event/notification processing the err could be a types.ErrNotFound even if the doc exists. So to answer the original question: yes, this could still be a recoverable error for the did update?

vdr/ambassador.go
```diff
@@ -232,13 +232,29 @@ func (n *ambassador) handleUpdateDIDDocument(transaction dag.Transaction, propos
...
 		Debug("Handling DID document update")
-	// Resolve latest version of DID Document
-	currentDIDDocument, _, err := n.didStore.Resolve(proposedDIDDocument.ID, &types.ResolveMetadata{AllowDeactivated: true})
-	if err != nil {
-		return fmt.Errorf("unable to update DID document: %w", err)
+
+	// Resolve version of DID Document referred to by transaction
+	var currentDIDDocument *did.Document
+	var err error
+	for _, ref := range transaction.Previous() {
```
gerardsn, line 239: The `prevs` currently point to the `head` and the `doc to update`. Both refs could point to different versions of the same DID document and there is no way to keep them apart. So we cannot assume that the first ref that points to the DID doc is intended as the ref to the version of the DID doc to update.

This should probably never happen on our node™, but maybe on other implementations / (future) clustering solutions?

woutslakhorst: Actually we can, because if the tx points to two doc updates, then those were actually conflicted and the metadata should contain refs to both tx. Therefore the resolve will find the same version if either f the prevs was used.

vdr/ambassador.go
woutslakhorst: There's still a fallback to find the latest. If the controller has been updated in the race condition this could go wrong, but there's no way to prevent that or fix it automatically (apart from the lock on API lvl which is already there).

vdr/ambassador.go
gerardsn: I was thinking of the case where the two prevs are not conflicted. The update currently being processed is the one that will create the conflict with the current head.

For example:
tx1=DID create, [head==tx0]<=prevs
...
tx10=DID update of [tx1, head==tx9]<=prevs
tx11=DID update of [tx1, head==tx10]<=prevs

tx10 and tx11 will create a conflict, but that requires tx11 to be accepted by the VDR. If tx10 is the first ref in `prevs` of tx11, the current solution will try to update tx10 twice, while it should update tx1. Out of order processing in the VDR means that the state of a DID document is not deterministic after tx10 and tx11 have been processed.

vdr/ambassador.go
gerardsn: Exactly, that is what makes the error recoverable right? Since the order of processing cannot be guaranteed, an update that is invalid right now could be valid in the future.

vdr/ambassador.go
woutslakhorst: The VDR document store solves this. If tx10 or tx11 comes first doesn't matter. In either case a timeline is constructed and tx10 will come before tx11. Then tx11 consumes all previous states (tx10 and tx1). All nodes will mark tx11 as the latest state.

> tx10 and tx11 will create a conflict

They don't. tx11 consumes all previous source transactions.

Remember that this logic is for determining the controllers not for determining DID document state.

vdr/ambassador.go
woutslakhorst: It's for controller finding. If the update succeeds with a controller from the previous/current state that's fine as well. If it fails, it might have been a race condition and then the retry will catch it.

gerardsn: We are talking about the same thing here. Only recoverable errors are retried.

gerardsn (approved):

vdr/ambassador.go
gerardsn: edge case tracked in #1873

vdr/ambassador.go
```diff
@@ -232,13 +232,29 @@ func (n *ambassador) handleUpdateDIDDocument(transaction dag.Transaction, propos
...
+		currentDIDDocument, _, err = n.didStore.Resolve(proposedDIDDocument.ID, &types.ResolveMetadata{AllowDeactivated: true, SourceTransaction: &ref})
```
reinkrul, line 240: This was my first approach of fixing this issue, but I found out there were 2 issues with it:

- This also returns deactivated documents, which don't contain controllers or keys, which can lead to the same situation (if referring a deactivated document).
- It just takes the first document it finds in the prevs, which might not contain the controller the TX was signed with (e.g. 2 branches of the DID document, the controller being used only present in one of the branches, which isn't referred).

I think the correct way would be to find all prevs and merge the resulting controllers. And I think `resolveControllers` doesn't take all prevs into account as well?

The issue might be too edge case/hard to fix right now. Do we still allow this situation with the multi chain? If not, it might not be worth fixing 100% (although not looking for deactivated DID documents is probably an easy fix).

PR #1892, opened 2023-02-13
Audit: log events on custom 'audit' log level

### LINKED ISSUE #1888: Audit: do not let configured log level influence audit logging
Events are audit logged on `info` level, but if the `verbosity` config parameter is set to `warn` or higher, it will prevent auditable events from being logged.

The audit log entries should not be influenced by the `verbosity` config parameter.

reinkrul:
Needs some more test coverage, but first review round to see if it's not too hacky.

It creates a new logger, specifically for auditing. At first use, it copies the Formatter from the standard logger and decorates it with a Formatter that alters the log level using byte comparison. The order of default fields is hardcoded, so it only replaces the first occurence.

Fixes #1888

Text output:
```
time="2023-02-13T15:52:47+01:00" level=audit msg="Generating new key pair: kid" actor=test-actor event=CreateNewKey module=Crypto operation=TestModule.TestOperation
```

Colored text output:
```
AUDIT[0000] Generating new key pair: kid                  actor=test-actor event=CreateNewKey module=Crypto operation=TestModule.TestOperation
```

JSON output:
```
{"actor":"test-actor","event":"CreateNewKey","level":"audit","module":"Crypto","msg":"Generating new key pair: kid","operation":"TestModule.TestOperation","time":"2023-02-13T15:54:57+01:00"}
```

### DIFF
--- a/audit/audit.go
+++ b/audit/audit.go
@@ -19,8 +19,13 @@
 package audit
 
 import (
+	"bytes"
 	"context"
+	"encoding/json"
+	"fmt"
 	"github.com/sirupsen/logrus"
+	"strings"
+	"sync"
 )
 
 const (
@@ -31,6 +36,22 @@ const (
 	// CryptoSignJWSEvent occurs when signing a JWS.
 	CryptoSignJWSEvent = "SignJWS"
 )
+
+const auditLogLevel = "audit"
+
+var auditLoggerInstance *logrus.Logger
+var initAuditLoggerOnce = &sync.Once{}
+
+func auditLogger() *logrus.Logger {
+	initAuditLoggerOnce.Do(func() {
+		// Create new logger with custom Formatter, which makes sures the log level is always "audit".
+		// Also override the level for this level, to make sure it is not influenced by a lower log verbosity.
+		auditLoggerInstance = logrus.New()
+		auditLoggerInstance.SetFormatter(&auditFormatter{underlyingFormatter: logrus.StandardLogger().Formatter})
+		auditLoggerInstance.SetLevel(logrus.InfoLevel)
+	})
+	return auditLoggerInstance
+}
 
 // Info provides contextual information for auditable events.
 type Info struct {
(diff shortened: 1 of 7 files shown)


audit/audit.go
```diff
@@ -59,6 +82,45 @@ func InfoFromContext(ctx context.Context) *Info {
...
+	// See sirupsen/logrus@v1.9.0/text_formatter.go:19
+	entry.Level = logrus.InfoLevel
+
+	formattedEntry, err := a.underlyingFormatter.Format(entry)
+	if err != nil {
+		return nil, err
+	}
+	// Replace log level in formatted log message
+	switch a.underlyingFormatter.(type) {
+	case *logrus.JSONFormatter:
```
woutslakhorst, line 101: I would create a `NewAuditFormatter(logrus.StandardLogger().Formatter)` that returns a specific AuditFormatter. In the New func you can do the switch and then each formatter has a simple Format function.

reinkrul:
Ready for review

audit/audit_test.go
```diff
@@ -46,3 +48,64 @@ func TestLog(t *testing.T) {
...
+		})
+	})
+}
+
+func Test_newAuditFormatter(t *testing.T) {
+	t.Run("unsupported", func(t *testing.T) {
+		f, err := newAuditFormatter(&textAuditFormatter{})
+
+		assert.Nil(t, f)
+		assert.Error(t, err)
```
stevenvegt, line 67: what error is expected here?

audit/audit.go
```diff
@@ -32,6 +37,28 @@ const (
 	CryptoSignJWSEvent = "SignJWS"
```
stevenvegt, line 37: Why are these constants in not in the crypto package?

stevenvegt (changes requested):
Looks OK, perhaps explain why we are doing this, and link to the github issue from logrus that it's in maintenance mode.
Also, take a look at the crypto specific constants in the audit engine, looks like these should be in the package of the caller?

audit/audit.go
reinkrul: I chose managability (is that a word?) over code cohesion here, since having them all 1 place makes it much easier to see what events are audited.

stevenvegt (approved):

PR #1893, opened 2023-02-13
allow for custom AuthorizationTokenBuilder on API Client

### LINKED ISSUE #1891: Extend CreateHTTPClient so it can create dynamic API tokens
Required for Cmds and Registry admin

woutslakhorst:
closes #1891 

Using the client API lib for our APIs involved creating a `api.HTTPClient` for various packages and then calling its methods. It was not possible to inject an alternative TokenBuilder in this process. Especially the registry admin benefits from this since it needs to inject its own token builder.

the registry creates clients like:
```
vdrClient := vdrAPI.HTTPClient{
	ServerAddress: config.NutsNodeAddress,
	Timeout:       apiTimeout,
}
```

This PR allows for adding an AuthorizationTokenBuilder impl

note: fixed `data` dir creation from `config_test.go`

### DIFF
--- a/core/http_client.go
+++ b/core/http_client.go
@@ -74,7 +74,7 @@ func (w httpRequestDoerAdapter) Do(req *http.Request) (*http.Response, error) {
 // This does not use the generated client options for e.g. authentication,
 // because each generated OpenAPI client reimplements the client options using structs,
 // which makes them incompatible with each other, making it impossible to use write generic client code for common traits like authorization.
-func CreateHTTPClient(cfg ClientConfig) (HTTPRequestDoer, error) {
+func CreateHTTPClient(cfg ClientConfig, builder AuthorizationTokenBuilder) (HTTPRequestDoer, error) {
 	var result *httpRequestDoerAdapter
 	client := &http.Client{}
 	client.Timeout = cfg.Timeout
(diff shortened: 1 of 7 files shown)


beardedfoo (approved):

core/http_client.go
```diff
@@ -74,35 +74,71 @@ func (w httpRequestDoerAdapter) Do(req *http.Request) (*http.Response, error) {
 // This does not use the generated client options for e.g. authentication,
 // because each generated OpenAPI client reimplements the client options using structs,
 // which makes them incompatible with each other, making it impossible to use write generic client code for common traits like authorization.
-func CreateHTTPClient(cfg ClientConfig) (HTTPRequestDoer, error) {
+func CreateHTTPClient(cfg ClientConfig, builder AuthorizationTokenBuilder) (HTTPRequestDoer, error) {
```
reinkrul, line 77:
```suggestion
// If the given authorization token builder is non-nil, it calls it and passes the resulting token as bearer token with requests.
func CreateHTTPClient(cfg ClientConfig, builder AuthorizationTokenBuilder) (HTTPRequestDoer, error) {
```

reinkrul (approved):

PR #1897, opened 2023-02-16
Revert "Network: change node authentication behavior (#1879)"

gerardsn:
This reverts commit 60527f5fff72e30a9707fccfe5eaab5114171b8a.

This change would just mark connections that failed authenticated as unauthenticated. This means that one end of the connection can be authenticated while the other is not. 

Since nodes can only have a single connection to a peerID, the side that has an unauthenticated connection will not be able to setup an authenticated connection while the other one exists. This creates issues with private credentials, and it is not possible for the side requesting the private credential to see why it keeps failing.


gerardsn:
Fixed in connection management refactor

Issue #1900, opened 2023-02-17
Racy behavior during shutdown due to obscure asynchronous programming


reinkrul:
Components which start goroutines themselves typically look like this (e.g. our modules/engines):

- Have a non-blocking `Start()` function, which starts one or more goroutines. These are passed a context.
- Have a non-blocking `Stop()` function, which cancels the passed contexts to instruct goroutines to stop.

Since `Stop()` is non-blocking, it will not wait until all goroutines have finished processing.

An alternative approach we have, is (e.g. `v2.Protocol` and its `diagnosticsMan`/`listHandler`):

- Have a non-blocking `Start(ctx)` function, which starts one or more goroutines. Context is passed through
- There is no `Stop()` function; goroutines are instructed to stop by cancelling the context.

This leads to racy behavior during shutdown: dependencies (typically storage) will already have been shutdown, while dependent goroutines are still trying to do work. We haven't encountered any big issues or data corruption due to this behavior, currently it's only annoying due to it causing error logs. So it should be fixed.

Given Golang's convention of letting callers start a goroutine for your function, instead of starting it yourself, improvements include:

- The component that starts a goroutine is responsible blocking until it shuts down. This yields to the following design patterns:
  - When `Start()` is non-blocking (meaning it starts goroutines), it must provide a `Stop()` function that blocks until all goroutines have exited
  - Components with a `Start()` (blocking) but without `Stop()` will typically be passed a context to signal shutdown
  - When `Start()` is blocking, it must block until all goroutines have exited
  - When starting goroutines, use a `sync.WaitGroup` to wait for them to exit, e.g.:
     ```
    go func(w *sync.WaitGroup) {
			w.Add(1)
			defer w.Done()
			asyncFunc()
		}(w)
    ```
- When issues arise with long-running goroutines blocking shutdown, the `System.Shutdown` function might have to time-out on `Stop()` functions of modules that take too long (or at least log it).

reinkrul:
Instances of race conditions:
```
[...]

PR #1901, opened 2023-02-17
Network: fix goroutine leak on shutdown in v2 protocol

reinkrul:
Part of https://github.com/nuts-foundation/nuts-node/issues/1900

### DIFF
--- a/network/transport/v2/peer_diagnostics_test.go
+++ b/network/transport/v2/peer_diagnostics_test.go
@@ -39,10 +39,10 @@ func Test_PeerDiagnosticsManager(t *testing.T) {
 
 		ctx, cancel := context.WithCancel(context.Background())
 		defer cancel()
-		manager.start(ctx, 5*time.Millisecond)
+		go manager.start(ctx, 5*time.Millisecond)
 
 		// Wait for the calls
-		wg.Done()
+		wg.Wait()
 	})
 }
 
--- a/network/transport/v2/protocol.go
+++ b/network/transport/v2/protocol.go
@@ -24,6 +24,7 @@ import (
 	"fmt"
 	"github.com/nuts-foundation/go-stoabs"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/nuts-foundation/nuts-node/core"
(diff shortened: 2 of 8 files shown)


network/transport/v2/protocol.go
```diff
@@ -187,13 +189,28 @@ func (p *protocol) Configure(_ transport.PeerID) error {
...
+		p.routines.Add(1)
+		go func(w *sync.WaitGroup) {
+			defer w.Done()
+			p.diagnosticsMan.start(p.ctx, time.Duration(p.config.DiagnosticsInterval)*time.Millisecond)
+		}(p.routines)
 	}
 
-	p.listHandler = newTransactionListHandler(p.ctx, p.handleTransactionList)
-	p.listHandler.start()
+	ctx := p.ctx // use copy of pointer to avoid data race
```
woutslakhorst, line 202: copy of pointer still points to same instance.

network/transport/v2/protocol.go
```diff
@@ -187,13 +189,28 @@ func (p *protocol) Configure(_ transport.PeerID) error {
...
+	// Wrap listHandler function to supply the context to handleTransactionList.
+	// It would be prettier to pass a context around in all protocol message handlers (since most use context.Background() for database access now),
+	// but that is too big a change for now.
+	p.listHandler = newTransactionListHandler(p.ctx, func(peer transport.Peer, envelope *Envelope) error {
```
woutslakhorst, line 206: change the `handleFunc` signature so the ctx is passed as well and you don't need a copy. HandleTransactionList should reuse the context passed by the TransctionListHandler, not the context from the protocol.

woutslakhorst (changes requested):
👍 on using contexts to stop procedures

network/transport/v2/peer_diagnostics_test.go
```diff
@@ -39,7 +39,7 @@ func Test_PeerDiagnosticsManager(t *testing.T) {
 
 		ctx, cancel := context.WithCancel(context.Background())
 		defer cancel()
-		manager.start(ctx, 5*time.Millisecond)
+		go manager.start(ctx, 5*time.Millisecond)
 
 		// Wait for the calls
 		wg.Done()
```
gerardsn, line 45: unrelate to this PR, but this test is broken. it doesn't wait...

woutslakhorst (approved):

PR #1905, opened 2023-02-20
Network: fix racy behavior of backoff persist on shutdown

reinkrul:
The problem was that `openOutboundStream`, a goroutine which blocks until the gRPC connection is closed, tried to reset the backoff timer after the storage engine already shut down. The function is (through intermediates) called by `outboundConnector.start` in a goroutine, and while the component has a `stop()` function, it doesn't wait until the goroutine exits. So what happened, was:

- Connection is active (goroutine blocked by `openOutboundStream()`)
- Kill signal, tells gRPC connections (through context cancellation) to shut down
- `openOutboundStream()` unblocks
- Meanwhile storage engine already shut down, since network module didn't wait until all subcomponents were shut down
- `openOutboundStream()` tries to reset backoff, which fails and renders an error

Fixed this by letting `outboundConnector.stop()` (called by `grpcConnectionManager.Stop()`) block the connector loop has stopped. To let current connection attempts (e.g. attempts to dead hosts, which may take a few seconds to time-out) not block shutdown, `outboundConnector.tryConnect()` now gets passed the context used to cancel the connector loop.

Also added logging exemptions to avoid logging `context.Canceled` as errors (they're normal shutdown artifacts).

Requires https://github.com/nuts-foundation/nuts-node/pull/1901

### DIFF
--- a/network/transport/grpc/connection_manager.go
+++ b/network/transport/grpc/connection_manager.go
@@ -24,8 +24,6 @@ import (
 	"errors"
 	"fmt"
 	"github.com/nuts-foundation/go-did/did"
-	"net"
-
 	"github.com/nuts-foundation/go-stoabs"
 	"github.com/nuts-foundation/nuts-node/core"
 	"github.com/nuts-foundation/nuts-node/network/log"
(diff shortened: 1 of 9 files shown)


reinkrul:
PR will be superseded by refactor of dialing components; https://github.com/nuts-foundation/nuts-node/pull/1874 (in which it will also be fixed)

gerardsn:
> PR will be superseded by refactor of dialing components; #1874 (in which it will also be fixed)

`transport/v2` is not covered by #1874, `transport/grpc` still leaks go routines from `conn` (connection.go)

reinkrul:
Fixes in `transport/v2` were already on master.

reinkrul:
Fixes in `transport/v2` were already on master.

PR #1908, opened 2023-02-20
JWM support in NUTS, part 2. This concerns the (de/en)crypt_jwe metho…

rolandgroen:
This PR is the second part of the JWM support in NUTS. Plain JWM messages can be wrapped in JWS and JWE envelopes to ensure their origin with a signature (JWS) and be confidential by encrypting the message (JWE).

Screenshot 2022-09-22 at 16 14 29

In order to support JWM messages we need to encrypt(sign(plain)) these messages with NUTS. This second part of the JWM support consists of the encryption part.

The following considerations have been taken:

- The encrypt_jwe method takes a `to` field in in the request that can be either the DID reference OR the KID reference. Both are accepted. The implementation tries first to parse the value as DID reference, if it fails it assumes the value is a KID. Than it does a lookup and fails if the reference does not resolve.
- If the reference is a DID reference, the implementation looks up the DID and uses the first keyAgreement, if the reference is a keyID, the implementation looks up the keyId in the keyAgreement map. The KeyResolver interface has some, but not all methods to do so. Instead of duly adding the missing methods (ResolveKeyAgreementKey(kid, time), ResolveKeyAgreementKeyID(did)), which would trigger an overloading conflict ResolveKeyAgreementKey(did). I do not fully understand the semantics and history of the KeyResolver interface interface, it looks a bit like something that has been growing in an ad-hoc fashion, I might have taken an approach that is not inline with the philosophy of the API.
I have added the following:
  - RelationType enum
  - ResolveRelationKey(keyID string, validAt *time.Time, relationType RelationType) (crypto.PublicKey, error)
  - ResolveRelationKeyID(id did.DID, relationType RelationType) (ssi.URI, error)
- TBH, I don't know a whole lot about encryption algorithms, I've chosen:
  - RSA_OAEP_256 for RSA keys
  - ECDH_ES_A256KW for EC keys


woutslakhorst:
For cryptography, please consult https://www.sogis.eu/documents/cc/crypto/SOGIS-Agreed-Cryptographic-Mechanisms-1.2.pdf

reinkrul:
A little bad news; we had to rewrite the master branch due to an unsquashed merge commit, so you'll need to re-checkout the master branch locally and rebase your PR onto the current (squashed) master (effectively dropping the commits)

rolandgroen:
Yup, seen it. Working on it. I'm a bit o a GitHub noob, so it might become e new PR.

PR #1912, opened 2023-02-21
JWM support in NUTS, part 2. The (de/en)crypt_jwe methods

rolandgroen:
This PR is the second part of the JWM support in NUTS. Plain JWM messages can be wrapped in JWS and JWE envelopes to ensure their origin with a signature (JWS) and be confidential by encrypting the message (JWE).

Screenshot 2022-09-22 at 16 14 29

In order to support JWM messages we need to encrypt(sign(plain)) these messages with NUTS. This second part of the JWM support consists of the encryption part.

The following considerations have been taken:

- The encrypt_jwe method takes a `receiver` field in in the request that can be either the DID reference OR the KID reference. Both are accepted. The implementation tries first to parse the value as DID reference, if it fails it assumes the value is a KID. Than it does a lookup and fails if the reference does not resolve.
- If the reference is a DID reference, the implementation looks up the DID and uses the first keyAgreement, if the reference is a keyID, the implementation looks up the keyId in the keyAgreement map. The KeyResolver interface has some, but not all methods to do so. Instead of duly adding the missing methods (ResolveKeyAgreementKey(kid, time), ResolveKeyAgreementKeyID(did)), which would trigger an overloading conflict ResolveKeyAgreementKey(did). I do not fully understand the semantics and history of the KeyResolver interface interface, it looks a bit like something that has been growing in an ad-hoc fashion, I might have taken an approach that is not inline with the philosophy of the API.
I have added the following:
  - RelationType enum
  - ResolveRelationKey(keyID string, validAt *time.Time, relationType RelationType) (crypto.PublicKey, error)
  - ResolveRelationKeyID(id did.DID, relationType RelationType) (ssi.URI, error)
- TBH, I don't know a whole lot about encryption algorithms, I've chosen:
  - RSA_OAEP_256 for RSA keys
  - ECDH_ES_A256KW for EC keys


rolandgroen:
Comment from @woutslakhorst:

For cryptography, please consult https://www.sogis.eu/documents/cc/crypto/SOGIS-Agreed-Cryptographic-Mechanisms-1.2.pdf

crypto/api/v1/api.go
```diff
@@ -119,3 +149,63 @@ func (w *Wrapper) SignJws(ctx context.Context, request SignJwsRequestObject) (Si
...
+		keyID, err = w.K.ResolveRelationKeyID(*id, types.KeyAgreement)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	headers := encryptRequest.Headers
+	resolvedKid := keyID.String()
+	if headerKid, ok := headers[jws.KeyIDKey]; ok && resolvedKid != headerKid {
+		return nil, core.Error(400, "the provided header value for kid: %s does not match the resolved keyID: %s", headerKid, resolvedKid)
```
reinkrul, line 189: Prefer `core.InvalidInputError`

crypto/api/v1/api.go
```diff
@@ -87,6 +92,31 @@ func (signRequest SignJwsRequest) validate() error {
 	return nil
 }
 
+func (signRequest EncryptJweRequest) validate() error {
+	url, err := did.ParseDIDURL(signRequest.To)
+	if err != nil {
+		return err
+	}
+	if url.Empty() {
+		return errors.New("missing to")
```
reinkrul, line 101: I think `to` is a bit weird in JWE terms (I believe it comes from `JWM`?), so `receiver` might be a better name?

crypto/jwx.go
```diff
@@ -84,6 +91,63 @@ func (client *Crypto) SignJWS(ctx context.Context, payload []byte, headers map[s
 	return signJWS(payload, headers, privateKey, detached)
 }
 
+// EncryptJWE encrypts a signed payload using the provided public key and key identifier.
+func (client *Crypto) EncryptJWE(ctx context.Context, payload []byte, headers map[string]interface{}, publicKey interface{}) (string, error) {
+	audit.Log(ctx, log.Logger(), audit.CryptoEncryptJWEEvent).Infof("Encrypting a JWE")
```
reinkrul, line 96:
```suggestion
	audit.Log(ctx, log.Logger(), audit.CryptoEncryptJWEEvent).Info("Encrypting a JWE")
```

crypto/jwx.go
```diff
@@ -84,6 +91,63 @@ func (client *Crypto) SignJWS(ctx context.Context, payload []byte, headers map[s
...
+	}
+
+	audit.Log(ctx, log.Logger(), audit.CryptoDecryptJWEEvent).Infof("Decrypting a JWE with kid: %s", kid)
+
+	body, err = jwe.Decrypt([]byte(message), protectedHeaders.Algorithm(), privateKey)
+	if err != nil {
+		return nil, nil, err
+	}
+	headers = make(map[string]interface{})
+	if uh := msg.UnprotectedHeaders(); uh != nil {
```
reinkrul, line 124: do unprotected headers make sense? Could it be dangerous to have them returned (to the API client) in 1 header map? (then the client can't differentiate between protected and unprotected headers, which might lead to false assumptions).

If we don't want to complicate the API by returning them in a separate header map, do we need them for the foreseen use cases? If not, could be better to just not support it for now (until someone requests it)

crypto/jwx.go
```diff
@@ -249,6 +313,38 @@ func signJWS(payload []byte, protectedHeaders map[string]interface{}, privateKey
 	return string(data), nil
 }
 
+func EncryptJWE(payload []byte, protectedHeaders map[string]interface{}, publicKey interface{}) (message string, err error) {
+
```
reinkrul, line 317:
```suggestion
```

crypto/jwx.go
```diff
@@ -338,6 +434,35 @@ func SignatureAlgorithm(key crypto.PublicKey) (jwa.SignatureAlgorithm, error) {
...
+	switch ptr.(type) {
+	case *crypto.PublicKey:
+		return defaultEcEncryptionAlgorithm, nil
+	case *rsa.PublicKey:
+		return defaultRsaEncryptionAlgorithm, nil
+	case *ecdsa.PublicKey:
+		return defaultEcEncryptionAlgorithm, nil
+	default:
+		return "", fmt.Errorf(`could not determine signature algorithm for key type '%T'`, key)
```
reinkrul, line 462: No need for backtiks here
```suggestion
		return "", fmt.Errorf("could not determine signature algorithm for key type '%T'", key)
```

crypto/jwx.go
```diff
@@ -338,6 +434,35 @@ func SignatureAlgorithm(key crypto.PublicKey) (jwa.SignatureAlgorithm, error) {
 	}
 }
 
+func EncryptionAlgorithm(key crypto.PublicKey) (jwa.KeyEncryptionAlgorithm, error) {
```
reinkrul, line 437: I think this doesn't need to be exported

crypto/jwx.go
```diff
@@ -338,6 +434,35 @@ func SignatureAlgorithm(key crypto.PublicKey) (jwa.SignatureAlgorithm, error) {
...
+	if key == nil {
+		return "", errors.New("no key provided")
+	}
```
reinkrul, line 440: I'd say move this check to EncryptJWE, make this func private

vdr/types/interface.go
```diff
@@ -139,3 +146,13 @@ type DocManipulator interface {
 	// It returns an ErrDIDNotManagedByThisNode if the DID document is not managed by this node.
 	AddVerificationMethod(ctx context.Context, id did.DID, keyUsage DIDKeyFlags) (*did.VerificationMethod, error)
 }
+
+type RelationType uint
+
+const (
+	Authentication       RelationType = 1
```
reinkrul, line 153: You can "automate" this using iota; https://yourbasic.org/golang/iota/

vdr/didservice/resolvers.go
```diff
@@ -176,6 +182,24 @@ func (r KeyResolver) ResolveSigningKey(keyID string, validAt *time.Time) (crypto
...
+		return doc.Authentication, nil
+	case types.AssertionMethod:
+		return doc.AssertionMethod, nil
+	case types.KeyAgreement:
+		return doc.KeyAgreement, nil
+	case types.CapabilityInvocation:
+		return doc.CapabilityInvocation, nil
+	case types.CapabilityDelegation:
+		return doc.CapabilityDelegation, nil
+
```
reinkrul, line 197:
```suggestion
```

vdr/didservice/resolvers.go
```diff
@@ -176,6 +182,24 @@ func (r KeyResolver) ResolveSigningKey(keyID string, validAt *time.Time) (crypto
...
+	default:
+		return nil, errors.New(fmt.Sprintf("unable to locate RelationType %v", relationType))
```
reinkrul, line 199:
```suggestion
		return nil, fmt.Errorf("unable to locate RelationType %v", relationType)
```

reinkrul (changes requested):

crypto/jwx.go
rolandgroen: Agree not to support it for now, this also results in less code.

crypto/jwx.go
rolandgroen: Agree.

crypto/api/v1/api.go
rolandgroen: Renamed to receiver

crypto/jwx.go
rolandgroen: Done

rolandgroen:
@reinkrul Ik heb alle aanpassingen doorgevoerd, maar op een of andere manier blijft gitlab melden dat niet alle verzochte aanpassingen zijn gedaan.

reinkrul:
@rolandgroen I'll look into it at Tuesday.

crypto/jwx.go
```diff
@@ -84,6 +90,42 @@ func (client *Crypto) SignJWS(ctx context.Context, payload []byte, headers map[s
 	return signJWS(payload, headers, privateKey, detached)
 }
 
+// EncryptJWE encrypts a signed payload using the provided public key and key identifier.
```
reinkrul, line 93: the payload isn't necesarrily signed? (only when used in the context of JWM)

crypto/jwx.go
```diff
@@ -84,6 +90,42 @@ func (client *Crypto) SignJWS(ctx context.Context, payload []byte, headers map[s
...
+func (client *Crypto) EncryptJWE(ctx context.Context, payload []byte, headers map[string]interface{}, publicKey interface{}) (string, error) {
+	audit.Log(ctx, log.Logger(), audit.CryptoEncryptJWEEvent).Info("Encrypting a JWE")
+	return EncryptJWE(payload, headers, publicKey)
+}
+
+// DecryptJWE decrypts a signed message using the associated private key from the kid header.
```
reinkrul, line 99: the payload isn't necesarrily signed? (only when used in the context of JWM)

reinkrul (approved):
Some small remarks on godoc

reinkrul:
Maybe @gerardsn or @stevenvegt can review this as well?

crypto/jwx.go
rolandgroen: Correct, removed the signed word from the comment.

crypto/jwx.go
rolandgroen: Correct, removed the signed word from the comment.

woutslakhorst:
Currently sogis.eu/documents/cc/crypto/SOGIS-Agreed-Cryptographic-Mechanisms-1.2.pdf defines that only RSA encryption is allowed for asymmetric encryption. We'll have to keep that in mind during review.

reinkrul:
RSA keys currently can't be created by the Nuts node (defaults to EC, no way to change it). To be compliant with the SOGIS guideline, RSA keys must be supported in the crypto `New` operation (also see https://github.com/nuts-foundation/nuts-node/issues/1948). But our of scope for now.

crypto/api/v1/api.go
```diff
@@ -119,3 +157,71 @@ func (w *Wrapper) SignJws(ctx context.Context, request SignJwsRequestObject) (Si
...
+
+	jwe, err := w.C.EncryptJWE(ctx, encryptRequest.Payload, headers, key)
+	if err != nil {
+		return nil, err
+	}
+	return EncryptJwe200TextResponse(jwe), err
+}
+
+// DecryptJwe handles api calls for decrypting JWE messages
+func (w *Wrapper) DecryptJwe(ctx context.Context, request DecryptJweRequestObject) (DecryptJweResponseObject, error) {
```
stevenvegt, line 217: Missing tests in api_test.go

crypto/api/v1/api.go
```diff
@@ -119,3 +157,71 @@ func (w *Wrapper) SignJws(ctx context.Context, request SignJwsRequestObject) (Si
...
+		return nil, core.InvalidInputError("invalid encrypt request: %w", err)
+	}
+	receiver := encryptRequest.Receiver
+	id, err := did.ParseDIDURL(receiver)
+	if err != nil {
+		return nil, err
+	}
+	var key crypt.PublicKey
+	var keyID ssi.URI
+	if id.DID.IsURL() {
```
stevenvegt, line 174: Perhaps extract this key fetching code into its own helper function?

stevenvegt (changes requested):
First review part, need more time to complete, but now you can already work on it.

crypto/api/v1/api.go
```diff
@@ -119,3 +149,63 @@ func (w *Wrapper) SignJws(ctx context.Context, request SignJwsRequestObject) (Si
...
+		}
+		keyID, err = w.K.ResolveRelationKeyID(*id, types.KeyAgreement)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	headers := encryptRequest.Headers
+	resolvedKid := keyID.String()
+	if headerKid, ok := headers[jws.KeyIDKey]; ok && resolvedKid != headerKid {
```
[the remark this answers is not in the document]
stevenvegt: I already removed this. So ignore.

crypto/api/v1/api.go
```diff
@@ -87,6 +92,31 @@ func (signRequest SignJwsRequest) validate() error {
 	return nil
 }
 
+func (signRequest EncryptJweRequest) validate() error {
+	url, err := did.ParseDIDURL(signRequest.Receiver)
+	if err != nil {
+		return err
+	}
+	if url.Empty() {
```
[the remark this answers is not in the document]
stevenvegt: I removed this already.

crypto/api/v1/api.go
```diff
@@ -87,6 +92,31 @@ func (signRequest SignJwsRequest) validate() error {
...
+		return errors.New("missing receiver")
+	}
+	if signRequest.Headers == nil {
+		return errors.New("missing headers")
+	}
+	if signRequest.Payload == nil {
+		return errors.New("missing payload")
+	}
```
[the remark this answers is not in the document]
rolandgroen: Test coverage has been improved.

crypto/api/v1/api.go
```diff
@@ -87,6 +92,31 @@ func (signRequest SignJwsRequest) validate() error {
...
+
+	return nil
+}
+func (signRequest DecryptJweRequest) validate() error {
```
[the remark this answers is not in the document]
rolandgroen: Test coverage has been improved.

crypto/api/v1/api.go
rolandgroen: Test coverage has been improved.

crypto/api/v1/api.go
rolandgroen: Created helper function.

rolandgroen:
Boosted test coverage, made sure the service returns 400 on invalid user input. I'm a bit confused by the types.ErrNotFound and types.ErrKeyNotFound. Do they return 400 or 500 class errors? For now I wrapped them in a core.InvalidInputError.

reinkrul:
Feature is scheduled for v5.2, I added the draft release notes

stevenvegt:
I wonder, are there "standard" tests for these JWE messages we can embed so we know we create the messages in a compatible format?

stevenvegt:
We should make this an experimental feature by adding a note to the API specification. This allows us to make changes in minor releases instead of waiting to a major.

vdr/didservice/resolvers.go
```diff
@@ -202,7 +234,14 @@ func (r KeyResolver) ResolveKeyAgreementKey(id did.DID) (crypto.PublicKey, error
 // ExtractAssertionKeyID returns a assertionMethod ID from the given DID document.
 // it returns types.ErrKeyNotFound is no assertionMethod key is present.
 func ExtractAssertionKeyID(doc did.Document) (ssi.URI, error) {
-	keys := doc.AssertionMethod
+	return ExtractRelationKeyID(doc, types.AssertionMethod)
+}
+
+func ExtractRelationKeyID(doc did.Document, relationType types.RelationType) (ssi.URI, error) {
```
stevenvegt, line 240: Needs comments

vdr/didservice/resolvers.go
stevenvegt, line 1: Make sure to write tests for this new logic

crypto/jwx.go
stevenvegt, line 1: Make sure to add some tests for the error conditions (missing keys, parsing errors etc.)

docs/_static/crypto/v1.yaml
```diff
@@ -62,7 +62,70 @@ paths:
                 example: "aa==.bb==.cc=="
         default:
           $ref: '../common/error_response.yaml'
+  /internal/crypto/v1/encrypt_jwe:
+    post:
+      summary: "encrypt a payload and headers with the public key of the given DID into a JWE object"
+      description: |
```
stevenvegt, line 68: Can you link to the RFC. What is supported, what not. Which encryption algorithms are used (by default).

docs/_static/crypto/v1.yaml
stevenvegt, line 1: Can you add a warning that this is experimental logic and might change in minor updates. This allows us to update the logic without having to release a major version.

vdr/didservice/resolvers.go
```diff
@@ -176,6 +182,23 @@ func (r KeyResolver) ResolveSigningKey(keyID string, validAt *time.Time) (crypto
 	return result.PublicKey()
 }
 
+func resolveRelationship(doc *did.Document, relationType types.RelationType) (relationships did.VerificationRelationships, err error) {
```
stevenvegt, line 185: Needs docs and tests

vdr/types/interface.go
```diff
@@ -139,3 +146,13 @@ type DocManipulator interface {
 	// It returns an ErrDIDNotManagedByThisNode if the DID document is not managed by this node.
 	AddVerificationMethod(ctx context.Context, id did.DID, keyUsage DIDKeyFlags) (*did.VerificationMethod, error)
 }
+
+type RelationType uint
```
stevenvegt, line 150: Needs docs

stevenvegt (changes requested):
Almost there! I altered some of the api tests and added some review comments about missing docs and test coverage.

stevenvegt:
e2e tests succeed.

docs/pages/release_notes.rst
```diff
@@ -7,14 +7,15 @@ Release notes
...
+Hazelnut update (v5.2.0)
+***********************
+
+Release date: **DRAFT**
 
-- Some VDR OpenAPI operations specified ``application/json+did-document`` as Content-Type, while they actually returned ``application/json``.
-  This inconsistency is fixed by changing the OpenAPI specification to ``application/json``.
-- Diagnostics now show the conflicted document count for DID Documents the node controls. See monitoring documentation for more detail.
-- ``network.connections.outbound_connectors`` on ``/status/diagnostics`` has been moved to ``/internal/v1/network/addressbook``.
-  Previously it showed only failing connections, now it shows all addresses it will try to connect to (regardless it's already connected to them or not).
```
reinkrul, line 16: I don't think this was meant to be removed

docs/pages/release_notes.rst
```diff
@@ -7,14 +7,15 @@ Release notes
 (To be released)
```
reinkrul, line 7: This section can go now as well

docs/_static/crypto/v1.yaml
```diff
@@ -64,9 +64,10 @@ paths:
           $ref: '../common/error_response.yaml'
   /internal/crypto/v1/encrypt_jwe:
     post:
-      summary: "encrypt a payload and headers with the public key of the given DID into a JWE object"
+      summary: "Encrypt a payload and headers with the public key of the given DID into a JWE object"
       description: |
         Encrypt a payload and headers with the public key of the given DID into a JWE object
+        Note: this feature is experimental and can be changed in a future minor release.
```
reinkrul, line 70:
```suggestion
        
        Note: this feature is experimental and might be changed in a future minor release without prior notice.
```

reinkrul (changes requested):

stevenvegt (approved):

reinkrul (approved):

Issue #1911, opened 2023-02-20
Lost issuer DB prevents revoking of issued credentials


woutslakhorst:
issued credentials are stored in the issuer db and the vcr. If you lose your issuer db, the network will send you a copy to the vcr. At that point you can see and use the vc but you can no longer revoke it.

"lost" credentials recovered from the network should find their way back into the issuer DB.

reinkrul:
As designed; database should be backed up. But we could introduce a "rebuild" feature; finding all credentials issued by specified issuer, put them in issuer DB.

Maybe we could reconsider the issuer DB and make it more resilient, either by completely eliminating it (always work on VCs received over network, which may not be feasible for private VCs), or making it more like "volatile" storage.

woutslakhorst:
> But we could introduce a "rebuild" feature; finding all credentials issued by specified issuer, put them in issuer DB.

Required since a backup could be missing some VCs but the network already has them. It's hole in the 'feature'.

reinkrul:
The easiest fix is to not check for credential existence. Fix on v5 and release patch version.

PR #1913, opened 2023-02-21
VCR: Don't check VC existence when revoking

reinkrul:
Fixes https://github.com/nuts-foundation/nuts-node/issues/1911

Also needs to be applied to master branch.

### DIFF
--- a/vcr/issuer/issuer.go
+++ b/vcr/issuer/issuer.go
@@ -168,12 +168,6 @@ func (i issuer) buildVC(credentialOptions vc.VerifiableCredential) (*vc.Verifiab
 }
 
 func (i issuer) Revoke(credentialID ssi.URI) (*credential.Revocation, error) {
-	// first find it using a query on id.
-	credentialToRevoke, err := i.store.GetCredential(credentialID)
-	if err != nil {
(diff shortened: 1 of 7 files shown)


vcr/issuer/issuer.go
```diff
@@ -168,12 +168,6 @@ func (i issuer) buildVC(credentialOptions vc.VerifiableCredential) (*vc.Verifiab
 }
 
 func (i issuer) Revoke(credentialID ssi.URI) (*credential.Revocation, error) {
-	// first find it using a query on id.
```
stevenvegt, line 171: Add a comment why you don't have to search in the store for the future Rein.

stevenvegt (changes requested):

reinkrul:
Failing end-to-end test is the secret storage (Vault) test, which is feature not availaible in v5.0.x. Running test suite locally with that specific test disabled.

stevenvegt (changes requested):
so, it publishes a revocation about every possible identifier without checking for the existence. Is that smart? There is a risk that a wrong identifier will not raise any errors, however, no credential is revoked.

stevenvegt (approved):

PR #1919, opened 2023-02-22
Network: dynamic service discovery

### LINKED ISSUE #690: Implement dynamic service discovery
Using RFC015 `NutsComm` endpoints.

gerardsn:
closes #690


network/interface.go
```diff
@@ -56,6 +57,8 @@ type Transactions interface {
 	// WithPersistency returns a SubscriberOption for persistency. It allows the DAG KVStore to be used as persistent store for notifications.
 	// The notifications will then have ACID properties
 	WithPersistency() SubscriberOption
+	// ServiceDiscovery should be called by the VDR to let the network know it has processed and verified a document (update) for the DID.
+	ServiceDiscovery(updatedDID did.DID)
```
reinkrul, line 61: I'd expect a verb in a function (except when it just returns something), e.g. `DiscoveryServices()`

network/network.go
```diff
@@ -58,6 +58,8 @@ const (
 	// health check keys
 	healthTLS        = "tls"
 	healthAuthConfig = "auth_config"
+	// newNodeConnectionDelay specifies how long the node should delay connecting to new NutsComm addresses.
```
reinkrul, line 61: maybe a bit more background info on why

network/network.go
```diff
@@ -81,6 +83,7 @@ type Network struct {
 	didDocumentFinder   types.DocFinder
 	eventPublisher      events.Event
 	storeProvider       storage.Provider
+	assumeNewNode       bool // node has assumed it has not performed initial sync with the network
```
reinkrul, line 86:
```suggestion
        // assumeNewNode indicates the node hasn't initially sync'd with the network.
	assumeNewNode       bool
```

network/network.go
```diff
@@ -284,6 +287,28 @@ func (n *Network) Configure(config core.ServerConfig) error {
...
+	if !n.config.EnableDiscovery {
+		return
+	}
+	document, _, err := n.didDocumentResolver.Resolve(updatedDID, nil)
+	if err != nil {
+		// VDR store is down. Any missed updates are resolved on node restart.
+		// This can happen when the VDR is receiving lots of DID updates, such as during the initial sync of the network.
+		log.Logger().WithError(err).
+			WithField(core.LogFieldDID, updatedDID.String()).
+			Error("service discovery could not read DID document after an update")
```
reinkrul, line 300:
```suggestion
			Error("Service discovery could not read DID document after an update")
```

network/network.go
```diff
@@ -284,6 +287,28 @@ func (n *Network) Configure(config core.ServerConfig) error {
...
+		return
+	}
+	nodeDID, err := n.nodeDIDResolver.Resolve()
+	if err != nil {
+		// This can only occur when the autoNodeDIDResolver is used (non-strict mode)
+		log.Logger().WithError(err).Error("could not resolve own node DID for service discovery")
```
reinkrul, line 306:
```suggestion
		log.Logger().WithError(err).Error("Could not resolve own node DID for service discovery")
```

network/network.go
```diff
@@ -369,30 +394,48 @@ func (n *Network) connectToKnownNodes(nodeDID did.DID) error {
...
-						WithError(err).
-						WithField(core.LogFieldDID, node.ID.String()).
-						Warn("Failed to extract NutsComm address from service")
-					continue inner
-				}
+		n.connectToDID(nodeDID, node)
+	}
+
+	// Assume this is a new node when no NutsComm endpoints are found.
+	n.assumeNewNode = len(otherNodes) == 0
```
reinkrul, line 401: `connectToDID()` uses this field, which is called above, so shouldn't it be set before the loop?

(indicates a missing test?)

network/network.go
```diff
@@ -369,30 +394,48 @@ func (n *Network) connectToKnownNodes(nodeDID did.DID) error {
...
+	if n.assumeNewNode {
+		log.Logger().Infof("assuming this is a new node, discovered NutsComm addresses are processed with a %s delay", newNodeConnectionDelay)
```
reinkrul, line 403:
```suggestion
		log.Logger().Infof("Assuming this is a new node, discovered NutsComm addresses are processed with a %s delay", newNodeConnectionDelay)
```

network/network.go
```diff
@@ -369,30 +394,48 @@ func (n *Network) connectToKnownNodes(nodeDID did.DID) error {
...
+	}
+
+	return nil
+}
+
+func (n *Network) connectToDID(nodeDID did.DID, node did.Document) {
+	if !nodeDID.Empty() && node.ID.Equals(nodeDID) {
+		// Found local node, do not discover.
```
reinkrul, line 411: add note for clustering?

reinkrul (changes requested):

network/network.go
gerardsn: what would be in the note? 
The nodes in the cluster should still not connect to their own DID using the NutsComm address?

network/network.go
gerardsn: if `len(otherNodes) == 0` the loop is ran 0 times.
Will move it up to avoid confusion

network/interface.go
gerardsn: changed to `DiscoverServices()`

network/network.go
```diff
@@ -368,31 +396,45 @@ func (n *Network) connectToKnownNodes(nodeDID did.DID) error {
...
+					Debug("Failed to extract NutsComm address from service")
+				continue inner
+			}
+			log.Logger().
+				WithField(core.LogFieldDID, node.ID.String()).
+				WithField(core.LogFieldNodeAddress, nutsCommUrl.Host).
+				Debug("Discovered Nuts node")
+			if n.assumeNewNode {
+				// Connect to NutsComm addresses with a delay.
+				n.connectionManager.Connect(nutsCommUrl.Host, node.ID, newNodeConnectionDelay)
```
reinkrul, line 432: Maybe a bit odd, that `assumeNewNode` is set on startup; when the node has been running for a while it isn't a new node anymore; but new addresses will be 5 minutes delayed before being connected to.

Maybe `newNodeConnectionDelay` should be a fixed time, e.g. `time.Now().Add(5 * time.Minute)`, which is then substracted in the `Connect()` call here?

Then after 5 minutes, new nodes are being connected to immediately.

Or simplify it, if `boottime + 5 minutes` has passed, pass `0`

reinkrul (changes requested):

network/network.go
gerardsn: I've applied the last suggestion.

Not sure if this actually relevant. A node that has never been restarted is an observer node. I don't think the delay is going to make a difference there.

reinkrul: Sort of agree; but maybe it will work a bit nicer in workshop environments. We'll see.

reinkrul (approved):

Issue #1858, opened 2023-01-27
Core: HealthCheckable should accept context


stevenvegt:
This allows cancellation of the request.

woutslakhorst:
Due to this being called via an API call?

reinkrul:
Not applicable for now.

PR #1920, opened 2023-02-23
Status: pass context for healthchecks to allow timing out on external resources

reinkrul:
For instance, external crypto providers (Vault or external). Or when the client already cancelled the request.

Fixes https://github.com/nuts-foundation/nuts-node/issues/1858

### DIFF
--- a/core/status/engine.go
+++ b/core/status/engine.go
@@ -21,6 +21,7 @@ package status
 
 import (
 	"bytes"
+	"context"
 	"fmt"
 	"github.com/sirupsen/logrus"
 	"gopkg.in/yaml.v3"
@@ -37,6 +38,8 @@ const moduleName = "Status"
 const diagnosticsEndpoint = "/status/diagnostics"
 const statusEndpoint = "/status"
 const checkHealthEndpoint = "/health"
+
+const healthCheckTimeout = 3 * time.Second
 
 type status struct {
 	system    *core.System
@@ -116,7 +119,7 @@ func (s *status) collectDiagnostics() map[string][]core.DiagnosticResult {
 }
 
 func (s *status) checkHealth(ctx echo.Context) error {
-	result := s.doCheckHealth()
+	result := s.doCheckHealth(ctx.Request().Context())
 	responseCode := 200
 	if result.Status != core.HealthStatusUp {
 		var failures []string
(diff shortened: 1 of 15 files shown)


core/status/engine_test.go
```diff
@@ -138,14 +144,45 @@ func Test_status_healthChecks(t *testing.T) {
...
 	t.Run("no engines", func(t *testing.T) {
 		s := status{system: core.NewSystem()}
 
-		result := s.doCheckHealth()
+		result := s.doCheckHealth(ctx)
 
 		assert.Equal(t, core.HealthStatusUp, result.Status)
 		assert.Empty(t, result.Details)
 	})
+	t.Run("concurrent checking", func(t *testing.T) {
```
stevenvegt, line 156: What does this do? What should a developer do when it fails?

stevenvegt (commented):
Do we need to check that a context.Cancel works the way as expected?

core/status/engine.go
```diff
@@ -132,16 +135,40 @@ func (s *status) checkHealth(ctx echo.Context) error {
...
-			for name, result := range checker.CheckHealth() {
-				results[strings.ToLower(checker.Name()+"."+name)] = result
-			}
+		if !ok {
+			// does not support health checking
+			return
 		}
+		checkCount++
+		go func(checker core.HealthCheckable) {
+			subCtx, cancel := context.WithTimeout(ctx, healthCheckTimeout)
```
stevenvegt, line 155: What is the difference between setting a timeout on the http.client which performs a healthcheck, and adding a timeout to the context?

reinkrul: In practice nothing, as long as the time-outs are configured the same.

But this way, the caller (health check engine) has control over the time-out. And the API client could cancel the health check by aborting the HTTP request (very niche feature not actually used, probably). But still :-p

crypto/storage/vault/vault.go
```diff
@@ -60,6 +61,7 @@ func DefaultConfig() Config {
 // logicaler is an interface which has been implemented by the mockVaultClient and real vault.Logical to allow testing vault without the server.
 type logicaler interface {
 	Read(path string) (*vault.Secret, error)
+	ReadWithDataWithContext(ctx context.Context, path string, data map[string][]string) (*vault.Secret, error)
```
gerardsn, line 64: when should `ReadWithDataWithContext` be used over `Read`? 
`data` seems to be unused, can we use `ReadWithContext` instead?

gerardsn (changes requested):

core/status/engine.go
```diff
@@ -132,16 +135,40 @@ func (s *status) checkHealth(ctx echo.Context) error {
...
+			defer cancel()
+			result := checker.CheckHealth(subCtx)
```
woutslakhorst, line 157: What if the checker ignores context? Should't your outer loop include a select on return channel and timeout? eg, handle timeouts here and not in each individual checker?

woutslakhorst (changes requested):

reinkrul:
Does not solve an actual issue, you can always configure the Vault client timeout to achieve the same status.

Maybe if we diagnose more external resources.

woutslakhorst:
> Does not solve an actual issue, you can always configure the Vault client timeout to achieve the same status.
> Maybe if we diagnose more external resources.

What if somebody does not add a timeout to its custom external key storage? Then the node hangs....

PR #1922, opened 2023-02-24
Auth: only connect to HTTPS endpoints when in strict mode

reinkrul:
Other outbound HTTP connections the Nuts node will make are:

- Downloading IRMA schemes; this is secured by the IRMA server
- Downloading CRLs, I couldn't find definitive evidence this should only be done over HTTPS. Under control of the respective CAs, so should be safe according to their CSPs.
- Downloading JSON-LD schemas; under control of the operator (`jsonld.contexts.remoteallowlist`) (good enough?)

To allow this change, I had to make the following changes:
- Move `RequestAccessToken` HTTP invocation (on remote Nuts node) from API to service layer (also leads to cleaner code)
- Now codeclimate (rightfully) complained about `oauth.go` being too large, realized that it implements multiple OAuth2 roles. Split it in `AuthorizationServer` and `RelyingParty`.
- Also: removed unused `GetOAuthEndpointURL `
- Also: moved Auth's (OAuth) API client to `api/client` to circumvent cyclic dependencies


docs/pages/deployment/configuration.rst
```diff
@@ -85,4 +85,6 @@ As a general safety precaution ``auth.contractvalidators`` ignores the ``dummy``
 requesting an access token from another node on ``/n2n/auth/v1/accesstoken`` does not return any error details,
 ``http.default.cors.origin`` does not allow a wildcard (``*``),
 json-ld context can only be downloaded from trusted domains configured in ``jsonld.contexts.remoteallowlist``,
-and the ``internalratelimiter`` is always on.
\ No newline at end of file
+and the ``internalratelimiter`` is always on.
+
+Interacting with remote Nuts nodes requires HTTPS: it will refuse to connect to plain HTTP endpoints when in strict mode.
```
gerardsn, line 90: this is very vague. "internal" endpoints (/internal, /health, /status, ...) do not require https

gerardsn (commented):
how does this work with TLS offloading in proxy?

docs/pages/deployment/configuration.rst
gerardsn: aah, missed the `remote` part

reinkrul:
> how does this work with TLS offloading in proxy?

I'd about accessing remote endpoints, regardless of how the remote node handles TLS offloading you'll be connecting to an HTTPS endpoint. We don't support outgoing proxies.

woutslakhorst (changes requested):
Using the engine level interface on the Wrapper is not a good idea. This creates spaghetti. I think it would be cleaner to move client logic from the API layer to the services layer where the configuration has been loaded properly.

reinkrul:
I agree, will be a larger refactor though.

auth/api/v1/api.go
```diff
@@ -23,10 +23,10 @@ import (
 	"fmt"
```
woutslakhorst, line 23: sort imports, the imports on this file are a mess

auth/api/v1/api.go
```diff
@@ -58,6 +59,13 @@ const (
 type Wrapper struct {
 	Auth               auth.AuthenticationServices
 	CredentialResolver vcr.Resolver
+	strictMode         bool
```
woutslakhorst, line 62: unused

auth/api/v1/api.go
```diff
@@ -58,6 +59,13 @@ const (
...
+}
+
+// Configure is called by the system to configure the API wrapper.
+func (w *Wrapper) Configure(config core.ServerConfig) error {
```
woutslakhorst, line 66: Can be removed now?

core/engine.go
```diff
@@ -129,15 +129,29 @@ func (system *System) Configure() error {
...
 			err = m.Configure(*system.Config)
 			coreLogger.Debugf("Configured %s", engineName(engine))
 		}
 		return err
 	})
+	if err != nil {
+		return nil
+	}
+	// Configure API routers
+	for _, router := range system.Routers {
```
woutslakhorst, line 145: May be removed now?

woutslakhorst (changes requested):

core/engine.go
reinkrul: Forgot to revert, good one

auth/services/oauth/relying_party.go
```diff
@@ -0,0 +1,156 @@
...
+		httpClient.Transport = &http.Transport{
+			TLSClientConfig: s.httpClientTLS,
+		}
+	}
+	authClient, err := client.NewHTTPClient("", s.httpClientTimeout, client.WithHTTPClient(httpClient), client.WithRequestEditorFn(core.UserAgentRequestEditor))
+	if err != nil {
+		return nil, fmt.Errorf("unable to create HTTP client: %w", err)
+	}
+	authServerEndpoint, err := url.Parse(authorizationServerEndpoint)
+	if err != nil {
```
stevenvegt, line 88: better to check the scheme after the parsing?

stevenvegt (changes requested):
Move the scheme check after the url parsing, so you don't have to perform these string operations

auth/api/v1/client/types.go
```diff
@@ -0,0 +1,36 @@
+/*
+ * Copyright (C) 2021 Nuts community
```
woutslakhorst, line 2: Yes, we won 2 years

woutslakhorst (approved):

Issue #1931, opened 2023-02-28
Auth: Inconsistency on check of organisation credentials


jelmerterwal:
When using the drawup contract API and using the responding 'message' part for a customer that has an '&' in their 'OrganizationCredential' it will introduce problems when verifying the credentials at the other side (other vendors node).

Example response;
```
Token request validation failed, legal entity mismatch (expected name=xxx & xxx, actual name=xxx &amp; xxx, expected city=xx, actual city=xx)
```

reinkrul:
Due to Mustache template rendering in `auth/contract/template.go:99`; variables with `&` gets replaced with `&amp;`

reinkrul:
Fixed by https://github.com/nuts-foundation/nuts-node/pull/1937

PR #1936, opened 2023-02-28
Auth: do not escape parameters in rendered IRMA contracts

reinkrul:
Fixes https://github.com/nuts-foundation/nuts-node/issues/1931

Also apply on master.


auth/contract/contract_test.go
```diff
@@ -222,4 +223,12 @@ func TestParseContractString(t *testing.T) {
 		assert.NotNil(t, err)
 		assert.True(t, errors.Is(err, ErrInvalidContractText))
 	})
+	t.Run("ok", func(t *testing.T) {
```
stevenvegt, line 226: I would make this explicitly an ampersand test and rename it to something like: "ok - it parses a contract text containing ampersand characters"

auth/contract/contract_test.go
```diff
@@ -222,4 +223,12 @@ func TestParseContractString(t *testing.T) {
...
+		rawText := "NL:BehandelaarLogin:v1 Ondergetekende geeft toestemming aan Demo EHR om namens Toon & Zoonen en ondergetekende het Nuts netwerk te bevragen. Deze toestemming is geldig van dinsdag, 1 oktober 2019 13:30:42 tot dinsdag, 1 oktober 2019 14:30:42."
+		signedContract, err := ParseContractString(rawText, StandardContractTemplates)
+
+		require.NoError(t, err)
+		require.NotNil(t, signedContract)
+		assert.Equal(t, "", signedContract.Params[""])
```
stevenvegt, line 232: why checking the empty string? Would make sense to check if the params contain the actual "Toon & Zoon" information in the `legal_entity`.

stevenvegt (changes requested):

stevenvegt:
Tests are failing because the e2e tests now include external crypto storage, but the V5 branch doesn't.

stevenvegt (approved):

reinkrul:
I tested it with the IRMA app, using the current v5.0.9 version the IRMA app shows the contract text with `&amp;`, with this patch it shows it as `&`. Looks OK.

gerardsn:
manual run of e2e tests is successful

PR #1937, opened 2023-03-01
Auth: do not escape parameters in rendered IRMA contracts

reinkrul:
Upstream patch of https://github.com/nuts-foundation/nuts-node/pull/1936


stevenvegt (approved):
LGTM

PR #1938, opened 2023-03-01
ProtocolV2: Payload recipients are matched on peerID + DID

gerardsn:
#1874 changed connection management from peerID to nodeDID+peerID. This change made it possible to have duplicate peerIDs in the connections list which hypothetically meant that a payload could end up at a different node than the one requesting it.


gerardsn:
#1940 extends this to respond directly on the connection a message came from

network/network_integration_test.go
```diff
@@ -741,6 +741,70 @@ func TestNetworkIntegration_PrivateTransaction(t *testing.T) {
...
+			mutex.Lock()
+			defer mutex.Unlock()
+			for _, receivedDoc := range receivedTransactions["node2"] {
+				if tx.Ref().Equals(receivedDoc.Ref()) {
+					return true, nil
+				}
+			}
+			return false, nil
+		}, time.Second)
+		arrivedAtEve := test.WaitForNoFail(t, func() (bool, error) {
```
reinkrul, line 793: did you test the test (did it fail before you fixed the vulnerability)?

network/network_integration_test.go
```diff
@@ -741,6 +741,70 @@ func TestNetworkIntegration_PrivateTransaction(t *testing.T) {
...
+			for _, receivedDoc := range receivedTransactions["node3"] {
+				if tx.Ref().Equals(receivedDoc.Ref()) {
+					return true, nil
+				}
+			}
+			return false, nil
+		}, time.Second)
+
+		// make sure eve does not have the payload
+		assert.False(t, arrivedAtEve)  // is true
```
reinkrul, line 805: weird comments? (is true/ is false)

network/transport/v2/handlers.go
```diff
@@ -133,24 +134,30 @@ func (p *protocol) handleTransactionPayloadQuery(ctx context.Context, peer trans
 		WithField(core.LogFieldTransactionRef, hash.FromSlice(msg.TransactionRef)).
 		Trace("Handling TransactionPayloadQuery")
 
+	// get peer's connection, authentication status is checked later on
+	connection := p.connectionList.Get(grpc.ByConnected(), grpc.ByNodeDID(peer.NodeDID), grpc.ByPeerID(peer.ID))
```
reinkrul, line 138: does this also work for unauthenticated connections? (bootstrapping?)

network/transport/v2/handlers.go
```diff
@@ -133,24 +134,30 @@ func (p *protocol) handleTransactionPayloadQuery(ctx context.Context, peer trans
...
 			// Transaction not found
-			return p.send(peer, emptyResponse)
+			return connection.Send(p, &Envelope{Message: emptyResponse}, false)
 		}
 		return err
 	}
 	if len(tx.PAL()) > 0 {
 		// Private TX, verify connection
-		if !peer.Authenticated {
+		if !peer.Authenticated || !connection.Peer().Authenticated { // peer == connection.Peer()
```
reinkrul, line 154: if they're the same, why is the extrea check required?

reinkrul (changes requested):

network/network_integration_test.go
gerardsn: first commit was test only, and yes it did fail

network/network_integration_test.go
gerardsn: removed. leftover from comment from when it was still failing

network/transport/v2/handlers.go
gerardsn: yes, but public payloads are attached to their transactions so this is for private TXs only.
#1940 changes this to respond on the connection the query came from.

network/transport/v2/handlers.go
gerardsn: being extra safe. removed in #1940

reinkrul: "being extra safe" sounds like unsure of program flow (or skills). But if it's removed in the next PR, ok.

reinkrul (approved):

PR #1945, opened 2023-03-03
Prevent go routine leak from connections

gerardsn:
wip
part of #1900 

Tests fail on a deadlock: In gRPC the client is expected to close the stream. So, gRPC's `ServerStreams` provides no way to close the stream/connection. This makes it very hard to close an inbound connection. 

One solution proposed in a similar GitHub issue is to wrap `net.Listener` to keep track of all inbound connections. This allows selectively closing an inbound stream by closing the corresponding connection (closes all streams on the connection), or just close all connections when shutting down the node. See commit 2 for a PoC of this. This works in `network_integration_test.go`, but needs some work in the `connection_manager_test.go` to close the mocked streams that have no corresponding connections.


gerardsn:
network will be removed in the future

PR #1954, opened 2023-03-06
VDR: Disable ephemeral keys in DID creation

### LINKED ISSUE #1947: Keep initial private key of controlled DIDs
When a new DID document is created and has no self-control, the initial key is not stored (ephemeral key). This is a good practice, however, since we might want to change the network tx format, keeping this key allows us to migrate existing DIDs.

Should be a simple code change:

https://github.com/nuts-foundation/nuts-node/blob/31dd66a42b806e61291cec4a7deb6bfd23d4452d/vdr/didservice/creator.go#L136-L140

Lets discuss.

stevenvegt:
Fixes #1947

### DIFF
--- a/vdr/didservice/creator_test.go
+++ b/vdr/didservice/creator_test.go
@@ -29,10 +29,11 @@ import (
 	"github.com/golang/mock/gomock"
 	"github.com/lestrrat-go/jwx/jwk"
 	"github.com/nuts-foundation/go-did/did"
-	nutsCrypto "github.com/nuts-foundation/nuts-node/crypto"
-	"github.com/nuts-foundation/nuts-node/vdr/types"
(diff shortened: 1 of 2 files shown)


vdr/didservice/creator_test.go
```diff
@@ -123,6 +124,7 @@ func TestCreator_Create(t *testing.T) {
 		})
 
 		t.Run("using ephemeral key creates different keys for assertion and DID", func(t *testing.T) {
+			t.Skip("Disabled while ephemeral keys are not used")
```
gerardsn, line 127: refer to an issue for future reference?

gerardsn (approved):

reinkrul (approved):

PR #1958, opened 2023-03-08
DIDman: Strict API interfaces

reinkrul:
Still causes compilation issues, which I fixed in the codegen in https://github.com/deepmap/oapi-codegen/pull/999, but which needs to be released. For the time being, maybe just fix it manually (it's changing 1 character), we don't alter the API too often.

### DIFF
--- a/didman/api/v1/api.go
+++ b/didman/api/v1/api.go
@@ -20,14 +20,14 @@
 package v1
 
 import (
+	"context"
 	"errors"
-	"fmt"
-	httpModule "github.com/nuts-foundation/nuts-node/http"
+	"github.com/labstack/echo/v4"
+	"github.com/nuts-foundation/nuts-node/audit"
 	"net/http"
 	"net/url"
 	"strings"
 
-	"github.com/labstack/echo/v4"
 	ssi "github.com/nuts-foundation/go-did"
 	"github.com/nuts-foundation/go-did/did"
 	"github.com/nuts-foundation/nuts-node/core"
@@ -37,7 +37,7 @@ import (
 	"github.com/nuts-foundation/nuts-node/vdr/types"
 )
 
-var _ ServerInterface = (*Wrapper)(nil)
+var _ StrictServerInterface = (*Wrapper)(nil)
 var _ core.ErrorStatusCodeResolver = (*Wrapper)(nil)
 
 // Wrapper implements the generated interface from oapi-codegen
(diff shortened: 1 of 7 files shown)


gerardsn:
title should be `Didman` instead of `VDR`?

reinkrul:
Things left to do:
- [ ] Find out how to get request Content-Type header (failing test)
- [ ] Test `Routes`

didman/api/v1/api.go
```diff
@@ -77,81 +77,88 @@ func (w *Wrapper) ResolveStatusCode(err error) int {
...
+	id, err := did.ParseDID(request.Did)
 	if err != nil {
-		return err
+		return nil, err
 	}
 
-	if len(strings.TrimSpace(request.Type)) == 0 {
-		return core.InvalidInputError("invalid value for type")
+	props := request.Body
+	if len(strings.TrimSpace(props.Type)) == 0 {
```
gerardsn, line 106: I prefer the explicitness of `request.Body.Type`

didman/api/v1/api.go
```diff
@@ -162,72 +169,55 @@ func (w *Wrapper) GetCompoundServices(ctx echo.Context, didStr string) error {
...
```
gerardsn, line 179: same here

didman/api/v1/api_test.go
```diff
@@ -34,462 +32,395 @@ import (
...
 
 	t.Run("ok", func(t *testing.T) {
-		ctx := newMockContext(t)
-		ctx.didman.EXPECT().DeleteEndpointsByType(audit.ContextWithAuditInfo(), *parsedID, endpointType)
-		ctx.echo.EXPECT().NoContent(http.StatusNoContent)
+		test := newMockContext(t)
+		test.didman.EXPECT().DeleteEndpointsByType(audit.ContextWithAuditInfo(), *parsedID, endpointType)
 
-		err := ctx.wrapper.DeleteEndpointsByType(ctx.echo, idStr, endpointType)
+		_, err := test.wrapper.DeleteEndpointsByType(ctx, request)
```
gerardsn, line 188: check that response is not nil

didman/api/v1/api.go
```diff
@@ -162,72 +169,55 @@ func (w *Wrapper) GetCompoundServices(ctx echo.Context, didStr string) error {
...
+		return nil, err
 	}
 
-	// default json
-	return ctx.JSON(http.StatusOK, EndpointResponse{Endpoint: endpoint})
+	if request.Params.Accept == nil || *request.Params.Accept != "text/plain" {
+		return GetCompoundServiceEndpoint200JSONResponse{Endpoint: endpoint}, nil
+	} else {
+		return GetCompoundServiceEndpoint200TextResponse(endpoint), nil
+	}
```
gerardsn, line 220: explicitly test for `text/plain` or default to json outside the if statement reduces cognitive complexity

didman/api/v1/api_test.go
```diff
@@ -34,462 +32,395 @@ import (
...
-		assert.EqualError(t, err, "failed to parse v1.CompoundServiceProperties: b00m!")
+		assert.Equal(t, http.StatusBadRequest, test.wrapper.ResolveStatusCode(err))
+		assert.Nil(t, response)
 	})
 }
 
 func TestWrapper_GetCompoundServices(t *testing.T) {
-	idStr := "did:nuts:1#1"
-	id, _ := did.ParseDIDURL(idStr)
+	targetDID := did.MustParseDID("did:nuts:1")
```
gerardsn, line 327: test has changed by removing the fragment

docs/_static/didman/v1.yaml
gerardsn, line 1: nice find! I was looking for a solution for this

gerardsn (changes requested):
just some minor things

didman/api/v1/api_test.go
gerardsn: the GetCompoundServices endpoint parses it as `id, err := did.ParseDIDURL(request.Did)` though. Is that wrong too then?

reinkrul: fixed

gerardsn (approved):

Issue #1929, opened 2023-02-27
Network: add more info to outbound connector stats


reinkrul:
- Next connection attempt
- Last occurred error

woutslakhorst:
Outbound connectors should get their own diagnostics endpoint? It's getting way too long. Might already be an issue for that somewhere....

reinkrul:
I had a dejavu and thought this was already implemented (separate endpoint), but couldn't find it. It must be _somewhere_

PR #1966, opened 2023-03-08
Network: move outbound_connectors to separate diagnostics page

reinkrul:


network/api/v1/api.go
```diff
@@ -53,6 +55,25 @@ func (a *Wrapper) Routes(router core.EchoRouter) {
 			return audit.StrictMiddleware(f, network.ModuleName, operationID)
 		},
 	}))
+
+	const diagnosticsBasePath = "/status/diagnostics/network"
```
gerardsn, line 59: should be added to `docs/_static/monitoring/v1.yaml` so it can be found in the API documentation

docs/pages/release_notes.rst
```diff
@@ -16,6 +16,8 @@ Release date: **DRAFT**
 - Added audit logging for cryptographic operations (creating a new key pair, signing, decrypting).
   Refer to the documentation for more information.
 - Added new API authentication method, in which the administrator configures authorized public keys and the API client is responsible for signing JWT using the private key. This new API authentication is preferred over the current method, which will be removed in the next major release.
+- ``network.connections.outbound_connectors`` diagnostic has been moved to ``/status/diagnostics/network/addressbook``.
```
gerardsn, line 19:
```suggestion
- ``network.connections.outbound_connectors`` on ``/status/diagnostics`` has been moved to ``/status/diagnostics/network/addressbook``.
```

network/api/v1/api.go
```diff
@@ -53,6 +55,25 @@ func (a *Wrapper) Routes(router core.EchoRouter) {
...
+			results := provider()
+			var output interface{}
+			if len(results) == 1 {
+				// If only 1 entry, flatten
+				output = results[0]
+			} else {
+				output = results
+			}
+			buf := new(bytes.Buffer)
+			_ = yaml.NewEncoder(buf).Encode(output)
```
gerardsn, line 72: `application/yaml` results in file download. I prefer the default to be json, same as diagnostics

core/engine.go
```diff
@@ -219,6 +219,14 @@ type Diagnosable interface {
 	Diagnostics() []DiagnosticResult
 }
 
+// MappedDiagnosable allows the implementer to return a map of diagnostics providers,
+// which can be used by the caller to retrieve specific diagnostics, rather than everything (like Diagnosable provides).
+// The map keys should be URL paths, to aid mapping to HTTP APIs.
+type MappedDiagnosable interface {
```
gerardsn, line 225: This adds a lot of complexity, do we expect any other use cases?

gerardsn (changes requested):
Yay! diagnostics fits on 1 screen again

 ---- 
`grpc/stats.go#ContactsStats.Name()` should return something like `contacts` or `address_book`

network/api/v1/api.go
gerardsn: nvm. content-type is `text/plain`

network/api/v1/api.go
reinkrul: Not sure, that means we're going to spec it, thus having to maintain backwards compatibility?

core/engine.go
reinkrul: I considered just registering a router and calling a separate diagnostics func for addressbook, but that leads to this special addressbook diagnostic func being propagated through all layers (Network -> ConnectionManager -> Addressbook), although Network doesn't care about gRPC impl's addressbook. So an abstracted concept is cleaner.

network/api/v1/api.go
gerardsn: `/status/diagnostics` is on there too and we do not maintain backwards compatibility for that.

I added the monitoring API for discoverability of endpoints without the intention of it being fully specced. `/status`, `/health` and probably `/metrics` should probably not change, so maybe we need make a distinction between machine readable monitoring and programmer debugging convenience where the latter can be changed without notice?

core/engine.go
gerardsn: fair.

network/transport/grpc/connection_manager.go
```diff
@@ -64,6 +64,8 @@ var MaxMessageSizeInBytes = defaultMaxMessageSizeInBytes
 // defaultInterceptors aids testing
 var defaultInterceptors []grpc.StreamServerInterceptor
 
+var _ transport.ConnectionManager = (*grpcConnectionManager)(nil)
```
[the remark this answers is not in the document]
reinkrul: True, removed.

network/api/v1/api.go
reinkrul: Will add it

core/engine.go
gerardsn: This still doesn't feel right. It moves routing decisions (url paths) to the diagnostics collectors. To me that sounds just as bad as the other solution. Maybe @woutslakhorst has any ideas?

gerardsn (commented):
Just the one question remaining

core/engine.go
reinkrul: A design-wise cleaner solution would involve a Visitor-esque pattern (like API's `Routes()`) so components can register it themselves, e.g.:

```go
type DiagnosticProvider interface {
    Register(name string, provider func() []DiagnosticResult)
}
```

Which are then called by the `Status` engine and exposed as HTTP endpoints.

But currently we just have 1 separate diagnostics endpoint and it's unclear whether when, or even if, there will be more. So the added abstractions and indirections only add complexity which don't add any value i.m.o.

So yes, I agree from a design point of view, but it doesn't add any actual value right now.

woutslakhorst (commented):
`status` and `health` are automated endpoints to check process status.
`diagnostics` are meant to give some debug/basic info on a node. Some of the numbers are also exposed as metrics. We might consider the addressbook status an API that is beyond simple diagnostics and therefore a *full* member of the network API? Then it has it's place in the network API spec and `api.go`

reinkrul:
> Relation to #1929?

#1929 should only be implemented after this PR is merged (to keep the diagnostics page as short as possible).

reinkrul:
> `status` and `health` are automated endpoints to check process status. `diagnostics` are meant to give some debug/basic info on a node. Some of the numbers are also exposed as metrics. We might consider the addressbook status an API that is beyond simple diagnostics and therefore a _full_ member of the network API? Then it has it's place in the network API spec and `api.go`

I agree

reinkrul:
@gerardsn @woutslakhorst  I reworked it to an actual OpenAPI endpoint of Network.

network/transport/grpc/addressbook_test.go
```diff
@@ -144,28 +144,9 @@ func TestAddressBook_remove(t *testing.T) {
 	assert.Len(t, ab.contacts, 1)
 }
 
-func TestAddressBook_Diagnostics(t *testing.T) {
```
gerardsn, line 147: `all()` is no longer tested properly

gerardsn (changes requested):

network/api/v1/api.go
```diff
@@ -125,6 +125,23 @@ func (a *Wrapper) GetPeerDiagnostics(_ context.Context, _ GetPeerDiagnosticsRequ
...
+	for _, contact := range a.Service.AddressBook() {
+		result := Contact{
+			Address:     contact.Address,
+			Attempts:    int(contact.Attempts),
+			LastAttempt: contact.LastAttempt,
+		}
+		if !contact.DID.Empty() {
+			result.Did = new(string)
+			*result.Did = contact.DID.String()
+		}
```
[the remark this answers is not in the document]
reinkrul: Is it? I was thinking of bootstrap nodes?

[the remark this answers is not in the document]
gerardsn: You are correct. I read this wrong

docs/_static/network/v1.yaml
```diff
@@ -211,6 +211,21 @@ paths:
                   $ref: '#/components/schemas/EventSubscriber'
         default:
           $ref: '../common/error_response.yaml'
+  /internal/network/v1/addressbook:
```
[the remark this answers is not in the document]
reinkrul: I don't think so, since we plan to release this in v5.2 and we could just fix the other change in v5.2 as well. And even then, the things to change are an addition, so they're backwards compatible.

network/transport/grpc/addressbook.go
```diff
@@ -38,26 +37,24 @@ type AddressBook interface {
 	// get the contact if it exists. Returns nil and false if it does not.
 	get(peer transport.Peer) (*contact, bool)
 	// all returns a copy of the slice of contacts.
-	all() []*contact
+	all() []transport.Contact
```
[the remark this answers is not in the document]
reinkrul: I find it hard to find a good name, which isn't very artificial. E.g. `export()`.

I can rename it to `list()` but it's hardly any better. Any suggestions?

network/transport/grpc/addressbook_test.go
reinkrul: fixed

docs/_static/network/v1.yaml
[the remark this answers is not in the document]
gerardsn: As long as the other issue is fixed as well its fine I guess.

docs/_static/network/v1.yaml
```diff
@@ -277,6 +292,26 @@ components:
...
+        - attempts
+      properties:
+        address:
+          description: Address of the node.
+          type: string
+        did:
+          description: DID of the node.
+          type: string
+        attempts:
+          description: Number of times the node has been contacted, but the connection didn't succeed.
```
gerardsn, line 309: Currently this is the number of connection attempts since the node has (re-)started.

network/transport/grpc/addressbook.go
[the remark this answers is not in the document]
gerardsn: I had the same issue. Perhaps `stats()` or `contacts()`? `all` just sounds its going to do something similar to `update`, `get` and `remove`. Perhaps the clarifying this in the docstring is enough.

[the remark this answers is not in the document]
reinkrul: I think `stats()` is the least offensive option

network/transport/grpc/addressbook.go
```diff
@@ -38,26 +37,24 @@ type AddressBook interface {
 	// get the contact if it exists. Returns nil and false if it does not.
 	get(peer transport.Peer) (*contact, bool)
 	// all returns a copy of the slice of contacts.
```
gerardsn, line 39:
```suggestion
	// stats returns all known contacts with their connection statistics
```

gerardsn (approved):
not sure why e2e are not starting

PR #1979, opened 2023-03-10
Rename crypto external storage url to address

### LINKED ISSUE #1977: Incorrect setting/documentation for external crypto address
Docs speak about `crypto.external.address` but config in code is called `crypto.config.url`, address is more in line with other config params.

stevenvegt:
Fixes #1977


stevenvegt:
e2e tests depending on https://github.com/nuts-foundation/nuts-go-e2e-test/pull/60

reinkrul:
It's actually more an URL than an address (which is generally just an IP, maybe port i.m.o.). But I think this is the case for more properties, so at least it's more consistent.

When did we introduce this, v5.0 or v5.1? If v5.0, it's a breaking change (thus shouldn't be in v5.1).

I couldn't find it in the release notes (something to add).

reinkrul (commented):
It's actually more an URL than an address (which is generally just an IP, maybe port i.m.o.). But I think this is the case for more properties, so at least it's more consistent.

When did we introduce this, v5.0 or v5.1? If v5.0, it's a breaking change (thus shouldn't be in v5.1).

I couldn't find it in the release notes (something to add).

reinkrul:
PR for release notes: https://github.com/nuts-foundation/nuts-node/pull/1980

reinkrul (approved):

stevenvegt:
e2e test pass but are failing in the cleanup phase.

PR #1982, opened 2023-03-13
Chore: goimports and fix copyrights

reinkrul:
According to @woutslakhorst we use `goimports` style import formatting, so I ran it.


reinkrul:
Generated files are formatted as well, it seems `goimports` can't ignore them.

PR #1994, opened 2023-03-17
fix finding issued credentials

### LINKED ISSUE #1910: Finding all issued credentials always returns empty result
The search is defined in the issuer store as:

```
func (s leiaIssuerStore) SearchCredential(jsonLDContext ssi.URI, credentialType ssi.URI, issuer did.DID, subject *ssi.URI) ([]vc.VerifiableCredential, error) {
	query := leia.New(leia.Eq(leia.NewJSONPath("issuer"), leia.MustParseScalar(issuer.String()))).
		And(leia.Eq(leia.NewJSONPath("type"), leia.MustParseScalar(credentialType.String()))).
		And(leia.Eq(leia.NewJSONPath("@context"), leia.MustParseScalar(jsonLDContext.String())))
```

The passed context is empty, so context must match an empty string which will fail. 

Search is only used from the API as a debug tool.

woutslakhorst:
fixes #1910 

The problem was that the query part for the JSON-LD context did an equal to `""`.

Changed it to only include that query part if the context is actually non-empty.
Requiring the `@context` to be non-nil for VCs is overkill since every VC will have a JSON-LD context.

### DIFF
--- a/vcr/api/v2/api.go
+++ b/vcr/api/v2/api.go
@@ -173,7 +173,7 @@ func (w *Wrapper) SearchIssuedVCs(ctx echo.Context, params SearchIssuedVCsParams
 		return core.InvalidInputError("invalid credentialType: %w", err)
 	}
 
-	foundVCs, err := w.VCR.Issuer().SearchCredential(ssi.URI{}, *credentialType, *issuerDID, subjectID)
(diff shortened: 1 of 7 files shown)


woutslakhorst:
> Since it is always passed an empty JSON-LD context URI, just remove the search parameter altogether? (now it is dead code)

Was only 1 usage indeed, removed param

reinkrul (approved):

PR #1995, opened 2023-03-17
fix correct loading order of data in issuer store

### LINKED ISSUE #1909: Missing index for shelf, rebuilding at startup
This is logged on info level.

After checking the code, the index for the issuer store is added after the handleRestore. Because this check does an indexIterate, it'll not find anything and thinks it needs to rebuild it. It'll rebuild it from the backup shelf. This is not a real problem just additional processing at startup. So no hot-startup.

Switching the index creation with the handleRestore should fix it.

Don't know what caused this, might have never worked....

woutslakhorst:
fixes #1909 

Indices had to be created first.
Added test shows that existing data is now correctly detected and no loading of the backup is done.

### DIFF
--- a/vcr/issuer/leia_store_test.go
+++ b/vcr/issuer/leia_store_test.go
@@ -388,7 +388,7 @@ func Test_leiaIssuerStore_handleRestore(t *testing.T) {
 func TestNewLeiaIssuerStore(t *testing.T) {
 	t.Run("bug test for https://github.com/nuts-foundation/nuts-node/issues/1909", func(t *testing.T) {
 		ctrl := gomock.NewController(t)
-		mockStore := stoabs.NewMockKVStore(ctrl)
+		backupMockStore := stoabs.NewMockKVStore(ctrl)
 		backupStorePath := path.Join(t.TempDir(), "backup-issued-credentials.db")
 		emptyBackupStore, err := bbolt.CreateBBoltStore(backupStorePath)
 		dbPath := path.Join(t.TempDir(), "issuer.db")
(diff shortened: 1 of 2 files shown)


vcr/issuer/leia_store_test.go
```diff
@@ -376,6 +376,43 @@ func TestLeiaIssuerStore_handleRestore(t *testing.T) {
 	})
 }
 
+func TestNewLeiaIssuerStore(t *testing.T) {
+	t.Run("bug test for https://github.com/nuts-foundation/nuts-node/issues/1909", func(t *testing.T) {
+		ctrl := gomock.NewController(t)
+		mockStore := stoabs.NewMockKVStore(ctrl)
```
reinkrul, line 382: rename to `backupMockStore`

vcr/issuer/leia_store_test.go
```diff
@@ -376,6 +376,43 @@ func TestLeiaIssuerStore_handleRestore(t *testing.T) {
...
+		require.NoError(t, err)
+		vc := vc.VerifiableCredential{}
+		_ = json.Unmarshal([]byte(jsonld.TestCredential), &vc)
+		require.NoError(t, store.StoreCredential(vc))
+		require.NoError(t, store.Close())
+
+		// now create a new store with a mock backup and show that ReadShelf only called once.
+		// no additional calls
+		reader := stoabs.NewMockReader(ctrl)
+		reader.EXPECT().Iterate(gomock.Any(), gomock.Any()).DoAndReturn(func(argCB interface{}, argT interface{}) error {
```
reinkrul, line 398: Had to dive into the code to see why this works, it's because `backupStorePresent` calls Iterate and ReadShelf, if restore happens it'll be called again by the restore. But it's quite complicated, so maybe add a comment as to why the mock behavior it is as it is.

reinkrul (changes requested):
Not in scope, but I noticed this function is quite expensive, since it doesn't exit the `Iterate` function when it finds a result, thus it always loops over all items?

```go
func (s leiaIssuerStore) backupStorePresent(backupShelf string) bool {
	backupPresent := false

	_ = s.backupStore.ReadShelf(context.Background(), backupShelf, func(reader stoabs.Reader) error {
		return reader.Iterate(func(key stoabs.Key, value []byte) error {
			backupPresent = true
			return nil
		}, stoabs.BytesKey{})
	})

	return backupPresent
}
```

woutslakhorst:
> ```go
> backupStorePresent
> ```

We could add a `First()` to stoabs?

reinkrul:
> > ```go
> > backupStorePresent
> > ```
> 
> We could add a `First()` and/or `isEmpty()` to stoabs?

That, or the "hacky" version would involve returning a specific error and checking whether the returned error is the one. Or a boolean which you set from inside the callback.

woutslakhorst:
> > > ```go
> > > backupStorePresent
> > > ```
> > 
> > 
> > We could add a `First()` and/or `isEmpty()` to stoabs?
> 
> That, or the "hacky" version would involve returning a specific error and checking whether the returned error is the one. Or a boolean which you set from inside the callback.

The hacky version would add error log entries (I checked). I'll add some things to stoabs

woutslakhorst:
@reinkrul it now uses the new Empty function

PR #1996, opened 2023-03-20
Crypto: storage funcs now pass context

### LINKED ISSUE #1853: Crypto: storage API funcs should pass context
To allow cancellation, and other features, e.g. request tracing (e.g. through HTTP headers).

reinkrul:
Fixes #1853


network/network.go
```diff
@@ -110,7 +110,7 @@ func (n *Network) CheckHealth() map[string]core.Health {
 		}
 	}
 	// auth_config checks that the node is correctly configured to be authenticated by others
-	nodeDID, err := n.nodeDIDResolver.Resolve()
+	nodeDID, err := n.nodeDIDResolver.Resolve(context.TODO())
```
woutslakhorst, line 113: This would be part of passing ctx through the health interface? Please link an issue if so.

http/engine.go
```diff
@@ -330,7 +330,7 @@ func (h Engine) applyBindMiddleware(echoServer EchoServer, path string, excludeP
 		log.Logger().Infof("Enabling token authentication for HTTP interface: %s%s", address, path)
 		echoServer.Use(middleware.JWTWithConfig(middleware.JWTConfig{
 			KeyFunc: func(_ *jwt.Token) (interface{}, error) {
```
woutslakhorst, line 332: Should be changed for `ParseTokenFunc` which is passed a ctx. Link an issue if not for this PR.

woutslakhorst (changes requested):

http/engine.go
reinkrul: Not really possible since ParseTokenFunc can't return the signing key and is not supposed to validate it. "Fixed" it by moving the signing key fetching to node initialization phase. Then it also happens only once, instead of every time a token is validated.

woutslakhorst (approved):

PR #1997, opened 2023-03-20
update diagnostics with owned did conflicts

### LINKED ISSUE #241: Extend VDR diagnostics with own conflicting DID Documents
Now it only shows the global # of conflicted docs. This should be extended with the DIDs of your own conflicted DID documents.

woutslakhorst:
closes #241 

VDR diagnostics now reads:

```
    vdr:
        did_documents_count: 5
        conflicted_did_documents:
            total_count: 2
            owned_count: 0
```

Where `owned_count` indicates your docs that have something wrong.

### DIFF
--- a/docs/pages/deployment/monitoring.rst
+++ b/docs/pages/deployment/monitoring.rst
@@ -116,6 +116,14 @@ Explanation of ambiguous/complex entries in the diagnostics:
 * ``vdr.conflicted_did_documents.total_count`` holds the total number of DID documents that are conflicted (have parallel updates). This may indicate a stolen private key
 * ``vdr.conflicted_did_documents.owned_count`` holds the number of conflicted DID documents you control as a node owner
 
+Conflicted DID documents
+************************
+
+Conflicted DID documents are the result of parallel updates of a DID document.
+It can be caused by a race condition during DID API calls, having a cluster of nodes when this is not supported or if a private key has been stolen.
+To rule out this last cause, every node must make sure their conflicted document count equals ``0``.
+There's an API to get the actual conflicted DID documents. A DID document can be *fixed* by simply updating it.
+
 Metrics
 *******
 
--- a/vdr/vdr.go
+++ b/vdr/vdr.go
@@ -132,7 +132,8 @@ func (r *VDR) Diagnostics() []core.DiagnosticResult {
 		totalCount++
 		controllers, err := r.didDocResolver.ResolveControllers(doc, &types.ResolveMetadata{Hash: &metadata.Hash})
 		if err != nil {
-			return err
+			log.Logger().Errorf("failed to resolve controller for %s: %v", doc.ID, err)
+			return nil
 		}
 		for _, controller := range controllers {
 			for _, vr := range controller.CapabilityInvocation {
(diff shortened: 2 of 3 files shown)


woutslakhorst:
> Why don't we add the list of (your) DIDs that conflict, just for ease of use?

There's an API for that. Same conversation as the connector diagnostics: Check diagnostics, is a number not as it should be? Go use other APIs to debug

docs/pages/deployment/monitoring.rst
```diff
@@ -108,6 +113,8 @@ Explanation of ambiguous/complex entries in the diagnostics:
 * ``vcr.issuer.issued_credentials_count`` holds the total number of credentials issued by the local node
 * ``vcr.issuer.revoked_credentials_count`` holds the total number of revoked credentials issued by the local node
 * ``vcr.verifier.revocations_count`` holds the total number of revoked credentials (public and private VCs)
+* ``vdr.conflicted_did_documents.total_count`` holds the total number of DID documents that have parallel updates. This may indicate a stolen private key
```
reinkrul, line 116:
```suggestion
* ``vdr.conflicted_did_documents.total_count`` holds the total number of DID documents that are conflicted (have parallel updates). This may indicate a stolen private key
```

vdr/vdr.go
```diff
@@ -124,13 +124,56 @@ func (r *VDR) ConflictedDocuments() ([]did.Document, []types.DocumentMetadata, e
...
-	count, _ := r.store.ConflictedCount()
+	totalCount := 0
+	ownedCount := 0
+
+	// uses dedicated storage shelf for conflicted docs, does not loop over all documents
+	err := r.store.Conflicted(func(doc did.Document, metadata types.DocumentMetadata) error {
+		totalCount++
+		controllers, err := r.didDocResolver.ResolveControllers(doc, &types.ResolveMetadata{Hash: &metadata.Hash})
+		if err != nil {
+			return err
```
reinkrul, line 135: I think this causes the iterator to stop at this point, causing subsequent conflicted documents not to be addressed?

vdr/vdr.go
```diff
@@ -124,13 +124,56 @@ func (r *VDR) ConflictedDocuments() ([]did.Document, []types.DocumentMetadata, e
...
+		log.Logger().Errorf("Failed to resolve conflicted documents diagnostics: %v", err)
+	}
+
 	docCount, _ := r.store.DocumentCount()
 
+	// release notes
+	// documentation on how to fix
+
+	// to go from int+error to interface{}
+	countOrError := func(count int, err error) interface{} {
```
reinkrul, line 157: interesting approach, applicable to other spots where we just log the error, as well?

reinkrul (changes requested):

vdr/vdr.go
woutslakhorst: Only if the return type is `interface{}`

vdr/vdr.go
woutslakhorst: Changed to log and return nil (continue)

reinkrul (approved):

PR #1999, opened 2023-03-22
Auth: fix failing test

gerardsn:


woutslakhorst (approved):

woutslakhorst (changes requested):
Test fails on circle CI.
I suggest changing the test so it checks the error starts with "could not parse validFrom", that should be sufficient for the API caller.

woutslakhorst (approved):

PR #2002, opened 2023-03-23
OIDC4VCI Issuer initiated credential offer prototype

stevenvegt:
Phase 0:
Prototype of issuer initiated credential offer.

### DIFF
--- a/auth/oidc4vci/flow.puml
+++ b/auth/oidc4vci/flow.puml
@@ -1,0 +1,48 @@
+@startuml
+'https://plantuml.com/sequence-diagram
+
+autonumber
+
+box Issuer
+    participant Issuer
+    participant "Authorization Server" as IssuerAS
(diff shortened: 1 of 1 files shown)


reinkrul:
Next steps:
- Compatibility:
  - Discover client endpoints using metadata, decide how to do client discovery (maybe register client metadata URL in DID document?)
  - Make Credential Issuer Metadata URLs configurable (now hardcoded http://localhost:1323)
- Security:
  - Create/Verify CredentialRequest proof (decide if we need it?)
- Redesign integration into Nuts node, taking into account:
  - Retry of: credential offer delivery (issuer), retrieving credential (wallet)
  - How to dispatch retrieving credentials (now a runaway goroutine in `wallet.go`)
  - Design lifecycle for offered VCs (only issue VC when they're actually requested by the wallet)
    - This probably needs storage as well
- Write RFC

PR #2004, opened 2023-03-24
refactor tls.Config creation on grpc connection manager

### LINKED ISSUE #1959: Merge server and client tls.Config in the grpc connection manager
Currently the connection manager setup creates a separate config to use in inbound and outbound connections. They both use the same config, but may need some customization incase TLS is offloaded.

Maintaining 1 config should be less susceptible to errors

Both configs always use the same CA's and client cert == server cert

still needs to work with TLS offloading

woutslakhorst:
closes #1959

It might be confusing that two tls.Config structs are being created, but this is actual a requirement because dialing out and listening to connections are two different things and therefore need their own signature.

I refactored some things so it might make more sense.

@gerardsn ?

### DIFF
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -140,10 +140,13 @@ func (auth *Auth) Configure(config core.ServerConfig) error {
 			return err
 		}
 
-		tlsConfig = createTLSConfig(clientCertificate, trustStore)
-
 		validator := crl.NewValidator(trustStore.Certificates())
-		validator.Configure(tlsConfig, config.TLS.GetCRLMaxValidityDays())
+		tlsConfig = &tls.Config{
+			Certificates:          []tls.Certificate{clientCertificate},
+			RootCAs:               trustStore.CertPool,
+			MinVersion:            core.MinTLSVersion,
+			VerifyPeerCertificate: validator.VerifyPeerCertificateFunction(config.TLS.GetCRLMaxValidityDays()),
+		}
 
 		auth.crlValidator = validator
 	}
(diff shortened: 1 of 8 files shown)


gerardsn:
The issue with having two is that it is easier to make mistakes when updating 1, but forgetting to update the other. (Like not configuring the crl validation for one of them.) But putting the config creator funcs next to each other minimizes chances of that happening I guess.

Just to double check. The reason for not merging them is because they have different responsibilities? (Question, not a suggestion.) Because the change below seems to work just fine:
```
func newServerTLSConfig(config Config) *tls.Config {
	return newTLSConfig(config)
}

func newClientTLSConfig(config Config) *tls.Config {
	return newTLSConfig(config)
}

func newTLSConfig(config Config) *tls.Config {
	return &tls.Config{
		RootCAs:               config.trustStore,
		ClientCAs:             config.trustStore,
		ClientAuth:            tls.RequireAndVerifyClientCert,
		Certificates:          []tls.Certificate{*config.clientCert}, // serverCert == clientCert when TLS is not offloaded. This Config is not used when configured for TLS offloading
		MinVersion:            core.MinTLSVersion,
		VerifyPeerCertificate: config.crlValidator.VerifyPeerCertificateFunction(config.maxCRLValidityDays),
	}
}

woutslakhorst:
It works now because we use the same certificates for server and client. If this would change it'll no longer work due to only having a single `certificates` attribute

gerardsn (approved):
Alright, the PR minimizes the chances of forgetting to update one of the two configs, so I think that is sufficient to complete the goal of the issue.

PR #2008, opened 2023-03-28
x509: show offending URL in CRL validator sync errors

reinkrul:
It currently only shows `CRL database is outdated, certificate revocation can't be checked`, this enriches this with the reason and URL of the CRL that caused the error.


woutslakhorst (approved):

reinkrul:
> Does it need a backport to v5.1?

Would be good, I think

PR #2010, opened 2023-03-29
Auth: log OAuth2 error responses on INFO client-side

reinkrul:
So it is generally logged without having to set the node to DEBUG.


reinkrul:
I looked into other OAuth2 implementations (o.a. Spring Security), they give more detail in their error messages. Maybe we should reconsider that, giving some helpful message when the client caused the error (in other words, validation error)?

woutslakhorst:
That would be revisiting the debate on security vs debug-ability.  Previous decision was in favour of security. There should be a strong case where returning the error in production is helpful.

reinkrul:
Probably better to do more sanity checks client-side (same validations that happen server-side)

PR #2011, opened 2023-03-29
CRL: replace validator

### LINKED ISSUE #1883: CRL: Validator.IsSynced should check status for specific CRLs, not all
It is used to check the validator is up-to-date (CRLs downloaded) when checking revocation status; when the CRL is too old, revocation status checking fails. But it checks all CRLs, not just for the relevant certificates (the certificate (being checked)'s tree).

In theory this could lead to the Nuts node rejecting a certificate, because it could not update the CRL of another, unrelated certificate. In practice, however, the number of CA certificates in the trust store is (or should) be carefully curated, only being PKIoverheid and its intermediate CAs (which is a small handful). Thus, the chance that it occurs is quite small (but blocking, nonetheless).

### LINKED ISSUE #1987: CRL: should try to update more frequently
`validator.updateCRL` only updates the CRL after its `Next Update` time has expired. (so when the CRL is invalid and `validator.IsSynced` will return false)

from [RFC5280](https://www.rfc-editor.org/rfc/rfc5280#section-5.1.2.5)
```
5.1.2.5. Next Update

   This field indicates the date by which the next CRL will be issued.
   The next CRL could be issued before the indicated date, but it will
   not be issued any later than the indicated date. CRL issuers SHOULD
   issue CRLs with a nextUpdate time equal to or later than all previous
   CRLs.  ...
```

The Next Update on pkiO certificates is very infrequently
```
DomPrivateServicesLatestCRL-G1.crl:
	This Update at 2022-11-03 14:22:42 +0000 UTC
	Next Update at 2023-11-02 14:22:41 +0000 UTC
PrivateRootLatestCRL-G1.crl:
	This Update at 2022-11-03 11:59:29 +0000 UTC
	Next Update at 2023-11-02 11:59:28 +0000 UTC
```

Worst case (for pkiO) the current flow fails to mark a certificate as revoked for a year.

gerardsn:
closes #1883 
closes #1987

In addition fixes:
The current implementation of the crl validator only checks for CRL distribution points that are listed on the certificates of CAs. However, certificates point to the CRL that will revoke them, not the CRL that they issue. So, leaf certificates (non CA) point to the most interesting CRL distribution points, but these are currently ignored.

Fixing #1987 also allows this PR to deprecate `tls.crl.maxvaliditydays`. Since CRLs are now updated every hour, and all CRLs in the pkiOverheid chain have a validity of at least 1 day, failing to update the CRL before it has reached its `NextUpdate` time allow a hard-fail immediately.

backport to v5.0 and v5.1


crl/validator.go
```diff
@@ -20,346 +20,420 @@ package crl
...
+	// check that all cert issuers are in the truststore, and add missing crl distribution points.
+	for _, chain := range config.Certificates {
+		certificates, err := parseCertificatesSlice(chain.Certificate)
+		if err != nil {
+			return err
+		}
+		for _, cert := range certificates {
+			issuer := v.truststore[cert.Issuer.String()]
+			if issuer == nil {
+				// This indicates a mismatch between crlValidator truststore and tls.Config. This is a programming error.
```
reinkrul, line 195: It could be the truststore is missing a certificate, which is a configuration error?

In that case, we might have to revisit the error message. Or, the truststore should check it. But if it starts failing because an unrelated cert (maybe without CRL) is missing, it might be a breaking change for some parties?

crl/validator.go
```diff
@@ -20,346 +20,420 @@ package crl
...
+				return errors.New("tls.Config contains certificate from issuer that is not in the truststore: " + cert.Subject.String())
+			}
+			// add any previously unknown CRL distribution points.
+			// (Leaf cert is likely to contain CRLDistributionPoints not found in the truststore.)
+			if len(cert.CRLDistributionPoints) > 0 {
+				for _, endpoint := range cert.CRLDistributionPoints {
+					if v.crls[endpoint] == nil {
+						v.crls[endpoint] = newRevocationList(issuer)
```
reinkrul, line 203: multiple issuers can use the same CRL distribution endpoint (e.g. QuoVadis PKIo Private intermediate CAs (e.g. Personen, Services) share the same CRL), so I'm missing an append flow here

crl/validator.go
```diff
@@ -20,346 +20,420 @@ package crl
...
-func crlHasExpired(crl *x509.RevocationList, deadline time.Time) bool {
-	return !deadline.Before(crl.NextUpdate)
+// validateChain validates that none of the certificates in the chain is revoked.
+// Certificated in chain are assumed ordered from leaf to root certificate.
+func (v *validator) validateChain(chain []*x509.Certificate) error {
+	var cert *x509.Certificate
+	var err error
+	for i := range chain {
+		cert = chain[len(chain)-1-i]
+		// check in reverse order to prevent CRL expiration errors due to revoked CAs no longer issuing CRLs
```
reinkrul, line 233: is the order consistent?

crl/validator.go
```diff
@@ -20,346 +20,420 @@ package crl
...
-
-	v.listsLock.RLock()
-	defer v.listsLock.RUnlock()
-
-	// Check if all CRLs have been downloaded, but if ignore CRL endpoints that are only used by expired certificates.
-	for endpoint, dependingCAs := range endpoints {
-		if !anyCertificateActive(dependingCAs) {
-			continue
+func (v *validator) validateCert(cert *x509.Certificate) error {
+	if nowFunc().After(cert.NotAfter) {
```
reinkrul, line 242: I think this is already checked by golang/tls?

(otherwise, we're missing a NotBefore check..)

crl/validator.go
```diff
@@ -20,346 +20,420 @@ package crl
...
+		return ErrCertExpired
+	}
+	for _, endpoint := range cert.CRLDistributionPoints {
+		crl, ok := v.getCRL(endpoint)
+		if ok && crl.revoked[cert.SerialNumber.String()] != nil {
+			// revocation takes precedence over an internal error.
+			return ErrCertRevoked
+		}
+		if !ok || crl.lastUpdated.IsZero() {
```
reinkrul, line 251: this is in case the CRL isn't downloaded yet? It can just fail i.m.o., since the CRLs are updated as the first thing when starting the node?

`Start()` could even block until the inital CRL download has finished (with a sane time-out, of course, to avoid blocking the node bootstrap).

If we don't block, initial calls might fail (or we block-wait here until the initial download has finished).

crl/validator.go
```diff
@@ -20,346 +20,420 @@ package crl
...
+			// TODO: async or not? This method call is the result of a client request running in its own go routine. We can wait for the crl to be downloaded...
+			// Currently the request fails on the first request using a certificate with an unknown CRl endpoint
+			go v.updateCRL(endpoint, nil)
+			return ErrCRLMissing
 		}
-		// downloaded?
-		if _, ok := v.lists[endpoint]; !ok {
-			return false
+		if !nowFunc().Before(crl.list.NextUpdate) {
+			// CA expiration is checked earlier in the chain
```
reinkrul, line 258: we had a configurable offset before (max time CRL is allowed to be out. of date)

crl/validator.go
```diff
@@ -20,346 +20,420 @@ package crl
...
+		return err
+	}
 
-		certificates, err := x509.ParseCertificates(raw)
-		if err != nil {
-			return err
+	// update when it is a new clr
+	if current == nil || current.list.Number == nil || current.list.Number.Cmp(crl.Number) < 0 {
+		// parse revocations
+		revoked := make(map[string]*pkix.RevokedCertificate, len(crl.RevokedCertificates))
```
reinkrul, line 358: No need to store `*pkix.RevokedCertificate`, cann make `revoked` just `map[string]bool`

crl/validator.go
```diff
@@ -20,346 +20,420 @@ package crl
...
-		var verifiedCerts []*x509.Certificate
 
-		for _, chain := range verifiedChains {
-			for _, verifiedCert := range chain {
-				if verifiedCert.IsCA {
-					verifiedCerts = append(verifiedCerts, verifiedCert)
+		go func(ep string, rl *revocationList, wg *sync.WaitGroup) {
+			err := v.updateCRL(ep, rl)
+			if err != nil {
+				// Connections containing a certificate pointing to this CRL will be accepted until its current.list.NextUpdate.
```
reinkrul, line 327: a CRL not being able to be updated always deserves logging I think?

Why only if `rl` (bad var name btw) is set?

reinkrul (changes requested):
I'm getting "The Big Rewrite that will solve everything" (but which will introduce numerous new bugs) vibes about this PR

crl/validator.go
gerardsn: It is, but `Valdiate` can be (and is) used outside the context of the standard library's tls handshake.

crl/validator.go
gerardsn: tls handshake provides them sorted from leaf to CA. Best effort approach imo

crl/validator.go
gerardsn: I don't think there can ever be more than one issuer per distribution point. It does not make sense to me to host more than 1 CRL on an endpoint.

All TSP (KPN, QuoVadis, ...) issue their own CRL on a CRLDistributionPoint they control (that is listed on the certs they issue), and the TSP all point to CRL `http://crl.pkioverheid.nl/DomPrivateServicesLatestCRL-G1.crl` that is issued by `Staat der Nederlanden Private Services CA - G1` because that is the issuer of the TSP's.

I think we can safely assume that subjects always point to the CRL of their issuer, because that is the CRL that will revoke them. (Theoretically the spec provides other options, but those do not apply to the pkiOverheid chain and I have not found a chain that deviates from this so far.)

crl/validator.go
gerardsn: We use the same truststore for the crl validator and network layer. So this can never fail until we deviate from this pattern. At this point the CRLs behavior should be reevaluated. So this is prevents a future bug.

crl/validator.go
```diff
@@ -20,346 +20,420 @@ package crl
...
+	// started is true once Start is called. Will be set to false if the context is cancelled.
+	started bool
+	// httpClient downloads the CRLs
+	httpClient *http.Client
+	// truststore maps Certificate.Subject.String() to their certificate.
+	// Used for CRL signature checking. Immutable once Start() has been called.
+	truststore map[string]*x509.Certificate
+	// crlChan protects crls from race conditions, both are private
+	// All operations on crls should be in the func sent to the channel.
+	crlChan chan func()
```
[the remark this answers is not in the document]
gerardsn: its an experiment to apply the go proverb: [Do not communicate by sharing memory; instead, share memory by communicating](https://go.dev/doc/effective_go#sharing)

crl/validator.go
gerardsn: I consider the offset to be a bug. The CRLs have an expiration time (`NextUpdate`), and are updated long before they expire. We needed the offset because the current validator does not download the new one until the previous one had expired and was technically invalid. This PR fixes that, so the offset is no longer needed.

crl/validator.go
gerardsn: It is always logged, I just don't want to spam the logs with errors when it fails every once in a while? Maybe reduce the max failure interval to 4 hours? (CSP of all TSP claim very high uptime, so maybe it can always be logged at error level?)

crl/validator.go
gerardsn: Yes, the CRL is not downloaded yet.

The issue is that we do not know the all the CRLDistributionPoints at startup. We know those in the truststore, but the leaf certificates contain some missing CRL endpoints (the TSPs CRL endpoints). We have 1 of them from our own cert, but the pkiO chain has 5 TSPs, so some of them we just need to encounter at runtime.

crl/validator.go
```diff
@@ -20,346 +20,420 @@ package crl
...
-	v.certificatesLock.RLock()
-	defer v.certificatesLock.RUnlock()
-
-	issuerName := crl.Issuer.String()
-
-	certificate, ok := v.certificates[issuerName]
-	if !ok {
-		return errors.New("CRL signature could not be validated against known certificates")
+func (v *validator) Validate(certificate *x509.Certificate) error {
+	if !v.started {
```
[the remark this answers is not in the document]
gerardsn: so you use a nil check to see if it started? I'm not sure how that solves the race condition, because you are still setting the value of a field?

crl/validator.go
gerardsn: changed `rl` -> `crl`

crl/validator.go
gerardsn: changed to `nowFunc().Before(cert.NotBefore) || nowFunc().After(cert.NotAfter) ` and error to `"certificate is not valid"`

crl/validator.go
gerardsn: Ahh, I see your point. We probably need a validation step in the trust configuration to see if the provided cert form a chain to some root?

crl/validator.go
gerardsn: Now always logs on error level when it is missing and fails to download

gerardsn:
> I'm getting "The Big Rewrite that will solve everything" (but which will introduce numerous new bugs) vibes about this PR

CRL validation seems to be hard to get right 100% correct (there's a reason it's not in the standard library). This PR probably has some bugs as too, but at least it will do more than the current validator, since this actually checks if certificates are revoked (and not just the CAs as the current validator is doing).

woutslakhorst (changes requested):
The generate script throws a bunch of errors:

```
...
Using configuration from root-ca.conf
cat: data/certs/04.pem: No such file or directory
cat: data/certs/05.pem: No such file or directory
cat: data/certs/06.pem: No such file or directory
cat: data/certs/07.pem: No such file or directory
data/intA.crl: No such file or directory
data/intB.crl: No such file or directory
```

Must it be run from crl/test/? It also appears it expects some directories that aren't there.

gerardsn:
> Must it be run from crl/test/? It also appears it expects some directories that aren't there.

it must currently be run from crl/test/, but the errors sound more like:
```
# Errors?!?!
MacOS by default uses `LibreSSL` which doesn't work with this script.
Check the version using `openssl version`. This is tested with `OpenSSL 1.1.1t  7 Feb 2023` distributed by `Homebrew`
```

crl/test/README.md
```diff
@@ -0,0 +1,72 @@
+
+# Generate chain
+`sh generate.sh` creates the trust chain using the configuration in `openssl.conf`. 
+
+All sub/intermediate CAs use the same config and therefore same certificate administration. 
+Revocations are currently only generated correctly because of the order of revocation and CRL generation.
+
+# Errors?!?!
+MacOS by default uses `LibreSSL` which doesn't work with this script.
```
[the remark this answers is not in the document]
gerardsn: I'm not sure what the exact issue is. All I know is that it does not apply the config file correctly. Added this to the readme

crl/validator.go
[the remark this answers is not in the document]
gerardsn: `started` is now obsolete with the sync.Map

crl/validator.go
gerardsn: Now returns an error when multiple CAs are found for an endpoint

woutslakhorst:
script works

auth/services/x509/uzi_validator.go
```diff
@@ -173,13 +174,15 @@ func NewUziValidator(env UziEnv, contractTemplates *contract.TemplateStore, crlV
...
-		crlValidator = crl.NewValidator(append(roots[:], intermediates...))
+		crlValidator = crl.New(append(roots[:], intermediates...))
 	}
 
 	validator = &UziValidator{
 		validator:         NewJwtX509Validator(roots, intermediates, validUziSigningAlgs(), crlValidator),
 		contractTemplates: contractTemplates,
 	}
+	// TODO: context is never cancelled
+	crlValidator.Start(context.TODO())
```
beardedfoo, line 185: What will we replace this with?

crl/validator.go
```diff
@@ -20,346 +20,388 @@ package crl
...
+		if nowFunc().Before(current.issuer.NotBefore) || nowFunc().After(current.issuer.NotAfter) {
+			log.Logger().Warnf("Trust store contains invalid certificate: %s (s/n: %s)", current.issuer.Subject.String(), current.issuer.SerialNumber.String())
+			return true
 		}
-
-		for _, certificate := range certificates {
-			if v.IsRevoked(certificate.Issuer.String(), certificate.SerialNumber) {
-				return fmt.Errorf("certificate with issuer '%s' is revoked: %s", certificate.Issuer.String(), certificate.Subject.String())
+		wg.Add(1)
+		go func(ep string, crl *revocationList, wg *sync.WaitGroup) {
```
beardedfoo, line 321: ep?

crl/validator.go
```diff
@@ -20,346 +20,388 @@ package crl
...
+		endpoint, isString := endpoint2.(string)
+		current, isCRL := current2.(*revocationList)
+		if !isString || !isCRL {
+			// should never happen
+			log.Logger().
+				WithField("endpoint", fmt.Sprintf("%v", endpoint2)).
+				WithField("CRL", fmt.Sprintf("%v", current2)).
+				Error("crl validator is invalid")
+			// TODO: should this just panic?
+			return true
```
beardedfoo, line 314: I like to see panics on code that is "unreachable"

beardedfoo (changes requested):
This seems pretty good in general. There's just a couple small comments I wrote that could be addressed

crl/validator.go
gerardsn: https://github.com/nuts-foundation/nuts-node/issues/2025

crl/validator.go
reinkrul: I think you're right, I probably misread it (on https://www.quovadisglobal.com/download-roots-crl/#private)

crl/validator.go
reinkrul: If we remove the option, remember to add it to the release notes

crl/validator.go
reinkrul: Panicing in a goroutine just causes that goroutine to exit, not sure if it even gets logged by default (you'd need a `recover()` clause)

crl/validator.go
```diff
@@ -20,346 +20,388 @@ package crl
...
 		}
-
-		certificates, err := x509.ParseCertificates(raw)
-		if err != nil {
-			return err
+		if nowFunc().Before(current.issuer.NotBefore) || nowFunc().After(current.issuer.NotAfter) {
```
reinkrul, line 316: "before" is a bit obscure, theoretically the certificate might not be valid yet (and become valid in future). But I very highly doubt it's ever used that way in practice

crl/validator.go
```diff
@@ -20,346 +20,388 @@ package crl
...
-		go func(endpoint string) {
-			defer wc.Done()
-
-			if err := v.updateCRL(endpoint); err != nil {
-				errorsChan <- err
-			}
-		}(endpoint)
+// downloadCRL downloads and parses the CRL
+func (v *validator) downloadCRL(endpoint string) (*x509.RevocationList, error) {
+	response, err := v.httpClient.Get(endpoint)
```
reinkrul, line 373: what is the time-out?

crl/validator.go
```diff
@@ -20,346 +20,388 @@ package crl
...
+		}
+	}()
 
-	var syncErrors []error
-
-	for err := range errorsChan {
-		syncErrors = append(syncErrors, err)
+	data, err := io.ReadAll(response.Body)
+	if err != nil {
+		return nil, fmt.Errorf("downloading CRL: %w", err)
```
reinkrul, line 385:
```suggestion
		return nil, fmt.Errorf("reading CRL response: %w", err)
```

crl/validator.go
```diff
@@ -20,346 +20,388 @@ package crl
...
-		if _, ok := v.lists[endpoint]; !ok {
-			return false
+
+		// initial download if missing
+		if crl.lastUpdated.IsZero() {
+			// Pause validate to download CRL. Client requests run in their own go routine, so we can afford to wait.
+			err := v.updateCRL(endpoint, crl)
+			if err != nil {
+				log.Logger().WithError(err).
+					WithField("Subject", cert.Subject.String()).
```
reinkrul, line 248: include endpoint as well

reinkrul (changes requested):
I feel this change deserves a RC1 release for v5.2, so parties can check it still works on ACC

reinkrul:
Much easier to understand now.

crl/validator.go
```diff
@@ -20,346 +20,388 @@ package crl
...
-	Sync() error
-	// SyncLoop periodically calls Sync
-	SyncLoop(ctx context.Context)
-	// IsSynced returns whether all the CRLs are downloaded and are not outdated (based on the offset)
-	IsSynced(maxOffsetDays int) bool
-	// VerifyPeerCertificateFunction returns a tls.Config.VerifyPeerCertificate function based on given config
-	VerifyPeerCertificateFunction(maxValidityDays int) func(rawCerts [][]byte, verifiedChains [][]*x509.Certificate) error
-	// IsRevoked checks whether the certificate was revoked. It does not check if the CRL IsSynced
-	IsRevoked(issuer string, serialNumber *big.Int) bool
+	// Start downloading CRLs. Cancelling the context will stop the Validator.
```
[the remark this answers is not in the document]
gerardsn: cancelling the context only stops the update process, but it is still possible to validate certs after this until the CRLs expire. So nothing breaks, and all go routines will be closed.

crl/validator.go
gerardsn: 10 seconds? added:
```
// After how long a CRL download attempt is stopped.
// Must be short-ish since the update can happen during a validation request.
const syncTimeout = 10 * time.Second
```

crl/validator.go
gerardsn: Opted not to panic since it only kills the `syncLoop`, so that would put us back at not validating anything.

crl/validator.go
```diff
@@ -20,346 +20,388 @@ package crl
...
-	v.listsLock.RUnlock()
-
-	if !ok || crlHasExpired(crl, nowFunc()) {
-		return v.downloadCRL(endpoint)
+	// add CRL distribution points
+	for _, certificate := range val.truststore {
+		issuer, ok := val.truststore[certificate.Issuer.String()]
+		if !ok {
+			err := fmt.Errorf("certificate's issuer is not in the trust store: subject=%s, issuer=%s", certificate.Subject.String(), certificate.Issuer.String())
+			panic(err)
```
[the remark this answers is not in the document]
gerardsn: The problem is that (our) constructors almost never return an error. Giving this one an error limits the locations where this could be called from. The location the constructor is called from in the connection manager for example cannot handle errors (inside a `ConfigOption`). It panics in `grpc/config.go` when `SetValidatePeerCertificateFunc` fails for the same reason. We can add errors all the way upstream until the error can be handled gracefully, but that just makes the code very complicated. So this seemed the lesser of two evils to me, but I'm open to discus this.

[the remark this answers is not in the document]
reinkrul: Maybe it should be passed to the gRPCConnectionManager?

crl/validator.go
```diff
@@ -20,346 +20,388 @@ package crl
...
-
-	v.listsLock.RLock()
-	defer v.listsLock.RUnlock()
-
-	// Check if all CRLs have been downloaded, but if ignore CRL endpoints that are only used by expired certificates.
-	for endpoint, dependingCAs := range endpoints {
-		if !anyCertificateActive(dependingCAs) {
-			continue
+func (v *validator) validateCert(cert *x509.Certificate) error {
+	if nowFunc().Before(cert.NotBefore) || nowFunc().After(cert.NotAfter) {
```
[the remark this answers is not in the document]
gerardsn: removed

crl/validator.go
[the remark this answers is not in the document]
gerardsn: panics can not happen due to user error, only programming error. added comments

crl/validator.go
```diff
@@ -20,346 +20,388 @@ package crl
...
+	// add CRL distribution points
+	for _, certificate := range val.truststore {
+		issuer, ok := val.truststore[certificate.Issuer.String()]
+		if !ok {
+			err := fmt.Errorf("certificate's issuer is not in the trust store: subject=%s, issuer=%s", certificate.Subject.String(), certificate.Issuer.String())
+			panic(err)
+		}
+		err := val.addEndpoints(issuer, certificate.CRLDistributionPoints)
+		if err != nil {
+			panic(err)
```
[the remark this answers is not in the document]
gerardsn: panics can not happen due to user error, only programming error.  It can fail due to invalid certificates (we assume real CAs behave correctly), or delegated CRL signing. The crl validator currently cannot handle delegated CRL signing, so panic is fine.

crl/validator.go
```diff
@@ -20,346 +20,388 @@ package crl
...
+		}
 	}
+	return nil
+}
 
-	crl, err := x509.ParseRevocationList(data)
-	if err != nil {
-		return fmt.Errorf("unable to parse downloaded CRL (url=%s): %w", endpoint, err)
+func (v *validator) SetValidatePeerCertificateFunc(config *tls.Config) error {
+	parseCertificatesSlice := func(rawCerts [][]byte) ([]*x509.Certificate, error) {
```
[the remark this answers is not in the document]
gerardsn: users cannot set `InsecureSkipVerify` and we don't

crl/validator.go
```diff
@@ -20,346 +20,388 @@ package crl
...
+		}
 	}
 
-	v.listsLock.Lock()
-	defer v.listsLock.Unlock()
+	config.VerifyPeerCertificate = func(rawCerts [][]byte, verifiedChains [][]*x509.Certificate) error {
+		// TODO: should we check rawCerts, or verified chains?
+		// rawCerts contains raw certificate data presented by the peer during the tls handshake.
+		// It is used together with the tls.Config to generate verifiedChains, but may contain additional certs that are not in a verified chain.
+		// So the question is, do we reject a client if it sends ANY invalid certificate, even if it is not part of a verifiedChain?
```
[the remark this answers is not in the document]
gerardsn: We decided to use `rawCerts` to prevent attackers from sending a bunch of certificates hoping that one would make it into the `verifiedChains`

reinkrul (approved):

crl/validator.go
gerardsn: changed to `endpoint`

auth/services/x509/uzi_validator.go
gerardsn: see the [make uzi crl validator cancellable](https://github.com/nuts-foundation/nuts-node/pull/2011/commits/9e501ffb2be629c4572d56f679a8d8acb887db23) commit. 
Big changes for little impact. Although I do think that this is a better approach.

crl/validator.go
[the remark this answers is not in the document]
gerardsn: @reinkrul see [don't panic](https://github.com/nuts-foundation/nuts-node/pull/2011/commits/49d2e185010de10b8d36ebcc6c169d80801e6ea5) commit for changes needed to do this. Makes the code better imo, maybe just needs to be in a separate issue.

gerardsn:
todo
- [ ] remove `tls.crl.maxvaliditydays` from config (deprecate to prevent errors?)
- [ ] add release note about the above
- [ ] pending review on [don't panic](https://github.com/nuts-foundation/nuts-node/pull/2011/commits/49d2e185010de10b8d36ebcc6c169d80801e6ea5) commit, newly returned errors need to be added to test coverage.

PR #2024, opened 2023-04-05
use transport.Peer in conversation instead of PeerID

### LINKED ISSUE #1950: ProtocolV2 gossip/conversation manager should not manage peers on peerID
connection management recently changed from peerID to peerID+nodeDID #1874. 

ProtocolV2 assumed peerIDs are unique, which is no longer true. Message sending/handeling has been updated to reflect this (#1938 #1940). All that remains is the gossip and conversation manager.

woutslakhorst:
fixes #1950 

replaced peerID in conversation map with `Peer.String()`.

### DIFF
--- a/network/transport/types.go
+++ b/network/transport/types.go
@@ -67,9 +67,6 @@ func (p Peer) ToFields() logrus.Fields {
 // Usable as map index, not usable for presentation.
 func (p Peer) Key() string {
 	// address is included since 2 connections may exist for a peer (inbound/outbound)
-	if p.NodeDID.Empty() {
-		return fmt.Sprintf("%s@%s", p.ID, p.Address)
-	}
 	return fmt.Sprintf("%s(%s)@%s", p.ID, p.NodeDID.String(), p.Address)
 }
 
--- a/network/transport/types.go
+++ b/network/transport/types.go
@@ -63,6 +63,16 @@ func (p Peer) ToFields() logrus.Fields {
 	}
 }
 
+// Key returns a unique key for this Peer including PeerID and NodeDID.
+// Usable as map index, not usable for presentation.
+func (p Peer) Key() string {
+	// address is included since 2 connections may exist for a peer (inbound/outbound)
+	if p.NodeDID.Empty() {
+		return fmt.Sprintf("%s@%s", p.ID, p.Address)
+	}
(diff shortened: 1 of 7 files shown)


gerardsn (approved):
LGTM for the conversation manager part. Does not fix the Gossip manager part of the original issue.

network/transport/v2/conversation.go
```diff
@@ -82,14 +82,14 @@ type conversationManager struct {
 	mutex                  sync.RWMutex
 	conversations          map[string]*conversation
 	validity               time.Duration
-	lastPeerConversationID map[transport.PeerID]conversationID
+	lastPeerConversationID map[string]conversationID
```
reinkrul, line 85: using `String()` of `Peer` is a bit dangerous, since it was initially intended to provide a loggable/readable representation of `Peer`, rather than a unique hash of the thing. When someone wants to change the format (to further improve readability, add detail, or less detail) it might break. Format also changes when there's an (authnticated) node DID, so a bit odd for a map key. Maybe better to introduce a separate function (e.g. `Key()` or `Hash()`).

Also, `Address` does not seem to be immutable (it is set on `network/transport/grpc/addressbook.go:95`)

reinkrul (changes requested):

network/transport/types.go
```diff
@@ -63,6 +63,16 @@ func (p Peer) ToFields() logrus.Fields {
 	}
 }
 
+// Key returns a unique key for this Peer including PeerID and NodeDID.
+// Usable as map index, not usable for presentation.
+func (p Peer) Key() string {
+	// address is included since 2 connections may exist for a peer (inbound/outbound)
+	if p.NodeDID.Empty() {
+		return fmt.Sprintf("%s@%s", p.ID, p.Address)
```
reinkrul, line 71: What if I make a peer ID:

```
169EAA84-FB3A-4D25-8340-6576787A0D84(did:nuts:some-did)
```

Then I can still mangle with the connection manager?

Safer to not have conditional formatting, but have a set format (just always print nodeDID inbetween `(, )`), even when its empty. Also simpler/less code to test.

reinkrul (changes requested):

reinkrul (approved):

PR #2028, opened 2023-04-07
Core: validate truststore on load

### LINKED ISSUE #2025: Core: LoadTrustStore should verify certificates form a chain
Currently it just loads the certificates without checking anything. 

Validating the chain(s) at this point prevents confusing errors down stream such as
- recent example where a missing intermediate CAs resulted in unexpected connection errors
- panicing CRL validator when the certificates do not form a chain to some root

gerardsn:
closes #2025

moved pkiOverheid certs from network/test to crl/test since it was not used in network

As a consequence of this PR, the node will fail to start when the truststore contains a certificate that is currently not valid.


core/tls.go
```diff
@@ -102,5 +110,24 @@ func LoadTrustStore(trustStoreFile string) (*TrustStore, error) {
...
+// validate returns an error if one of the certificates is invalid or does not form a chain to some root
+func validate(store *TrustStore) error {
+	opts := x509.VerifyOptions{
+		Intermediates: NewCertPool(store.IntermediateCAs),
+		Roots:         NewCertPool(store.RootCAs),
+		CurrentTime:   nowFunc(), // leaving this empty defaults to the current time. Cannot not check time.
+	}
+	for _, cert := range store.Certificates() {
+		if _, err := cert.Verify(opts); err != nil {
+			return err
```
reinkrul, line 129: does it fail when one of the trust store certificate's is expired? I don't think we want that. Since that might cause failures just because the next day arrived

reinkrul (changes requested):

core/tls.go
gerardsn: Yes it fails when a cert has expired. However, this check is only performed at startup, so it wil not cause random crashes on a node that is already running. In that case it will start producing warnings every hour from the CRL validator. Failing to start with an insecure setup sounds like a save default?

reinkrul: It's not insecure; certificates will be rejected because the CA certs are expired. I think we should just log a warning. Can't we skip the expiry check?

reinkrul (approved):

Issue #2047, opened 2023-04-12
Support new authentication means


woutslakhorst:
The means allows for an organization to issue a credential to itself with a very limited lifespan. The contents includes the user as employee. The signing session will create the credential and create a presentation. The presentation is returned as token that can be entered in the `usi` field.

- [x] add specification to RFCs
- [x] alter nuts json-ld context with new `assuranceLevel` field. ['low','substantial','high'] #2048
[...]

PR #2062, opened 2023-04-17
Start signing session for EmployeeIdentity

woutslakhorst:
part of #2047 

PR contains the scaffold and support for starting the signing session.


auth/api/v1/api.go
```diff
@@ -134,6 +134,7 @@ func (w Wrapper) CreateSignSession(ctx echo.Context) error {
 	createSessionRequest := services.CreateSessionRequest{
 		SigningMeans: string(requestParams.Means),
 		Message:      requestParams.Payload,
+		Params:       requestParams.Params,
```
reinkrul, line 137: It was already there, but unused? Odd..

auth/services/contract/notary.go
```diff
@@ -214,6 +213,13 @@ func (n *notary) Configure() (err error) {
 		n.verifiers[uzi.VerifiablePresentationType] = uziVerifier
 	}
 
+	if _, ok := cvMap[selfsigned.ContractFormat]; ok {
+		ss := selfsigned.NewSessionStore(n.vcr)
```
reinkrul, line 217: my spouse would be offended by this short var name (well maybe not, but I don't like these short names).

If you don't like writing you could write;

```
n.verifiers[selfsigned.VerifiablePresentationType] = n.signers[selfsigned.ContractFormat] = selfsigned.NewSessionStore(n.vcr)
```

auth/services/selfsigned/test/vc.json
```diff
@@ -0,0 +1,39 @@
...
+  "credentialSubject": [
+    {
+      "@type": "Organization",
+      "id": "did:nuts:8NYzfsndZJHh6GqzKiSBpyERrFxuX64z6tE5raa7nEjm",
+      "member": {
+        "identifier": "user@example.com",
+        "member": {
+          "familyName": "Tester",
+          "initials": "T",
+          "type": "Person"
```
reinkrul, line 23: shouldn't this be `@type`? If not, `@type` and `type` are the same thing;

"When using a JSON-LD context (see Section [5.3 Extensibility](https://www.w3.org/TR/vc-data-model/#extensibility)), this specification aliases the @type keyword to type to make the JSON-LD documents more easily understood. While application developers and document authors do not need to understand the specifics of the JSON-LD type system, implementers of this specification who want to support interoperable extensibility, do."

(https://www.w3.org/TR/vc-data-model/#types)

auth/services/selfsigned/signer.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+	// load params directly into session
+	marshalled, err := json.Marshal(params)
+	// only functions or other weird constructions can cause an error here. No need for custom error handling.
+	if err != nil {
+		return nil, err
+	}
+	// impossible to get an error here since both the pointer and the data is under our control.
+	_ = json.Unmarshal(marshalled, &s)
+	v.sessions[sessionID] = s
```
reinkrul, line 97: should we check for existence of sessionID?

auth/services/selfsigned/signer.go
```diff
@@ -0,0 +1,103 @@
...
+
+func (v sessionStore) SigningSessionStatus(sessionID string) (contract.SigningSessionResult, error) {
+	s, ok := v.sessions[sessionID]
+	if !ok {
+		return nil, services.ErrSessionNotFound
+	}
+	var vp *vc2.VerifiablePresentation
+
+	if s.status == SessionCompleted {
+		issuer := did.MustParseDID(s.Employer) // todo panic << validate in API?
```
reinkrul, line 47: or when starting session, and convert it to DID there, so it can't fail here (just have to cast)

auth/services/selfsigned/signer.go
```diff
@@ -0,0 +1,103 @@
...
+		expirationData := time.Now().Add(24 * time.Hour)
+		credentialOptions := vc2.VerifiableCredential{
+			Context:           []ssi.URI{credential.NutsV1ContextURI},
+			Type:              []ssi.URI{vc2.VerifiableCredentialTypeV1URI(), ssi.MustParseURI(credentialType)},
+			Issuer:            issuer.URI(),
+			IssuanceDate:      time.Now(),
+			ExpirationDate:    &expirationData,
+			CredentialSubject: s.credentialSubject(),
+		}
+		verifiableCredential, err := v.vcr.Issuer().Issue(context.TODO(), credentialOptions, false, false)
```
reinkrul, line 57: needs audit info in context, otherwise it'll panic later (just pass on the ctx from API)

auth/services/selfsigned/signer.go
```diff
@@ -0,0 +1,103 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("issue VC failed: %w", err)
+		}
+		proofOptions := proof.ProofOptions{
+			Created:      time.Now(),
+			Challenge:    &s.contract,
+			ProofPurpose: "",
```
reinkrul, line 64: empty purpose?

auth/services/selfsigned/signer.go
```diff
@@ -0,0 +1,103 @@
...
+		}
+		vp, err = v.vcr.Holder().BuildVP(context.TODO(), []vc2.VerifiableCredential{*verifiableCredential}, proofOptions, &issuer, true)
```
reinkrul, line 66: `vc2` is ugly, pass ctx

auth/services/selfsigned/types.go
```diff
@@ -0,0 +1,133 @@
...
+// SessionCreated represents the session state after creation
+const SessionCreated = "created"
+
+// SessionInProgress represents the session state after rendering the html
+const SessionInProgress = "in-progress"
+
+// SessionCompleted represents the session state after the user has accepted the contract
+const SessionCompleted = "completed"
+
+type SessionStore interface {
```
reinkrul, line 43: I find "Session" and "Store" low-level concepts (implementation details of services), especially given it doesn't have any "session" or "store"-esque functions (e.g. "startSession", "readX", "readY"), so I think this type should be named something like "SelfSignedMeans". But it's already in the `selfsigned` package, so just "Means" or "Service"?

auth/services/selfsigned/types.go
```diff
@@ -0,0 +1,133 @@
...
+	contract.Signer
+	contract.VPVerifier
+}
+
+// SessionStore is a contract signer and verifier that always succeeds
+// The SessionStore signer is not supposed to be used in a clustered context unless consecutive calls arrive at the same instance
+type sessionStore struct {
+	sessions map[string]session
```
reinkrul, line 51: access is racy, use concurrent map?

auth/services/selfsigned/signer.go
```diff
@@ -0,0 +1,103 @@
...
+	return signingSessionResult{
+		id:                     sessionID,
+		status:                 s.status,
+		request:                s.contract,
+		verifiablePresentation: vp,
+	}, nil
+}
+
+func (v sessionStore) StartSigningSession(rawContractText string, params map[string]interface{}) (contract.SessionPointer, error) {
+	sessionBytes := make([]byte, 16)
```
reinkrul, line 81: entries are created, but never deleted?

And maybe we need a max. number of active sessions (to avoid DoS attacks through memory consumption?) (can be quite high, probably)

auth/services/selfsigned/types.go
```diff
@@ -0,0 +1,133 @@
...
+		vcr:      vcr,
+	}
+}
+
+// session contains the contract text and session signing status
+type session struct {
+	contract string
+	status   string
+	Employer string
+	Employee Employee `json:"employee"`
```
reinkrul, line 68: wait, we're marshalling "session"? (can't find where). Better to separate that (we can marshal Employee, but take it from the session then marshal it, to avoid accidentally exposing internal session state)

auth/services/selfsigned/types.go
```diff
@@ -0,0 +1,133 @@
...
+		sessions: map[string]session{},
+		vcr:      vcr,
+	}
+}
+
+// session contains the contract text and session signing status
+type session struct {
+	contract string
+	status   string
+	Employer string
```
reinkrul, line 67: this doesn't have a JSON tag, so gets odd casing. Example of accidental leaking through marshalling session state?

auth/services/selfsigned/types.go
```diff
@@ -0,0 +1,133 @@
...
+
+type Employee struct {
+	Identifier string `json:"identifier"`
+	RoleName   string `json:"roleName"`
+	Initials   string `json:"initials"`
+	FamilyName string `json:"familyName"`
+}
+
+type sessionPointer struct {
+	sessionID string `json:"sessionID"`
```
reinkrul, line 79: unused JSON tags?

reinkrul (changes requested):
This PR feels rushed (maybe needed some more love)

README.rst
```diff
@@ -194,7 +194,7 @@ The following options can be configured on the server:
...
-    auth.contractvalidators                   [irma,uzi,dummy]                                                                                                                                                                                                                                                                                                     sets the different contract validators to use
+    auth.contractvalidators                   [irma,uzi,selfsigned,dummy]                                                                                                                                                                                                                                                                                                     sets the different contract validators to use
```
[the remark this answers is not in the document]
woutslakhorst: that was fixed long ago

docs/_static/auth/v1.yaml
```diff
@@ -307,7 +307,7 @@ components:
       properties:
         means:
           type: string
-          enum: [ irma, dummy ]
+          enum: [ irma, selfsigned, dummy ]
```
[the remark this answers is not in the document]
woutslakhorst: not related to this PR

auth/api/v1/api.go
woutslakhorst: There wasn't any need up to now

auth/services/contract/notary.go
woutslakhorst: I don't like long lines of code.

auth/services/selfsigned/test/vc.json
woutslakhorst: Just followed the example given on JSON-LD playground. But will change it to `type`

auth/services/selfsigned/signer.go
woutslakhorst: It's a 16 byte secure rand with a several minute lifetime

auth/services/selfsigned/signer.go
woutslakhorst: This is an open question. I guess it should be "authentication", but also see #2047. One of the findings is that we don't support anything other than 'assertionMethod'

auth/services/selfsigned/signer.go
woutslakhorst: dunno why Intellij keeps on naming it vc2 instead of just vc...

auth/services/selfsigned/types.go
woutslakhorst: it's service now.

auth/services/selfsigned/signer.go
woutslakhorst: primary removal will be done by another PR that implements the front channel. Will add an item to #2047 to cleanup abandoned sessions.

Sessions can only be started through internal API's, so DoS attacks are out of scope.

auth/services/selfsigned/types.go
woutslakhorst: Only for unmarshalling. But I splitted the two.

auth/services/selfsigned/types.go
woutslakhorst: indead, removed.

auth/services/selfsigned/types.go
woutslakhorst: I'll add it to the other issue/item (cleaner procedure) because the tests for that will then actually fail.

reinkrul (approved):

Issue #2066, opened 2023-04-19
Network: remove AutoDIDResolver


reinkrul:
People are generally not aware that the node might auto-detect a node DID, which then is often the cause of connection issued (authentication failures), since those DIDs are often not set up correctly (e.g. incorrect NutsComm address, or a DID that hasn't been synced at all). At the same time, it's hard to disable the feature: you can only do so by not configuring a node DID and putting it into strict mode (which forces you to configure your node in a safer way, which you might not want in dev environments).

Better to get people more comfortable handling their node DID and have them explicitly configure it in their `nuts.yaml`, so I propose removing the feature.

woutslakhorst:
fine with me

PR #2067, opened 2023-04-19
Network: remove autoresolver for node DID

reinkrul:
Fixes https://github.com/nuts-foundation/nuts-node/issues/2066

Not allowed in strict mode, so impact (backwards compatibility) should be very limited.


docs/pages/release_notes.rst
```diff
@@ -3,6 +3,15 @@
...
 #############
 
+************************
+Hazelnut update (v5.3.0)
+************************
+
+Release date: **DRAFT**
+
+- Automatically resolving of node DIDs has been removed, since it caused more confusion than it simplified things.
+  It was only meant for workshop/demo purposes and not allowed in strict mode, so the impact should be very limited.
```
gerardsn, line 13: This doesn't really help me in fixing any issues I run into after an upgrade.

gerardsn (changes requested):
backup restore e2e test needs to be fixed since it relies on the auto resolver to set the nodeDID

gerardsn (approved):

PR #2075, opened 2023-04-21
EmployeeIdentity signing means validation

woutslakhorst:
part of #2047 

Adds validation for presentations created with the employee identity


auth/services/selfsigned/types.go
```diff
@@ -141,3 +145,39 @@ func (s session) credentialSubject() []interface{} {
...
+}
+
+type employeeIdentityCredentialMember struct {
+	Identifier string                                 `json:"identifier"`
+	Member     employeeIdentityCredentialMemberMember `json:"member"`
+	RoleName   string                                 `json:"roleName"`
+	Type       string                                 `json:"type"`
+}
+
+type employeeIdentityCredentialMemberMember struct {
```
reinkrul, line 162: too many members?

auth/services/selfsigned/validator.go
```diff
@@ -19,42 +19,170 @@
...
+		return
+	}
+	vc := vcs[0]
+	signingMethod := proof.VerificationMethod
+	signingMethod.Fragment = ""
+	if vc.Issuer.String() != signingMethod.String() {
+		resultErr = newVerificationError("signer must be credential issuer")
+		return
+	}
+	bytes, _ = json.Marshal(vc.CredentialSubject[0])
```
reinkrul, line 122: I think you can use `vc.UnmarshalSubject()`

auth/services/selfsigned/validator.go
```diff
@@ -19,42 +19,170 @@
...
+	if vc.Issuer.String() != credentialSubject.ID {
+		resultErr = newVerificationError("signer must be credentialSubject")
+		return
+	}
+
+	return credentialSubject, proof, nil
+}
+
+func validateRequiredAttributes(credentialSubject employeeIdentityCredentialSubject) error {
+	// check for mandatory attrs
```
reinkrul, line 133: shouldn't these be validated on the VC already?

reinkrul (changes requested):

auth/services/selfsigned/types.go
woutslakhorst: Unfortunate result of the VC model

auth/services/selfsigned/validator.go
woutslakhorst: These are the requirements from the spec (rfc that has to be written). It doesn't validate stuff like issuer, context and id but mostly `credentialSubject` requirements.

Or do you mean it should be under vcr/credential/validator? << where credentials are validated that enter the server wallet from the network?

auth/services/selfsigned/validator.go
woutslakhorst: Also added check for exactly 1 credentialSubject

auth/services/selfsigned/validator.go
reinkrul: The last one, indeed. I could've been more clear.

woutslakhorst: So we validate identity means (and VPs) inside auth. The attestations (VCs over network) are validated within VCR.

reinkrul (approved):

PR #2081, opened 2023-04-25
Release notes: added parallel update fix for v5.1.0

reinkrul:


stevenvegt:
How does this relate to the 5.0.2 notes?

```Coconut update (v5.0.2)[](https://nuts-node.readthedocs.io/en/stable/pages/release_notes.html#coconut-update-v5-0-2)

Release date: 2022-11-30

This patch release fixes the following:

* Synchronize calls to DIDMan to avoid parallel calls from clients creating conflicted DID documents
```

reinkrul:
> How does this relate to the 5.0.2 notes?
> 
> ```
> 
> Release date: 2022-11-30
> 
> This patch release fixes the following:
> 
> * Synchronize calls to DIDMan to avoid parallel calls from clients creating conflicted DID documents
> ```

My hypothesis is that 5.0.2 fixed it client-side, but the cases are from older TXs or TXs that were created by out-of-date nodes?

stevenvegt:
Ah, the 5.0.2 prevents it from happening, and this is about correctly handling the situation if it has occurred.

docs/pages/release_notes.rst
```diff
@@ -43,6 +43,9 @@ Release date: 2023-03-15
   It uses a separate service (like a sidecar in Kubernetes) which implements a standardized API.
   The feature is still experimental, but will become the recommended backend for storing private keys in the next major release.
   See `Storage Configuration <https://nuts-node.readthedocs.io/en/latest/pages/deployment/storage-configuration.html#external-store-api>`_ for more information.
+- Fixed situations in which parallel updates of a DID documents lead to the node not being able to sync with the network,
```
stevenvegt, line 46: A bit unclear if this prevents parallel updates (which it does not) or it handles parallel updates in a different way so the node can still connect to the network.

Perhaps a line like this explains it better:
`Improve transaction processing during startup. This fixes a rare situation caused by parallel updates which leads to an inconsistent state and prevents the node from connecting to the network.
`

reinkrul: Not true, it's not just during startup and you'll still be able to connect to the network.

But rewritten it to make it (hopefully) more clear and to the point.

woutslakhorst (approved):

PR #2082, opened 2023-04-25
Docs: encode + in backup/restore for covenience

reinkrul:


gerardsn (dismissed):
This still doesn't work as copy paste since you need to insert the address:port, but it does make this a lot harder to read. I prefer adding a note/comment reminding users that this should be URL encoded

gerardsn:
> I prefer adding a note/comment reminding users that this should be URL encoded

 I just realized this is already mentioned right below. Perhaps just turn that into a `note::`?

reinkrul:
Turned existing hint into a note.

woutslakhorst (approved):

PR #2095, opened 2023-05-01
PKI: Move default certs to central location

gerardsn:
previous cert in /test/pki had expired. Moved /network/test data to /test/pki since it was used in several locations


stevenvegt:
Looks good, can you add a README to the directory with the certs and truststore and explain the commands used to generate the certs so we can rerun this the next time it expires?

stevenvegt (approved):

PR #2098, opened 2023-05-02
Network: add NutsComm connect-to-self health check 

### LINKED ISSUE #2087: Network: add connect-to-self healthcheck
This asserts the NutsComm address matches the node's TLS certificate, to see if it's set up correctly.
This also checks that the correct certificate chain is configured.

Should be just opening a (TLS) socket (no HTTP/gRPC involved).

gerardsn:
closes #2087

### DIFF
--- a/network/network.go
+++ b/network/network.go
@@ -25,6 +25,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"net"
 	"strings"
 	"sync/atomic"
 	"time"
@@ -86,7 +87,8 @@ type Network struct {
 	eventPublisher      events.Event
 	storeProvider       storage.Provider
 	// assumeNewNode indicates the node hasn't initially sync'd with the network.
-	assumeNewNode bool
+	assumeNewNode  bool
+	selfTestDialer tls.Dialer
 }
 
 // CheckHealth performs health checks for the network engine.
@@ -110,25 +112,9 @@ func (n *Network) CheckHealth() map[string]core.Health {
 		}
 	}
 
-	// auth_config checks that the node is correctly configured to be authenticated by others
-	if n.nodeDID.Empty() {
-		results[healthAuthConfig] = core.Health{
-			Status:  core.HealthStatusUp,
-			Details: "no node DID",
-		}
-		return results
-	}
+	// healthAuthConfig checks that the node is correctly configured to be authenticated by others
+	results[healthAuthConfig] = n.checkNodeDIDHealth(context.TODO(), n.nodeDID)
 
-	if err := n.validateNodeDID(context.TODO(), n.nodeDID); err != nil {
-		results[healthAuthConfig] = core.Health{
-			Status:  core.HealthStatusDown,
-			Details: err.Error(),
-		}
-	} else {
-		results[healthAuthConfig] = core.Health{
-			Status: core.HealthStatusUp,
-		}
-	}
 	return results
 }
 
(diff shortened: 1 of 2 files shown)


network/network_test.go
```diff
@@ -694,9 +760,24 @@ func TestNetwork_validateNodeDID(t *testing.T) {
...
+		cxt.network.certificate = certificate
+		cxt.network.nodeDID = *nodeDID
+		cxt.docResolver.EXPECT().Resolve(*nodeDID, nil).MinTimes(1).Return(completeDocument, &vdrTypes.DocumentMetadata{}, nil)
+		cxt.network.selfTestDialer.NetDialer.Timeout = 10 * time.Millisecond
+
+		health := cxt.network.checkNodeDIDHealth(ctx, *nodeDID)
 
-		assert.EqualError(t, err, "none of the DNS names in TLS certificate match the NutsComm service endpoint (nodeDID=did:nuts:test, NutsComm=grpc://not.nuts.nl:5555)")
+		assert.Equal(t, core.HealthStatusUnknown, health.Status)
+		assert.Contains(t, health.Details, "cannot connect to own NutsComm grpc://nuts.nl:5555: dial tcp")
```
gerardsn, line 780: this will try to connect to `grpc://nuts.nl:5555` every time the test runs, not sure if this matters

reinkrul, line 780: why not use a localhost address, or `example.com`?

gerardsn: before it gets to the connection part of the test it verifies the NutsComm address. Reserved addresses (example.com), IP addresses, and localhost will not pass this check.

network/network.go
```diff
@@ -154,6 +140,13 @@ func NewNetworkInstance(
 		didDocumentFinder:   didDocumentFinder,
 		eventPublisher:      eventPublisher,
 		storeProvider:       storeProvider,
+		selfTestDialer: tls.Dialer{
+			NetDialer: &net.Dialer{
+				Timeout:   time.Second, // TODO: what is a reasonable timeout for this? It runs every couple of seconds when the /health endpoint is called
+				LocalAddr: nil,         // TODO: should this be filled for logging purposes or something?
+			},
+			Config: &tls.Config{InsecureSkipVerify: true}, // set during Configure
```
reinkrul, line 148: Why set this if (especially with an insecure value) always overridden? Then I'd prefer to have it unset.

network/network.go
```diff
@@ -154,6 +140,13 @@ func NewNetworkInstance(
 		didDocumentFinder:   didDocumentFinder,
 		eventPublisher:      eventPublisher,
 		storeProvider:       storeProvider,
+		selfTestDialer: tls.Dialer{
+			NetDialer: &net.Dialer{
+				Timeout:   time.Second, // TODO: what is a reasonable timeout for this? It runs every couple of seconds when the /health endpoint is called
+				LocalAddr: nil,         // TODO: should this be filled for logging purposes or something?
```
reinkrul, line 146: not sure

network/network.go
```diff
@@ -154,6 +140,13 @@ func NewNetworkInstance(
 		didDocumentFinder:   didDocumentFinder,
 		eventPublisher:      eventPublisher,
 		storeProvider:       storeProvider,
+		selfTestDialer: tls.Dialer{
+			NetDialer: &net.Dialer{
+				Timeout:   time.Second, // TODO: what is a reasonable timeout for this? It runs every couple of seconds when the /health endpoint is called
```
reinkrul, line 145: I'd expect a typical TCP/TLS handshake to take 100-ish ms. So should be okay-ish, but we might have to adjust it later on.

reinkrul (changes requested):

network/network.go
gerardsn: This is a self-check that only verifies if a connection can be made without any data exchange, so probably not that insecure? My question is _if_ we want to verify certificates in this self-check or only that the endpoint can be reached?

woutslakhorst:
Issue updated with potential blocking challenge

network/network.go
gerardsn: removed for now

network/network.go
gerardsn: NutsComm is already verified agains the certificate in the node. 
If we verify the certificate, the dialed address is matched against the certificate.

PR #2102, opened 2023-05-03
Docs: add selfsigned means to default contract validators

gerardsn:
`selfsigned` was manually added to the README where it will disappear after the next `make gen-readme`. Adding it as a default contract validator includes it to the generated documentation.

We can also add it in the cmd flag description if it shouldn't be a default


reinkrul (approved):

auth/config.go
```diff
@@ -42,7 +42,7 @@ func DefaultConfig() Config {
 		},
 		HTTPTimeout:         30,
 		ClockSkew:           5000,
-		ContractValidators:  []string{"irma", "uzi", "dummy"},
+		ContractValidators:  []string{"irma", "uzi", "selfsigned", "dummy"},
```
reinkrul, line 45: @stevenvegt does order matter here, should we add it to the end? (I think you fixed weirdness with Koanf slices a long time ago, right?)

PR #2113, opened 2023-05-08
make sure username claim is present for dummy, irma and selfsigned

### LINKED ISSUE #2111: Access token includes username field
`username` is a standard introspection field defined in: [rfc7662](https://www.rfc-editor.org/rfc/rfc7662)
Currently it's only filled for the `selfsigned` means.
If we fill it from IRMA (with email), dummy (with *dummy*) and UZI (with #UZI) as well, it can be used as default auditing identifier.

woutslakhorst:
closes #2111 

the `username` claim will now always be present in the access token and introspection.
note: not for UZI
note: needs doc update, but waiting for other PR. Added to to #2047


reinkrul (approved):
Consideration, the OAS states for this field:

> Identifier uniquely identifying the End-User's account in the issuing system.

Is correct-ish enough?

PR #2117, opened 2023-05-08
VCR: validate credentialSubject

reinkrul:


gerardsn (commented):
Is this always going to be the case? If not this change may complicate the introduction of new features.

For the VDR there is validator for documents received through the network, and a validator that performs some extra sanity checks for data (about to be) created by the node. Does it make sense to have a similar separation here?

reinkrul:
I see no case in which a VC doesn't need to have a DID as subject?

@woutslakhorst @stevenvegt ?

woutslakhorst (changes requested):
Not according to spec,
but may be added to organizationCred and authCred

woutslakhorst (approved):
And even way less code :)

PR #2121, opened 2023-05-08
Core: use /health endpoint in dockerfile

### LINKED ISSUE #2116: Healthcheck should use /health endpoint
Currently the health check in the Dockerfile still points to `/status`

Current statuscodes on `/health` endpoint map UP to 200, and UNKOWN/DOWN to 503
Status UNKNOWN should return statuscode 200 to prevent load balancers from considering it unhealthy and not send traffic to it and prevent orchestrators from killing/restarting the node

- [ ] UNKNOWN return statuscode 200
- [ ] replace `/status` with `/health` in Dockerfile

gerardsn:
closes #2116

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -34,7 +34,7 @@ RUN apk update \
 COPY --from=builder /opt/nuts/nuts /usr/bin/nuts
 
 HEALTHCHECK --start-period=30s --timeout=5s --interval=10s \
-    CMD curl -f http://localhost:1323/status || exit 1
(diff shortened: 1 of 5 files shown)


reinkrul (approved):

reinkrul:
Add to release notes that the health check URL changed (could be considered breaking if someone bound `/status` to an alternative interface, but not very likely).

reinkrul (approved):

reinkrul (approved):

PR #2126, opened 2023-05-09
Docs: add nodedid to backup-restore procedure

### LINKED ISSUE #1927: VDR: private transactions fail if the node's DID document cannot be found
strict mode + unable find the node's own DID document is a fatal error that kills the node. 

In non-strict mode it is not fatal, which results in: cannot find own DID doc -> don't know if private transactions is addressed to me -> a failed event for every private transaction. This also clutters the diagnostics page with these messages:

```
- type: transaction
  hash: e4f301b21c7b2dc04917c60eb55f84dba1d9677425a78c220e7dac90dd74fdc4
  retries: 101
  latest: 2023-02-25T09:47:28.48363339Z
  transaction: {}
  payload: []
  error: 'failed to decrypt PAL header (tx=e4f301b21c7b2dc04917c60eb55f84dba1d9677425a78c220e7dac90dd74fdc4): unable to find the DID document'
```

This can occur when the configured DID is incorrect, or in my case, I deleted `network/data.db` (DAG) + `vdr/didstore.db` (VDR) to trigger resynchronization but left the `NodeDID` in the config.

gerardsn:
closes #1927

### DIFF
--- a/docs/pages/deployment/backup-restore.rst
+++ b/docs/pages/deployment/backup-restore.rst
@@ -86,7 +86,7 @@ To restore a backup, follow the following steps:
 
 .. note::
 
-    When making the API calls, make sure you use the proper URL escaping (`%2B` for `+` and `%3B` for `;`).
+    When making the API calls, make sure you use the proper URL escaping (``%2B`` for ``+`` and ``%3B`` for ``;``).
     Reprocess calls return immediately and will do the work in the background.
 
 BBolt
--- a/docs/pages/deployment/backup-restore.rst
+++ b/docs/pages/deployment/backup-restore.rst
@@ -53,7 +53,7 @@ Refer to the `Redis documentation <https://redis.io/docs/manual/persistence/>`_ 
 Other
 =====
 
-Additionaly, the list of trusted VC issuers must be backed up as well. 
+Additionally, the list of trusted VC issuers must be backed up as well.
 Trusted issuers of VCs are stored in  ``vcr/trusted_issuers.yaml`` inside the ``datadir`` directory.
 If the contents of this file is your primary store for trusted issuers (you're not managing them in an external administrative system), make sure to make a backup.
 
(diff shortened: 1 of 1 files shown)


reinkrul (approved):

reinkrul (approved):

gerardsn:
The node must be kept offline during the reprocess procedure. This requires changing 4 fields in the network config (`bootstrapnodes`, `grpcaddr`, and `nodedid` must be empty, and `enablediscovery==false`). Those changes need to reverted after the reprocess is complete to get the node back into production mode.

- Adding a `network.offline` config flag that automatically applies the above changes and seems like a user friendlier approach than the current flow
- The only indication that the reprocess is complete (signal to restart) is a recently added log message that is not in a tagged version yet

woutslakhorst:
> Adding a network.offline config flag that automatically applies the above changes and seems like a user friendlier approach than the current flow

An API to change the state of the node to offline might be a more friendly approach?

reinkrul:
Reprocess should take node offline until it finishes?

gerardsn:
An API is nicer but also requires a larger change

Hypothesizing what it would need

**logic**
- kill protocols and connection manager to close all connections and take the network offline
- add status to `/health`
- status at startup can be configured with `network.offline` (if not offline at startup you would still have to remove the nodeDID from the config in a backup-restore procedure)
- disallow any API call that creates data published to the network? (OpenID flow can still work?)

**API**
- single endpoint
- POST to change status
- GET returns status for load balancer grpc health check?

gerardsn:
Taking the node offline for a reprocess requires much of the same logic. It feels prone to errors if you request to reprocess multiple types at the same time. What determines when it goes back online? We may end up with the API anyway to fix the status.

gerardsn:
Created #2149 to continue the discussion, and merging this as it describes the current situation

PR #2129, opened 2023-05-09
Docs: add .readthedocs.yaml to fix failing build

gerardsn:
Our read the docs builds have been failing since last week. This is caused by a dependency update, but since we do not fix all dependencies it's not obvious which one caused it. 

The dep update no longer accepts the outdated version of (open)ssl that is apparently used by default in the build environment. The simplest fix appears to be forcing RTD to use a newer version of python (now uses 3.7, version 3.10+ require the supported version of openssl) which is easily managed by this default config.

I've added a hidden page to RTD to test if this works
https://nuts-node.readthedocs.io/en/docs-readthedocs-yaml/

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -1,0 +1,29 @@
+# .readthedocs.yaml
+# Read the Docs configuration file
+# See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
+
+# Required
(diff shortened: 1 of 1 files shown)


reinkrul:
Nice find!

reinkrul (approved):

PR #2130, opened 2023-05-09
Docs: add gen-readme to gen-docs

gerardsn:
recently we've been forgetting to regenerate the readme. This is an attempt to make the process a bit simpler.
Should we add `make gen-docs` as a github action?


reinkrul (approved):

reinkrul:
This will probably be enough fix

PR #2132, opened 2023-05-10
Fix service deletion

### LINKED ISSUE #2131: Deleting a service fails if it is referenced
API call such as`DELETE {{nodeaddress}}/internal/didman/v1/service/did:nuts:FJ2WaKNaf6jKZ9tgJoNCmQKh3xoe9vFEBzj59WSkmSEV%23G4QaQAzVABv6x5RTKNHwSAonJQW2F4hcb4Y71NWdMMiD`

Which is a serviceEndpoint, fails if the service is referenced in a compound service.

stevenvegt:
Fix #2131.

This is WIP, needs tests and cleanup.

## The bug:
Given a service with id `did:nuts:123#abc` of type `oauth`.
During a DIDMan `deleteService` call it is checked if the service is in use.
Instead of the following reference format: `did:nuts:123/serviceEndpoint?type=oauth`
The following was used `did:nuts:123#abc` in the check.
Which always returned false
Then DIDMan removes the service from the DID Document and tries to update the DID Document.
Then the Updater performs a validation step on the DID Document which fails because there is a compound service with a reference to a service that does not exists.

### DIFF
--- a/didman/didman.go
+++ b/didman/didman.go
@@ -26,6 +26,7 @@ import (
 	"errors"
 	"fmt"
 	"net/url"
+	"strings"
 	"sync"
 
 	ssi "github.com/nuts-foundation/go-did"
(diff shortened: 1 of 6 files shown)


gerardsn:
The service endpoints check was designed to consist of 2 parts:
- `Didman` was supposed to check if the service to be deleted was referenced by another service on its own or any upstream did document and throw a `ErrServiceInUse` if true.
- `VDR.Update` validates that all services on the updated did document can be resolved to an absolute endpoint within 5 steps.

The check in didman was broken, but since the VDR also checks the updated document it resulted in a confusing error. 

I've amended this PR to let didman only check the did document that is being updated. There was some discussion around this allowing the creation of invalid upstream references. IMO we should not check upsteam references since
- it requires looping the VDR
- it prevents the service from being deleted if referenced by another did (think lost private keys, or intentionally/attack)
- the newly introduced UpdateService API does not validate all upstream references either, which would be to complex since it requires following all (possibly multiple) upstream references to verify that they are all still resolvable within 5 steps.

gerardsn:
@stevenvegt I can't request a review from you since you opened the PR

woutslakhorst (approved):

reinkrul (approved):

PR #2137, opened 2023-05-10
back port expired certificates fix to v5.1 - simpler fix for #2095

gerardsn:


reinkrul (approved):

reinkrul (approved):

gerardsn:
> What do we need to fix on v5.1? Only backport fixes on v5.2 I'd say?

because this version is still supported (?) and I do not want to debug issues that are already fixed when there is a bug fix that actually matters

PR #2141, opened 2023-05-10
always write index data for VDR, never check on duplicate

woutslakhorst:
Backport to master and other V5 versions.


gerardsn (commented):
with some comments these files would be a lot easier to review

beardedfoo:
> with some comments these files would be a lot easier to review

My feedback would be the same. Adding some lengthy comments about *why* we are doing these inserts etc., what we're trying to work around, etc. would be good.

gerardsn (approved):

PR #2147, opened 2023-05-11
Didman: added UpdateEndpoint and UpdateCompoundService

reinkrul:
Given https://github.com/nuts-foundation/nuts-node/issues/2131, it becomes very unpractical to update an absolute endpoint (service with a URL) when it's referenced by other services: it would mean parties have to:

1. delete all references first,
2. then delete the service itself,
3. then create the service with the updated endpoint,
4. finally recreate all references.

That it didn't hurt much up until now is because the "is being referenced"-check was always broken (at least for references from other documents), but it still applies to for references within the same document.

Furthermore, each hackaton people are asking how to update a service so having an API for that really helps.

### DIFF
--- a/didman/api/v1/api.go
+++ b/didman/api/v1/api.go
@@ -94,29 +94,44 @@ func (w *Wrapper) Routes(router core.EchoRouter) {
 	}))
 }
 
-// AddEndpoint handles calls to add a service. It only checks params and sets the correct return status code.
-// didman.AddEndpoint does the heavy lifting.
-func (w *Wrapper) AddEndpoint(ctx context.Context, request AddEndpointRequestObject) (AddEndpointResponseObject, error) {
-	id, err := did.ParseDID(request.Did)
+func (w *Wrapper) addOrUpdateEndpoint(
+	ctx context.Context, requestDID string, properties EndpointProperties,
+	operation func(ctx context.Context, id did.DID, serviceType string, endpoint url.URL) (*did.Service, error),
+) (*did.Service, error) {
+	id, err := did.ParseDID(requestDID)
 	if err != nil {
 		return nil, err
 	}
 
-	if len(strings.TrimSpace(request.Body.Type)) == 0 {
+	if len(strings.TrimSpace(properties.Type)) == 0 {
 		return nil, core.InvalidInputError("invalid value for type")
 	}
 
-	u, err := url.Parse(request.Body.Endpoint)
+	endpoint, err := url.Parse(properties.Endpoint)
 	if err != nil {
 		return nil, core.InvalidInputError("invalid value for endpoint: %w", err)
 	}
+	return operation(ctx, *id, properties.Type, *endpoint)
+}
 
-	endpoint, err := w.Didman.AddEndpoint(ctx, *id, request.Body.Type, *u)
+// AddEndpoint handles calls to add a service. It only checks params and sets the correct return status code.
+// didman.AddEndpoint does the heavy lifting.
+func (w *Wrapper) AddEndpoint(ctx context.Context, request AddEndpointRequestObject) (AddEndpointResponseObject, error) {
+	endpoint, err := w.addOrUpdateEndpoint(ctx, request.Did, *request.Body, w.Didman.AddEndpoint)
 	if err != nil {
 		return nil, err
 	}
-
 	return AddEndpoint200JSONResponse(*endpoint), nil
+}
+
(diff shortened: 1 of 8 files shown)


didman/api/v1/api.go
```diff
@@ -94,31 +94,50 @@ func (w *Wrapper) Routes(router core.EchoRouter) {
...
 		return nil, err
 	}
-
 	return AddEndpoint200JSONResponse(*endpoint), nil
 }
 
+// UpdateEndpoint handles calls to update a service. It only checks params and sets the correct return status code.
+// didman.UpdateEndpoint does the heavy lifting.
+func (w *Wrapper) UpdateEndpoint(ctx context.Context, request UpdateEndpointRequestObject) (UpdateEndpointResponseObject, error) {
+	if request.Body.Type != "" && request.Body.Type != request.Type {
```
stevenvegt, line 130: duplicate logic with addOrUpdateEndpoint checks?

didman/api/v1/api.go
```diff
@@ -94,31 +94,50 @@ func (w *Wrapper) Routes(router core.EchoRouter) {
 	}))
 }
 
-// AddEndpoint handles calls to add a service. It only checks params and sets the correct return status code.
-// didman.AddEndpoint does the heavy lifting.
-func (w *Wrapper) AddEndpoint(ctx context.Context, request AddEndpointRequestObject) (AddEndpointResponseObject, error) {
-	id, err := did.ParseDID(request.Did)
+func (w *Wrapper) addOrUpdateEndpoint(
```
stevenvegt, line 97: Function params gives extra mental complexity to prevent just a few duplicate lines.

didman/api/v1/api.go
```diff
@@ -94,31 +94,50 @@ func (w *Wrapper) Routes(router core.EchoRouter) {
...
 		return nil, err
 	}
 
-	if len(strings.TrimSpace(request.Body.Type)) == 0 {
+	if len(strings.TrimSpace(properties.Type)) == 0 {
 		return nil, core.InvalidInputError("invalid value for type")
 	}
 
-	u, err := url.Parse(request.Body.Endpoint)
+	endpoint, err := url.Parse(properties.Endpoint)
```
stevenvegt, line 110: consider using https://pkg.go.dev/net/url#ParseRequestURI which is more strict.

didman/didman.go
```diff
@@ -195,6 +215,34 @@ func (d *didman) AddCompoundService(ctx context.Context, id did.DID, serviceType
...
 
+func (d *didman) UpdateCompoundService(ctx context.Context, id did.DID, serviceType string, endpoints map[string]ssi.URI) (*did.Service, error) {
+	unlockFn := d.callSerializer.Lock(id.String())
+	defer unlockFn()
+
+	log.Logger().
+		WithField(core.LogFieldDID, id.String()).
+		WithField(core.LogFieldServiceType, serviceType).
+		WithField(core.LogFieldServiceEndpoint, endpoints).
+		Debug("Adding compound updated")
```
stevenvegt, line 226: What is "Adding compound updated"?

didman/didman.go
```diff
@@ -524,6 +572,33 @@ func (d *didman) addService(ctx context.Context, id did.DID, serviceType string,
...
+	if err != nil {
+		return nil, err
+	}
+	service := &did.Service{
+		Type:            serviceType,
+		ServiceEndpoint: serviceEndpoint,
+	}
+	service.ID = generateIDForService(id, *service)
+
+	serviceFound := false
```
stevenvegt, line 586: Can you add a comment what this serviceFound does? Perhaps add a function description as well and under which conditions it might error.

didman/api/v1/api.go
reinkrul: First complain about duplicate logic, then complain about generic functions to prevent code duplication? :-p

didman/didman.go
reinkrul: changed variable name to `serviceToBeUpdatedFound`, function is quite self-explanatory i.m.o?

didman/api/v1/api.go
reinkrul: Dangerously out of scope change (changing existing behavior)

didman/api/v1/api.go
reinkrul: That check isn't there in `addOrUpdateEndpoint`. You mean with `UpdateCompoundService`? That's a different type, so it looks the same, but according to Go, it isn't

stevenvegt (approved):

Issue #2056, opened 2023-04-13
Wallet API should check DID is under control of node when accepting credential offer


reinkrul:
Currently, it just starts handling it, will probably fail when signing credential request proof.

Should fail fast.

Issue #2054, opened 2023-04-13
OIDC4VCI: OIDCIssuer must check requested issuer is under control of the node


reinkrul:
Meaning, it has private keys associated with the issuer DID. Now it just answers with incorrect info (metadata) or it will fail due to missing key material. Should be a preliminary check for all funcs.

reinkrul:
Already needs to be checked when serving metadata, but that leads to a DoS vulnerability if we simply check for private key material existence: an attacker could cause 1 request to be routed through the Nuts node, maybe Vault proxy, to the backing Vault instance. Caching results partly helps, but would still be vulnerable to request random (non-existing) DIDs?

woutslakhorst:
Doesn't requesting access token suffer from the same attack (note that the API is a /n2n API)?

PR #2154, opened 2023-05-15
OpenID4VCI: Check whether holder/issuer DID is tenant of the node

reinkrul:
Open for discussion.

Fixes https://github.com/nuts-foundation/nuts-node/issues/2056
Fixes https://github.com/nuts-foundation/nuts-node/issues/2054

Kinda hard feature, but reusable in other cases (e.g. show list of tenants in diagnostics or API).

Do you think this is the right approach?

Found it hard to determine where it belongs (which engine).

### DIFF
--- a/vdr/vdr.go
+++ b/vdr/vdr.go
@@ -60,13 +60,14 @@ type VDR struct {
 
 // NewVDR creates a new VDR with provided params
 func NewVDR(config Config, cryptoClient crypto.KeyStore, networkClient network.Transactions, store didstore.Store, eventManager events.Event) *VDR {
+	resolver := didservice.Resolver{Store: store}
 	return &VDR{
 		config:            config,
 		network:           networkClient,
 		store:             store,
 		didDocCreator:     didservice.Creator{KeyStore: cryptoClient},
-		didDocResolver:    didservice.Resolver{Store: store},
-		documentOwner:     newCachingDocumentOwner(privateKeyDocumentOwner{keyResolver: cryptoClient}),
+		didDocResolver:    resolver,
+		documentOwner:     newCachingDocumentOwner(privateKeyDocumentOwner{keyResolver: cryptoClient}, resolver),
 		networkAmbassador: NewAmbassador(networkClient, store, eventManager),
 		keyStore:          cryptoClient,
 	}
@@ -123,13 +124,6 @@ func (r *VDR) ConflictedDocuments() ([]did.Document, []types.DocumentMetadata, e
 }
 
 func (r *VDR) IsOwner(ctx context.Context, id did.DID) (bool, error) {
-	// First perform a cheap DID existence check (subsequent checks are more expensive)
-	_, _, err := r.store.Resolve(id, nil)
-	if errors.Is(err, types.ErrNotFound) || errors.Is(err, types.ErrDeactivated) {
-		return false, nil
-	} else if err != nil {
-		return false, fmt.Errorf("unable to check ownership of DID: %w", err)
-	}
 	return r.documentOwner.IsOwner(ctx, id)
 }
 
--- a/vdr/did_owner.go
+++ b/vdr/did_owner.go
@@ -20,6 +20,8 @@ package vdr
 
 import (
 	"context"
+	"errors"
+	"fmt"
 	"github.com/nuts-foundation/go-did/did"
 	"github.com/nuts-foundation/nuts-node/crypto"
 	"github.com/nuts-foundation/nuts-node/vdr/types"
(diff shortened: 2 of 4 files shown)


woutslakhorst (changes requested):
First check the VDR for existence of DIDs. Then if valid DID, proceed with true/false caching.

vcr/tenants.go
gerardsn, line 1: copyright

vcr/tenants.go
```diff
@@ -0,0 +1,75 @@
...
+//   - Positive matches never change (would indicate a DID deactivation, and it being used afterwards).
+//   - Negative matches never change (would indicate an attacker trying DIDs, or a bug/misconfiguration) until restart.
+//
+// Since it does some assumptions it should not be used as authorization mechanism,
+// since it does not unmark as tenant when a DID is deactivated.
+type cachedTenantRegistry struct {
+	keyResolver     crypto.KeyResolver
+	tenantsList     *sync.Map
+	negativeMatches *sync.Map
+	updateLock      *sync.Mutex
```
gerardsn, line 26: `updateLock` seems unused?

vcr/tenants.go
```diff
@@ -0,0 +1,75 @@
...
+func newCachedTenantsRegistry(keyResolver crypto.KeyResolver) *cachedTenantRegistry {
+	return &cachedTenantRegistry{
+		keyResolver:     keyResolver,
+		updateLock:      &sync.Mutex{},
+		tenantsList:     new(sync.Map),
+		negativeMatches: new(sync.Map),
+	}
+}
+
+func (t *cachedTenantRegistry) IsProbableTenant(ctx context.Context, id did.DID) (bool, error) {
```
gerardsn, line 38: no unit tests and all other tests appear to be mocked

vcr/tenants.go
```diff
@@ -0,0 +1,75 @@
...
+		if idx == -1 {
+			// Not a valid key ID
+			continue
+		}
+		tenantDID := key[:idx]
+		if isAsString == tenantDID {
+			// Positive match
+			isATenant = true
+		}
+		t.tenantsList.Store(tenantDID, true)
```
gerardsn, line 66: this doesn't seem right

gerardsn (changes requested):
I think this belongs in the `VDR`

reinkrul:
Discussed with @gerardsn that this is the right direction:

- "Owning" can mean 2 things:
  - You have a key of a DID with which you can do something with it (e.g. issue VC or present a VC)
  - You can alter it (by having a specific key or being controller of it)
- The first one is relevant for this issue. But "tenant" is a bad name (since it implies multi-tenancy, which we don't support).
-

reinkrul:
Changed to `DocumentOwner.IsOwner()` and moved to VDR. Split caching in separate impl.

vdr/vdr.go
```diff
@@ -118,6 +122,17 @@ func (r *VDR) ConflictedDocuments() ([]did.Document, []types.DocumentMetadata, e
 	return conflictedDocs, conflictedMeta, err
 }
 
+func (r *VDR) IsOwner(ctx context.Context, id did.DID) (bool, error) {
+	// First perform a cheap DID existence check (subsequent checks are more expensive)
+	_, _, err := r.store.Resolve(id, nil)
```
woutslakhorst, line 127: although cheap, it's not cached. My guess is that you can skip this step and make the average performance faster. The only reason to do this, is if you expect a lot of invalid requests. Also the Resolve will be probably be done again later in the flow.

I think it would be beneficial if the Resolve operation has it's own internal cache? >> new issue?

woutslakhorst (commented):
Failing tests?

vdr/vdr.go
reinkrul: Problem is not checking DID existence would introduce a DoS vulnerability for the negative result caching, since an attacker can just post random `did:nuts` strings, growing the negative result cache indefinitely.

When the store is in BBolt, it's already in-memory thus won't get much faster then it already is? (Maybe less locking).

gerardsn: Should this move to `cachingDocumentOwner`? You probably want this check for every type of cached document owner.

woutslakhorst: BBolt does not keep everything in cache, also this isn't true for other storage. That's why I suggested it to become a new issue.

woutslakhorst (approved):

vdr/vdr.go
reinkrul: That's right. But i.m.o. it was still worth adding, since in the recommended production deployment (VDR storage in Redis, private keys in Vault through the Vault Proxy) querying the VDR storage is most probably faster than querying Vault.

PR #2158, opened 2023-05-15
PKI: turn PKI into an engine

gerardsn:


README.rst
```diff
@@ -236,6 +235,9 @@ The following options can be configured on the server:
...
     network.v2.gossipinterval                 5000                                                                                                                                                                                                                                                                                                                 Interval (in milliseconds) that specifies how often the node should gossip its new hashes to other nodes.
+    **PKI**
+    pki.maxupdatefailhours                    4                                                                                                                                                                                                                                                                                                                    maximum number of hours that a denylist update can fail
```
reinkrul, line 239: Descriptions normally start with a capital

auth/auth.go
```diff
@@ -154,11 +148,14 @@ func (auth *Auth) Configure(config core.ServerConfig) error {
 			RootCAs:      trustStore.CertPool,
 			MinVersion:   core.MinTLSVersion,
 		}
-		if err = validator.SetValidatePeerCertificateFunc(tlsConfig); err != nil {
+		if err = auth.pkiValidator.AddTruststore(trustStore.Certificates()); err != nil {
```
reinkrul, line 151: The truststore config is defined on root? Shouldn't this truststore initialization be performed by the PKI engine?

auth/services/notary/notary.go
```diff
@@ -187,15 +187,13 @@ func (n *notary) Configure() error {
 			return err
 		}
 
-		pkiCfg := pkiconfig.Config{
-			MaxUpdateFailHours: 4,
-		}
-
-		n.uziCrlValidator, err = pki.NewValidator(pkiCfg, truststore.Certificates())
+		// seed pkiValidator with uzi certificate chain
+		err = n.pkiValidator.AddTruststore(truststore.Certificates())
```
reinkrul, line 191: I'm thinking... shouldn't there be multiple PKI validators, depending on the context? When making n2n calls, you have a different truststore as when verifying UZI signatures? Or doesn't it matter?

reinkrul (changes requested):

auth/services/notary/notary.go
gerardsn: The PKI validator only verifies the revocation status. If the presented certificate is part of the correct chain is (still) determined per use case. On connections this is part of the tls.Config, and in UZI this is validated against the UZI trust config. By merging the validators we centralize the CRL logic, deduplicate CRL checking (currently CRLs are downloaded twice), and make it possible to add a health check to monitor all of this.

auth/auth.go
gerardsn: We should probably also move `tls.go` and related code to this engine, but this and the config change are larger changes that need their own issue/PR

beardedfoo (approved):
LGTM

PR #2169, opened 2023-05-17
added Start() to auth means signers for session cleanup

### LINKED ISSUE #2168: Add cleanup procedure for removing abandoned sessions. Make session map access concurrent.

woutslakhorst:
closes #2168 

Auth means signers now have a `Start(ctx)` method that could start a go routine. The selfsigned means starts a go routine for session cleanup. It's stopped through the context.

mutex was also described in issue, but was already implemented.


auth/services/selfsigned/store.go
```diff
@@ -65,3 +74,30 @@ func (s *memorySessionStore) Delete(sessionID string) {
...
+		}
+	}()
+}
+
+func (s *memorySessionStore) evict() {
+	s.lock.Lock()
+	defer s.lock.Unlock()
+
+	for k, v := range s.sessions {
+		if v.ExpiresAt.Before(time.Now()) {
```
reinkrul, line 99: Now certain states can't be retrieved anymore by the REST client (e.g. expired), leading them to be functionally dead code, maybe evict it a minute or so after expiry instead of straight away.

auth/services/selfsigned/store.go
```diff
@@ -65,3 +74,30 @@ func (s *memorySessionStore) Delete(sessionID string) {
 	defer s.lock.Unlock()
 	delete(s.sessions, sessionID)
 }
+
+func (s *memorySessionStore) Start(ctx context.Context) {
```
reinkrul, line 78: theoretically you'd need a `Stop()` function which is called on shutdown and waits until the goroutine has finished. Otherwise you could have runaway goroutines or corrupt state (although there's no persistent state here).

reinkrul (changes requested):

gerardsn (changes requested):

auth/services/selfsigned/store.go
woutslakhorst: I'll add a 10 minute delay

gerardsn (approved):

auth/services/notary/notary.go
```diff
@@ -214,6 +214,15 @@ func (n *notary) Configure() error {
 	return nil
 }
 
+func (n *notary) Start(ctx context.Context) {
+	if n.uziCrlValidator != nil {
+		n.uziCrlValidator.Start(ctx)
+	}
```
gerardsn, line 220: removed in #2158 
```suggestion
```

auth/services/selfsigned/store.go
reinkrul: Yes, and then the Stop function should wait until the go routine being stopped has actually finished. But there no resources to corrupt (by early termination) here

PR #2171, opened 2023-05-17
added field description to token introspection

### LINKED ISSUE #2170: Add note on `username` claim in docs. It's always present and should be used over email/identifier.

woutslakhorst:
closes #2170 

Added note to OAS and auth getting started page.

### DIFF
--- a/docs/pages/getting-started/5-authentication.rst
+++ b/docs/pages/getting-started/5-authentication.rst
@@ -521,7 +521,7 @@ The result will return the following identifying properties if user authenticati
 - family_name
 - assurance_level
 
-``username`` can be used for an unique identifier. It will be filled with an email address or an employer identifier.
+``username`` can be used as a unique user identifier. It will be filled with an email address or other identifier, issued by the user's employer.
 
 .. note::
 
--- a/docs/pages/getting-started/5-authentication.rst
+++ b/docs/pages/getting-started/5-authentication.rst
@@ -513,7 +513,7 @@ The POST body is ``application/x-www-form-urlencoded`` encoded
 
     token=<jwt>
 
-The result will return the following identifying properties if user authentication was present in the access token request:
+The result will return the following identifying properties if user authentication was present in the access token:
 
 - username
 - initials
--- a/docs/pages/getting-started/5-authentication.rst
+++ b/docs/pages/getting-started/5-authentication.rst
@@ -517,11 +517,14 @@ The result will return the following identifying properties if user authenticati
 
 - username
 - initials
-- prefix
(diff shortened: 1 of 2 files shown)


docs/pages/getting-started/5-authentication.rst
```diff
@@ -496,3 +496,33 @@ It will return a structure similar to:
...
+
+    POST /internal/auth/v1/accesstoken/introspect
+
+The POST body is ``application/x-www-form-urlencoded`` encoded
+
+.. code-block::
+
+    token=<jwt>
+
+The result will return the following identifying properties if user authentication was present in the access token request:
```
reinkrul, line 516:
```suggestion
The result will return the following identifying properties if user authentication was present in the access token:
```

docs/pages/getting-started/5-authentication.rst
```diff
@@ -496,3 +496,33 @@ It will return a structure similar to:
...
+
+- username
+- initials
+- prefix
+- family_name
+- assurance_level
+
+``username`` can be used for an unique identifier. It will be filled with an email address or an employer identifier.
```
reinkrul, line 524:
```suggestion
``username`` can be used as a unique user identifier. It will be filled with an email address or other identifier, issued by the user's employer.
```

docs/pages/getting-started/5-authentication.rst
```diff
@@ -496,3 +496,33 @@ It will return a structure similar to:
...
+
+.. code-block::
+
+    token=<jwt>
+
+The result will return the following identifying properties if user authentication was present in the access token request:
+
+- username
+- initials
+- prefix
```
reinkrul, line 520: Not present in employeeid? Also missing role name.

reinkrul (changes requested):

docs/pages/getting-started/5-authentication.rst
```diff
@@ -496,3 +496,33 @@ It will return a structure similar to:
     }
 
 The ``validity`` will indicate its validity. An expired contract is considered invalid.
+
+Audit log requirements
```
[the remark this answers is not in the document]
woutslakhorst: might be true, but we only have 1 page on authentication and it might be best to leave it that way, otherwise documentation gets scattered to much?

docs/pages/getting-started/5-authentication.rst
woutslakhorst: True, so I added a note on the minimal set and that other props may be present.

reinkrul (approved):

PR #2180, opened 2023-05-22
change all refs from selfsigned to employeeid

woutslakhorst:
urls used `employeeid` but other references used `selfsigned`. All (but the packages) now use `employeeid` as the name of the means.


gerardsn:
urls are part of an internal service, why not change those?

woutslakhorst:
@stevenvegt , your thoughts?

reinkrul (approved):
Not opposed to the changes, but the Package is still called selfsigned, does that make sense?

stevenvegt (approved):
LGTM

PR #2182, opened 2023-05-22
fix make gen-docs

gerardsn:


makefile
```diff
@@ -74,7 +74,7 @@ gen-protobuf:
 
 gen-docs:
 	go run ./docs docs
-	gen-readme
+	make gen-readme
```
woutslakhorst, line 77: `gen-docs: gen-readme`

then the `gen-docs` target depends on `gen-readme`

woutslakhorst (changes requested):

makefile
gerardsn: that would processes the targets in the wrong order. `go run ./docs docs` generates files that are used by `gen-readme`

gerardsn: added comment to clarify this

PR #2183, opened 2023-05-23
make assuranceLevel optional

woutslakhorst:
should have been optional, otherwise you get an empty value and the test data is no longer valid.


gerardsn (approved):

reinkrul:
Omitempty?

PR #2185, opened 2023-05-24
Allow literal \n in trustedsigner

beardedfoo:
This change makes configuring the denylist a lot simpler


reinkrul:
The new lines in PEM files are just there for readability: I think they are always 64 characters long (except the last line)?

We could also just try to parse it as base64 DER encoded when there is no -----BEGIN CERT (etc) header and/or newlines present. Which is basically PEM encoding, if I remember correctly.

PR #2196, opened 2023-05-26
PKI: do not check if CA in AddTruststore(chain)

gerardsn:
stable and dev network root certificates are x509 v1. Checking if a v1 cert is a CA is non trivial resulting in the following log messages:
```
time="2023-05-26T13:29:30Z" level=warning msg="Did not add CRL Distribution Points" error="certificate's issuer is not in the trust store: subject=CN=Nuts Development Network Root CA, issuer=CN=Nuts Development Network Root CA" module=pki
time="2023-05-26T13:29:30Z" level=warning msg="Did not add CRL Distribution Points" error="certificate's issuer is not in the trust store: subject=CN=Nuts Development Network Root CA, issuer=CN=Nuts Development Network Root CA" module=pki
```
pki validator now just adds the entire chain, which may include end user certs. They are only used for CRL signature verification so should have no impact.

### DIFF
--- a/pki/validator.go
+++ b/pki/validator.go
@@ -210,7 +210,7 @@ func (v *validator) validateCert(cert *x509.Certificate) error {
 		// add distribution endpoint if unknown
 		if !ok {
 			var issuer *x509.Certificate
-			issuer, ok = v.getCA(cert.Issuer.String())
+			issuer, ok = v.getCert(cert.Issuer.String())
 			if !ok {
 				return ErrCertUntrusted
 			}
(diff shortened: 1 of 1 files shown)


woutslakhorst (changes requested):
should be fine, but could the code below work?

```
// isCA returns true if cert is a CA certificate
// it checks against both v1 and v2 CAs
func isCA(cert *x509.Certificate) bool {
	return cert.IsCA || cert.KeyUsage&x509.KeyUsageCertSign != 0
}
```

woutslakhorst (approved):

PR #2202, opened 2023-05-30
Wrap denylist errors in tls.CertificateVerificationError

beardedfoo:
@gerardsn what further changes would make sense in the connection manager? What kind of test should be implemented?


gerardsn (changes requested):
This should move to validator.Validate or validate.SetVerifyPeerFunc. The latter only applies to tls.Config, so that would place the error in the correct context

gerardsn:
@beardedfoo can you review this

reinkrul (approved):

Issue #2204, opened 2023-05-31
Remove purposeOfUseClaimDeprecated from access token request


woutslakhorst:
the `purposeOfUseClaim` field can be removed from access token request. `bearerToken.service` no longer needs to check for both fields.

PR #2205, opened 2023-05-31
add extra purposeOfUse claim to access token request

### LINKED ISSUE #1861: Typo in purposeOfUseClaim
```
const purposeOfUseClaim = "purposeOfUseClaim"
```
That looks like a find/replace went wrong. According to spec it must be `purposeOfUse`. 

When changing this, the check must both check for the old and the new otherwise non-updated nodes can no longer get a token from updated nodes.

Introspection will return it under the `service` field.

woutslakhorst:
fixes #1861

This current field name is a type and not according to spec. 

removal of deprecated claim is done in #2204 (v6)

### DIFF
--- a/auth/services/oauth/authz_server_test.go
+++ b/auth/services/oauth/authz_server_test.go
@@ -418,7 +418,7 @@ func TestService_validatePurposeOfUse(t *testing.T) {
 		assert.NoError(t, err)
 	})
 
-	t.Run("error - only correct claim", func(t *testing.T) {
+	t.Run("ok - only correct claim", func(t *testing.T) {
 		ctx := createContext(t)
 		tokenCtx := validContext()
 		tokenCtx.jwtBearerToken.Remove(purposeOfUseClaimDeprecated)
--- a/auth/services/oauth/authz_server_test.go
+++ b/auth/services/oauth/authz_server_test.go
@@ -408,7 +408,7 @@ func TestService_validatePurposeOfUse(t *testing.T) {
 		assert.EqualError(t, err, "no purposeOfUse given")
 	})
 
-	t.Run("error - only deprecated claim", func(t *testing.T) {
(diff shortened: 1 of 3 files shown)


auth/services/oauth/authz_server_test.go
```diff
@@ -397,15 +397,36 @@ func TestService_validateSubject(t *testing.T) {
...
+		ctx := createContext(t)
+		tokenCtx := validContext()
+		tokenCtx.jwtBearerToken.Remove(purposeOfUseClaim)
+
+		err := ctx.oauthService.validatePurposeOfUse(tokenCtx)
+
+		assert.NoError(t, err)
+	})
+
+	t.Run("error - only correct claim", func(t *testing.T) {
```
reinkrul, line 421:
```suggestion
	t.Run("ok - only correct claim", func(t *testing.T) {
```

auth/services/oauth/authz_server_test.go
```diff
@@ -397,15 +397,36 @@ func TestService_validateSubject(t *testing.T) {
...
 		tokenCtx := validContext()
+		tokenCtx.jwtBearerToken.Remove(purposeOfUseClaimDeprecated)
 		tokenCtx.jwtBearerToken.Remove(purposeOfUseClaim)
 
 		err := ctx.oauthService.validatePurposeOfUse(tokenCtx)
 
 		assert.EqualError(t, err, "no purposeOfUse given")
 	})
+
+	t.Run("error - only deprecated claim", func(t *testing.T) {
```
reinkrul, line 411:
```suggestion
	t.Run("ok - only deprecated claim", func(t *testing.T) {
```

reinkrul (changes requested):

reinkrul (approved):

PR #2214, opened 2023-05-31
helm chart improvements

henk-hofs-pink:
added support to specify an existingSecret and the secretKey in the helm chart


henk-hofs-pink:
Additionally you might want to look into being able to mount secrets as volume/file so it's not exposed as an env var inside the container/application

henk-hofs-pink:
I'm working on some additional improvements, i'll first test it locally and open a new PR

reinkrul:
> I'm working on some additional improvements, i'll first test it locally and open a new PR

👍

PR #2215, opened 2023-06-01
helm chart improvements:

henk-hofs-pink:
1.  added option to use existingSecret for redis/sentinel 
2. using releasename and namespace for pv and pvc so it can be deployed on the same cluster in different namespaces 
3. disabled strictmode in default values.yaml as tls is also disabled in the default values.yaml
4. updated chart version

### DIFF
--- a/charts/nuts-node/values.yaml
+++ b/charts/nuts-node/values.yaml
@@ -82,8 +82,18 @@ tolerations: []
 
 affinity: {}
 
+# if you want to use an existing secret for the redis password, uncomment below
+#storage:
+#  redis: # for normal redis password
+#    existingSecret: nuts-node-redis-auth
+#    existingSecretPasswordKey: redis-password
+#    sentinel: # for redis sentinel password
+#      existingSecret: nuts-node-redis-auth
+#      existingSecretPasswordKey: redis-password
+
 nuts:
   config:
+    strictmode: false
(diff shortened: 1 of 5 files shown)


reinkrul:
@beardedfoo

beardedfoo:
I'm not comfortable with TLS and strictmode being disabled by default. This inverts the behaviour of nuts-node and conflicts with the mission of "secure by default". Can we invert this behaviour in the helm chart?

Otherwise, everything looks good to me!

woutslakhorst:
@henk-hofs-pink see comment from beardedfoo

henk-hofs-pink:
@woutslakhorst  @beardedfoo  updated values to use strictmode and tls.
Had to use dummy for contractvalidators, let me know if that's acceptable.

charts/nuts-node/values.yaml
```diff
@@ -122,7 +127,7 @@ nuts:
         - ReadWriteOnce
       storageClassName: ""
     persistedVolumeClaim:
-      enabled: true
+      enabled: t`rue
```
woutslakhorst, line 130: rogue character

woutslakhorst (changes requested):
The setup with `strictmode=true` and the `dummy` validator combination will, as default, generate an error on startup. Use:
```
- irma
- uzi
- employeeid
```

woutslakhorst (approved):

PR #2230, opened 2023-06-07
OpenID4VCI: CredentialOffer.Grants was incorrectly wrapped in an array

reinkrul:


vcr/oidc4vci/types.go
```diff
@@ -84,7 +84,7 @@ type CredentialOffer struct {
 	// Credentials defines the credentials offered by the issuer to the wallet.
 	Credentials []map[string]interface{} `json:"credentials"`
 	// Grants defines the grants offered by the issuer to the wallet.
-	Grants []map[string]interface{} `json:"grants"`
+	Grants map[string]interface{} `json:"grants"`
```
gerardsn, line 87: spec says it is optional

gerardsn (approved):

vcr/oidc4vci/types.go
reinkrul: For the sake of simple Golang structure (nil checking for pointers), probably easiest to keep it non-optional here? (then it is returned as empty map)

PR #2235, opened 2023-06-09
Network: add maximum connection lifetime

### LINKED ISSUE #2018: Network: close connection if certificate is revoked while connected
Currently the node only checks the validity of a certificate during the TLS handshake.

Connections are never re-authenticated or closed as long as both nodes are online. This means that a connection could easily last for another week/month if one of the involved certificates is revoked.

gRPC adds a bit of complexity here, since it does not provide a way to close inbound connections. See #1945 for a possible direction for this issue.

gerardsn:
closes #2018 
Closing a connection once the used certificate is revoked/deny listed as proposed in the issue adds a lot of complexity. 

The current solution sets a maximum lifetime before the connection will be closed on the server side (15min) and client side (20 min). Since the network was designed to have connections open all the time, there are some race conditions in the disconnect vs sending of messages. The server side has a graceful closure using gRPC to wrap-up any ongoing communication resulting in less issues. We need to test how big of an issue this really is.

Also added some integration tests

### DIFF
--- a/network/transport/grpc/test.go
+++ b/network/transport/grpc/test.go
@@ -161,8 +161,8 @@ func SetPeerID(t *testing.T, manager transport.ConnectionManager, id transport.P
 	cm.config.peerID = id
 }
 
-// SetServerKeepaliveParams sets the MaxConnectionAge to 1 sec and MaxConnectionAgeGrace to 1 nSec on the keepalive.ServerParameters and restores the default during cleanup.
-func SetServerKeepaliveParams(t *testing.T) {
+// SetServerKeepaliveParamsForTest sets the MaxConnectionAge to 1 sec and MaxConnectionAgeGrace to 1 nSec on the keepalive.ServerParameters and restores the default during cleanup.
+func SetServerKeepaliveParamsForTest(t *testing.T) {
 	defaultAge := serverKeepaliveParams
 	serverKeepaliveParams = keepalive.ServerParameters{
 		MaxConnectionAge:      time.Second,
@@ -173,8 +173,8 @@ func SetServerKeepaliveParams(t *testing.T) {
 	})
 }
 
-// SetClientKeepaliveParams sets the clientMaxConnectionAge to 1 sec and restores the default during cleanup.
-func SetClientKeepaliveParams(t *testing.T) {
+// SetClientKeepaliveParamsForTest sets the clientMaxConnectionAge to 1 sec and restores the default during cleanup.
+func SetClientKeepaliveParamsForTest(t *testing.T) {
 	defaultAge := clientMaxConnectionAge
 	clientMaxConnectionAge = time.Second
 	t.Cleanup(func() {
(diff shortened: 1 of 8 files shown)


beardedfoo (approved):

network/transport/grpc/test.go
```diff
@@ -158,3 +160,24 @@ func SetPeerID(t *testing.T, manager transport.ConnectionManager, id transport.P
 	}
 	cm.config.peerID = id
 }
+
+// SetServerKeepaliveParams sets the MaxConnectionAge to 1 sec and MaxConnectionAgeGrace to 1 nSec on the keepalive.ServerParameters and restores the default during cleanup.
+func SetServerKeepaliveParams(t *testing.T) {
```
reinkrul, line 165: put something like "Test" in the name

network/transport/grpc/test.go
```diff
@@ -158,3 +160,24 @@ func SetPeerID(t *testing.T, manager transport.ConnectionManager, id transport.P
...
+		MaxConnectionAge:      time.Second,
+		MaxConnectionAgeGrace: time.Nanosecond,
+	}
+	t.Cleanup(func() {
+		serverKeepaliveParams = defaultAge
+	})
+}
+
+// SetClientKeepaliveParams sets the clientMaxConnectionAge to 1 sec and restores the default during cleanup.
+func SetClientKeepaliveParams(t *testing.T) {
```
reinkrul, line 177: here as well

network/transport/grpc/tls_offloading.go
```diff
@@ -49,17 +49,23 @@ type tlsOffloadingAuthenticator struct {
 }
 
 func (t *tlsOffloadingAuthenticator) intercept(srv interface{}, serverStream grpc.ServerStream, _ *grpc.StreamServerInfo, handler grpc.StreamHandler) error {
+	// Det certificate from header
```
reinkrul, line 52: more like a comment for the type itself

reinkrul (approved):
Try to keep the denylist improvements in a different PR next time I think.

Anything to consider when behind a proxy? Does it cause error logging when disconnecting?

network/transport/grpc/tls_offloading.go
gerardsn: The name of this thing is weird to me. It does not authenticate anything, it just extracts the certificate from the header and sets in on the grpc.Peer.

gerardsn:
> Try to keep the denylist improvements in a different PR next time I think.
Sorry, the change was part of an integration test, but then wasn't.

> Anything to consider when behind a proxy? 
Not that I know of.

> Does it cause error logging when disconnecting?
The disconnect itself does not. If the disconnect is mid-conversation there are some errors. I synced a new node after submitting this PR and the errors eventually resulted in bbolt being permanently locked for me, only a restart could fix it. This part needs some work before merging. Any suggestions?

gerardsn:
obsoleted by #2238

PR #2236, opened 2023-06-09
Docs: add proxy_buffering and client_max_body_size to example NGINX config

reinkrul:
Otherwise, bootstrapping nodes are disconnected (gRPC will report `RST_STREAM` error) due to (too) large pages being transmitted.

### DIFF
--- a/docs/pages/deployment/tls-configuration.rst
+++ b/docs/pages/deployment/tls-configuration.rst
@@ -84,7 +84,8 @@ For `NGINX <https://www.nginx.com/>`_ the proxy configuration could look as foll
           ssl_verify_depth          1;
 
           location / {
-            # for large messages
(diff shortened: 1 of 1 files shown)


gerardsn (commented):
The note below this example should then be removed?

I imagine we are als hitting this limit on bootstrap nodes: https://nginx.org/en/docs/http/ngx_http_upstream_module.html#keepalive_requests

woutslakhorst:
It would be indeed interesting to test these limits

gerardsn:
Either gRPC or the connection manager will reconnect. 

I think every message sent is a request, so each chunk in a range request (transactionList msg) probably counts as an individual request.

However, even if the nodes are in sync, the gossip and diagnostics broadcasts produce 720 requests/hour each with the default settings. This should result in a reconnect every ±42min?

woutslakhorst (changes requested):
The note below says the same

woutslakhorst (approved):

PR #2238, opened 2023-06-12
Network: re-validate peers after denylist update

### LINKED ISSUE #2018: Network: close connection if certificate is revoked while connected

gerardsn:
closes #2018  alternative solution to #2235 

DenyList.Subscribers get pinged when the Denylist is updated (revalidation is not performed when Denylist is not configured). What to do with this information is determined by the subscriber. The connection manager verifies the certificate of all peers and disconnects those that are no longer valid.

diagnostics also show the last time the certificate was updated, which is zero if the denylist is not used
```
network:
    connections:
        certificates_last_validated: 0001-01-01T00:00:00Z
        connected_peers:
            - id: ...
```

### DIFF
--- a/network/transport/grpc/connection.go
+++ b/network/transport/grpc/connection.go
@@ -221,13 +221,10 @@ func (mc *conn) startReceiving(protocol Protocol, stream Stream) {
 		defer atomic.AddInt32(activeGoroutines, -1)
 		for {
 			message := protocol.CreateEnvelope()
-			err := stream.RecvMsg(message)
-			select {
-			case <-mc.ctx.Done():
-				// disconnect: drop message and stop receiving
+			err := stream.RecvMsg(message) // blocking
+			if mc.ctx.Err() != nil {
+				// connection has been closed: drop message and stop receiving
 				return
-			default:
-				//non-blocking
 			}
 			if err != nil {
 				errStatus, isStatusError := status.FromError(err)
--- a/network/transport/grpc/connection.go
+++ b/network/transport/grpc/connection.go
@@ -81,9 +81,6 @@ type Connection interface {
 	// IsConnected returns whether the connection is active or not.
 	IsConnected() bool
 
-	// IsProtocolConnected returns whether the given protocol is active on the connection.
-	IsProtocolConnected(protocol Protocol) bool
-
 	// IsAuthenticated returns whether teh given connection is authenticated.
 	IsAuthenticated() bool
 
@@ -152,9 +149,8 @@ func (mc *conn) waitUntilDisconnected() {
 		mc.mux.RUnlock()
 		return
 	}
-	done := mc.ctx.Done()
 	mc.mux.RUnlock()
-	<-done
+	<-mc.ctx.Done()
 }
 
 func (mc *conn) verifyOrSetPeerID(id transport.PeerID) bool {
@@ -210,10 +206,10 @@ func (mc *conn) registerStream(protocol Protocol, stream Stream) bool {
 	mc.startSending(protocol, stream)
 
 	// A connection can have multiple active streams, but if one of them is closed, all of them should be closed, also closing the underlying connection.
-	go func(cancel func()) {
+	go func() {
 		<-stream.Context().Done()
-		cancel()
-	}(mc.cancelCtx)
+		mc.cancelCtx()
+	}()
 
 	return true
 }
@@ -221,11 +217,18 @@ func (mc *conn) registerStream(protocol Protocol, stream Stream) bool {
 func (mc *conn) startReceiving(protocol Protocol, stream Stream) {
 	peer := mc.Peer() // copy Peer, because it will be nil when logging after disconnecting.
 	atomic.AddInt32(&mc.activeGoroutines, 1)
-	go func(activeGoroutines *int32, cancel func()) {
+	go func(activeGoroutines *int32) {
 		defer atomic.AddInt32(activeGoroutines, -1)
 		for {
 			message := protocol.CreateEnvelope()
 			err := stream.RecvMsg(message)
+			select {
+			case <-mc.ctx.Done():
+				// disconnect: drop message and stop receiving
+				return
+			default:
+				//non-blocking
+			}
 			if err != nil {
 				errStatus, isStatusError := status.FromError(err)
 				if errors.Is(err, io.EOF) || (isStatusError && errStatus.Code() == codes.Canceled) {
@@ -241,7 +244,7 @@ func (mc *conn) startReceiving(protocol Protocol, stream Stream) {
 						Warn("Peer connection error")
 				}
 				mc.status.Store(errStatus)
-				cancel()
+				mc.cancelCtx()
 				break
 			}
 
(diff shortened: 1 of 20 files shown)


pki/interface.go
```diff
@@ -31,8 +32,31 @@ var (
...
+	// Update fetches a new copy of the denylist
+	Update() error
+
+	// URL returns the URL of the denylist
+	URL() string
+
+	// ValidateCert returns an error if a certificate should not be used
+	ValidateCert(cert *x509.Certificate) error
+
+	// Subscribe calls all subscribers when the denylist is updated
```
woutslakhorst, line 56: I don't think the comment is correct?

pki/test.go
```diff
@@ -0,0 +1,32 @@
+package pki
```
woutslakhorst, line 1: copyright

pki/test.go
```diff
@@ -0,0 +1,32 @@
...
+
+import (
+	"crypto/x509"
+	"testing"
+	"time"
+)
+
+// TestDenylistWithCert sets a new Denylist on the Validator and adds the certificate.
+// This is useful in integrations tests etc.
+func TestDenylistWithCert(t *testing.T, val Validator, cert *x509.Certificate) {
```
woutslakhorst, line 11: Change the `Test` prefix, I thought it was a test, but it's a helper

network/transport/grpc/connection_test.go
```diff
@@ -139,8 +123,8 @@ func Test_conn_startSending(t *testing.T) {
 			return atomic.LoadInt32(&connection.activeGoroutines) == 0, nil
 		}, 5*time.Second, "waiting for all goroutines to exit")
 
-		// err status is set on connection. Due to EOF it's an unknown error
-		assert.Equal(t, codes.Unknown, connection.status.Load().Code())
+		// Last received message is dropped and no status is set. Default value is OK.
```
woutslakhorst, line 126: This is a bit weird, status is now the same as being operational. Shouldn't it be a disconnected status somehow?

network/transport/grpc/stats.go
```diff
@@ -92,6 +93,25 @@ func (o ownPeerIDStatistic) String() string {
 	return o.peerID.String()
 }
 
+// lastCertificateValidation contains the Unix timestamp of the most recent certificate validation of all peers.
+type lastCertificateValidation struct {
```
woutslakhorst, line 97: It's a statistic, maybe the name should reflect it

woutslakhorst (changes requested):

pki/interface.go
gerardsn: now
> Subscribe registers a callback that is triggered everytime the denylist is updated

pki/test.go
gerardsn: now `SetNewDenylistWithCert`

network/transport/grpc/stats.go
gerardsn: now `lastCertificateValidationStatistic`

network/transport/grpc/connection_test.go
gerardsn: I don't know what the right solution is here.

connection.status is a grpc Status that reflects the state of the last message received. (it is a bit of a hack for the backoff) no-status is considered equal to `OK` (this situation I think)

When the connection's context is cancelled it means that the receiving side closed the connection. `conn.startReceiving` ignores any future message (including any possible errors) in this case and the status of the last processed incoming message was still ok. 

Also, `Unauthenticated` is the only status we act on and does not apply here, all other statuses are handled the same as the current situation.

gerardsn: This test is weird anyway. It supposedly tests `conn.startSending`, but the only thing it tests is error handling of `conn.startReceiving` after disconnect with a stub stream that simulates stream closure with an EOF message

woutslakhorst (approved):

network/transport/grpc/connection.go
```diff
@@ -210,22 +206,29 @@ func (mc *conn) registerStream(protocol Protocol, stream Stream) bool {
...
+	go func(activeGoroutines *int32) {
 		defer atomic.AddInt32(activeGoroutines, -1)
 		for {
 			message := protocol.CreateEnvelope()
 			err := stream.RecvMsg(message)
+			select {
+			case <-mc.ctx.Done():
+				// disconnect: drop message and stop receiving
+				return
+			default:
```
reinkrul, line 229: interesting pattern, but note that `RecvMsg` is blocking anyway, so you could just check `ctx.Err() != nil`?

network/transport/grpc/connection_manager.go
```diff
@@ -544,6 +542,26 @@ func (s *grpcConnectionManager) authenticate(nodeDID did.DID, peer transport.Pee
...
 
+// revalidatePeers verifies for all peers the x509.Certificate provided during TLS handshake is still valid.
+func (s *grpcConnectionManager) revalidatePeers() {
+	var err error
+	now := nowFunc()
+	s.lastCertificateValidation.Store(&now)
+	s.connections.forEach(func(conn Connection) {
+		peerCert := conn.Peer().Certificate
+		if nowFunc().After(peerCert.NotAfter) {
+			log.Logger().WithError(errors.New("certificate expired while in use")).WithFields(conn.Peer().ToFields()).Info("Disconnected peer")
```
reinkrul, line 553: why check this? This will be checked by the Golang's TLS when setting up the connection anyway?

pki/denylist.go
```diff
@@ -224,10 +203,19 @@ func (b *denylistImpl) Update() error {
 	// Log when the denylist is updated
 	logger().Debug("Denylist updated successfully")
 
+	// Notify all subscribers synchronously
+	for _, sub := range b.subscribers {
+		sub()
```
reinkrul, line 208: can `conn.disconnect()` block when the other side is a slow responder (closing TCP involves sending packets to thw other side)? Since subscribers are called synchronously?

pki/interface.go
```diff
@@ -48,6 +72,10 @@ type Validator interface {
 	// AddTruststore adds all CAs to the truststore for validation of CRL signatures. It also adds all CRL Distribution Endpoints found in the chain.
 	// CRL Distribution Points encountered during operation, such as on end user certificates, are only added to the monitored CRLs if their issuer is in the truststore.
 	AddTruststore(chain []*x509.Certificate) error
+
+	// SubscribeDenied registers a callback that is triggered everytime the denylist is updated.
+	// This can be used to revalidate all certificates on long-lasting connections by calling Validate on them again.
+	SubscribeDenied(f func())
```
reinkrul, line 78: `SubscribeUpdated`?

reinkrul (changes requested):

network/transport/grpc/connection_manager.go
gerardsn: because we're re-validating the peer's certificate on a connection that may last for a month? The only things that could have changed since it was first accepted is that it is banned or expired.

pki/interface.go
gerardsn: You only subscribe to updates on the denylist, so I wanted to have something indicating that in the name.

reinkrul (approved):

Issue #2112, opened 2023-05-08
Revisit actions required to enable OpenID4VCI


reinkrul:
Document what a vendor needs to do to start using (receiving, issuing) VCs over OpenID4VCI. Should be as little as possible.

A vendor currently needs to register a wallet metadata URL on each of its customer's DID documents, which is going to be very cumbersome when a vendor has more than a handful of customers. They also can't be updated transactionally, so keeping it synchronized is going to be tricky.

Therefore, we might want to introduce some conventions, e.g. when a DID document does not have a wallet metadata URL, look at its controllers for a wallet metadata base URL and inject/append the customer's DID into it to construct it (the wallet metadata URL).

reinkrul:
Discussed with @gerardsn:

- We want parties to start using OpenID4VCI as soon as possible, with the fallback to issuing over the network. Otherwise, we won't know if actually works and subsequently don't know if we can deprecate/remove the private TX fallback.
- Thus, enabling it should be as simple and easy as possible. Having to register additional DID services is probably too big of a hurdle.
- But, we want a DID service for discovery of OpenID4VCI issuers and wallets. Probably a metadata URL which can be used as entry point for discovering additional OAuth2 services. Maybe in the form of a well-known endpoint.
- The current OAuth2 Access Token flow in the Auth module should be migrated (or merged) with the new OpenID4VCI implementation, ultimately (optionally?) removing/replacing the current `oauth` service registered in the eOverdracht compound services (and others).
- The new DID service endpoint would be almost the same (at least the base URL) as the existing `oauth` service endpoint, but routing to `/n2n/identity` instead of `/n2n/auth/...`, thus could be derived. But it could differ (in some unforeseen situation) and might not be reachable due to strict reverse proxy config (proxying only the existing `/n2n` URL).

Suggestion:
- Introduce a new DID service endpoint for discovery of OAuth2/OpenID4VCI services (to be specified)
- At startup the new DID service is auto-registered, if OpenID4VCI is enabled:
  - Try to resolve the new DID service on the Nuts node's DID documents
  - If not present, derive it from the `oauth` service. If it is reachable (self-dial), register the DID service
  - If it can't be derived or isn't reachable, log an error (and health check should indicate OpenID4VCI isn't operational)

reinkrul:
We could check stable/acc/prd to find common patterns (e.g. `:8443`).

reinkrul:
Registry Admin Demo needs to be able to register the metadata URL.

[...]

PR #2239, opened 2023-06-13
OpenID4VCI: Golden Hammer for fixing services in DID services 

reinkrul:
for OpenID4VCI support

Requires https://github.com/nuts-foundation/nuts-node/pull/2240
Fixes https://github.com/nuts-foundation/nuts-node/issues/2112


reinkrul:
@woutslakhorst @gerardsn a sneak peek for your enjoyment (need to add some more test coverage).

woutslakhorst (commented):
Naming it 'golden hammer' is really Nuts...

reinkrul:
> Naming it 'golden hammer' is really Nuts...

Keeping true to our core values.

woutslakhorst (commented):
Is it possible to migrate without using NodeDID?

reinkrul:
> Is it possible to migrate without using NodeDID?

No, since the service is to be registered on the vendor's DID document (thus it needs a node DID). As issuer, you could issue without node DID since the credential offer contains your identifier, which can be used to lookup your metadata (which can be resolved runtime using the TLS certificate).

But the issuer needs the credential offer endpoint of the wallet, which is specified by the wallet metadata. But the issuer needs to know the URL where to find the metadata.

We _could_ just ignore node DID altogether and just register the service base URL on any owned DID (meaning of which the private keys are present), that would work for discovering wallets. Then we need to decide whether we want to register the URL on the care organization documents, or just a reference to the vendor's DID document (for easier administration later on). Determining the vendor's DID document is done by following the NutsComm service reference, which should be configured as node DID. But still, we could go by without it:

1. Make a list of owned DIDs
2. Filter documents without `node-http-services-base-url` service
3. Sort: nodes without `NutsComm` service go first (so "vendor DID documents" go first).
4. Try to resolve metadata (`TLSIdentifierResolver`), if successful, continue
5. No `NutsComm` service reference? Register `node-http-services-base-url` service it on the DID document itself.
6. Has `NutsComm` service reference? Register `node-http-services-base-url` reference to ref'd DID document, if present there.

You could also just ignore `NutsComm` here and always register the complete URL, but that could cause administrative issues when the endpoint changes.

An edge case (race condition) in which it sort of breaks is when the `NutsComm` reference is registered too late on the care organization, then the process above might've already registered an absolute URL on a care organization DID document instead of a reference to the vendor's. This might lead to issues when the vendor changes its URL. But we might want to add a health check that verifies the availability of a node's OpenID4VCI endpoints.

reinkrul:
@woutslakhorst I removed usage of node DID

golden_hammer/module.go
```diff
@@ -0,0 +1,275 @@
...
+			endpointToRegister = *identifier
+		}
+		_, err := h.didmanAPI.AddEndpoint(h.ctx, document.ID, types.BaseURLServiceType, endpointToRegister)
+		if err != nil {
+			log.Logger().WithError(err).
+				Errorf("Unable to register DID services base URL (did=%s): %s", document.ID, endpointToRegister.String())
+		} else {
+			h.fixedDocumentDIDs[document.ID.String()] = true
+			log.Logger().Infof("Registered DIDs services base URL (did=%s): %s", document.ID, endpointToRegister.String())
+		}
```
gerardsn, line 173: log mania on first run for vendors with many DIDs. didman also logs twice (1 debug, 1 info) that it added an endpoint.

gerardsn (changes requested):
We should probable set a deadline for when this thing is removed?

Just some minor things. Points below might not be worth fixing.
1. The DIDs are found based on owned private keys, but I don't think the key is ever logged. This could result in hard to debug errors. I don't see an easy fix that doesn't spam the logs though.
2. the `cachingDocumentOwner` stores that an identity does not exist on this node if requested before `hammerTime` is over? Since caching is in memory only it is currently fixed by a restart.

golden_hammer/module.go
```diff
@@ -0,0 +1,275 @@
...
+		didmanAPI:         didmanAPI,
+		documentOwner:     documentOwner,
+		docResolver:       docResolver,
+		fixedDocumentDIDs: map[string]bool{},
+	}
+}
+
+// GoldenHammer is a module that fixes a node's DID configuration.
+// Its name is intentionally weird, since the module should not exist.
+// In future all fixes it does should be deprecated and removed after they've become obsolete.
```
woutslakhorst, line 58: Document which version this is to be introduced and which to be removed.

golden_hammer/module.go
```diff
@@ -0,0 +1,275 @@
...
+			}
+		}
+	}
+}
+
+// registerServiceBaseURLs registers the node's services HTTP base URL on its DIDs.
+// This base URL is used to discover HTTP services (for now only OpenID4VCI wallet/issuer metadata).
+//   - Make a list of owned DIDs
+//   - Filter documents without node-http-services-base-url service
+//   - Sort: nodes without NutsComm service go first (so "vendor DID documents" go first).
```
woutslakhorst, line 128: Vendors also have a `NutsComm` service

golden_hammer/module.go
```diff
@@ -0,0 +1,275 @@
...
+		if containsService(*document, types.BaseURLServiceType) {
+			h.fixedDocumentDIDs[id.String()] = true
+			continue
+		}
+		// This document needs fixing
+		documents = append(documents, *document)
+	}
+	// Sort: since care organization DIDs refer to vendor DIDs through NutsComm service,
+	// vendor DIDs should be fixed first. Meaning DIDs without NutsComm service should go first.
+	sort.Slice(documents, func(i, j int) bool {
```
woutslakhorst, line 205: again, vendors have a NutsComm service. Only diff is that they have a concrete endpoint not a ref.

golden_hammer/module.go
woutslakhorst: and an audit log as well? So debug is enough

woutslakhorst (changes requested):
Just a thought: if one owned DID succeeds, they'll probably all succeed?

docs/pages/deployment/verifiable-credentials.rst
```diff
@@ -1,10 +1,51 @@
-.. _custom-credentials-configuration:
+.. _verifiable-credentials-configuration:
+
+Verifiable Credentials
+######################
+
+One of the most important features of the Nuts node is handling Verifiable Credentials.
+This chapter describes the various configuration aspects of credentials in the Nuts node.
+
+Issuing and receiving over OpenID4VCI
```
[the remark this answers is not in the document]
reinkrul: I don't think end-users are interested in the RFCs (only if implementing them theirselves)

golden_hammer/module.go
reinkrul: Will just add a single log message when it has fixed anything

reinkrul:
> Just a thought: if one owned DID succeeds, they'll probably all succeed?

Most probably

golden_hammer/module.go
```diff
@@ -0,0 +1,275 @@
...
+	}
+	var endpoint string
+	_ = service.UnmarshalServiceEndpoint(&endpoint)
+	return endpoint
+}
+
+func containsService(document did.Document, serviceType string) bool {
+	service := getService(document, serviceType)
+	return service != nil
+}
```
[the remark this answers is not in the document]
reinkrul: There's a similar function `ByServiceType` (matcher) which has same behavior, but using a matcher-type function is kinda weird here (although it works) code-wise, and `ByServiceType` is not explicit about matching references or not (it's just a comment), so prone to bugs in future, when it changes.

Moving it there doesn't make much sense either, since we plan to remove this module as soon as it isn't required any more, making the util funcs moved to `didservice` dead code.

golden_hammer/module.go
```diff
@@ -0,0 +1,275 @@
...
+			continue
+		}
+		// This document needs fixing
+		documents = append(documents, *document)
+	}
+	// Sort: since care organization DIDs refer to vendor DIDs through NutsComm service,
+	// vendor DIDs should be fixed first. Meaning DIDs without NutsComm service should go first.
+	sort.Slice(documents, func(i, j int) bool {
+		return containsService(documents[i], transport.NutsCommServiceType)
+	})
```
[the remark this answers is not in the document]
reinkrul: fixed

golden_hammer/module.go
reinkrul: fixed

golden_hammer/module.go
reinkrul: fixed

reinkrul:
> The DIDs are found based on owned private keys, but I don't think the key is ever logged. This could result in hard to debug errors. I don't see an easy fix that doesn't spam the logs though.
Should be done by inspecting the DID document itself, making sure there's an `assertionMethod` key available. Which by default, it always there. If there's log spam because a party has their own more strict vendor/org DID key setup, we could add functionality to check the existence of specific keys instead of just any DID's private key.
> the cachingDocumentOwner stores that an identity does not exist on this node if requested before hammerTime is over? Since caching is in memory only it is currently fixed by a restart.
Not sure how this can happen, since `hammerTime()` both (ListOwned and IsOwned) work on private keys present, just that `IsOwned` is cached. Maybe in a weird Vault hickup. We could always add a TTL to cache entries if it proves to be an issue. But if there's a hickup in Vault, people might be restarting their Nuts node anyways..?

gerardsn (approved):
Hammer time!

woutslakhorst (approved):

PR #2240, opened 2023-06-13
OpenID4VCI: DID service with base URL for wallet/issuer discovery

reinkrul:
This PR:
- Replaces the wallet metadata DID URL service with an HTTP services base URL, using which both wallet and issuer can be discovered
- It removes the `vcr.oidc4vci.url` property, which is now resolved
- The URL is now resolved using the discovery URL (point 1), or fallback to trying to derive it from the TLS certificate if not present. This allows an issuer to issue over OpenID4VCI without registering the DID service.
- It resolves the metadata using a HTTP HEAD operation, which only returns the HTTP headers as if a GET request was performed. This way, the metadata doesn't need to be actually served (which would lead to a chicken-egg problem).

Part of https://github.com/nuts-foundation/nuts-node/issues/2112

TODO:
- [ ] Caching of DID node HTTP service base URLs (implement caching layer for ServiceResolver?)
- [x] What to do about refusing to serve over plain HTTP? The auto-fixer won't support HTTP (just HTTPS), so might not be an actual problem. Clients should refuse to talk to wallets/issuers over HTTP.


reinkrul:
Made draft again, extending local OpenID4VCI identifier with resolving it using the TLS certificate

woutslakhorst:
Last checkbox is covered/addressed by #2290

vcr/api/oidc4vci/v0/api.go
```diff
@@ -119,18 +115,46 @@ func (w Wrapper) Routes(router core.EchoRouter) {
 	}))
 }
 
+func (w Wrapper) getWallet(ctx context.Context, holderDIDAsString string) (holder.OIDCWallet, error) {
```
woutslakhorst, line 118: method is used in other file so expect it to be there. 
`holderDIDAsString string` is a bit overkill? `holder string` is enough info for anybody to know they should pass the string format?
The method name is non-descriptive. Also due to the `holder.OIDCWallet` naming. Would `getHandler(..) (holder.OpenIDHandler, error)` be more clear?

vcr/api/oidc4vci/v0/api.go
```diff
@@ -119,18 +115,46 @@ func (w Wrapper) Routes(router core.EchoRouter) {
...
+	holderDID, err := w.validateDIDIsOwned(ctx, holderDIDAsString)
+	if err != nil {
+		return nil, err
+	}
+	return w.VCR.GetOIDCWallet(ctx, holderDID)
+}
+
+func (w Wrapper) getIssuer(ctx context.Context, issuerAsString string) (openid4vci.Issuer, error) {
```
woutslakhorst, line 126: same as above, this is the IssuerHandler?

vcr/issuer/issuer.go
```diff
@@ -47,36 +47,42 @@ import (
...
+		trustConfig:   trustConfig,
+		vcrStore:      vcrStore,
 	}
 }
 
 type issuer struct {
 	store            Store
 	networkPublisher Publisher
-	oidcIssuer       openid4vci.Issuer
+	openidIssuerFn   func(ctx context.Context, id did.DID) (openid4vci.Issuer, error)
```
woutslakhorst, line 78: Code smell, this might be hiding some spaghetti-like construction.
The referenced method uses the `walletResolver` and the `DocumentOwner`. You already have the walletResolver, only need the DocumentOwner service. Then there's no need to back-reference to the VCR.

vcr/issuer/issuer.go
```diff
@@ -119,19 +125,23 @@ func (i issuer) Issue(ctx context.Context, credentialOptions vc.VerifiableCreden
...
 					Info("Published credential over OIDC4VCI")
 				return createdVC, nil
-			} else if !errors.Is(err, vdr.ErrServiceNotFound) {
-				// An error occurred, but it's not because the DID doesn't support OIDC4VCI. Fallback to publishing over Nuts network.
+			} else {
 				log.Logger().
 					WithField(core.LogFieldCredentialID, createdVC.ID.String()).
-					WithError(err).
-					Warnf("Could publish credential over OIDC4VCI, fallback to publish over Nuts network")
+					Info("Wallet or issuer does not support OpenID4VCI, fallback to publish over Nuts network")
```
woutslakhorst, line 144: This should only happen if the wallet doesn't support it?

vcr/oidc4vci/service/identifiers.go
```diff
@@ -0,0 +1,164 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *
+ */
+
+package service
```
woutslakhorst, line 19: rather generic package name.
Can't it be placed one lever higher, under `oidc4vci`?

vcr/oidc4vci/service/identifiers.go
```diff
@@ -0,0 +1,164 @@
...
+
+// IdentifierResolver defines the interface for resolving OpenID4VCI identifiers (of wallet and issuer).
+// The identifier is the base URL of the issuer or wallet, at which well-known endpoints can be found.
+type IdentifierResolver interface {
+	Resolve(id did.DID) (string, error)
+}
+
+var _ IdentifierResolver = DIDIdentifierResolver{}
+
+// DIDIdentifierResolver is a IdentifierResolver that uses resolves identifiers from DID documents.
```
woutslakhorst, line 44: -uses

vcr/oidc4vci/service/identifiers.go
```diff
@@ -0,0 +1,164 @@
...
+type tlsIdentifierResolver struct {
+	underlying       IdentifierResolver
+	config           *tls.Config
+	cachedIdentifier *atomic.Pointer[string]
+	// lastAttempt is the time at which the last attempt to resolve the identifier from the TLS certificate was made.
+	// It is used to prevent spamming the local node, since it could be called on each OpenID4VCI request.
+	lastAttempt *atomic.Pointer[time.Time]
+}
+
+func (t tlsIdentifierResolver) Resolve(id did.DID) (string, error) {
```
woutslakhorst, line 96: Weird caching. The primary resolve loads a DID Document from database which is a heavier operation that checking the certificate data which is already in-memory?

vcr/oidc4vci/service/identifiers.go
```diff
@@ -0,0 +1,164 @@
...
+var tlsIdentifierResolverPort = 443
+
+func SetTLSIdentifierResolverPort(port int) {
+	tlsIdentifierResolverPort = port
+}
+
+var _ IdentifierResolver = tlsIdentifierResolver{}
+
+type tlsIdentifierResolver struct {
+	underlying       IdentifierResolver
```
woutslakhorst, line 88: This pass-through resolver will become obsolete when all DID Documents are migrated?

vcr/oidc4vci/service/identifiers.go
```diff
@@ -0,0 +1,164 @@
...
+	candidateHosts := append(t.config.Certificates[0].Leaf.DNSNames, t.config.Certificates[0].Leaf.Subject.CommonName)
+	for _, host := range candidateHosts {
+		candidateURLs = append(candidateURLs, fmt.Sprintf("https://%s:%d", host, tlsIdentifierResolverPort))
+	}
+
+	// Resolve URLs
+	httpTransport := http.DefaultTransport.(*http.Transport).Clone()
+	httpTransport.TLSClientConfig = t.config
+	httpClient := &http.Client{
+		Timeout:   2 * time.Second,
```
woutslakhorst, line 135: extremely short, if you miss a single `SYN` package, timeout is triggered. Recommend 5

vcr/oidc4vci/service/identifiers_test.go
```diff
@@ -0,0 +1,185 @@
...
+			}
+			w.Header().Set("Content-Type", "application/json")
+			if strings.Contains(r.URL.Path, "/.well-known/openid-credential-issuer") {
+				w.WriteHeader(http.StatusOK)
+			} else {
+				w.WriteHeader(http.StatusNotFound)
+			}
+		}))
+		httpServer.TLS = tlsConfig.Clone()
+		httpServer.StartTLS()
```
woutslakhorst, line 137: Does the server need TLS?

woutslakhorst (changes requested):
As a general remark, the oidc4vci packages are scattered and the naming doesn't really imply what you're using. We might have to refactor some thing to get everything in a logical place when phase 1 issues have been resolved?

vcr/oidc4vci/service/identifiers_test.go
reinkrul: We decided to only check port 443, so I assumed we only resolve identifier on HTTPS endpoints.

vcr/oidc4vci/service/identifiers.go
reinkrul: I wanted to put the rest under a new `client` package but didn't want to do that in this PR (to cater a sort of generic OpenID4VCI library);

- `openid4vci/client`
- `openid4vci/services`

vcr/issuer/issuer.go
reinkrul: Or when the issuer doesn't support it, meaning it didn't configure the base URL in its DID document, and it can't be resolved using the TLS certificate.

vcr/oidc4vci/service/identifiers.go
reinkrul: Now I think of it, `TLSIdentifierResolver` shouldn't ideally look into the DID document, but just try the TLS certificate. And then we introduce a `LocalIdentifierResolver` which tries `DIDIdentifierResolver` first, then `TLSIdentifierResolver` if it fails. But that might be a bit too overengineerd (thus, leave it like this)?

vcr/api/oidc4vci/v0/api.go
reinkrul: I agree `OIDCWallet` is a lame term, `Wallet` would be nicer to read. `Handler` or `Wallet`; i.m.o. a holder has a wallet, thus a `holder.Wallet` type sounds logical. But the wallet is OpenID4VCI specific. So then `OpenIDHandler` might be not so bad after all (and the `holder` can have a more high-level `Wallet` type in future.

gerardsn (dismissed):
I struggle to properly review PRs like this, not sure if its the openid4vci part or our code

vcr/api/oidc4vci/v0/api.go
reinkrul: Fixed

e2e-tests/README.md
```diff
@@ -32,14 +32,14 @@ When there are multiple in a test, some will use `IMAGE_NODE_A` and the other `I
...
 ./run-tests.sh
 ```
 
-And than run again with the values swapped:
+And then run again with the values swapped:
 ```console
-$ IMAGE_NODE_A=nutsfoundation/nuts-node:v5.0.0 \
+IMAGE_NODE_A=nutsfoundation/nuts-node:v5.0.0 \
 IMAGE_NODE_B=nutsfoundation/nuts-node:v4.3.0 \
 ./run-tests.sh
```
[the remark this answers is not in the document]
reinkrul: How does that emphasize that? Makes copy/pasting harder imo?

vdr/types/common.go
```diff
@@ -53,6 +53,10 @@ var ErrServiceNotFound = errors.New("service not found in DID Document")
 // ErrServiceReferenceToDeep is returned when a service reference is chain is nested too deeply.
 var ErrServiceReferenceToDeep = errors.New("service references are nested to deeply before resolving to a non-reference")
 
+// BaseURLServiceType is type of the DID service which holds the base URL of the node's HTTP services,
+// exposed to other Nuts nodes. E.g. OpenID4VCI or OAuth2 endpoints.
+const BaseURLServiceType = "node-http-services-baseurl"
```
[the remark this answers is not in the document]
reinkrul: Generic

e2e-tests/README.md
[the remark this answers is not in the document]
gerardsn: I'm not the right person to ask, just commenting on why it was there. For some the brain automatically strips a `\` character at the end, but not a `$` at the beginning I guess...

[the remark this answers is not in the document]
woutslakhorst: I have trouble with those prompts as well. Makes it harder to copy-paste

woutslakhorst (approved):
Typo in vcr/issuer/issuer.go#NewIssuer comment

PR #2253, opened 2023-06-15
Migrate e2e tests to nuts-node repo

reinkrul:
Pair it with the source code, so that they always match (add/change e2e test and nuts-node code in 1 commit). No need for a separate repo.


nuts.yaml
gerardsn, line 1: file should not be here?

.gitignore
```diff
@@ -28,4 +28,4 @@ data
 /*.db
 
 /nuts-node
-nuts.yaml
+./nuts.yaml
```
gerardsn, line 31: if the goal is to include nuts.yaml in the e2e directory is not better to add `!nuts.yaml` in the e2e gitignore? (not sure if that actually works)

gerardsn (changes requested):
Are there any changes to files in the e2e-test directory? reviewing 15k lines is a bit much if nothing has changed.

---
`make fix-copyright` probably does something here

.gitignore
reinkrul: I'll test it

reinkrul:
> Are there any changes to files in the e2e-test directory? reviewing 15k lines is a bit much if nothing has changed.
> 
> `make fix-copyright` probably does something here

IRMA wasn't actually used, so I removed the files.

e2e-tests/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+And than run again with the values swapped:
+```console
+$ IMAGE_NODE_A=nutsfoundation/nuts-node:v5.0.0 \
+IMAGE_NODE_B=nutsfoundation/nuts-node:v4.3.0 \
+./run-tests.sh
+```
+
+> **Note**  
+> The tests are updated to the latest node version, they might break when you use an older node version.
```
gerardsn, line 49: can be removed

e2e-tests/README.md
```diff
@@ -0,0 +1,49 @@
+This repository contains end-to-end tests for the [nuts-node](https://github.com/nuts-foundation/nuts-node).
```
gerardsn, line 1: can be removed

gerardsn, line 1:
```suggestion
```

e2e-tests/README.md
gerardsn, line 49:
```suggestion
```

gerardsn (approved):

PR #2261, opened 2023-06-19
replace direct String return with specific error writer

### LINKED ISSUE #1974: HTTP: Authz failure leads to response committed warnings in log
When authentication using token_v2 fails, it logs the authz error (which is expected):

```
stable-stable-nuts-node-1               | time="2023-03-10T12:49:19+01:00" level=audit msg="Access denied: expected issuer (demo-apps) does not match iss" actor=172.20.0.11 event=AccessDenied module=http operation=tokenV2.middleware
stable-stable-nuts-node-1               | time="2023-03-10T12:49:19+01:00" level=info msg="HTTP request" method=GET module=http remote_ip=172.20.0.11 status=401 uri="/internal/didman/v1/did/did:nuts:7eaDstczhnigW9uW1JRzYF2u5KhTntFTnee8vGDHokD/contactinfo"
stable-stable-nuts-node-1               | time="2023-03-10T12:49:19+01:00" level=warning msg="Operation failed" error=Unauthorized operationID="<nil>" 
```

But then it logs a warning, which is unexpected:

```
stable-stable-nuts-node-1               | time="2023-03-10T12:49:19+01:00" level=warning msg="Unable to send error back to client, response already committed" error=Unauthorized
```

woutslakhorst:
fixes #1974 

the auth middleware no longer does `echoCtx.String(code, message)` but now relies on the error writer in the context to do the final output.

### DIFF
--- a/http/tokenV2/middleware.go
+++ b/http/tokenV2/middleware.go
@@ -392,10 +392,17 @@ func authenticationCredential(context echo.Context) string {
 	return fields[1]
 }
 
+type unauthorizedErrorWriter struct{}
+
+func (p unauthorizedErrorWriter) Write(echoContext echo.Context, _ int, _ string, _ error) error {
+	// always return 401 unauthorized regardless of error. It's a security error, so we don't want to leak any information
+	return echoContext.String(http.StatusUnauthorized, http.StatusText(http.StatusUnauthorized))
+}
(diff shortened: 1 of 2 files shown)


reinkrul (approved):
good thinking

PR #2267, opened 2023-06-21
only enable latest tag for docker if build tag equals latest github tag

### LINKED ISSUE #2167: Docker Hub tags are stale
The generalized and latest tags (e.g. `v5` and `latest`) are pointed to the last release, not the highest released version. Thus, if `v5.3.0` gets released and then bugfix version `v5.2.9`, the latter is pointed to by the `v5` and `latest`. But the should point to the highest released version, meaning `v5.3.0`.

woutslakhorst:
fixes #2167 

`latest` was always created, now it's only enabled if the tag being build is the latest github tag.

### DIFF
--- a/.github/workflows/build-images.yaml
+++ b/.github/workflows/build-images.yaml
@@ -26,11 +26,18 @@ jobs:
           echo "::set-output name=git_branch::$(echo ${GITHUB_REF#refs/heads/})"
           echo "::set-output name=git_version::$(git name-rev --tags --name-only $(git rev-parse HEAD))"
 
+      - uses: actions-ecosystem/action-get-latest-tag@v1
+        id: get-latest-tag
+        with:
+          semver_only: true
+
       - name: Print version params
         run: |
           echo "Commit: ${{ steps.version.outputs.git_commit }}"
           echo "Branch: ${{ steps.version.outputs.git_branch }}"
           echo "Version: ${{ steps.version.outputs.git_version }}"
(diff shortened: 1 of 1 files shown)


reinkrul (approved):

.github/workflows/build-images.yaml
```diff
@@ -45,7 +52,7 @@ jobs:
             # generate v5 tag
             type=semver,pattern={{major}}
           flavor: |
-            latest=auto
+            latest=${{ github.ref == steps.get-latest-tag.outputs.tag }}
```
reinkrul, line 55: I find it hard to understand, but `latest` should be either `true`, `false`, or `auto` according to https://github.com/docker/metadata-action#latest-tag ?

gerardsn (commented):
the `v5` tag had the same issue (but probably limited to releases starting with v5). Can we apply the same fix there?

.github/workflows/build-images.yaml
woutslakhorst: yes and auto is broken, so now with scripting set to false/true

woutslakhorst:
> the `v5` tag had the same issue (but probably limited to releases starting with v5). Can we apply the same fix there?

Not supported

gerardsn:
so delete `v5`?

woutslakhorst:
I'll merge for now since this is better then it was. Will address `v5` version later

PR #2272, opened 2023-06-22
OIDC4VCI: compare offered and received credential types for equality

### LINKED ISSUE #2050: Wallet must check issued VC is of the requested type
Just to be sure.

When the wallet performs a Credential Request, it contains the type if the credential it wants. It should make sure the issued credential is of the expected type.

gerardsn:
closes #2050

### DIFF
--- a/vcr/holder/oidc4vci_wallet.go
+++ b/vcr/holder/oidc4vci_wallet.go
@@ -31,6 +31,7 @@ import (
 	"github.com/nuts-foundation/nuts-node/audit"
 	"github.com/nuts-foundation/nuts-node/core"
 	"github.com/nuts-foundation/nuts-node/crypto"
+	"github.com/nuts-foundation/nuts-node/jsonld"
 	"github.com/nuts-foundation/nuts-node/vcr/log"
 	"github.com/nuts-foundation/nuts-node/vcr/oidc4vci"
 	vcrTypes "github.com/nuts-foundation/nuts-node/vcr/types"
@@ -50,8 +51,7 @@ var nowFunc = time.Now
 var _ OIDCWallet = (*wallet)(nil)
 
 // NewOIDCWallet creates an OIDCWallet that tries to retrieve offered credentials, to store it in the given credential store.
-func NewOIDCWallet(did did.DID, identifier string, credentialStore vcrTypes.Writer, signer crypto.JWTSigner, resolver vdr.KeyResolver,
-	clientTimeout time.Duration, clientTLSConfig *tls.Config) OIDCWallet {
+func NewOIDCWallet(did did.DID, identifier string, credentialStore vcrTypes.Writer, signer crypto.JWTSigner, resolver vdr.KeyResolver, clientTimeout time.Duration, clientTLSConfig *tls.Config, jsonldReader jsonld.Reader) OIDCWallet {
 	return &wallet{
 		did:                 did,
 		identifier:          identifier,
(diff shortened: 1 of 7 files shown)


vcr/holder/oidc4vci_wallet.go
```diff
@@ -170,6 +184,57 @@ func (h wallet) HandleCredentialOffer(ctx context.Context, offer oidc4vci.Creden
 	return nil
 }
 
+func supportedCredentialFormat(offeredCredential map[string]interface{}) bool {
+	switch offeredCredential["format"] {
```
reinkrul, line 188: Wouldn't
```
return offeredCredential["format"] == oidc4vci.VerifiableCredentialJSONLDFormat
```
suffice? :-p

vcr/holder/oidc4vci_wallet.go
```diff
@@ -158,8 +167,13 @@ func (h wallet) HandleCredentialOffer(ctx context.Context, offer oidc4vci.Creden
 			StatusCode: http.StatusInternalServerError,
 		}
 	}
-	// TODO: Wallet should make sure the VC is of the expected type
-	//       See https://github.com/nuts-foundation/nuts-node/issues/2050
+	if err = credentialTypesMatchOffer(h.jsonldReader, *credential, offer.Credentials[0]); err != nil {
```
reinkrul, line 170: Should you expand `offer.Credentials[0]` as well before processing it?

vcr/holder/oidc4vci_wallet.go
```diff
@@ -170,6 +184,57 @@ func (h wallet) HandleCredentialOffer(ctx context.Context, offer oidc4vci.Creden
...
+		// TODO: can credentialDefinition contain invalid values that makes this panic?
+		expectedTypes := document.ValueAt(jsonld.NewPath("@type"))
+
+		document, err = reader.Read(credential)
+		if err != nil {
+			return fmt.Errorf("invalid credential: %w", err)
+		}
+		receivedTypes := document.ValueAt(jsonld.NewPath("@type"))
+
+		if !equal(expectedTypes, receivedTypes) {
```
reinkrul, line 214: I think

`VerifiableCredential, NutsAuthorizationCredential` is equal to `NutsAuthorizationCredential, VerifiableCredential`.

Not very probable that it would differ, though, so probably OK (maybe add a note)

gerardsn: I was not sure about this. According to the JSON spec an array is ordered, so I do not expect them to change from offer to issuance. The main question I had was how is this handled when fields are defined in 2 context, is the one that is applied selected based on the order of contexts, types (meaning that order must be preserved), or something else?

vcr/holder/oidc4vci_wallet.go
gerardsn: Why do you prefer to do it there? It is currently expanded inside the function as it is a detail from the format we are using.

vcr/holder/oidc4vci_wallet.go
gerardsn: pff.... to easy. 

I will take it out and add a new PR that adds format specific validation. This solves the other todo I added.

vcr/holder/oidc4vci_wallet.go
reinkrul: Order is irrelevant, but is preserved. So probably OK like this (super-edge-case I'd say)

vcr/holder/oidc4vci_wallet.go
reinkrul: Nvm, I misread

reinkrul (approved):

gerardsn:
The `credential_definition` in an offer uses `types` according to spec instead of `@type` or `type`, this means we can't expand the offer to compare the types. Any ideas on how to move forward on this? @woutslakhorst @reinkrul

gerardsn:
`types` vs `type` issue fixed in #2286

PR #2273, opened 2023-06-22
e2e-tests: don't use go.mod replace, breaks

reinkrul:


gerardsn (commented):
go.mod still contains `module github.com/nuts-foundation/nuts-go-e2e-test` and this is also used by `main.go`, could this be the source for the issues?

I'm  concerned that removing this replacement results in the e2e test not failing because it is not using the correct version.

reinkrul:
Alternatively: https://github.com/nuts-foundation/nuts-node/pull/2275

PR #2287, opened 2023-06-26
Feature/2058/oidc4vci/definitions

### LINKED ISSUE #2058: OIDCIssuer must have a list of offered credential types
Places of use:

- List in OIDCIssuer metadata (`credentials_supported`)
- Check when credential is requested

woutslakhorst:
closes #2058 

Credential definitions are loaded from assets and optionally from a location on disk. From disk the entire dir (and subdir) is read and all json files are parsed and added to the definitions.

The definitions are loaded with `New()` and cached until restart


reinkrul:
How does this fit into the future issuance flow? E.g. do we limit issuance over OpenID4VCI to configured VC definitions? The depending on how flexible the dynamic issuance flow is going to be (probably not a fully-fledged business-rule-driven credential engine), this configuration might be sufficient or not.

reinkrul (commented):
I think this is very much subject to change, so not sure whether we should allow it to be configured in that case. Or add a warning in the flag docs that it might change (in a breaking manner) without further notice.

woutslakhorst:
> How does this fit into the future issuance flow? E.g. do we limit issuance over OpenID4VCI to configured VC definitions? The depending on how flexible the dynamic issuance flow is going to be (probably not a fully-fledged business-rule-driven credential engine), this configuration might be sufficient or not.

Credential types are not created on the fly, but are the result of a careful design session. How can you otherwise trust a specific credential and its issuer?

reinkrul:
Then I would still opt for a notice in the flag's docs.

reinkrul (approved):

vcr/oidc4vci/types.go
```diff
@@ -138,3 +140,15 @@ type TokenResponse struct {
 	// TokenType defines the type of the token issued as described in [RFC6749].
 	TokenType string `json:"token_type"`
 }
+
+// Config holds the config for the OIDC4VCI credential issuer and wallet
+type Config struct {
+	// DefinitionsDIR defines the directory where the credential definitions are stored
```
reinkrul, line 146: actually, additional definitions?

vcr/issuer/openid4vci/issuer.go
```diff
@@ -376,3 +436,14 @@ func generateCode() string {
 	}
 	return base64.URLEncoding.EncodeToString(buf)
 }
+
+func deepcopy(src []map[string]interface{}) []map[string]interface{} {
```
reinkrul, line 440: Not really a deep copy (just of the keys), could use `mohae/deepcopy.Copy()` which is already a (transitive?) dependency

woutslakhorst: and the slice.
I didn't want to use that dep, not really needed here.

PR #2288, opened 2023-06-27
e2e tests: vault was moved to hashicorp/vault

reinkrul:


gerardsn (approved):
do we have docs pointing to the old location?

reinkrul:
> do we have docs pointing to the old location?

No

PR #2293, opened 2023-06-28
Helm: do not set new chart as latest release

gerardsn:
The action sets the newest chart as the latest release on GitHub by default


reinkrul (approved):
Nice

PR #2294, opened 2023-06-29
VCR Search: don't log each individual invalid search result

reinkrul:
Right now it logs a line for every invalid VCs, which might also be a revoked credential. Or untrusted. We're probably not interested in revoked credentials, _maybe_ in untrusted credentials (to find out why you can't find a certain VC). So best option to both keep the errors but also declutter the DEBUG log is to just log the error counts, individual errors on TRACE.

Example DEBUG log:

```
Filtered 13 invalid VCs from search results (more info on TRACE): 'EOF' (1 times), 'credential is revoked' (2 times), 'credential issuer is untrusted' (10 times)
```

### DIFF
--- a/vcr/search.go
+++ b/vcr/search.go
@@ -115,6 +115,6 @@ func formatFilteredVCsLogMessage(verifyErrors map[string]int) string {
 		sort.Strings(parts)
 		numFiltered += count
 	}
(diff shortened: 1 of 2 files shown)


woutslakhorst (approved):
nice

PR #2301, opened 2023-06-30
Pass didStore instead of service wrappers

woutslakhorst:
This is a small refactor to reduce some parameter passing to different engines. The docFinder, docResolver, keyResolver and serviceResolver can all use the store directly.

The didstore is now passed, since it's the resource to share. The resolvers are just functional wrappers.

Now all resolvers have an interface under vdr/types and an implementation in didservice


woutslakhorst:
I'm wondering if we can remove the resolver interfaces and just mock the didstore all the time....

gerardsn (approved):
I don't see the real advantages yet besides making things a bit more consistent.

If didstore.Store was not an interface but a struct it could implement all of these resolvers and access to the store could be limited by resolver interfaces.

woutslakhorst:
> If didstore.Store was not an interface but a struct it could implement all of these resolvers and access to the store could be limited by resolver interfaces.

It is an interface, we just added another layer of interfaces on top of it...

PR #2302, opened 2023-06-30
Cleanup: remove old network.tls properties from e2e tests

reinkrul:
Replaced by `tls` root props.


gerardsn (approved):
We are not running v4 nodes any more in tests I suppose

PR #2303, opened 2023-06-30
Cleanup: use --wait in e2e tests

reinkrul:
No need for our own bash-fu.


gerardsn (approved):
nice!
is `--wait-timeout int          timeout waiting for application to be running|healthy.` the same with a timeout?

reinkrul:
Not sure

PR #2304, opened 2023-06-30
OpenID4VCI: validate credential offer and request

### LINKED ISSUE #2037: Validate requested credential definition and format
When a wallet requests a credential using `CredentialRequest`, it sends a credential definition and format. These need to be validated:

- The only supported format is `ldp_vc`, reject requests with other formats
- Verify the requested credential is the one offered (and verify credentialSubject, bound to pre-authorized code).

See https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#name-credential-request and https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#name-credential-offer-3

gerardsn:
closes #2037 and cleanup from #2050

- [x] ~set credentialSubject in CredentialRequest.~ already validated dring authorization, credentialSubject is not used
- [x] tests

### DIFF
--- a/jsonld/scalar.go
+++ b/jsonld/scalar.go
@@ -29,8 +29,6 @@ type Scalar interface {
 	fmt.Stringer
 	// Value returns the underlying value (string, float, true or false)
 	Value() interface{}
-	// Equal compares the value of both Scalar
-	Equal(o Scalar) bool
 }
 
 // StringScalar is the string version of a Scalar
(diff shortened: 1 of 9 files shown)


gerardsn:
from [the spec](https://openid.bitbucket.io/connect/editors-draft/openid-4-verifiable-credential-issuance-1_0.html#name-format-identifier-2)
> When the format value is ldp_vc, entire Credential Offer, Authorization Details, Credential Request and Credential Issuer metadata, including credential_definition object, MUST NOT be processed using JSON-LD rules.
> 
> @context value in the credential_definition object could be used by the Wallet to check whether it supports a certain VC or not. If necessary, the Wallet could apply JSON-LD processing to the Credential issued by the Credential Issuer.

Currently `VC`'s and `credential_definitions` are compared using JSON-LD processing which is apparently not allowed. Does this mean we should just compare the credential `type` fields directly?

gerardsn:
closed in favor of #2308

PR #2305, opened 2023-07-03
OpenID4VCI: fix unknown nonce error

reinkrul:
Should be `invalid_or_missing_proof` according to https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#name-credential-error-response


woutslakhorst (approved):
Specs do not really describe the specific checks we do, so it doesn't really matter which error to return.

gerardsn (changes requested):
updated in draft versions  to `invalid_proof` https://openid.bitbucket.io/connect/editors-draft/openid-4-verifiable-credential-issuance-1_0.html#name-credential-error-response

reinkrul:
Revert

gerardsn (approved):

PR #2306, opened 2023-07-03
e2e-tests: removed prepare.sh scripts with env variable node DIDs

reinkrul:
Made e2e tests simpler and faster.


e2e-tests/nuts-network/private-transactions/node-A/nuts.yaml
```diff
@@ -1,4 +1,4 @@
-verbosity: debug
+verbosity: trace
```
gerardsn, line 1: was this intentional?

gerardsn (approved):

e2e-tests/nuts-network/private-transactions/node-A/nuts.yaml
reinkrul: I'll revert it

PR #2308, opened 2023-07-04
OpenID4VCI: Add OfferedCredential and CredentialDefinition types

gerardsn:
new approach to validation of credential and credential_definition #2037 and #2050
1. Add `OfferedCredential` and `CredentialDefinition` types (this PR) to make validation and parsing easier
2. Add Format specific CredentialDefinition validation on `CredentialRequest` and `CredentialOffer`
3. Compare CredentialDefinition with received/issued VC


vcr/oidc4vci/validators.go
```diff
@@ -0,0 +1,86 @@
...
+		return errors.New("invalid credential_definition: missing @context field")
+	}
+	if len(definition.Type) == 0 {
+		return errors.New("invalid credential_definition: missing type field")
+	}
+	if definition.CredentialSubject != nil {
+		if isOffer {
+			return errors.New("invalid credential_definition: credentialSubject not allowed in offer")
+		}
+		// TODO: add credentialSubject validation.
```
woutslakhorst, line 42: a todo

vcr/oidc4vci/validators.go
```diff
@@ -0,0 +1,86 @@
...
+	if len(credential.Context) < len(definition.Context) || !isSubset(credential.Context, definition.Context) {
+		return errors.New("credential does not match credential_definition: context mismatch")
+	}
+
+	// compare types. fails when definition.Type contains duplicates
+	if len(credential.Type) != len(definition.Type) || !isSubset(credential.Type, definition.Type) {
+		return errors.New("credential does not match credential_definition: type mismatch")
+	}
+
+	// TODO: compare credentialSubject
```
woutslakhorst, line 63: another todo

vcr/issuer/openid.go
```diff
@@ -230,6 +230,20 @@ func (i *openidHandler) OfferCredential(ctx context.Context, credential vc.Verif
 func (i *openidHandler) HandleCredentialRequest(ctx context.Context, request oidc4vci.CredentialRequest, accessToken string) (*vc.VerifiableCredential, error) {
 	// TODO: Verify requested format and credential definition
```
woutslakhorst, line 231: remove todo

vcr/issuer/openid.go
```diff
@@ -230,6 +230,20 @@ func (i *openidHandler) OfferCredential(ctx context.Context, credential vc.Verif
...
 	//       See https://github.com/nuts-foundation/nuts-node/issues/2037
+	if request.Format != oidc4vci.VerifiableCredentialJSONLDFormat {
+		return nil, oidc4vci.Error{
+			Err:        fmt.Errorf("credential request: unsupported format '%s'", request.Format),
+			Code:       oidc4vci.UnsupportedCredentialType,
+			StatusCode: http.StatusInternalServerError,
```
woutslakhorst, line 237: should be bad request.

woutslakhorst (changes requested):
Just some comments

vcr/oidc4vci/types.go
```diff
@@ -97,11 +98,29 @@ type CredentialOffer struct {
 	// CredentialIssuer defines the identifier of the credential issuer.
 	CredentialIssuer string `json:"credential_issuer"`
 	// Credentials defines the credentials offered by the issuer to the wallet.
-	Credentials []map[string]interface{} `json:"credentials"`
+	Credentials []OfferedCredential `json:"credentials"`
 	// Grants defines the grants offered by the issuer to the wallet.
 	Grants map[string]interface{} `json:"grants"`
 }
 
+// OfferedCredential defines an individual credential offered. We currently do not support 'JSON string' offers.
```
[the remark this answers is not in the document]
gerardsn: changed to `OfferedCredential defines a single entry in the credentials array of a CredentialOffer. ...`

vcr/oidc4vci/validators.go
gerardsn: we currently do not use the credentialSubject, and I am not sure what to validate here. We could add some basic validation but that might give us more issues later on if the validation turns out to be incorrect.

vcr/holder/openid.go
```diff
@@ -149,18 +163,18 @@ func (h openidHandler) HandleCredentialOffer(ctx context.Context, offer oidc4vci
...
 			Code:       oidc4vci.ServerError,
 			StatusCode: http.StatusInternalServerError,
 		}
 	}
-	if err = credentialTypesMatchOffer(h.jsonldReader, *credential, offer.Credentials[0]); err != nil {
+	if err = oidc4vci.ValidateDefinitionWithCredential(*credential, *offeredCredential.CredentialDefinition); err != nil {
 		return oidc4vci.Error{
 			Err:        fmt.Errorf("received credential does not match offer: %w", err),
-			Code:       oidc4vci.UnsupportedCredentialType,
+			Code:       oidc4vci.InvalidRequest,
```
reinkrul, line 177: change is correct?

vcr/oidc4vci/validators.go
```diff
@@ -0,0 +1,86 @@
...
+			return errors.New("invalid credential_definition: credentialSubject not allowed in offer")
+		}
+		// TODO: add credentialSubject validation.
+	}
+	return nil
+}
+
+// ValidateDefinitionWithCredential confirms that the vc.VerifiableCredential is defined by the CredentialDefinition.
+// CredentialDefinition is assumed to be valid, see ValidateCredentialDefinition.
+func ValidateDefinitionWithCredential(credential vc.VerifiableCredential, definition CredentialDefinition) error {
```
reinkrul, line 49: are those `isSubset` applicable? Is it allowed the returned credential is richer than the one offered?

reinkrul (changes requested):

vcr/oidc4vci/types.go
```diff
@@ -97,11 +98,29 @@ type CredentialOffer struct {
...
+type OfferedCredential struct {
+	// Format specifies the credential format.
+	Format string `json:"format"`
+	// CredentialDefinition contains the 'credential_definition' for the Verifiable Credential Format flows.
+	CredentialDefinition *CredentialDefinition `json:"credential_definition,omitempty"`
+}
+
+// CredentialDefinition defines the 'credential_definition' for Format VerifiableCredentialJSONLDFormat
+// Specified by https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#name-vc-secured-using-data-integ
+type CredentialDefinition struct {
```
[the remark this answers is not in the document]
gerardsn: oidc4vci spec defines credential_definition for non-JSON-LD credentials too. It also specifies that the credential_definition MUST NOT be subjected to JSON-LD processing. Using a different struct seems like a better choice for that.

vcr/holder/openid.go
gerardsn: spec doesn't define a response to the credential offer, so its whatever we define to be correct. IMO it is correct.

vcr/oidc4vci/validators.go
gerardsn: the `credential_definition` basically contains a `type` and needs a `@context` to go with that. Once its a full VC it may contain additional contexts for signatures / proofs etc. It is unclear to me if those additional contexts need to be present in the original `credential_definition`, so this allows both.

vcr/oidc4vci/validators.go
```diff
@@ -0,0 +1,86 @@
...
+package oidc4vci
+
+import (
+	"errors"
+	ssi "github.com/nuts-foundation/go-did"
+	"github.com/nuts-foundation/go-did/vc"
+)
+
+// ValidateCredentialDefinition validates the CredentialDefinition according to the VerifiableCredentialJSONLDFormat format
+func ValidateCredentialDefinition(definition *CredentialDefinition, isOffer bool) error {
```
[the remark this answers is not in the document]
gerardsn: doesn't that just move the `nil` check outside this function before calling the validator on the `CredentialDefinition`?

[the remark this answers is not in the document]
gerardsn: I've made it a function on the `CredentialDefinition`, but it still contains the nil check.

vcr/oidc4vci/validators.go
reinkrul: I guess it wont hurt much

reinkrul (approved):

vcr/oidc4vci/validators.go
woutslakhorst: then it's not a todo for the code. If we need to remember something, it should be listed in an issue not in the code.

vcr/oidc4vci/validators.go
woutslakhorst: If we need to remember something, it should be listed in an issue not in the code.

vcr/oidc4vci/validators.go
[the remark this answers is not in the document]
woutslakhorst: That's because of the pointer receiver. Do we always need a definition on an offer. Then the offer can also remove the pointer from the definition. You'll get an empty struct but validation would fail for that as well. Less pointers equals easier life.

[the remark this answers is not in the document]
gerardsn: the `CredentialDefinition` is optional since it is only part of the verifiable credential flows, not de mDL flows. We only support `ldp_vc` for now where it is required.

vcr/oidc4vci/validators.go
gerardsn: created issue #2320

vcr/oidc4vci/validators.go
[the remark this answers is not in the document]
gerardsn: Unless you don't care about the specific error and just let it fail on missing @context. I prefer the current solution as it is closest to the spec and therefor easiest to understand/extend.

woutslakhorst (approved):

PR #2309, opened 2023-07-04
add xor tree repair loop

### LINKED ISSUE #2295: Auto-fix incorrect XOR/IBLT values
There's a small change data is lost when using Redis. Most of the data can be retrieved using other nodes, but the XOR and IBLT value are forever broken.

A background proces can simply walk the dag, per 512 LC values, and recalculate/check values. If the process does this for pages that are already full, the change is concurrent access is relatively small.

Note: XOR and IBLT data is segmented over pages. Pages are only updated if a historic transaction is processed. During normal operation only data is added to the last page.

woutslakhorst:
closes #2295 

The dag state now starts a go routine that loops over parts of the dag when triggered. The trigger is a multiple XOR comparison failure detected from gossip messages. All connected nodes will need to have a different XOR before it's triggered. When triggered it handles a page of TX per 10 seconds.

The fix does not run as migration on startup. Although performance is good enough (20k tx/s), it exposes healthy nodes to an alternate code path which was previously not present. With this solution only nodes with a broken xor tree will be exposed to this code, reducing the risk of introducing new bugs for healthy nodes.

iblt trees are not repaired as discussed in #2295

### DIFF
--- a/network/dag/consistency.go
+++ b/network/dag/consistency.go
@@ -1,0 +1,161 @@
+/*
+ * Copyright (C) 2023 Nuts community
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *
+ */
+
+package dag
+
+import (
+	"context"
+	"github.com/nuts-foundation/go-stoabs"
+	"github.com/nuts-foundation/nuts-node/network/dag/tree"
+	"github.com/nuts-foundation/nuts-node/network/log"
+	"sync"
+	"time"
+)
+
+type circuit int
+
+const (
+	circuitGreen circuit = iota
+	circuitYellow
+	circuitRed
+)
+
+// xorTreeRepair is responsible for repairing the XOR tree. Its loop is triggered when the network layer detects differences in XOR values with all other nodes.
+// It will loop over all pages of the XOR tree and recalculates the XOR value with the transactions in the database.
+// This repair is needed because current networks have nodes that have a wrong XOR value. How this happens is not yet known, it could be due to DB failures of due to failures in older versions.
+// The fact is that we can fix the state relatively easy.
+// The loop checks a page (512 LC values) per 10 seconds and continues looping until the network layer signals all is ok again.
+type xorTreeRepair struct {
+	ctx          context.Context
+	cancel       context.CancelFunc
+	ticker       *time.Ticker
+	currentPage  uint32
+	state        *state
+	circuitState circuit
+	mutex        sync.Mutex
+}
+
+func newXorTreeRepair(state *state) *xorTreeRepair {
+	return &xorTreeRepair{
+		state:  state,
+		ticker: time.NewTicker(10 * time.Second),
+	}
+}
+
+func (f *xorTreeRepair) start() {
+	f.ctx, f.cancel = context.WithCancel(context.Background())
+	go f.loop()
+}
+
+func (f *xorTreeRepair) shutdown() {
+	if f.cancel != nil {
+		f.cancel()
+	}
+}
+
+func (f *xorTreeRepair) loop() {
+	for {
+		select {
+		case <-f.ctx.Done():
+			return
+		case <-f.ticker.C:
+			f.checkPage()
+		}
+	}
+}
+
+func (f *xorTreeRepair) checkPage() {
+	f.mutex.Lock()
+	defer f.mutex.Unlock()
+
+	// ignore run if circuit is not red
+	if f.circuitState < circuitRed {
+		return
+	}
+
+	currentLC := f.state.lamportClockHigh.Load()
+	lcStart := f.currentPage * PageSize
+	lcEnd := lcStart + PageSize
+
+	// initialize an XOR tree
+	calculatedXorTree := tree.New(tree.NewXor(), PageSize)
+
+	// acquire global lock
+	err := f.state.graph.db.Write(context.Background(), func(txn stoabs.WriteTx) error {
+		txs, err := f.state.graph.findBetweenLC(txn, lcStart, lcEnd)
+		if err != nil {
+			return err
+		}
+		for _, tx := range txs {
+			calculatedXorTree.Insert(tx.Ref(), tx.Clock())
+		}
+
+		// Get XOR leaf from current XOR tree
+		xorTillEnd, _ := f.state.xorTree.getZeroTo(lcEnd - 1)
+		if lcStart != 0 {
+			xorTillStart, _ := f.state.xorTree.getZeroTo(lcStart - 1)
+			_ = xorTillEnd.Subtract(xorTillStart)
+		}
+
+		// Subtract the calculated tree, should be empty if the trees are equal
+		_ = xorTillEnd.Subtract(calculatedXorTree.Root())
+		if !xorTillEnd.Empty() {
+			// it's not empty, so replace the leaf in the current XOR tree with the calculated one
+			err = f.state.xorTree.tree.Replace(lcStart, calculatedXorTree.Root())
+			if err != nil {
+				return err
+			}
+			log.Logger().Warnf("detected XOR tree mismatch for page %d, fixed using recalculated values", f.currentPage)
+		}
+
+		// Now we do the same for the IBLT tree as stated in
+		// https://github.com/nuts-foundation/nuts-node/issues/2295
+		// we skip the iblt tree for now, since the chance for it to become corrupt is incredibly low.
+		// there can only be a problem with duplicate entries, not with missing entries.
+		// the xor tree already has an effect when it's missing entries.
+		// fixing the iblt tree is a copy of the code above (but with ibltTree instead of xorTree).
+
+		return nil
+	})
+	if err != nil {
+		log.Logger().Warnf("failed to run xorTreeRepair check: %s", err)
+	}
+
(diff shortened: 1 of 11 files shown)


network/transport/v2/handlers.go
```diff
@@ -300,6 +302,7 @@ func (p *protocol) handleGossip(ctx context.Context, connection grpc.Connection,
 	// If the XORs are not equal and the peer is behind, still request the missing refs if there are any.
 	tempXor := xor.Xor(refs...)
 	if tempXor.Equals(peerXor) || (msg.LC < clock && len(refs) > 0) {
+		p.state.CorrectStateDetected()
```
gerardsn, line 305: add comment on why this is the correct state but the state on L266 is not. This is not trivial

gerardsn (changes requested):

reinkrul:
> All connected nodes will need to have a different XOR before it's triggered

How does this fix the local node if only the local XOR is broken?

network/dag/consistency.go
```diff
@@ -0,0 +1,160 @@
...
+
+const (
+	circuitGreen circuit = iota
+	circuitYellow
+	circuitRed
+)
+
+// xorTreeRepair is responsible for repairing the XOR tree. Its loop is triggered when the network layer detects differences in XOR values with all other nodes.
+// It will loop over all pages of the XOR tree and recalculates the XOR value with the transactions in the database.
+// This repair is needed because current networks have nodes that have a wrong XOR value. How this happens is not yet known, it could be due to DB failures of due to failures in older versions.
```
reinkrul, line 40:
```suggestion
// This repair is needed because current networks have nodes that have a wrong XOR value. How this happens is not yet known, it could be due to DB failures or due to bugs in older versions.
```

reinkrul:
Should this be an invokable/triggerable operation as well (e.g. by removing certain files or sending a specific API call), in case it is suspected to be broken but it doesn't trigger for whatever reason (e.g. other node has the same defect.

network/dag/consistency.go
```diff
@@ -0,0 +1,160 @@
...
+// The fact is that we can fix the state relatively easy.
+// The loop checks a page (512 LC values) per 10 seconds and continues looping until the network layer signals all is ok again.
+type xorTreeRepair struct {
+	ctx          context.Context
+	cancel       context.CancelFunc
+	ticker       *time.Ticker
+	currentPage  uint32
```
reinkrul, line 47: Racy access (by both ticker and `stateOK` for instance)

network/dag/consistency.go
```diff
@@ -0,0 +1,160 @@
...
+
+	currentLC := f.state.lamportClockHigh.Load()
+	lcStart := f.currentPage * PageSize
+	lcEnd := lcStart + PageSize
+
+	// initialize an XOR tree
+	calculatedXorTree := tree.New(tree.NewXor(), PageSize)
+
+	// acquire global lock
+	err := f.state.graph.db.Write(context.Background(), func(txn stoabs.WriteTx) error {
```
reinkrul, line 98: how quick/fast is this? The funcs to be called by externals are now blocked by the mutex.(although they are blocked by the write TX anyways..?)

network/dag/consistency.go
```diff
@@ -0,0 +1,160 @@
...
+				return
+			case <-f.ticker.C:
+				f.checkPage()
+			}
+		}
+	}()
+}
+
+func (f *xorTreeRepair) shutdown() {
+	if f.cancel != nil {
```
reinkrul, line 76: ticket should be stopped

network/dag/consistency.go
```diff
@@ -0,0 +1,160 @@
...
+
+		// Subtract the calculated tree, should be empty if the trees are equal
+		_ = xorTillEnd.Subtract(calculatedXorTree.Root())
+		if !xorTillEnd.Empty() {
+			// it's not empty, so replace the leaf in the current XOR tree with the calculated one
+			err = f.state.xorTree.tree.Replace(lcStart, calculatedXorTree.Root())
+			if err != nil {
+				return err
+			}
+			log.Logger().Warnf("detected XOR tree mismatch for page %d, fixed using recalculated values", f.currentPage)
```
reinkrul, line 122: could it just be it's temporarily out-of-sync with the other nodes due "normal" out-of-date (missing a few TXs) and being slow for a moment?

network/dag/consistency.go
```diff
@@ -0,0 +1,160 @@
...
+		// https://github.com/nuts-foundation/nuts-node/issues/2295
+		// we skip the iblt tree for now, since the chance for it to become corrupt is incredibly low.
+		// there can only be a problem with duplicate entries, not with missing entries.
+		// the xor tree already has an effect when it's missing entries.
+		// fixing the iblt tree is a copy of the code above (but with ibltTree instead of xorTree).
+
+		return nil
+	})
+	if err != nil {
+		log.Logger().Warnf("failed to run xorTreeRepair check: %s", err)
```
reinkrul, line 135:
```suggestion
		log.Logger().WithError(err).Warnf("Failed to run xorTreeRepair check.")
```

network/dag/interface.go
```diff
@@ -93,6 +93,11 @@ type State interface {
 	//	- highest lamport clock in the DAG
 	// A requested clock of math.MaxUint32 will return the iblt of the entire DAG
 	IBLT(reqClock uint32) (tree.Iblt, uint32)
+
+	// IncorrectStateDetected is called when the xor and LC value from a gossip message do NOT match the local state.
+	IncorrectStateDetected()
```
reinkrul, line 98: 2 node mismatches = repair?

network/dag/consistency.go
```diff
@@ -0,0 +1,160 @@
...
+	currentPage  uint32
+	state        *state
+	circuitState circuit
+	mutex        sync.Mutex
+}
+
+func newXorTreeRepair(state *state) *xorTreeRepair {
+	return &xorTreeRepair{
+		state:  state,
+		ticker: time.NewTicker(10 * time.Second),
```
reinkrul, line 56: given almost 30000 TXs in stable, it would take `30000/512*10/60 ~= 10 minutes` to fix it, worst case. Is that OK?

reinkrul (changes requested):

network/dag/tree/tree.go
```diff
@@ -252,6 +287,10 @@ func (t *tree) Root() Data {
 }
 
 func (t *tree) ZeroTo(clock uint32) (Data, uint32) {
+	if clock < 0 {
+		return t.prototype.New(), 0
+	}
```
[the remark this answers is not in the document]
woutslakhorst: leftover code. Removed it.

network/dag/consistency.go
```diff
@@ -0,0 +1,160 @@
...
+		f.currentPage++
+	}
+}
+
+func (f *xorTreeRepair) stateOK() {
+	f.mutex.Lock()
+	defer f.mutex.Unlock()
+
+	f.circuitState = circuitGreen
+	f.currentPage = 0
```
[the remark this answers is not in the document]
woutslakhorst: It'll only be triggered if you are in sync with another node, so your LC must be the same as another one.

But since it doesn't really matter, the reset can be removed.

network/transport/v2/handlers.go
woutslakhorst: actually it should also have been on L266, just forgot it.

woutslakhorst:
> Should this be an invokable/triggerable operation as well (e.g. by removing certain files or sending a specific API call), in case it is suspected to be broken but it doesn't trigger for whatever reason (e.g. other node has the same defect.

Another node would have to have the exact same defect. Extremely unlikely. Then you can always turn off 1 node or nuke the DAG.

network/dag/consistency.go
woutslakhorst: tested with -race and no race conditions have been found. Both checkPage and stateOk use the mutex.

network/dag/consistency.go
woutslakhorst: takes around 50-100ms per check.

network/dag/consistency.go
woutslakhorst: defer close from within loop

network/dag/consistency.go
woutslakhorst: only when you also have experienced down time. In a high traffic scenario, it could be triggered once. The checker starts at 0 so it'll start checking the page that has no traffic to avoid racy fixes.

network/dag/interface.go
woutslakhorst: Yes, could become interesting if 2 nodes are out of sync.... Could set it to 5 or even 10

network/dag/consistency.go
woutslakhorst: Opposed to the 6 months of additional network messages every 10s.....

You want different gossip messages to arrive within a single interval so they can set the state to green again.

network/dag/consistency.go
```diff
@@ -0,0 +1,160 @@
...
+			}
+			log.Logger().Warnf("detected XOR tree mismatch for page %d, fixed using recalculated values", f.currentPage)
+		}
+
+		// Now we do the same for the IBLT tree as stated in
+		// https://github.com/nuts-foundation/nuts-node/issues/2295
+		// we skip the iblt tree for now, since the chance for it to become corrupt is incredibly low.
+		// there can only be a problem with duplicate entries, not with missing entries.
+		// the xor tree already has an effect when it's missing entries.
+		// fixing the iblt tree is a copy of the code above (but with ibltTree instead of xorTree).
```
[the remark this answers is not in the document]
woutslakhorst: I'm a bit weary about bugs in this code, destroying an xor leaf which is already broken is not a big deal. Destroying a iblt leaf has bigger consequences.

network/dag/consistency.go
gerardsn: perhaps good to do a performance test with this? 100ms on an interval of a page/10s = ~1% DB downtime. It could have a significant impact during bursts of data (auto publishing OpenID services), but becomes less relevant as more transactions move off the dag.

gerardsn (approved):

network/dag/consistency.go
reinkrul: That's right, I saw that later and forgot to remove the second comment 👍

reinkrul (approved):

reinkrul:
Band-aid time!

PR #2323, opened 2023-07-07
Add backup for private credentials

### LINKED ISSUE #2317: VCR: Verifier store must have backup shelf as well
It doesn't have one because it never needed one: all VCs were on the DAG meaning it could be restored from the DAG. But when receiving VCs over OpenID4VCI, they're not backed up (since they're not on the DAG).

The verifier store needs a backup shelf (and restore procedure) as well, it needs to be documented in the backup/restore documentation and might be mentioned in the release notes as well (since there's an additional file to be backed up).

woutslakhorst:
closes #2317 

- extracted leia backup functionality from issuer_store to own file


woutslakhorst:
As you can see, there's quite some logic needed to separate JSON and JSON-LD collections. This is due to the interface on leia which has a `JSONCollection` and `JSONLDCollection` method. We could change that interface to a `Collection(type, name)` type of method and handle any invalid combinations inside leia.

The different query types would remain in the backup config.

woutslakhorst:
See nuts-foundation/go-leia#26

reinkrul:
Would be cool to revisit the backup/restore procedure as well

storage/leia.go
```diff
@@ -0,0 +1,300 @@
...
+	HandleRestore() error
+}
+
+type kvBackedLeiaStore struct {
+	store               leia.Store
+	backup              stoabs.KVStore
+	collectionConfigSet map[string]LeiaBackupConfiguration
+}
+
+// NewKVBackedLeiaStore creates a wrapper around a leia.Store that uses a stoabs.KVStore as backup for any documents added.
```
reinkrul, line 44:
```suggestion
// NewKVBackedLeiaStore creates a wrapper around a leia.Store that uses a stoabs.KVStore as backup. Write operations (add/delete/update) are first performed on the backup store, then on the leia store.
```

storage/leia.go
```diff
@@ -0,0 +1,300 @@
...
+func (k *kvBackedLeiaStore) AddConfiguration(config LeiaBackupConfiguration) {
+	k.collectionConfigSet[config.CollectionName] = config
+}
+
+func (k *kvBackedLeiaStore) HandleRestore() error {
+	// leia indices have been added, so the collection names have been added to the collectionConfigSet.
+	// Loop over this set to check if the backup store contains any documents for these collections and add them to the leia.Store if the named collection is empty there.
+	for _, config := range k.collectionConfigSet {
+		if err := k.handleRestore(config); err != nil {
+			return err
```
reinkrul, line 87: useful to add configset name to error?

storage/leia.go
```diff
@@ -0,0 +1,300 @@
...
+	case JSONCollectionType:
+		collection = k.store.JSONCollection(config.CollectionName)
+	default:
+		return errors.New("unknown collection type")
+	}
+	storePresent := storePresent(collection, config)
+
+	if backupPresent && storePresent {
+		// both are filled => normal operation, done
+		return nil
```
reinkrul, line 108: Although this didn't change, could it not be the store is not populated completely, and the restore is still required?

storage/leia.go
```diff
@@ -0,0 +1,300 @@
...
+
+func (k kvBackedCollection) DocumentCount() (int, error) {
+	return k.underlying.DocumentCount()
+}
+
+func (k *kvBackedLeiaStore) backupStorePresent(backupShelf string) bool {
+	backupPresent := false
+
+	_ = k.backup.ReadShelf(context.Background(), backupShelf, func(reader stoabs.Reader) error {
+		isEmpty, err := reader.Empty()
```
reinkrul, line 283: kinda dangerous to have an error default to backupPresent = false?

storage/leia.go
```diff
@@ -0,0 +1,300 @@
...
+		return err
+	})
+
+	return backupPresent
+}
+
+func storePresent(collection leia.Collection, config LeiaBackupConfiguration) bool {
+	issuedPresent := false
+	query := leia.New(leia.NotNil(config.SearchQuery))
+	_ = collection.IndexIterate(query, func(key []byte, value []byte) error {
```
reinkrul, line 294: same for this one

reinkrul (changes requested):

storage/leia.go
```diff
@@ -0,0 +1,300 @@
...
+}
+
+type kvBackedLeiaStore struct {
+	store               leia.Store
+	backup              stoabs.KVStore
+	collectionConfigSet map[string]LeiaBackupConfiguration
+}
+
+// NewKVBackedLeiaStore creates a wrapper around a leia.Store that uses a stoabs.KVStore as backup for any documents added.
+// The backup store is not closed when Close is called. The leia.Store is closed when Close is called.
```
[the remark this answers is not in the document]
woutslakhorst: because all stoabs stores are closed from the storage engine/a level higher.

storage/leia.go
woutslakhorst: There's no way to detect that other then the node didn't start on the previous run due to an error.

storage/leia.go
woutslakhorst: If it's a technical error, it'll try to fill the backup which also gives an error or it will rewrite some of the same values (hashes as keys).

reinkrul (approved):
✅ for the sake of avoiding scope creep

woutslakhorst:
@reinkrul replaced leia with v4 lib, that removed some ugly code. One last look?

Issue #2315, opened 2023-07-05
Reuse http connections for offering and receiving VCs


woutslakhorst:
(maybe also oauth via /n2n)

Currently http connections are not reused which means that for 100 vc/s, around 60k connections/file descriptors are needed. 1024 is usually a normal limit.

reinkrul:
> 100 vc/s, around 60k connections/file descriptors are needed

how is that?

gerardsn:
the performance test runs for 10min at 100 vc/s = 60k

woutslakhorst:
and 1 offer also means 1 credential request

reinkrul:
Apparently it's just a matter of making sure the response is fully read and closed.

https://golang.cafe/blog/how-to-reuse-http-connections-in-go.html

PR #2327, opened 2023-07-12
OpenID4VCI: Re-use http.Client for connection re-use

reinkrul:
Fixes https://github.com/nuts-foundation/nuts-node/issues/2315

We should probably test this using the performance test?


woutslakhorst:
Performance test says no:
Screenshot 2023-07-12 at 08 51 33

woutslakhorst:
I check the git hash from diagnostics to confirm the correct image was used to test.

woutslakhorst:
fixed:
Screenshot 2023-07-12 at 12 33 10

PR #2328, opened 2023-07-12
Remove DID restore from restore procedure

### LINKED ISSUE #2324: Backup/restore documentation is incorrect
- it lists didstore.db as something that needs to be restored but also a reprocess of dids is mentioned.
- the did reprocess is also the last in the order, it should be the first.

At some point we decided on a basic set of things to restore, just because it was easier to just tell people to make a backup of the entire data dir. This was also reflected with the restore (1:1). Seems this is incorrect/weird now.

What do we want? Only restoring 1 or 2 db files when making a backup for 5 seems weird (although functionally correct). Restoring all from file is uniform but without a reprocess could be broken (because different db files get a backup at a different time)

woutslakhorst:
closes #2324 

With the removal of the DID reprocess, the docs are consistent again.

### DIFF
--- a/docs/pages/deployment/backup-restore.rst
+++ b/docs/pages/deployment/backup-restore.rst
@@ -64,12 +64,11 @@ To restore a backup, follow the following steps:
 
 1. shutdown the node
 2. remove the following directories from the ``datadir``: ``events``, ``network``, ``vcr``, and ``vdr``
-3. remove the ``network.nodedid`` from your configuration file
-4. keep node offline by removing ``network.bootstrapnodes`` and ``network.grpcaddr`` from the config file and set ``network.enablediscovery`` to ``false``
-5. follow the restore procedure for your storage (BBolt, Redis, Hashicorp Vault)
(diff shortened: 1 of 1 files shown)


reinkrul:
Clearer PR title please

reinkrul:
We chose to have an additional store backed up (`vdr/didstore.db`) versus an extra reprocess (which got removed from docs). Why did we choose having an extra file to backup, instead of rebuilding it? The latter sounds more robust?

reinkrul (commented):
See question

gerardsn:
I think this is branching off of #2323? last commit is fine, but I will wait until that is merged to review this

woutslakhorst:
> I think this is branching off of #2323? last commit is fine, but I will wait until that is merged to review this

oops

woutslakhorst:
@gerardsn changed branch point

gerardsn (commented):
step 3 is no longer needed either? (and part of step 11)

PR #2338, opened 2023-07-12
Docs: update deprecated NGINX http2 directive

gerardsn:
Not sure if we should already fix this.
The [http2 directive](https://nginx.org/en/docs/http/ngx_http_v2_module.html#http2) is introduced in v1.25.1 that was released on 2023-06-13. Updating this now might be causing more issues with older versions than it is fixing.


woutslakhorst (approved):
we provide an example. Older versions of the docs are still available.

PR #2339, opened 2023-07-12
VDR: CMD and API use same defaults when creating a new DID

### LINKED ISSUE #2337: Creating node DID using CLI does not register keyAgreement key
When performing `nuts vdr create-did`, the resulting DID does not have a keyAgreement key, making it unusable as node DID:

```
 level=fatal msg="Could not start the server" error="unable to start Network: DID document does not contain a keyAgreement key, register a keyAgreement key (did=...)"
```

Although fixable (see below), it's unpractical.

```
nuts vdr add-keyagreement (key-id)
```

gerardsn:
closes #2337


docs/pages/deployment/cli-reference.rst
```diff
@@ -422,7 +422,7 @@ Registers a new DID
       --capabilityInvocation   Pass 'false' to disable capabilityInvocation capabilities. (default true)
       --controllers strings    Comma-separated list of DIDs that can control the generated DID Document.
   -h, --help                   help for create-did
-      --keyAgreement           Pass 'true' to enable keyAgreement capabilities.
+      --keyAgreement           Pass 'false' to disable keyAgreement capabilities. (default true)
```
reinkrul, line 425: I find these sentences hard to read, because they're inverted (applies to the other verification method flags as well). I'd rather read something like: "Whether to register a keyAgreement key on the DID. (default true)".

reinkrul (changes requested):
Better, but we could improve the flag docs when we're at it.

docs/pages/deployment/cli-reference.rst
gerardsn: I agree on hard to read, but it does provide a consistent indication of the default value. The generated code now contains the default value as `(default true)` only when the value deviates from the default value for the type in go. So when bool=false the default value is missing from the docs which I find confusing.

reinkrul (approved):

PR #2345, opened 2023-07-14
Add denylist e2e test

beardedfoo:


e2e-tests/denylist/run-tests.sh
gerardsn, line 1: Standard strategy is to test 1 thing at a time. So I would split this in 2 tests that go in the subfolders in a `run-test.sh`. This file then just points to those two tests. See e2e-tests/nuts-network for an example

e2e-tests/denylist/run-tests.sh
```diff
@@ -0,0 +1,157 @@
...
+	# Connect to the nuts-node with a valid client certificate
+	#curl --fail --cert client-allowed.crt --key client-allowed.key "https://localhost:1323/status/diagnostics"
+	openssl s_client -connect localhost:15555 -cert client-allowed.crt -key client-allowed.key -CAfile truststore-development.pem -verify_return_error -tls1_2 < <(echo "Hello Nuts 🥜")
+	if [ $? -ne 0 ]; then
+		echo "ERROR: failed to contact nuts-node-defaults with valid certificate"
+		exitWithDockerLogs 1
+	fi
+
+	# Check the logs have the right contents
+	docker logs denylist-nuts-node-defaults-1 2>&1 | tail -n1 | grep 'Validated certificate'
```
gerardsn, line 47: adding `container_name: <some-name>` to a service in the docker-compose file sets a reliable name to use instead of `denylist-nuts-node-defaults-1`

e2e-tests/denylist/run-tests.sh
```diff
@@ -0,0 +1,157 @@
...
+
+sleep 5
+
+# Simply to log this
+openssl version
+
+echo "------------------------------------"
+echo "Connecting (defaults) with allowed cert.."
+echo "------------------------------------"
+for x in $(seq 100); do
```
gerardsn, line 37: do we need to run all these tests 100 times? If the test is triggered on ever commit it runs ofter enough I'd say

gerardsn (changes requested):
all other tests currently use the certificates and truststore listed under e2e-tests/tls-certs which makes it easier to make changes or regenerate expired certificates etc. I'd prefer to use those as much as possible or add a similar script here? (we may need to regenerate the existing certificates to be valid for localhost anyway)

the `client-blocked` certificate is probably an exception since it needs to be added to the denylist if refreshed. Instructions for this should go in a readme.

gerardsn (approved):

PR #2355, opened 2023-07-17
Only set issuer.openidHandlerFn when vcr.OpenID4VCI.Enabled

gerardsn:
without this the issuer panics when issuing if vcr.OpenID4VCI.Enabled=false because vcr.localWalletResolver == nil

### DIFF
--- a/vcr/test/openid4vci_integration_test.go
+++ b/vcr/test/openid4vci_integration_test.go
@@ -190,6 +190,20 @@ func TestOpenID4VCIDisabled(t *testing.T) {
 		data, _ := io.ReadAll(resp.Body)
 		assert.Equal(t, `{"detail":"openid4vci is disabled","status":404,"title":"Operation failed"}`, string(data))
 	})
+	t.Run("Issues over network", func(t *testing.T) {
+		credential := testCredential()
+		credential.Issuer = walletDID.URI()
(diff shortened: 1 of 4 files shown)


vcr/test/openid4vci_integration_test.go
```diff
@@ -190,6 +190,20 @@ func TestOpenID4VCIDisabled(t *testing.T) {
...
+		credential.ID, _ = ssi.ParseURI(walletDID.URI().String() + "#1")
+		credential.CredentialSubject = append(credential.CredentialSubject, map[string]interface{}{
+			"id":           walletDID.URI().String(),
+			"purposeOfUse": "test",
+		})
+
+		vcrService := system.FindEngineByName("vcr").(vcr.VCR)
+		_, err := vcrService.Issuer().Issue(audit.TestContext(), credential, true, false)
+
+		assert.ErrorContains(t, err, "unable to publish the issued credential")
```
gerardsn, line 205: not really happy with this assertion, but I could not get NutsComm registered on localhost.

reinkrul: We could always make it an e2e test

reinkrul (approved):

woutslakhorst (approved):

gerardsn:
The panic only occurs when issuing to a node that does support openid4vci from a node that has it disabled. I will add a new e2e test for this.

PR #2358, opened 2023-07-19
OpenID4VCI: Move OAuth2 paths out of openid4vci subpath

reinkrul:
Make paths simpler;
- OAuth2 endpoints should be directly under the DID (`/token`)
- We don't need the `/issuer` and `/wallet` prefixes for openid4vci endpoints, since they don't overlap

### DIFF
--- a/docs/_static/vcr/openid4vci_v0.yaml
+++ b/docs/_static/vcr/openid4vci_v0.yaml
@@ -3,8 +3,8 @@ info:
   title: OpenID4VCI Issuer API
   version: 0.0.0
   description: >
-    This API is EXPERIMENTAL and implements a draft version of OpenID 4 Verifiable Credential Issuance.
-    It is subject to change and should not be used in production.
+    This API implements OpenID 4 Verifiable Credential Issuance.
+    The specification is in draft and may change, thus this API might change as well.
 servers:
   - url: "http://localhost:1323"
 paths:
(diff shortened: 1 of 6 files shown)


vcr/issuer/openid_test.go
```diff
@@ -110,7 +110,7 @@ func Test_memoryIssuer_ProviderMetadata(t *testing.T) {
 
 	assert.Equal(t, openid4vci.ProviderMetadata{
 		Issuer:        "https://example.com/did:nuts:issuer",
-		TokenEndpoint: "https://example.com/did:nuts:issuer/oidc/token",
+		TokenEndpoint: "https://example.com/did:nuts:issuer/token",
```
woutslakhorst, line 113: How come the tests do not have `/n2n/identity` in it?

reinkrul: Since we didn't include it in the issuer identifier in the test (line 42)

docs/_static/vcr/openid4vci_v0.yaml
gerardsn, line 1: line 5
>  description: >
    This API is EXPERIMENTAL and implements a draft version of OpenID 4 Verifiable Credential Issuance.
    It is subject to change and should not be used in production.

should this still be here? We removed the warnings everywhere, but none of these endpoints should be used directly, only by the nuts node thought discovery of the `.well-known` paths

reinkrul: It shouldn't

gerardsn (approved):

PR #2363, opened 2023-07-19
Add openid4vci disabled issuance test

### LINKED ISSUE #2362: VC issuance is always tried via OpenID4VCI, even if disabled in config
The nuts node always tries to issue via OpenID4VCI, even if explicitly disabled in the config. 

The issuer uses `issuer.openidHandlerFn==nil` to assert if openid is disabled, but the value is always set. This causes nil dereference panics downstream if the receiving node does support openid4vci issuance.

- [ ] #2355 
- [ ] Add e2e test to prevent this being re-introduces

gerardsn:
closes #2362 

mix-and match from `nuts-network/private-transactions` and `openid4vci/issuer-initiated` tests

### DIFF
--- a/e2e-tests/openid4vci/network-issuance/run-test.sh
+++ b/e2e-tests/openid4vci/network-issuance/run-test.sh
@@ -1,0 +1,60 @@
+#!/usr/bin/env bash
+
+set -e
+
+source ../../util.sh
+
+echo "------------------------------------"
+echo "Setting up fresh environment"
+echo "------------------------------------"
+docker compose down
+docker compose rm -f -v
+rm -rf ./node-*/data
+mkdir ./node-A/data ./node-B/data  # 'data' dirs will be created with root owner by docker if they do not exit. This creates permission issues on CI.
+docker compose up --wait
+
+echo "------------------------------------"
+echo "Creating NodeDIDs, waiting for Golden Hammer to register base URLs..."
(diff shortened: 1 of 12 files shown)


e2e-tests/openid4vci/network-issuance/run-test.sh
```diff
@@ -0,0 +1,70 @@
...
+
+echo "------------------------------------"
+echo "Creating NodeDIDs, waiting for Golden Hammer to register base URLs..."
+echo "------------------------------------"
+export NODEA_DID=$(setupNode "http://localhost:11323" "nodeA:5555")
+printf "NodeDID for node A: %s\n" "$NODEA_DID"
+waitForTXCount "NodeB" "http://localhost:21323/status/diagnostics" 2 10 # 2 for setupNode, 0 for GoldenHammer
+export NODEB_DID=$(setupNode "http://localhost:21323" "nodeB:5555")
+printf "NodeDID for node B: %s\n" "$NODEB_DID"
+waitForTXCount "NodeB" "http://localhost:21323/status/diagnostics" 5 10 # 2 for setupNode, 1 for GoldenHammer
```
woutslakhorst, line 34: When setting up B, wait for TX count at A?

woutslakhorst (approved):
One minor thing the test probably won't notice.

PR #2371, opened 2023-07-20
VDR: Introduce slimmed-down DIDResolver interface for implementing other DID methods

reinkrul:
E.g., did:web and did:jwk


gerardsn (commented):
the `ResolveMetadata` and `DocumentMetadata` are still very specific to the Nuts DID as they uses the transaction hashes. Can this change resolve other DID methods?

reinkrul:
> the ResolveMetadata and DocumentMetadata are still very specific to the Nuts DID as they uses the transaction hashes. Can this change resolve other DID methods?

Yeah, we'll probably have to extend it to accommodate for e.g. what did:web returns. Maybe add a property bag. But we don't need the resolution metadata for now.

vdr/types/interface.go
```diff
@@ -29,15 +29,20 @@ import (
...
 	// If metadata is not provided the latest version is returned.
 	// If metadata is provided then the result is filtered or scoped on that metadata.
 	// It returns ErrNotFound if there are no corresponding DID documents or when the DID Documents are disjoint with the provided ResolveMetadata
 	// It returns ErrDeactivated if the DID Document has been deactivated and metadata is unset or metadata.AllowDeactivated is false.
 	// It returns ErrNoActiveController if all of the DID Documents controllers have been deactivated and metadata is unset or metadata.AllowDeactivated is false.
 	Resolve(id did.DID, metadata *ResolveMetadata) (*did.Document, *DocumentMetadata, error)
+}
+
+// DocResolver is the interface that groups all the DID Document read methods.
+type DocResolver interface {
```
woutslakhorst, line 44: `ResolveControllers` could just be a function. It's only used for the Nuts method or nested from a resolver. It's used for mocking as well, but that could just be replaced with a resolve mock with the same outcome?

woutslakhorst: Comment due to having DIDResolver and DocResolver for which you can't deduce from the name what they do or what the difference is.

woutslakhorst (changes requested):

vdr/types/interface.go
reinkrul: DocResolver is renamed to `didnuts.DIDResolver` in a following PR

reinkrul:
Removed DocResolver and moved ResolveControllers to its own function

woutslakhorst (approved):

PR #2380, opened 2023-07-25
Add support for did:jwk resolution

beardedfoo:
This adds support for the resolving did:jwk decentralized identifiers per https://w3c-ccg.github.io/did-method-key/


vdr/didjwk/resolver.go
```diff
@@ -0,0 +1,102 @@
...
+	// Split the JWK ID on ':' characters to get the base64 encoded JWK
+	idParts := strings.Split(id.String(), ":")
+	if len(idParts) != 3 {
+		return nil, nil, fmt.Errorf("malformed did:jwk, expected 3 parts (did:jwk:...): %s", id.ID)
+	}
+	b64EncodedJWK := idParts[2]
+
+	// Remove any '#0' fragment suffix from the string
+	// See https://github.com/quartzjer/did-jwk/blob/6520a0edc8fa8f37c09af99efe841d54c3ca3b3b/spec.md#to-create-the-did-url
+	b64EncodedJWK, _ = strings.CutSuffix(b64EncodedJWK, "#0")
```
reinkrul, line 60: There won't be a prefix, since then it would be a DID URL (not the DID being resolved). I don't think it's needed.

Resolving keys involves resolving the DID first, then looking up the key inside the document.

If you want to be sure you can check `IsURL()` (meaning it contains a fragment, query or part)  on the DID and return an error if so

vdr/didjwk/resolver.go
```diff
@@ -0,0 +1,102 @@
...
+
+	// Extract the public key from the JWK
+	publicRawKey, err := jwk.PublicRawKeyOf(key)
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to get PublicRawKeyOf(key): %w", err)
+	}
+
+	// Create a new DID verification method.
+	// See https://www.w3.org/TR/did-core/#verification-methods	
+	verificationMethod, err := did.NewVerificationMethod(id, godid.JsonWebKey2020, id, publicRawKey)
```
reinkrul, line 82: This remove additional headers from the JWK, do we need to retain them (does the spec say anything about that)? Also, the ID of the verification method must include the `#0` fragment

vdr/didjwk/resolver.go
```diff
@@ -0,0 +1,102 @@
...
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to create verification method: %w", err)
+	}
+
+	// Create an empty DID document
+	var document did.Document
+
+	// Set the document ID
+	newID, err := did.ParseDIDURL("did:jwk:" + b64EncodedJWK + "#0")
```
reinkrul, line 91: DID excludes fragment

vdr/didjwk/resolver.go
```diff
@@ -0,0 +1,102 @@
...
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to create document id: %w", err)
+	}
+	document.ID = *newID
+
+	// Add the verification method
+	document.AddAssertionMethod(verificationMethod)
```
reinkrul, line 98: i think it needs to be added as VM relationship depending on a JWK header

reinkrul (changes requested):

vdr/didjwk/resolver.go
beardedfoo: I'll look into the retaining headers question.

The second point here has been addressed.

vdr/didjwk/resolver.go
beardedfoo: Let's chat about it

reinkrul:
JWKs containing a private key should probably be rejected, as per spec (Security section)

https://github.com/quartzjer/did-jwk/blob/main/spec.md#security

beardedfoo:
> JWKs containing a private key should probably be rejected, as per spec (Security section)
> 
> 
> 
> https://github.com/quartzjer/did-jwk/blob/main/spec.md#security

I implemented this with some tests, including a utility function which may be good to send over as a PR for the jwx/jwk library.

vdr/didjwk/utils.go
```diff
@@ -0,0 +1,36 @@
+package didjwk
+
+import (
+	"fmt"
+	"reflect"
+
+        "github.com/lestrrat-go/jwx/jwk"
+)
+
+// rawPrivateKeyOf returns the private key component of a jwk.Key, or nil if one is not available (e.g. public key only JWK's). An error is returned if a public key is not contained in the JWK.
```
reinkrul, line 10: elaborate on how it checks there's no private key in it

vdr/didjwk/resolver.go
```diff
@@ -0,0 +1,102 @@
...
+
+	// Extract the public key from the JWK
+	publicRawKey, err := jwk.PublicRawKeyOf(key)
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to get PublicRawKeyOf(key): %w", err)
+	}
+
+	// Create a new DID verification method.
+	// See https://www.w3.org/TR/did-core/#verification-methods	
+	verificationMethod, err := did.NewVerificationMethod(id, godid.JsonWebKey2020, id, publicRawKey)
```
reinkrul, line 82:
```suggestion
 	keyID := id.WithoutURL()
 	keyID.Fragment = "0"
	verificationMethod, err := did.NewVerificationMethod(keyID, godid.JsonWebKey2020, id.WithoutURL(), publicRawKey)
```

vdr/didjwk/resolver.go
```diff
@@ -0,0 +1,102 @@
...
+
+	// Create an empty DID document
+	var document did.Document
+
+	// Set the document ID
+	newID, err := did.ParseDIDURL("did:jwk:" + b64EncodedJWK)
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to create document id: %w", err)
+	}
+	document.ID = *newID
```
reinkrul, line 95:
```suggestion
	document.ID = id.WithoutURL()
```

reinkrul (changes requested):

reinkrul (approved):
Format code, then OK

vdr/didjwk/utils.go
```diff
@@ -0,0 +1,36 @@
...
+func rawPrivateKeyOf(key jwk.Key) (any, error) {
```
reinkrul, line 11: test coverage is low because there's no associated `_test.go` for this file. I'd say just move it to `resolver.go`

beardedfoo:
@reinkrul one more revision for go fmt

reinkrul:
I already approved, so you can just merge. Although diff coverage is below threshold, I looked at the actual coverage and think it's fine.

Issue #2359, opened 2023-07-19
Handle presentation_definition queries


woutslakhorst:
A large part of the development effort for openid4vp is to be expected in handling `presentation_definition`s. A defnition must be converted to a query on the wallet. The result is a list of matching VCs. The resulting VCs and the initial definition should also produce a `definition_submission`. A submission maps the requested definition parts onto the specific credentials in the list.

To support this we need:
- a `presentation_definition` parser
- converting the definition object to a leia query on the wallet
- using the resulting VCs and the definition to construct a `presentation_submission`

woutslakhorst:
I'm not sure if a query needs to be generated from a definition. We only have to search through a wallet which should not include a whole lot of VCs. Expecting AuthCreds to be replaced by inflow auth checks.

woutslakhorst:
The JSONPath expressions are not compatible with JSON-LD. For example the expression `$.type == VerifiableCredential` does not expand to `https://www.w3.org/2018/credentials/v1#VerifiableCredential` by magic.

We could just fetch all credentials for a wallet (exclude AuthCreds) and assume a limited set of credentials in a wallet at all times.

woutslakhorst:
Done since a presentation_definition will loop over all VCs in a wallet.

PR #2381, opened 2023-07-26
Add basic PresentationDefinition to VC matching

woutslakhorst:
part of #2359 

This PR introduces the required types and matching functionality for PresentationDefinitions.
It does not handle searching for credential candidates in the wallet.

It does not have a full feature set, but should be able to handle all of our use-cases.

the `pe` package stands for PresentationExchange. The family of standards that encompasses `PresentationDefinition` and `PresentationSubmission`

The choice for handling matching outside of a query is due to the feature set of a PresentationDefinition. It uses JSON schema and ECMAScript Regex constructs that are very hard to impossible to implement for a DB query.
Therefore a DB query should be simple and return (to) many results, the matching then reduces the result set.


vcr/pe/matcher_test.go
```diff
@@ -0,0 +1,391 @@
...
+
+var testCredentialString = `
+{
+  "type": "VerifiableCredential",
+  "credentialSubject": {
+	"field": "value"
+  }
+}`
+
+func TestMatch(t *testing.T) {
```
reinkrul, line 92: You have many tests asserting small individual part, but just 1 for the function (`Match`) that'll actually be called by the rest of the code (which is to come). For logical test coverage and understand of the reader, it'd help to have more tests for `Match`. Since tests for `Match` don't change when you change the internals of how it works (all the other smaller non-exported functions), but the other small tests will.

vcr/pe/schema/main.go
```diff
@@ -0,0 +1,96 @@
...
+	"net/http"
+	"net/url"
+	"os"
+	"time"
+)
+
+const (
+	claimFormatDescriptorLocation  = "http://identity.foundation/claim-format-registry/schemas/presentation-definition-claim-format-designations.json"
+	presentationDefinitionLocation = "https://github.com/decentralized-identity/presentation-exchange/raw/main/schemas/v2.0.0/presentation-definition.json"
+	//presentationSubmissionLocation = "https://github.com/decentralized-identity/presentation-exchange/raw/main/schemas/v2.0.0/presentation-submission.json"
```
reinkrul, line 34: why commented?

vcr/pe/schema/main.go
```diff
@@ -0,0 +1,96 @@
...
+
+	g := generate.New(schemas...)
+
+	err := g.CreateTypes()
+	if err != nil {
+		fmt.Fprintln(os.Stderr, "Failure generating structs: ", err)
+		os.Exit(1)
+	}
+
+	f, err := os.OpenFile("../generated.go", os.O_CREATE|os.O_WRONLY|os.O_TRUNC, 0644)
```
reinkrul, line 48: missing this file

vcr/pe/schema/main.go
```diff
@@ -0,0 +1,96 @@
...
+	"time"
+)
+
+const (
+	claimFormatDescriptorLocation  = "http://identity.foundation/claim-format-registry/schemas/presentation-definition-claim-format-designations.json"
+	presentationDefinitionLocation = "https://github.com/decentralized-identity/presentation-exchange/raw/main/schemas/v2.0.0/presentation-definition.json"
+	//presentationSubmissionLocation = "https://github.com/decentralized-identity/presentation-exchange/raw/main/schemas/v2.0.0/presentation-submission.json"
+)
+
+func main() {
```
reinkrul, line 37: add to makefile I guess?

vcr/pe/matcher.go
```diff
@@ -0,0 +1,288 @@
...
+func matchCredential(descriptor InputDescriptor, credential vc.VerifiableCredential) (bool, error) {
+	// for each constraint in descriptor.constraints:
+	//   a vc must match the constraint
+	if descriptor.Constraints != nil {
+		return matchConstraint(descriptor.Constraints, credential)
+	}
+	return true, nil
+}
+
+// matchConstraint matches the constraint against the VC.
```
reinkrul, line 144: should we error out (or non-match and log a warning) when a VP request uses an unsupported contstraint?

reinkrul (changes requested):

vcr/pe/fields.go
```diff
@@ -0,0 +1,42 @@
...
+	Id             *string  `json:"id,omitempty"`
+	Optional       *bool    `json:"optional,omitempty"`
+	Path           []string `json:"path"`
+	Purpose        *string  `json:"purpose,omitempty"`
+	Name           *string  `json:"name,omitempty"`
+	IntentToRetain *bool    `json:"intent_to_retain,omitempty"`
+	Filter         *Filter  `json:"filter,omitempty"`
+}
+
+// Filter is a JSON Schema descriptor
```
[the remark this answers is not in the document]
woutslakhorst: They just call it a `JSON Schema` in the JSON schema spec. Will adjust

vcr/pe/fields.go
```diff
@@ -0,0 +1,42 @@
...
+type Filter struct {
+	// Type is the type of field: string, number, boolean, array, object
+	Type string `json:"type"`
+	// Const is a constant value to match, currently only strings are supported
+	Const *string `json:"const,omitempty"`
+	// Enum is a list of values to match
+	Enum *[]string `json:"enum,omitempty"`
```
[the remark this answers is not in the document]
woutslakhorst: result of code generation, will remove *

vcr/pe/matcher.go
```diff
@@ -0,0 +1,288 @@
...
+}
+
+// matchFormat checks if the credential matches the Format from the presentationDefinition.
+// if one of format['ldp_vc'] or format['jwt_vc'] is present, the VC must match that format.
+// If the VC is of the required format, the alg or proofType must also match.
+// vp formats are ignored.
+// This might not be fully interoperable, but the spec at https://identity.foundation/presentation-exchange/#presentation-definition is not clear on this.
+func matchFormat(format *PresentationDefinitionClaimFormatDesignations, credential vc.VerifiableCredential) bool {
+	if format == nil {
+		return true
```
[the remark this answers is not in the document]
woutslakhorst: Format is optional, if no Format is given you can return whatever type of VC you want.
I do think the spec will become a bit clearer on this in future versions.

vcr/pe/matcher.go
```diff
@@ -0,0 +1,288 @@
...
+	}
+
+	asMap := map[string]map[string][]string(*format)
+	// we're only interested in the jwt_vc and ldp_vc formats
+	if asMap["jwt_vc"] == nil && asMap["ldp_vc"] == nil {
+		return true
```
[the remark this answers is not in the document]
woutslakhorst: Same as previous comment, if nothing is requested regarding VCs, you can return any type of VC.

vcr/pe/matcher.go
```diff
@@ -0,0 +1,288 @@
...
+	for i, inputDescriptor := range presentationDefinition.InputDescriptors {
+		var mapping *InputDescriptorMappingObject
+		var err error
+		for _, credential := range vcs {
+			mapping, err = matchDescriptor(*inputDescriptor, credential)
+			if err != nil {
+				return PresentationSubmission{}, nil, err
+			}
+			if mapping != nil && matchFormat(presentationDefinition.Format, credential) {
+				mapping.Path = fmt.Sprintf("$.verifiableCredential[%d]", i)
```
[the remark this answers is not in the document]
woutslakhorst: Good catch, should be `index++`. Will add test to confirm.

vcr/pe/matcher.go
```diff
@@ -0,0 +1,288 @@
...
+	//   a vc must match the constraint
+	if descriptor.Constraints != nil {
+		return matchConstraint(descriptor.Constraints, credential)
+	}
+	return true, nil
+}
+
+// matchConstraint matches the constraint against the VC.
+// All Fields need to match according to the Field rules.
+// IsHolder, SameSubject, SubjectIsIssuer, Statuses are not supported for now.
```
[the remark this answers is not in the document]
woutslakhorst: They are under the §7 Feature section of the spec (https://identity.foundation/presentation-exchange/#features)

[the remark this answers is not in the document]
woutslakhorst: Features are optional

vcr/pe/matcher.go
```diff
@@ -0,0 +1,288 @@
...
+		if !match {
+			return false, nil
+		}
+	}
+	return true, nil
+}
+
+// matchField matches the field against the VC.
+// All fields need to match unless optional is set to true and no values are found for all the paths.
+func matchField(field Field, credential vc.VerifiableCredential) (bool, error) {
```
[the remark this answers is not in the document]
woutslakhorst: I'll add some to the doc for `Match`

vcr/pe/matcher.go
```diff
@@ -0,0 +1,288 @@
...
+
+	if filter.Pattern != nil && filter.Type == "string" {
+		re, err := regexp2.Compile(*filter.Pattern, regexp2.ECMAScript)
+		if err != nil {
+			return false, err
+		}
+		return re.MatchString(value.(string))
+	}
+
+	return true, nil
```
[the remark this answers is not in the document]
woutslakhorst: patterns can only be strings, so any other type will give an unmarshall error. Other cases are handled by the switch. So if you get here, you only want type to match. Will add comment.

vcr/pe/schema/main.go
woutslakhorst: I wanted to generate code from it, but it's broken.

vcr/pe/schema/main.go
woutslakhorst: You're not missing anything, trust me ;).
The generated code is more a suggestion (because it sucks), it can be used to generated code from updated JSON schema definitions with newer versions of the spec.

vcr/pe/schema/main.go
woutslakhorst: Nah, types.go doesn't use the generated code 1-on-1, more as a suggestion.

vcr/pe/matcher.go
woutslakhorst: that will be up to the caller of the function. This package just implements (some) features of the PresentationExchange spec. I guess the calling piece of code will log a warning/error and return false/403 when it happens.

vcr/pe/matcher_test.go
woutslakhorst: The smaller functions won't change unless the spec changes. They are a 1-on-1 copy of the bullet points of the spec. A `presentation_definition` is quite verbose so you probably want individual tests to use the struct directly which basically is the same as the smaller tests right not. I would rather add another larger real world test case when we have one.

vcr/pe/matcher.go
```diff
@@ -0,0 +1,288 @@
...
+// - ldp_vc format
+// - pattern, const and enum only on string fields
+// - number, boolean, array and string JSON schema types
+// It doesn't do the credential search, this should be done before calling this function.
+// The PresentationDefinition.Format should be altered/set if an envelope defines the supported format before calling.
+// The resulting PresentationSubmission has paths that are relative to the matching VCs.
+// The PresentationSubmission needs to be altered so the paths use "path_nested"s that are relative to the created VP.
+// ErrUnsupportedFilter is returned when a filter uses unsupported features.
+// Other errors can be returned for faulty JSON paths or regex patterns.
+func Match(presentationDefinition PresentationDefinition, vcs []vc.VerifiableCredential) (PresentationSubmission, []vc.VerifiableCredential, error) {
```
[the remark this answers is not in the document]
woutslakhorst: I'll try and check.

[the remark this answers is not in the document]
woutslakhorst: I guess it would help since the IDE would suggest it through code completion.

vcr/pe/matcher.go
reinkrul: So the matching is a suggestion, but the caller still has to decide whether to actually present the credentials? And if the requester specifies an unsupported constraint, a VC might be matched that doesn't match in reality.

Although this is a larger question: I can imagine this is not a big problem in a mobile wallet, since a person still has to approve the presentation, but on what grounds would a server wallet (without a human making decisions) decide when to/not to present the resulting VCs?

reinkrul (approved):

vcr/pe/matcher.go
woutslakhorst: This functionality will not return a non-match, it could return an empty list. Then the caller must log and return an error.

PR #2382, opened 2023-07-27
did:web support

reinkrul:
Uses https://github.com/nuts-foundation/nuts-node/pull/2373 (needs to be merged first)


vdr/didweb/web.go
```diff
@@ -0,0 +1,116 @@
...
+	defer httpResponse.Body.Close()
+	if !(httpResponse.StatusCode >= 200 && httpResponse.StatusCode < 300) {
+		return nil, nil, fmt.Errorf("did:web non-ok HTTP status: %s", httpResponse.Status)
+	}
+
+	ct, _, err := mime.ParseMediaType(httpResponse.Header.Get("Content-Type"))
+	if err != nil {
+		return nil, nil, fmt.Errorf("did:web invalid content-type: %w", err)
+	}
+	switch ct {
```
gerardsn, line 84: §2.4 of the [spec](https://w3c-ccg.github.io/did-method-web/#key-material-and-document-handling) expects most `/did.json` to be served as `application/json` (it's in the file extension), including LD documents. So LD documents are detected based on the presence of an `@context`. Can we process it using these rules?

vdr/didweb/web.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+	var document did.Document
+	err = document.UnmarshalJSON(data)
+	if err != nil {
+		return nil, nil, fmt.Errorf("did:web JSON unmarshal error: %w", err)
+	}
+
+	if !document.ID.Equals(id.WithoutURL()) {
+		return nil, nil, fmt.Errorf("did:web document ID mismatch: %s != %s", document.ID, id.WithoutURL())
+	}
```
gerardsn, line 113: this is not specced, but probably fine?

gerardsn (changes requested):
Section 2.3 of the spec contains
> The method specific identifier MUST match the common name used in the SSL/TLS certificate, and it MUST NOT include IP addresses.

vdr/didweb/web_test.go
gerardsn, line 1: add test with port number

reinkrul:
> Section 2.3 of the spec contains
> 
> > The method specific identifier MUST match the common name used in the SSL/TLS certificate, and it MUST NOT include IP addresses.

This is implemented, as the DID itself contains the domain name, from which the URL to load the document from is derived. Thus, when using HTTPS (which is required by the method spec) Go's TLS library handles verification of the host name (certificate.san matches certificate). No need to do it ourselves.

vdr/didweb/web.go
```diff
@@ -0,0 +1,116 @@
...
+	var path string
+	subpathIdx := strings.Index(id.ID, ":")
+	if subpathIdx == -1 {
+		path = "/.well-known/did.json"
+	} else {
+		// subpaths are encoded as / -> :
+		baseID = id.ID[:subpathIdx]
+		path = id.ID[subpathIdx:]
+		path = strings.ReplaceAll(path, ":", "/") + "/did.json"
+	}
```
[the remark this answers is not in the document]
reinkrul: https://w3c-ccg.github.io/did-method-web/#method-specific-identifier states a `:` must be followed by a path, so we should probably add check to reject DIDs with empty paths (`:: -> //` is an empty path, and a DID ending with a `:` also contains an empty path)

vdr/didweb/web.go
reinkrul: We _could_ process them (and it would probably make the Nuts Node a purer did:web resolver), but we don't need it for our use cases (see comment on line 86). And we don't do it for `did:nuts` documents anyways, so at least the behavior is the same for how we treat `did:web`.

vdr/didweb/web.go
reinkrul: I thought I read it somewhere, but I probably saw it in the reference implementation.

https://github.com/decentralized-identity/web-did-resolver/blob/master/src/resolver.ts#L40

vdr/didweb/web_test.go
reinkrul: All test actually use a port number, I'll add a test without one

vdr/didweb/web.go
```diff
@@ -0,0 +1,116 @@
...
+
+	unescapedID, err := url.PathUnescape(baseID)
```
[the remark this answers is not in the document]
reinkrul: For supporting percent-encoded characters in the DID. You could theoretically have no "ordinary" ANSI characters in the DID's ID/path, and have all of them in hex-encoding. Also required for supporting a port number (which uses percent-encoding for `:`, to avoid having it interpreted as a subpath).

vdr/didweb/web.go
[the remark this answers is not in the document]
gerardsn: sounds like a good check to add

vdr/didweb/web.go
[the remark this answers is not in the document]
gerardsn: The spec only requires restoring the port
> If the domain contains a port percent decode the colon.

encoding https://example.com/path as `did:web:example.com%2Fpath` is invalid according to `did:web` but passes these checks. I don't know how this should be handled or if it matters, just making an observation.

vdr/didweb/web.go
gerardsn: The spec also mentions rejecting any document containing `@context` that fail JSON-LD parsing. Ignoring this only means we are a bit relaxed on validation? Probably fine as long as we don't serve did:web documents later on that adopt this relaxed view.

gerardsn (changes requested):
We still need to verify that it does not contain an IP. the stdlib only validates that the information on the cert matches the request, which will pass if the IP is on the cert.

vdr/didweb/web.go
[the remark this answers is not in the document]
reinkrul: I will add a check that checks if the hostname in the parsed URL equals the method-specific ID, that should cover these cases

vdr/didweb/web.go
```diff
@@ -0,0 +1,116 @@
+package didweb
```
woutslakhorst, line 1: Hah, copyright

woutslakhorst (changes requested):
The spec doesn't say anything on caching, but should we add it anyway (new issue)?

reinkrul:
> The spec doesn't say anything on caching, but should we add it anyway (new issue)?

Caching is sure something to support, for out-of-scope for this issue i.m.o.


https://github.com/nuts-foundation/nuts-node/issues/2420

woutslakhorst (approved):

vdr/didweb/web.go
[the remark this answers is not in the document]
gerardsn: Am I just not seeing this check or have you decided against it?

gerardsn (commented):
just one question left

gerardsn (approved):

Issue #2311, opened 2023-07-05
Have VDR diagnostics conflict count use cached isOwned


woutslakhorst:
the diagnostics in the VDR loop over all conflicted did documents to get the controller and check is a key exists. It should be using the cached isOwned call now available in the vdr.

reinkrul:
For testing improvement:
- On my stable node, requesting diagnostics (checking 24 conflicts, none owned) took between 230ms and 290ms
- When using the cached IsOwned function: 

This is with keys on filesystems, when using external storage it should yield move improvement.

PR #2386, opened 2023-07-28
VDR: Use cached IsOwned function for finding owned conflicts

reinkrul:
Semantics are slightly different (now any private key matches) but it's for diagnostics only, so an extra false positive in a super-super edge case doesn't hurt.

Fixes https://github.com/nuts-foundation/nuts-node/issues/2311


reinkrul:
For testing improvement:
- On my stable node, requesting diagnostics (checking 24 conflicts, none owned) took between 230ms and 290ms
- When using the cached IsOwned function it gave slightly lower, but very comparable  results.

This is with keys on filesystems, when using external storage it should yield move improvement.

woutslakhorst (approved):

PR #2396, opened 2023-08-02
Enable did:jwk resolving

reinkrul:


gerardsn (commented):
do we need a test to check all resolvers are there? (very basic like cmd/root_test.go#Test_CreateSystem?)

reinkrul:
I'm thinking more of integration-style tests, but not sure how

reinkrul:
For did:web I wrote a simple test resolving a stubbed did:web document (see https://github.com/nuts-foundation/nuts-node/pull/2382). I will do the same for did:jwk after the did:web PR is merged.

End-to-end doesn't make much sense for now (at least for did:web), since we can't serve did:web from the Nuts node yet. did:jwk is mainly meant for mobile wallets, so an end-to-end test would involve mimicking one. But that part of an OpenID4VP flow, which we have yet to build.

gerardsn (approved):

PR #2407, opened 2023-08-07
ResolveControllers should return controllers that can be resolved

### LINKED ISSUE #2405: Investigate broken sync on master
v5.4.1 syncs perfectly but my 2nd development node stops at 523 TXs. (Uses redis)

woutslakhorst:
fixes #2405 

only broken on master, probably due to the resolver refactor.
Added a test case to reproduce.

The resolver#ResolveControllers returned an error if one of the controllers couldn't be resolved. This is always the case when you resolve with a sourceTX in the resolveMetadata. On the development network this will always fail at tx:555. The ResolveControllers method should not stop on ErrNotFound but it should continue with resolving the other controllers.

### DIFF
--- a/vdr/didnuts/resolver.go
+++ b/vdr/didnuts/resolver.go
@@ -101,7 +101,7 @@ func resolveControllers(resolver types.DIDResolver, doc did.Document, metadata *
 	// resolve all unresolved doc
 	for _, ref := range refsToResolve {
 		node, _, err := resolve(resolver, ref, metadata, depth)
-		if errors.Is(err, types.ErrDeactivated) || errors.Is(err, types.ErrNoActiveController) {
+		if errors.Is(err, types.ErrDeactivated) || errors.Is(err, types.ErrNoActiveController) || errors.Is(err, types.ErrNotFound) {
 			continue
 		}
 		if errors.Is(err, ErrNestedDocumentsTooDeep) {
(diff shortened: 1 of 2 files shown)


vdr/didnuts/resolver.go
```diff
@@ -101,7 +101,7 @@ func resolveControllers(resolver types.DIDResolver, doc did.Document, metadata *
 	// resolve all unresolved doc
 	for _, ref := range refsToResolve {
 		node, _, err := resolve(resolver, ref, metadata, depth)
-		if errors.Is(err, types.ErrDeactivated) || errors.Is(err, types.ErrNoActiveController) {
+		if errors.Is(err, types.ErrDeactivated) || errors.Is(err, types.ErrNoActiveController) || errors.Is(err, types.ErrNotFound) {
 			continue
 		}
 		if errors.Is(err, ErrNestedDocumentsTooDeep) {
```
gerardsn, line 107: Should `ErrNestedDocumentsTooDeep` get the same treatment instead of returning an error?

gerardsn (approved):

vdr/didnuts/resolver.go
woutslakhorst: Hmm complicated, I think it's slightly different. Can't oversee what the consequences would be

PR #2408, opened 2023-08-07
Only request 2 pages per conversation on full sync

### LINKED ISSUE #2406: Node requests too many TX at sync
When missing all TXs a node could request All from a peer. When there are more and more TXs, this could cause for too much strain on the peer. It should limit to a couple (2 or 3) of pages?

woutslakhorst:
fixes #2406 

When a node does a full sync it will request missing pages. This could result in a request of thousands of transactions. The peer sending those will not handle this request in a streaming way and will thus consume a lot of resources to fulfil the request. Syncing goes much smoother when only requesting 2 pages at a time. If more pages are requested, this PR will also change that the peer only sends 2.

### DIFF
--- a/network/transport/v2/handlers.go
+++ b/network/transport/v2/handlers.go
@@ -243,12 +243,12 @@ func (p *protocol) handleTransactionRangeQuery(ctx context.Context, connection g
 
 	// limit to two pages to reduce load
 	limit := msg.Start + 2*dag.PageSize
-	requested := msg.End
-	if requested > limit {
-		requested = limit
+	end := msg.End
+	if end > limit {
+		end = limit
 	}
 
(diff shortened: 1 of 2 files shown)


network/transport/v2/handlers.go
```diff
@@ -241,7 +241,14 @@ func (p *protocol) handleTransactionRangeQuery(ctx context.Context, connection g
 		return errors.New("invalid range query")
 	}
 
-	txs, err := p.state.FindBetweenLC(ctx, msg.Start, msg.End)
+	// limit to two pages to reduce load
+	limit := msg.Start + 2*dag.PageSize
+	requested := msg.End
```
gerardsn, line 246: Rename `requested` to `end`? Since `limit != requested`, or at least not always.

gerardsn (approved):

PR #2414, opened 2023-08-07
Fix/2402/retry events synchronously at startup

### LINKED ISSUE #2402: Failed events are retried asynchronous at startup
When having hundreds of events this causes DB timeouts because those events can't be handled asynchronously at that scale. At startup they should be handled synchronously.

woutslakhorst:
fixes #2402 

the call to retry is replaced with `notifyNow` for each event that failed in the past. The events are tried one by one. All events that still failed are then scheduled in the retry loop.

This will reduce the number of errors on startup due to DB lock errors. Events that still fail due to other causes still fail and will still be rescheduled.

### DIFF
--- a/network/dag/notifier.go
+++ b/network/dag/notifier.go
@@ -46,6 +46,9 @@ const (
 
 var timeFunc = time.Now
 
+// maxJitter is adjustable for testing purposes
+var maxJitter = time.Second
+
 // EventFatal signals that an Event receiver encountered a fatal error and that the Event should not be retried.
 type EventFatal struct {
 	Err error
@@ -360,8 +363,9 @@ func (p *notifier) retry(event Event) {
 		},
 			retry.Attempts(maxRetries-uint(initialCount)),
 			retry.MaxDelay(24*time.Hour),
+			retry.MaxJitter(maxJitter),
 			retry.Delay(delay),
-			retry.DelayType(retry.BackOffDelay),
+			retry.DelayType(retry.CombineDelay(retry.BackOffDelay, retry.RandomDelay)),
 			retry.Context(ctx),
 			retry.LastErrorOnly(true),
 			retry.OnRetry(func(n uint, err error) {
--- a/network/dag/notifier.go
+++ b/network/dag/notifier.go
@@ -37,7 +37,7 @@ import (
 const (
 	defaultRetryDelay      = time.Second
 	retriesFailedThreshold = 10
-	maxRetries             = 100
+	maxRetries             = 20
 	// TransactionEventType is used as Type in an Event when a transaction is added to the DAG.
 	TransactionEventType = "transaction"
 	// PayloadEventType is used as Type in an Event when a payload is written to the DB.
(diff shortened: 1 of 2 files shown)


network/dag/notifier.go
```diff
@@ -239,16 +239,31 @@ func (p *notifier) Run() error {
...
+	failedAtStartup := make([]Event, 0)
+	err := p.db.ReadShelf(p.ctx, p.shelfName(), func(reader stoabs.Reader) error {
 		return reader.Iterate(func(k stoabs.Key, v []byte) error {
 			event := Event{}
 			_ = json.Unmarshal(v, &event)
 
-			p.retry(event)
+			if err := p.notifyNow(event); err != nil {
+				failedAtStartup = append(failedAtStartup, event)
+			}
```
gerardsn, line 251: this changes some logic. Events with `Event.Retries >= maxRetries` should not be retried.

gerardsn (changes requested):
This adds a requirement on the startup order of the modules. Network MUST be started after all modules that subscribe to the notifier for this change to work (which it currently does).

The DB issue arrises when there are many events with the same number of `Event.Retries` since these all get the same initial delay in `p.retry`. This could also be solved by adding some jitter to the initial delay.

network/dag/notifier.go
woutslakhorst: > this changes some logic. Events with `Event.Retries >= maxRetries` should not be retried.

That is already the case @ startup

gerardsn: I think the checks are in the wrong order? It still retries when the maxRetries is reached. (might also fix the broken test)

gerardsn: hmm... I see what you're saying. Everything is retried at startup after the initial delay, which is 2^100 sec == never for events that hit an `EventFatal` error. `p.retry` is basically collecting go routines that sleep forever...?

gerardsn: time.Second^100 overflows and results in `0s` in go, so it retries immediately. I'd say that this is a different bug? 

Also, `retry.Attempts(maxRetries-uint(initialCount)),` in `p.retry` result in infinite retries when `initialCount== maxRetries`

mehhh

woutslakhorst: Added Jitter as well. Hopefully the test isn't going to random fail

gerardsn (approved):
Seems reasonable. Just a nasty race condition left according to the tests.

woutslakhorst:
@gerardsn reduced max retry to 20 to prevent overflow. Set jitter to equal initialDelay to prevent race conditions in test. Jitter and initialDelay were both 1s... Also helps in test where sometimes initialDelay is 1ns

network/dag/notifier.go
```diff
@@ -330,21 +347,24 @@ func (p *notifier) Notify(event Event) {
...
 	}
 
 	go func(ctx context.Context) {
-		// also an initial delay
-		time.Sleep(delay)
 		err := retry.Do(func() error {
 			return p.notifyNow(event)
 		},
-			retry.Attempts(maxRetries-uint(initialCount)),
+			retry.Attempts(attempts),
```
gerardsn, line 363: `attempts==0` means infinite Attempts

woutslakhorst: anything <0 and >max will just return the retry.

PR #2423, opened 2023-08-12
Flexible OAuth2 Server POC

reinkrul:
PoC for https://github.com/nuts-foundation/nuts-node/issues/2424

Missing bits:
- Linking up with existing OpenID4VCI handlers in VCR issuer/holder
- Session management, maybe allowing `/token` for code flow to be generalized
- Registering the new API in the Nuts Node

Stuff to discuss:
- Applicable approach?
- What module should it reside in
- What API package

### DIFF
--- a/vcr/api/oauth2/v0/api.go
+++ b/vcr/api/oauth2/v0/api.go
@@ -1,0 +1,75 @@
+package v0
+
+import (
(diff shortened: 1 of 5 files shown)


reinkrul:
Going to try an approach with OpenAPI first

Issue #2428, opened 2023-08-16
NutsEmployeeCredential does not need to be trusted


reinkrul:
The following error occurs:

```
"Unable to create access token, probably due to JWT grant token validation" authorizerDID="did:nuts:7yu19NYWPq97dnRyv16KnLnx3kAkUYHqbMkbTxJDMBg1" error="identity validation failed: verification error: invalid VC (id=did:nuts:53ie6BBv5JdcFrRw9bi4aaiKDXwHTi5wYnBqzgRjTGYd#7be1ac4f-5e69-4575-8944-4a9618e2e08b): credential issuer is untrusted" module=Auth requesterDID="did:nuts:53ie6BBv5JdcFrRw9bi4aaiKDXwHTi5wYnBqzgRjTGYd
```

This concerns the trust-status of the NutsEmployeeCredential.

We have an end-to-end test, but it uses just 1 Nuts node, which will always work since issuing the credential will automatically trust it.

It should rather derive trust from the fact that it was issued by a trusted care organization (meaning it has a NutsOrganizationCredential whose issuer is trusted).

PR #2429, opened 2023-08-17
Derive trust for NutsEmployeeCredential from NutsOrganizationCredential

reinkrul:
Fixes https://github.com/nuts-foundation/nuts-node/issues/2428

### DIFF
--- a/auth/services/selfsigned/validator.go
+++ b/auth/services/selfsigned/validator.go
@@ -141,7 +141,7 @@ func (v validator) verifyVP(vp vc.VerifiablePresentation, validAt *time.Time) (c
 
 	// #2428: NutsEmployeeCredential trust is derived from the fact that the issuer has a trusted NutsOrganizationCredential
 	searchTerms := []vcr.SearchTerm{
-		{IRIPath: jsonld.CredentialSubjectPath, Value: vc.Issuer.String()},
+		{IRIPath: jsonld.CredentialSubjectPath, Value: credentialSubject.ID},
 		{IRIPath: jsonld.OrganizationNamePath, Type: vcr.NotNil},
 		{IRIPath: jsonld.OrganizationCityPath, Type: vcr.NotNil},
 	}
--- a/auth/services/selfsigned/validator.go
+++ b/auth/services/selfsigned/validator.go
@@ -145,7 +145,7 @@ func (v validator) verifyVP(vp vc.VerifiablePresentation, validAt *time.Time) (c
 		{IRIPath: jsonld.OrganizationNamePath, Type: vcr.NotNil},
 		{IRIPath: jsonld.OrganizationCityPath, Type: vcr.NotNil},
 	}
-	nutsOrgCreds, err := v.vcr.Search(context.Background(), searchTerms, false, validAt)
+	nutsOrgCreds, err := v.vcr.Search(context.TODO(), searchTerms, false, validAt)
 	if err != nil {
 		resultErr = fmt.Errorf("unable to check NutsEmployeeCredential trust status using NutsOrganizationCredential: %w", err)
 		return
--- a/auth/services/selfsigned/validator.go
+++ b/auth/services/selfsigned/validator.go
@@ -19,13 +19,16 @@
 package selfsigned
 
 import (
+	"context"
 	"encoding/json"
 	"errors"
+	"fmt"
 	ssi "github.com/nuts-foundation/go-did"
 	"github.com/nuts-foundation/go-did/vc"
 	"github.com/nuts-foundation/nuts-node/auth/contract"
 	"github.com/nuts-foundation/nuts-node/auth/services"
 	"github.com/nuts-foundation/nuts-node/auth/services/selfsigned/types"
+	"github.com/nuts-foundation/nuts-node/jsonld"
 	"github.com/nuts-foundation/nuts-node/vcr"
 	"github.com/nuts-foundation/nuts-node/vcr/verifier"
 	"time"
@@ -92,7 +95,8 @@ func (v validator) VerifyVP(vp vc.VerifiablePresentation, validAt *time.Time) (c
 }
 
 func (v validator) verifyVP(vp vc.VerifiablePresentation, validAt *time.Time) (credentialSubject types.EmployeeIdentityCredentialSubject, proof vc.JSONWebSignature2020Proof, resultErr error) {
-	vcs, err := v.vcr.Verifier().VerifyVP(vp, true, validAt)
+	// #2428: NutsEmployeeCredential should be valid (signature), but does not need to be trusted.
+	vcs, err := v.vcr.Verifier().VerifyVP(vp, true, false, validAt)
 	if err != nil {
 		if errors.As(err, &verifier.VerificationError{}) {
 			resultErr = newVerificationError(err.Error())
@@ -132,6 +136,22 @@ func (v validator) verifyVP(vp vc.VerifiablePresentation, validAt *time.Time) (c
 	credentialSubject = credentialSubjects[0]
 	if vc.Issuer.String() != credentialSubject.ID {
 		resultErr = newVerificationError("signer must be credentialSubject")
+		return
+	}
+
+	// #2428: NutsEmployeeCredential trust is derived from the fact that the issuer has a trusted NutsOrganizationCredential
+	searchTerms := []vcr.SearchTerm{
+		{IRIPath: jsonld.CredentialSubjectPath, Value: vc.Issuer.String()},
+		{IRIPath: jsonld.OrganizationNamePath, Type: vcr.NotNil},
+		{IRIPath: jsonld.OrganizationCityPath, Type: vcr.NotNil},
+	}
+	nutsOrgCreds, err := v.vcr.Search(context.Background(), searchTerms, false, validAt)
+	if err != nil {
+		resultErr = fmt.Errorf("unable to check NutsEmployeeCredential trust status using NutsOrganizationCredential: %w", err)
+		return
+	}
+	if len(nutsOrgCreds) == 0 {
+		resultErr = newVerificationError("invalid NutsEmployeeCredential, issuer does not have a trusted NutsOrganizationCredential")
 		return
 	}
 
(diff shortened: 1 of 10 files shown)


vcr/verifier/interface.go
```diff
@@ -48,8 +48,8 @@ type Verifier interface {
 	RegisterRevocation(revocation credential.Revocation) error
 
 	// VerifyVP verifies the given Verifiable Presentation. If successful, it returns the credentials within the presentation.
-	// If verifyVCs is true, it will also verify the credentials inside the VP, checking their correctness, signature and trust status.
-	VerifyVP(presentation vc.VerifiablePresentation, verifyVCs bool, validAt *time.Time) ([]vc.VerifiableCredential, error)
+	// If verifyVCs is true, it will also verify the credentials inside the VP, checking their correctness, signature and trust status (unless verifyVCsTrusted is false).
+	VerifyVP(presentation vc.VerifiablePresentation, verifyVCs bool, verifyVCsTrusted bool, validAt *time.Time) ([]vc.VerifiableCredential, error)
```
gerardsn, line 52: As far as I can tell `verifyVCsTrusted ==  !allowUntrusted` (from `Verify()` in the same interface)? If so, I'd prefer to use `allowUntrusted` here too to make it easier to understand. (partially because I don't find `verifyVCsTrusted` very clear)

e2e-tests/auth/selfsigned/apps/selfsigned.go
```diff
@@ -91,12 +92,28 @@ func (s SelfSigned) GetSessionStatus(sessionID string) (string, *authAPI.Verifia
 	return response.JSON200.Status, response.JSON200.VerifiablePresentation, nil
 }
 
-func (s SelfSigned) RequestAccessToken(organizationDID string, purposeOfUse string, presentation *authAPI.VerifiablePresentation) (*authAPI.TokenIntrospectionResponse, error) {
+func (s SelfSigned) UntrustEmployeeCredential(issuerDID string) error {
```
gerardsn, line 95: only used in tests?

gerardsn: hmm.. having e2e-tests in the main repo does require you to look twice sometimes

vcr/verifier/interface.go
gerardsn: `allowUntrusted` is also used for this in `Finder.Search`

auth/services/selfsigned/validator.go
```diff
@@ -135,6 +139,22 @@ func (v validator) verifyVP(vp vc.VerifiablePresentation, validAt *time.Time) (c
...
+		{IRIPath: jsonld.OrganizationNamePath, Type: vcr.NotNil},
+		{IRIPath: jsonld.OrganizationCityPath, Type: vcr.NotNil},
+	}
+	nutsOrgCreds, err := v.vcr.Search(context.Background(), searchTerms, false, validAt)
+	if err != nil {
+		resultErr = fmt.Errorf("unable to check NutsEmployeeCredential trust status using NutsOrganizationCredential: %w", err)
+		return
+	}
+	if len(nutsOrgCreds) == 0 {
+		resultErr = newVerificationError("invalid NutsEmployeeCredential, issuer does not have a trusted NutsOrganizationCredential")
```
gerardsn, line 154: untrusted != invalid

auth/services/selfsigned/validator.go
```diff
@@ -135,6 +139,22 @@ func (v validator) verifyVP(vp vc.VerifiablePresentation, validAt *time.Time) (c
 		return
 	}
 
+	// #2428: NutsEmployeeCredential trust is derived from the fact that the issuer has a trusted NutsOrganizationCredential
+	searchTerms := []vcr.SearchTerm{
+		{IRIPath: jsonld.CredentialSubjectPath, Value: vc.Issuer.String()},
+		{IRIPath: jsonld.OrganizationNamePath, Type: vcr.NotNil},
+		{IRIPath: jsonld.OrganizationCityPath, Type: vcr.NotNil},
+	}
+	nutsOrgCreds, err := v.vcr.Search(context.Background(), searchTerms, false, validAt)
```
gerardsn, line 148:
```suggestion
	nutsOrgCreds, err := v.vcr.Search(context.TODO(), searchTerms, false, validAt)
```

auth/services/selfsigned/validator.go
```diff
@@ -135,6 +139,22 @@ func (v validator) verifyVP(vp vc.VerifiablePresentation, validAt *time.Time) (c
 		return
 	}
 
+	// #2428: NutsEmployeeCredential trust is derived from the fact that the issuer has a trusted NutsOrganizationCredential
+	searchTerms := []vcr.SearchTerm{
+		{IRIPath: jsonld.CredentialSubjectPath, Value: vc.Issuer.String()},
+		{IRIPath: jsonld.OrganizationNamePath, Type: vcr.NotNil},
+		{IRIPath: jsonld.OrganizationCityPath, Type: vcr.NotNil},
+	}
```
gerardsn, line 147: a filter on credential type option would have been nice

gerardsn (changes requested):

vcr/verifier/interface.go
reinkrul: I followed the `verify` pattern from `verifyVCs`, so at least to make it consistent within this function signature.

I think "allowUntrusted" was the wrong choice, since the default (true) should be "verifyTrusted" imo

auth/services/selfsigned/validator.go
reinkrul: We could say "any trusted organization credential", which this does project. Although we could filter, this is a bit more flexible than accepting just "NutsOrganizationCredential"?

auth/services/selfsigned/validator.go
reinkrul: I didn't want to have another "trust" in this sentence, making people think they still have to explicitly trust the NutsEmployeeCredential. Ideas?

vcr/verifier/interface.go
gerardsn: The default value for a bool is false, so when adding `allowUntrusted` to a struct it would actually default to the safe option.

auth/services/selfsigned/validator.go
gerardsn: just drop the "invalid" at the start? (and comma)

gerardsn: or `NutsEmployeeCredential rejected, ...`

vcr/verifier/interface.go
reinkrul: Ok I'll stick to `allowUntrustedVCs` then

auth/services/selfsigned/validator.go
```diff
@@ -135,6 +139,22 @@ func (v validator) verifyVP(vp vc.VerifiablePresentation, validAt *time.Time) (c
 		return
 	}
 
+	// #2428: NutsEmployeeCredential trust is derived from the fact that the issuer has a trusted NutsOrganizationCredential
+	searchTerms := []vcr.SearchTerm{
+		{IRIPath: jsonld.CredentialSubjectPath, Value: vc.Issuer.String()},
```
stevenvegt, line 144: Better to search for the subject here, since the subject is the `Organisation`. Yes, the credential is self-issued, but this is more a business rule.

vcr/verifier/interface.go
```diff
@@ -48,8 +48,8 @@ type Verifier interface {
 	RegisterRevocation(revocation credential.Revocation) error
 
 	// VerifyVP verifies the given Verifiable Presentation. If successful, it returns the credentials within the presentation.
-	// If verifyVCs is true, it will also verify the credentials inside the VP, checking their correctness, signature and trust status.
-	VerifyVP(presentation vc.VerifiablePresentation, verifyVCs bool, validAt *time.Time) ([]vc.VerifiableCredential, error)
+	// If verifyVCs is true, it will also verify the credentials inside the VP, checking their correctness, signature and trust status (if allowUntrustedVCs is true).
+	VerifyVP(presentation vc.VerifiablePresentation, verifyVCs bool, allowUntrustedVCs bool, validAt *time.Time) ([]vc.VerifiableCredential, error)
```
stevenvegt, line 52: Probably better to make this a list of types, because if the presentation contained both the org and employee credential, the org should be verified.

stevenvegt (changes requested):
Use the `credentialSubject.id` instead of the `issuer` for searching the org credentials. Also probably better to exclude specific types in the VP verification than just all of them.

vcr/verifier/interface.go
reinkrul: I think the potential use cases outclass any reasonable changes we can make to this signature without going wildly out of scope. E.g.; maybe not all VCs need to be verified?

The presentation type (and validator) specifies there's exactly 1 VC, which is the NutsEmployeeCredential. Redesign would probably lead to a flexible DSL/business rule engine to express these.

auth/services/selfsigned/validator.go
reinkrul: I think that property is not indexed, so search might be impacted. Actually, you'd want to assert both, but since the presentation and credential specify they must be the same, it doesnt really matter?

vcr/verifier/interface.go
```diff
@@ -48,8 +48,8 @@ type Verifier interface {
 	RegisterRevocation(revocation credential.Revocation) error
 
 	// VerifyVP verifies the given Verifiable Presentation. If successful, it returns the credentials within the presentation.
-	// If verifyVCs is true, it will also verify the credentials inside the VP, checking their correctness, signature and trust status.
-	VerifyVP(presentation vc.VerifiablePresentation, verifyVCs bool, validAt *time.Time) ([]vc.VerifiableCredential, error)
+	// If verifyVCs is true, it will also verify the credentials inside the VP, checking their correctness, signature and trust status (if allowUntrustedVCs is true).
```
gerardsn, line 51:
```suggestion
	// If verifyVCs is true, it will also verify the credentials inside the VP, checking their correctness, signature and trust status (unless allowUntrustedVCs is true).
```

gerardsn (approved):

auth/services/selfsigned/validator.go
stevenvegt: What I mean is, the value of the jsonld.CredentialSubjectPath should be the credentialSubject ID and not the Issuer. Since the subject of the EmployeeID credential is the same organisation as the nutsOrgCredential subject ID.

vcr/verifier/interface.go
stevenvegt: Yeah, perhaps? I originally designed this credential to be presented together with the orgCredential in one presentation. Somewhere this design got lost in translation/adaptation. Now we have to find the orgCredential from the store with custom logic. So yes, it is already weird indeed. 

> why call Verify on a presentation and not verify the VCs itself?
Verification of the VP is necessary because the credentials inside it must have the samen subject as the presentation issuer.

reinkrul:
The selfsigned validator checks that the VP does contain exactly 1 NutsEmployeeCredential, so we should be OK.

PR #2430, opened 2023-08-18
Experimental Auth v2 API

reinkrul:
Intended to incrementally implement the new OAuth2 flows (authz code, OpenID4VP, later on OpenID4VC).

It is disabled by default, config flag (`auth.v2apienabled`) hidden.


auth/api/auth/v2/session.go
```diff
@@ -0,0 +1,46 @@
...
+func (s *SessionManager) Get(id string) *Session {
+	session, ok := s.sessions.Load(id)
+	if !ok {
+		return nil
+	}
+	result := session.(Session)
+	return &result
+}
+
+type Session struct {
```
gerardsn, line 31: comment: [RFC6749 3.1 Authorization Endpoint](https://datatracker.ietf.org/doc/html/rfc6749#section-3.1)
>   Parameters sent without a value MUST be treated as if they were
   omitted from the request.  The authorization server MUST ignore
   unrecognized request parameters.  Request and response parameters
   MUST NOT be included more than once.

Meaning `param == ""` can be interpreted as the parameter is not present in the authorization request

(just nothing this as confirmation that "" == nil in this case)

auth/api/auth/v2/s2s_vptoken.go
```diff
@@ -0,0 +1,44 @@
...
+// then create a presentation submission given the definition which is posted to the token endpoint as vp_token.
+// The AS then returns an access token with the requested scope.
+// Requires:
+// - GET /presentation_definition?scope=... (returns a presentation definition)
+// - POST /token (with vp_token grant)
+type serviceToService struct {
+}
+
+func (s serviceToService) Routes(router core.EchoRouter) {
+	router.Add("GET", "/public/oauth2/:did/presentation_definition", func(echoCtx echo.Context) error {
```
gerardsn, line 22: the manually registered paths need to be updated?

docs/_static/auth/v2.yaml
```diff
@@ -0,0 +1,247 @@
+openapi: 3.0.0
+info:
+  title: Auth v2 API (experimental)
+  version: 0.0.0
+servers:
+  - url: "http://localhost:1323"
+paths:
+  "/iam/{did}/token":
```
gerardsn, line 8: comment: public paths need to be discussed later

auth/api/auth/v2/api.go
```diff
@@ -0,0 +1,162 @@
...
+		// TODO: Check parameters for right flow
+		// TODO: Do we actually need this? (probably not)
+	case "vp_token id_token":
+		// Options:
+		// - OpenID4VP+SIOP flow, vp_token is sent in Authorization Response
+		// TODO: Check parameters for right flow
+	default:
+		// TODO: This should be a redirect?
+		return nil, openid4vci.Error{
+			Code:       openid4vci.InvalidRequest,
```
gerardsn, line 139: should be error code
[4.1.2.1 Error Response](https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.2.1)
```
unsupported_response_type
      The authorization server does not support obtaining an
      authorization code using this method.
```
or add issue as reminder to validate errors?

gerardsn (commented):
comments labeled comment are just comments ;)

it feels like the current separation in files is not the right one. the authorized_code flow can also be used with OpenID4VCI&VP, but those have their own files. I don't see a better separation at this point, so probably fine for now.

we're still missing the metadata endpoints (they are on /n2n, which means that public endpoints /authorize and /token can only be discovered on a non-public api...)

the `pre-authorized_code` flow is currently in the VCR, but should probably move here ?

auth/api/auth/v2/api.go
reinkrul: Added TODO for errors; they shouldn't reside in the openid4vci package

docs/_static/auth/v2.yaml
reinkrul: done

auth/api/auth/v2/s2s_vptoken.go
reinkrul: This needs to move to api.go#Routes anyways

reinkrul:
> comments labeled comment are just comments ;)
> 
> it feels like the current separation in files is not the right one. the authorized_code flow can also be used with OpenID4VCI&VP, but those have their own files. I don't see a better separation at this point, so probably fine for now.

I agree.

> we're still missing the metadata endpoints (they are on /n2n, which means that public endpoints /authorize and /token can only be discovered on a non-public api...)

That's right, I'd suggest new metadata endpoints under `/iam` and move the VCI flows to this implementation later on.

> the `pre-authorized_code` flow is currently in the VCR, but should probably move here ?

Later on, I'd say.

gerardsn (approved):

auth/api/auth/v2/session.go
gerardsn: exactly, `"" != nil` but we can interpret it as such  instead of using annoying  `*string`

Issue #2347, opened 2023-07-14
did:web support


reinkrul:
# Why
Although Nuts is specified in RFCs, there are no (known) implementations other than "this" Nuts node.
This is a problem when the ecosystem contains parties that want to interact with Nuts node, but don't actually have a Nuts node. E.g.,
- issuing a care organization credential to a Nuts node's wallet,
- using a Microsoft Authenticator wallet to authenticate a caregiver,
- having the Nuts node issue a credential to a non-Nuts node (?)

# Resources
- did:web specification: https://w3c-ccg.github.io/did-method-web/
- For reference; resolver used in Universal Resolver: https://github.com/decentralized-identity/web-did-resolver

# Stuff to do

- Consuming did:web
   - Implement a resolver
   - Support VCs with web DID (issuer is a did:web, or credential subject is a did:web)
      - Probably good idea to disable it by default, at first
- Producing did:web
- Migrating did:nuts to did:web
   - Aliasing did:nuts with alsoKnownAs?
   - We can only start issuing VCs to did:web when all Nuts nodes have upgraded to a version with did:web support
   - As such, we should try to support consuming did:web as early as possible?
- Write RFC that adds did:web support for Nuts

# issues

 - [x] Expose all nuts did documents via its web url. This requires a publicUrl in the config. All */did.json calls should be intercepted (use middleware). 
 - [x] Write a did:web resolver in its basic form (no tenant id <> public key checking). #2351 
 - [ ] GoldenHammer Part deux: add AlsoKnownAs to DID Documents.

woutslakhorst:
The did:nuts method now also handles service discovery for the DIDs. Without pro-actively pushing DID changes to others, they won't know a DID exists or that it supports a service.

Supporting did:web allows non nuts network parties to use the nuts node as issuer/verifier.

reinkrul:
We should probably make things easier for ourselves and produce did:web in JSON format instead of JSON-LD

woutslakhorst:
> We should probably make things easier for ourselves and produce did:web in JSON format instead of JSON-LD

result wouldn't really be different apart from the `@context` array?

reinkrul:
> > We should probably make things easier for ourselves and produce did:web in JSON format instead of JSON-LD
> 
> result wouldn't really be different apart from the `@context` array?

Yes, and no array compacting (compacting an array with a single value to a scalar/object)

reinkrul:
[...]

PR #2433, opened 2023-08-20
Serve Nuts DID documents as did:web

reinkrul:
Things to figure out/discuss:
- What path do we want to serve did:web on? Now: `/iam/{id}/did.json`. Should we stick to `/public` (kinda ugly)?
  - Current solution: new `iam` path. `public` is ugly, `identity` looks nice but too shallow. `iam` means "I am" and "Identity and Access/Authorization Management", which is imo the whole spectrum. Same question applies to new OAuth2 server impl base path (see https://github.com/nuts-foundation/nuts-node/issues/2439).
- Do we want to have just the did:nuts `ID` in the path (omitting `did:nuts`) to keep the URL shorter? (current solution)
- How do we want to derive the did:web URL;
  - Do we use a `public` config property? (we already have one in Auth module)
  - (current solution) Derive it from the request URL (might require specific proxy rewrite instructions)
  - Both: use config if set, derive from request URL otherwise

Part of https://github.com/nuts-foundation/nuts-node/issues/2347


reinkrul:
I thought of making an e2e test, but probably more useful to have a use case-oriented e2e test later on in which a `did:nuts` DID is resolved through `did:web` as part of the flow.

gerardsn (changes requested):
I see a lot of (potential) issues in derivation of did:web. Not doing it seems like the best solution to me.

vdr/api/v1/api.go
```diff
@@ -226,3 +234,43 @@ func (a *Wrapper) DeactivateDID(ctx context.Context, request DeactivateDIDReques
...
+		}
+	}
+	isOwner, err := a.VDR.IsOwner(ctx, *nutsDID)
+	if err != nil {
+		log.Logger().WithError(err).Warnf("Could not check ownership of Nuts DID: %s", nutsDID.String())
+		return nil, serverError
+	}
+	if isOwner {
+		nutsDIDDocumentData, _ := nutsDIDDocument.MarshalJSON()
+		webDIDDocumentData := bytes.ReplaceAll(nutsDIDDocumentData, []byte(nutsDID.String()), []byte(webDID.String()))
```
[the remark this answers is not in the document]
reinkrul: `alsoKnownAs` is probably good to add, but go-did doesn't support it yet. I'll add it.

docs/_static/vdr/v1.yaml
```diff
@@ -234,7 +234,31 @@ paths:
           description: Verification Method was successfully deleted
         default:
           $ref: '../common/error_response.yaml'
-
+  /iam/{id}/did.json:
+    parameters:
+      - name: id
+        in: path
+        description: ID of the corresponding Nuts DID document without the `did:nuts:` prefix.
```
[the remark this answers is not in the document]
reinkrul: Wouldn't it be weird to specify a `did:nuts` DID, and having a `did:web` returned?

vdr/didweb/util.go
```diff
@@ -0,0 +1,109 @@
...
+		if path != "" {
+			return nil, errInvalidWebDIDURL
+		}
+	} else {
+		// In case of another path, we need to chop off the did.json part
+		path, cut = strings.CutSuffix(path, "/did.json")
+		if !cut || len(path) == 0 {
+			return nil, errInvalidWebDIDURL
+		}
+	}
```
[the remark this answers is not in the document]
reinkrul: I can live with that, but I want to specifically check for the well-known case (not being supported), in case we decide to change it

vdr/didweb/util.go
```diff
@@ -0,0 +1,109 @@
...
+
+	parts := strings.Split(path, "/")
+	for i, part := range parts {
+		part = percentEncodeString(part)
+		parts[i] = part
+	}
```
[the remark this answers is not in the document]
reinkrul: URL paths allow more characters (e.g. `+` or `@`) than allowed in the DID syntax.

docs/_static/vdr/v1.yaml
```diff
@@ -234,7 +234,31 @@ paths:
           description: Verification Method was successfully deleted
         default:
           $ref: '../common/error_response.yaml'
-
+  /iam/{id}/did.json:
```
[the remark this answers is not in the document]
reinkrul: The logic/feature really belongs in VDR i.m.o. (resolving DIDs), unless we consider it "iam" identity stuff, then we maybe could put it with `auth` for the sake of consistency?

vdr/didweb/util.go
```diff
@@ -0,0 +1,109 @@
...
+	str := "did:web:" + percentEncodeString(u.Host) + strings.Join(parts, ":")
```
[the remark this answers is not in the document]
reinkrul: Now only possible if not in strict mode, of it auth.publicurl contains an IP address (but then the TLS server certificate will be invalid). So not really worth it, I think

vdr/vdr.go
```diff
@@ -68,6 +70,37 @@ type VDR struct {
...
+	if err != nil {
+		return nil, err
+	}
+	isOwner, err := r.IsOwner(ctx, nutsDID)
+	if err != nil {
+		return nil, err
+	}
+	if isOwner {
+		nutsDIDDocumentData, _ := nutsDIDDocument.MarshalJSON()
+		webDIDDocumentData := bytes.ReplaceAll(nutsDIDDocumentData, []byte(nutsDID.String()), []byte(webDID.String()))
```
stevenvegt, line 89: Does this also rewrites the document controller?

docs/_static/auth/iam.yaml
```diff
@@ -6,6 +6,32 @@ servers:
...
+        - DID
+      responses:
+        "200":
+          description: DID has been found and returned.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/DIDDocument'
+        "404":
+          description: DID does not exist.
```
gerardsn, line 33: are we sticking with `../common/error_response.yaml` for the public APIs?

gerardsn (changes requested):
Just a couple of minor, perhaps nit-picky things.
Steven raised valid issue though

vdr/didweb/util.go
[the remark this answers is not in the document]
reinkrul: It appears you're right; although not recommended, some CAs allow it. I don't want to introduce breaking changes, but if we use `publicurl` for the did:web endpoint we have to. I'll bake a cake if it breaks.

vdr/vdr.go
reinkrul: It only rewrites the controller if the document is its own controller. But question is what value the controller has for `did:web`,  since `did:nuts` uses it for authentication DID document updates, which we don't have for `did:web`. The DID spec says for controller:

> When a [controller](https://www.w3.org/TR/did-core/#dfn-controller) property is present in a [DID document](https://www.w3.org/TR/did-core/#dfn-did-documents), its value expresses one or more [DIDs](https://www.w3.org/TR/did-core/#dfn-decentralized-identifiers). Any [verification methods](https://www.w3.org/TR/did-core/#dfn-verification-method) contained in the [DID documents](https://www.w3.org/TR/did-core/#dfn-did-documents) for those [DIDs](https://www.w3.org/TR/did-core/#dfn-decentralized-identifiers) SHOULD be accepted as authoritative, such that proofs that satisfy those [verification methods](https://www.w3.org/TR/did-core/#dfn-verification-method) are to be considered equivalent to proofs provided by the [DID subject](https://www.w3.org/TR/did-core/#dfn-did-subjects).

and:

> NOTE: Authorization vs authentication
Note that authorization provided by the value of controller is separate from authentication as described in [5.3.1 Authentication](https://www.w3.org/TR/did-core/#authentication). This is particularly important for key recovery in the case of cryptographic key loss, where the [DID subject](https://www.w3.org/TR/did-core/#dfn-did-subjects) no longer has access to their keys, or key compromise, where the [DID controller](https://www.w3.org/TR/did-core/#dfn-did-controllers)'s trusted third parties need to override malicious activity by an attacker. See [9. Security Considerations](https://www.w3.org/TR/did-core/#security-considerations) for information related to threat models and attack vectors.

I think we can probably leave it as it currently is?

auth/auth.go
```diff
@@ -116,6 +127,11 @@ func (auth *Auth) Configure(config core.ServerConfig) error {
 	if auth.config.PublicURL == "" && config.Strictmode {
 		return ErrMissingPublicURL
 	}
+	if auth.config.PublicURL != "" {
+		if _, err := url.Parse(auth.config.PublicURL); err != nil {
+			return fmt.Errorf("invalid auth.publicurl: %w", err)
+		}
+	}
```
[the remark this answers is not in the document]
reinkrul: fixed

docs/_static/auth/iam.yaml
reinkrul: I don't think so, since most endpoints' error cases are specified by an external spec

auth/api/iam/api.go
```diff
@@ -153,3 +181,38 @@ func (r Wrapper) HandleAuthorizeRequest(ctx context.Context, request HandleAutho
...
+
+func (r Wrapper) GetWebDID(ctx context.Context, request GetWebDIDRequestObject) (GetWebDIDResponseObject, error) {
+	const baseSuffix = "iam"
+	publicURL := r.Auth.PublicURL()
+	var baseURL url.URL
+	if publicURL == nil {
+		// Possible in non-strict mode
+		httpRequest := *ctx.Value(httpRequestContextKey).(*http.Request)
+		baseURL = *httpRequest.URL
+		baseURL.Host = httpRequest.Host
```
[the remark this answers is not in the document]
reinkrul: Depends on the proxy; if you omit a suffix slash in nginx, it retains the original host. One more reason to only do it in non-strict mode I guess and have a health check... We could easily test this in an e2e-test?

auth/api/iam/api.go
```diff
@@ -153,3 +181,38 @@ func (r Wrapper) HandleAuthorizeRequest(ctx context.Context, request HandleAutho
...
+		idx := strings.LastIndex(baseURL.Host, "/"+baseSuffix)
+		if idx != -1 {
+			// Would be very weird if this happens, routing error?
+			return nil, errors.New("could not derive base URL")
+		}
+		baseURL.Path = baseURL.Path[:idx+len("/"+baseSuffix)]
```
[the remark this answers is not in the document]
reinkrul: I don't remember how I got here... (wrote this weird code)

reinkrul:
Oh noes... The conficts...

auth/api/iam/api.go
```diff
@@ -44,6 +70,7 @@ func (r Wrapper) Routes(router core.EchoRouter) {
 				ctx.Set(core.ModuleNameContextKey, vcr.ModuleName+"/v2")
```
woutslakhorst, line 70: why the vcr module name?

auth/api/iam/types.go
```diff
@@ -0,0 +1,27 @@
+/*
+ * Nuts node
```
woutslakhorst, line 2: oddly formatted copyright

vdr/didweb/util.go
```diff
@@ -0,0 +1,94 @@
...
+import (
+	"errors"
+	"github.com/nuts-foundation/go-did/did"
+	"net/url"
+	"strings"
+)
+
+var errInvalidWebDIDURL = errors.New("URL does not represent a Web DID")
+
+// URLToDID converts a URL back to a Web DID (did:web). It removes /did.json and /.well-known/did.json from the path, if present.
```
woutslakhorst, line 30: Add a note on escaping. It requires unescaped urls?

woutslakhorst (changes requested):

vdr/didweb/util.go
reinkrul: When `url.URL` is properly used by the caller, it will always be correct? (so, either use `url.Parse` or set the fields properly)

gerardsn:
@reinkrul There are so many conflicts that I prefer to re-review this after a rebase

reinkrul:
I'll rebase first.

reinkrul:
Ready for re-review.

auth/api/iam/api.go
```diff
@@ -225,11 +229,29 @@ func (r Wrapper) GetOAuthAuthorizationServerMetadata(ctx context.Context, reques
...
+		return nil, err
+	}
+
+	document, err := r.vdr.DeriveWebDIDDocument(ctx, baseURL, *nutsDID)
+	if err != nil {
+		if didservice.IsFunctionalResolveError(err) {
+			return GetWebDID404Response{}, nil
+		}
+		log.Logger().WithError(err).Warnf("Could not resolve Nuts DID: %s", nutsDID.String())
+		return nil, errors.New("unable to resolve DID")
```
woutslakhorst, line 250: This is a 500 right? Then the logging should error not warn

woutslakhorst (changes requested):

vdr/vdr.go
```diff
@@ -69,6 +70,57 @@ type VDR struct {
...
+	}
+	if !isOwner {
+		log.Logger().
+			WithError(err).
+			WithField(core.LogFieldDID, nutsDID).
+			Info("Tried to derive did:web document from Nuts DID that is not owned by this node")
+		return nil, types.ErrNotFound
+	}
+
+	// Replace did:nuts DIDs with did:web, but only for owned DIDs
```
gerardsn, line 90: references in `did:nuts` format (`did:nuts:123/serviceEndpoint?type=oauth`) are not according to the did spec. The nuts node needs to implement did dereferencing according to spec if we want to be compatible, and we need to convert the did:nuts references for the did:web document. New issue probably

vdr/vdr.go
```diff
@@ -69,6 +70,57 @@ type VDR struct {
...
+		if controller.Method == didnuts.MethodName && !controller.IsURL() {
+			isOwner, err := r.IsOwner(ctx, controller)
+			if err != nil {
+				return nil, err
+			}
+			if isOwner {
+				ownedNutsDIDs = append(ownedNutsDIDs, controller)
+			}
+		}
+	}
```
gerardsn, line 107: [Q] The Controller is authorized to make changes to a DID, what does this mean on a did document that has no history like did:web?

We can verify the history through the linked did:nuts, but perhaps the Controller should be deleted from did:web?

gerardsn (commented):
Only need some clarification

reinkrul:
After discussion with @gerardsn, we decided to remove `service` and `controller` fields from the `did:web` documents. Ratio: we want to serve `did:web` documents, derived from their `did:nuts` counterparts, for integration with third party systems, e.g. credential issuers, verifiers and wallets. These won't support `did:nuts`, so need `did:web` documents to resolve key material. As such, they most probably don't need any services (since all services we now have are Nuts specific). The controllers on `did:nuts` documents are used to authenticate updates on DID documents, but this is not required for `did:web`, since documents aren't signed (and thus the `controller` field has no practical value).

When we do want to keep the `service` property, we should decide how we want to treat service references; e.g. resolve them to absolute URLs before returning them as `did:web`.

auth/api/iam/api.go
woutslakhorst: @reinkrul this hasn't been addressed yet

vdr/vdr.go
```diff
@@ -69,6 +70,43 @@ type VDR struct {
...
+	resultDIDDocumentData = bytes.ReplaceAll(resultDIDDocumentData, []byte(nutsDID.String()), []byte(webDID.String()))
+	var result did.Document
+	if err = result.UnmarshalJSON(resultDIDDocumentData); err != nil {
+		return nil, fmt.Errorf("did:web unmarshal error (%s): %w", nutsDID, err)
+	}
+	result.AlsoKnownAs = append(result.AlsoKnownAs, nutsDID.URI())
+	// did:web support is currently just for supporting third party systems resolving key material,
+	// so no need to retain services (which often refer to services in other did:nuts documents, complicating things).
+	result.Service = nil
```
gerardsn, line 104: Should this have a TODO with an issue?

gerardsn (approved):

vdr/vdr.go
reinkrul: I don't think so, since it's currently not a feature we plan to have? So not TODO per se?

Issue #2389, opened 2023-08-01
Create VC Wallet


woutslakhorst:
Under `vcr/holder`, the `wallet.go` can be added.

The wallet contains a stoabs KVStore with a shelf for each DID and values (vc.ID => vc).
The wallet must be filled by a migration where all non-AuthCreds that are owned by a node are added to the right wallet. This could be done from the VCR?
Then any VC (owned, non-AuthCred) received through the network (ambassador) or via OpenID4VCI must be added to the wallet.

The wallet should get an All(did) method. Maybe we can expose it through an API as well, but that would be a bonus?

reinkrul:
- Diagnostic showing the number of VCs in the wallet
- Re-add to wallet when DAG is reprocessed (probably works out of the box through ambassador)

PR #2442, opened 2023-08-24
OpenID4VP: Implement authorization request and handling

reinkrul:
Next things to work on, fitting into this PR:
- Support direct_post mode for sending the authorization response from wallet to verifier, instead of having it as parameters on the redirect back to the verifier.
- Replace hardcoded metadata URLs with the ones specified by https://github.com/nuts-foundation/nuts-node/pull/2443 
- Support VPs in JWT format (consuming and producing), for compatibility with non-Nuts nodes
- Match Presentation Definition with credentials in wallet, instead of just loading credentials from VCR store (requires  https://github.com/nuts-foundation/nuts-node/issues/2389)
- Do some styling of the HTML templates for the "WOW!" factor
- i18n HTML templates (we should at least support Dutch)
- Support presentation_definition and presentation_definition_uri, so the presentation definition can be specified there (by the verifier) instead of mapping it from scope.
- Validate redirect URI sent by verifier
- Implement id_token (needs design work)
- Implement /token path

### DIFF
--- a/auth/api/iam/api.go
+++ b/auth/api/iam/api.go
@@ -159,18 +159,7 @@ func (r Wrapper) HandleAuthorizeRequest(ctx context.Context, request HandleAutho
 	for key, value := range httpRequest.URL.Query() {
 		params[key] = value[0]
 	}
-	session := &Session{
-		// TODO: Validate client ID
-		ClientID: params[clientIDParam],
-		// TODO: Validate scope
-		Scope:       params[scopeParam],
-		ClientState: params[stateParam],
-		ServerState: map[string]interface{}{},
-		// TODO: Validate according to https://datatracker.ietf.org/doc/html/rfc6749#section-3.1.2
-		RedirectURI:  params[redirectURIParam],
-		OwnDID:       *ownDID,
-		ResponseType: params[responseTypeParam],
-	}
+	session := createSession(params, *ownDID)
 	if session.RedirectURI == "" {
 		// TODO: Spec says that the redirect URI is optional, but it's not clear what to do if it's not provided.
 		//       Threat models say it's unsafe to omit redirect_uri.
@@ -240,4 +229,20 @@ func (r Wrapper) GetOAuthAuthorizationServerMetadata(ctx context.Context, reques
 	identity := r.auth.PublicURL().JoinPath("iam", id.WithoutURL().String())
 
 	return GetOAuthAuthorizationServerMetadata200JSONResponse(authorizationServerMetadata(*identity)), nil
+}
+
+func createSession(params map[string]string, ownDID did.DID) *Session {
+	session := &Session{
+		// TODO: Validate client ID
+		ClientID: params[clientIDParam],
+		// TODO: Validate scope
+		Scope:       params[scopeParam],
+		ClientState: params[stateParam],
+		ServerState: map[string]interface{}{},
+		// TODO: Validate according to https://datatracker.ietf.org/doc/html/rfc6749#section-3.1.2
+		RedirectURI:  params[redirectURIParam],
+		OwnDID:       ownDID,
+		ResponseType: params[responseTypeParam],
+	}
+	return session
 }
--- a/auth/api/iam/api.go
+++ b/auth/api/iam/api.go
@@ -204,6 +204,7 @@ func (r Wrapper) HandleAuthorizeRequest(ctx context.Context, request HandleAutho
 		// TODO: Validate scope
 		Scope:       params[scopeParam],
 		ClientState: params[stateParam],
+		ServerState: map[string]interface{}{},
 		// TODO: Validate according to https://datatracker.ietf.org/doc/html/rfc6749#section-3.1.2
 		RedirectURI: params[redirectURIParam],
 		OwnDID:      *ownDID,
--- a/auth/api/iam/api.go
+++ b/auth/api/iam/api.go
@@ -28,10 +28,8 @@ import (
 	"github.com/nuts-foundation/nuts-node/auth"
 	"github.com/nuts-foundation/nuts-node/auth/log"
 	"github.com/nuts-foundation/nuts-node/core"
-	"github.com/nuts-foundation/nuts-node/jsonld"
 	"github.com/nuts-foundation/nuts-node/vcr"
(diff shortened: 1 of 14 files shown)


auth/api/iam/api.go
```diff
@@ -22,17 +26,30 @@ var assets embed.FS
...
 // Wrapper handles OAuth2 flows.
 type Wrapper struct {
-	VCR      vcr.VCR
-	Auth     auth.AuthenticationServices
-	sessions *SessionManager
+	VCR                     vcr.VCR
+	VDR                     vdr.VDR
+	JSONLD                  jsonld.JSONLD
+	Auth                    auth.AuthenticationServices
+	sessions                *SessionManager
```
gerardsn, line 33: can be all lowercase? (unexported?)

auth/api/iam/api.go
```diff
@@ -59,6 +79,17 @@ func (r Wrapper) Routes(router core.EchoRouter) {
 			return audit.StrictMiddleware(f, vcr.ModuleName+"/v2", operationID)
 		},
 	}))
+	// The following handler is of the OpenID4VCI wallet which is called by the holder (wallet owner)
+	// when accepting an OpenID4VP authorization request.
+	router.POST("/iam/:did/openid4vp_authz_accept", r.handlePresentationRequestAccept, audit.Middleware(vcr.ModuleName+"/v2", "openid4vp_authz_accept"))
```
gerardsn, line 84: add comment that these need to be added to client redirect_uris (part of security model to prevent request forgery after obtaining an authorization code)

auth/api/iam/api.go
```diff
@@ -153,3 +184,19 @@ func (r Wrapper) HandleAuthorizeRequest(ctx context.Context, request HandleAutho
 		//Description: "missing or invalid parameters",
 	}
 }
+
+func createSession(params map[string]string, ownDID did.DID) *Session {
+	session := &Session{
```
gerardsn, line 189: I expect different authorizations flows to use different parameters. Maybe just store the entire request? (or defer to different issue/PR)

auth/api/iam/presentation_definitions.go
```diff
@@ -0,0 +1,68 @@
...
+}
+
+func (n nutsPresentationDefinitionRegistry) ByScope(scope string) *pe.PresentationDefinition {
+	if scope != eOverdrachtOverdrachtsberichtScope {
+		return nil
+	}
+	return &pe.PresentationDefinition{
+		Format: &pe.PresentationDefinitionClaimFormatDesignations{
+			// TODO: Might have to support jwt_vc, but node doesn't support it yet.
+			"ldp_vc": {
```
gerardsn, line 28: It think the idea was to use `jwt_vp` for VPs, and `ldp_vc` for VCs?

auth/api/iam/openid4vp.go
```diff
@@ -2,75 +2,264 @@ package iam
...
 
-func (a openID4VP) Routes(router core.EchoRouter) {
-	router.Add(http.MethodPost, "/public/oauth2/:did/authz_consent", a.handleAuthConsent)
+// createPresentationRequest creates a new Authorization Request as specified by OpenID4VP: https://openid.net/specs/openid-4-verifiable-presentations-1_0.html.
+// It is sent by a verifier to a wallet, to request one or more verifiable credentials as verifiable presentation from the wallet.
+func (r *Wrapper) sendPresentationRequest(ctx context.Context, response http.ResponseWriter, scope string,
+	redirectURL url.URL, verifierIdentifier url.URL, walletIdentifier url.URL) error {
+	// TODO: Lookup wallet metadata for correct authorization endpoint. But for Nuts nodes, we derive it from the walletIdentifier
+	authzEndpoint := walletIdentifier.JoinPath("/authorize")
+	params := make(map[string]string)
```
gerardsn, line 42: can probably be simplified by marshaling a struct? TODO for now I think

auth/api/iam/api.go
```diff
@@ -59,6 +79,17 @@ func (r Wrapper) Routes(router core.EchoRouter) {
...
+	router.POST("/iam/:did/openid4vp_authz_accept", r.handlePresentationRequestAccept, audit.Middleware(vcr.ModuleName+"/v2", "openid4vp_authz_accept"))
+	// The following handler is of the OpenID4VP verifier where the browser will be redirected to by the wallet,
+	// after completing a presentation exchange.
+	router.GET("/iam/:did/openid4vp_completed", r.handlePresentationRequestCompleted, audit.Middleware(vcr.ModuleName+"/v2", "openid4vp_completed"))
+	// The following 2 handlers are used to test/demo the OpenID4VP flow.
+	// - GET renders an HTML page with a form to start the flow.
+	// - POST handles the form submission, initiating the flow.
+	router.GET("/iam/:did/openid4vp_demo", r.handleOpenID4VPDemoLanding)
+	router.POST("/iam/:did/openid4vp_demo", r.handleOpenID4VPDemoSendRequest)
 }
```
gerardsn, line 93: only 2/4 have audit middleware. Is this correct? (echo router has a [Group](https://echo.labstack.com/docs/routing#group) method. Probably doesn't apply here though)

gerardsn (approved):
It is hard to review since most is still a TODO, so only a few comments that may or may not need to be fixed now.

auth/api/iam/api.go
reinkrul: fixed

auth/api/iam/openid4vp.go
reinkrul: Later on, yes, when we support form post as well. Now they're URL query parameters, so a map is more practical to iterate over.

auth/api/iam/api.go
reinkrul: fixed

auth/api/iam/presentation_definitions.go
reinkrul: Yes, but we don't support them yet

auth/api/iam/api.go
reinkrul: Yeah, I'd refactor that when we introduce the next flow requiring these parameters. I think we should only store specified parameters, otherwise we're suspect to resource-exhaustion attacks? (attackers making up parameter names)

Issue #2357, opened 2023-07-18
Implement authorization request endpoint for openid4vp


woutslakhorst:
A client does a `/internal/auth/v2/create-access-token` request (part of other issue/pr). The node will request authorization via `GET <domain>/identity/<did>/authorize`. We currently only support getting an access token for a node. The openid4vp flow for a user is out of scope for this issue.

This issue is about the controller behind `/authorize` and the response it gives. Below is an example of the request expected to get at this endpoint:

## Examples

```
GET from https://my.verifier.com/client_id/authorize
?response_type=code
&scope=openid
&state=tracker%3Dvcfghhj
&client_id=did%3Akey%3Az2dmzD81cgPx8Vki7JbuuMmFYrWPgYoytykUZ3eyqht1j9KbsEYvdrjxMjQ4tpnje9BDBTzuNDP3knn6qLZErzd4bJ5go2CChoPjd5GAH3zpFJP5fuwSk66U5Pq6EhF4nKnHzDnznEP8fX99nZGgwbAh1o7Gj1X52Tdhf7U4KTk66xsA5r
&redirect_uri=openid%3A%2F%2F
&nonce=glkFFoisdfEui43
&client_metadata=%7B%22vp_formats_supported%22%3A%7B%22jwt_vp%22%3A%7B%22alg%22%3A%5B%22ES256%22%5D%7D%2C%22jwt_vc%22%3A%7B%22alg%22%3A%5B%22ES256%22%5D%7D%7D%2C%22response_types_supported%22%3A%5B%22vp_token%22%2C%22id_token%22%5D%2C%22authorization_endpoint%22%3A%22openid%3A%2F%2F%22%7D
```

and the result:
```
HTTP 302 Location: openid://
?client_id=did%3Aweb%3Aexample.com
&response_type=vp_token
&response_mode=direct_post
&redirect_uri=https%3A%2F%2Fexample.com%2Fdirect_post
&nonce=dtyr57brt
&client_id_scheme=did
&presentation_definition=...
[...]

PR #2443, opened 2023-08-24
OAuth add /.well-known/oauth-authorization-server endpoint

gerardsn:
WIP

discussion point: 
auth.publicurl should become a required config field (and possibly move to root of config).
Its value is required on all .well-known endpoints. (oauth-authorization-server, (oauth) client, openid-credential-issuer, openid-credential-wallet, did.json, ...) 

To me it makes a lot of sense that you have to configure the _public address_ of the node for it to work (vault and redis require it too). We could set the default to http://localhost:1323 in non-strict mode when no value is provided, but guessing what the address is just adds to much complexity and room for errors (will require support) for something that is not very hard to configure.

TODO
 - [x] tests
 - [ ] auth.PublicURL should not panic

### DIFF
--- a/auth/api/iam/api.go
+++ b/auth/api/iam/api.go
@@ -5,11 +5,13 @@ import (
 	"embed"
 	"errors"
 	"github.com/labstack/echo/v4"
+	"github.com/nuts-foundation/go-did/did"
 	"github.com/nuts-foundation/nuts-node/audit"
 	"github.com/nuts-foundation/nuts-node/auth"
 	"github.com/nuts-foundation/nuts-node/core"
 	"github.com/nuts-foundation/nuts-node/vcr"
 	"github.com/nuts-foundation/nuts-node/vcr/openid4vci"
+	vdr "github.com/nuts-foundation/nuts-node/vdr/types"
 	"net/http"
 	"sync"
 )
@@ -22,17 +24,18 @@ var assets embed.FS
 
 // Wrapper handles OAuth2 flows.
 type Wrapper struct {
-	VCR      vcr.VCR
-	Auth     auth.AuthenticationServices
+	vcr      vcr.VCR
+	vdr      vdr.DocumentOwner
+	auth     auth.AuthenticationServices
 	sessions *SessionManager
 }
 
-func New(authInstance auth.AuthenticationServices, vcrInstance vcr.VCR) *Wrapper {
-	sessionManager := &SessionManager{sessions: new(sync.Map)}
+func New(authInstance auth.AuthenticationServices, vcrInstance vcr.VCR, vdrInstance vdr.DocumentOwner) *Wrapper {
 	return &Wrapper{
-		sessions: sessionManager,
-		Auth:     authInstance,
-		VCR:      vcrInstance,
+		sessions: &SessionManager{sessions: new(sync.Map)},
+		auth:     authInstance,
+		vcr:      vcrInstance,
+		vdr:      vdrInstance,
 	}
 }
 
@@ -41,7 +44,7 @@ func (r Wrapper) Routes(router core.EchoRouter) {
 		func(f StrictHandlerFunc, operationID string) StrictHandlerFunc {
 			return func(ctx echo.Context, request interface{}) (response interface{}, err error) {
 				ctx.Set(core.OperationIDContextKey, operationID)
-				ctx.Set(core.ModuleNameContextKey, vcr.ModuleName+"/v2")
+				ctx.Set(core.ModuleNameContextKey, auth.ModuleName+"/v2")
 				// TODO: Do we need a generic error handler?
 				// ctx.Set(core.ErrorWriterContextKey, &protocolErrorWriter{})
 				return f(ctx, request)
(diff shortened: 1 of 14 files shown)


auth/api/iam/api.go
```diff
@@ -153,3 +156,29 @@ func (r Wrapper) HandleAuthorizeRequest(ctx context.Context, request HandleAutho
...
+		return nil, err
+	}
+
+	if id.Method != "nuts" {
+		return nil, errors.New("only did:nuts is supported")
+	}
+
+	owned, err := r.vdr.IsOwner(ctx, *id)
+	if err != nil {
+		// TODO: should this be not found?
```
reinkrul, line 173: yes. Maybe in combination with `IsFunctionalError`; DID resolution errors like NotFound, Deactivated etc can lead to 404, other technical errors (database error) could lead to 500 (and need to be logged)

auth/api/iam/api.go
```diff
@@ -153,3 +156,29 @@ func (r Wrapper) HandleAuthorizeRequest(ctx context.Context, request HandleAutho
...
+		return nil, err
+	}
+	if !owned {
+		// TODO: probably not the error we want?
```
reinkrul, line 177: 404

auth/api/iam/implementation_documentation/README
```diff
@@ -0,0 +1,5 @@
+THIS DIRECTORY WILL BE REMOVED WHEN THE OAUTH / OPENID4VC IMPLEMENTATIONS ARE COMPLETE
+
+This directory can be used to jolt down some notes / thoughts while building.
```
reinkrul, line 3: jot down?

auth/api/iam/metadata.go
```diff
@@ -0,0 +1,15 @@
+package iam
```
reinkrul, line 1: fix copyrights

auth/api/iam/metadata.go
```diff
@@ -0,0 +1,15 @@
...
+
+func authorizationServerMetadata(identity string) OAuthAuthorizationServerMetadata {
```
reinkrul, line 3: better pass url.URL and use `JoinPath` to avoid messing around with slashes?

auth/api/iam/types.go
```diff
@@ -0,0 +1,152 @@
...
+	// https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#name-sub-namespace-registration
+	grantTypePreAuthorizedCode grantType = "urn:ietf:params:oauth:grant-type:pre-authorized_code"
+)
+
+var grantTypesSupported = []grantType{grantTypeAuthorizationCode, grantTypeVPToken, grantTypePreAuthorizedCode}
+
+// algValuesSupported contains a list of supported cipher suites for jwt_vc_json & jwt_vp_json presentation formats
+// Recommended list of options https://www.iana.org/assignments/jose/jose.xhtml#web-signature-encryption-algorithms
+// TODO: validate list, should reflect current recommendations from https://www.ncsc.nl
+var algValuesSupported = []string{"PS256", "PS384", "PS512", "ES256", "ES384", "ES512"}
```
reinkrul, line 63: check with ARF

auth/api/iam/types.go
```diff
@@ -0,0 +1,152 @@
...
+//
+// OpenID4VCI defined formats are jwt_vc_json, jwt_vc_json-ld, ldp_vc
+//
+// OpenID4VP defines:
+//   - jwt_vc_json & jwt_vp_json   that should use values defined on https://www.iana.org/assignments/jose/jose.xhtml#web-signature-encryption-algorithms
+//   - ldp_vc      & ldp_vp		that should use valued defined on https://w3c-ccg.github.io/ld-cryptosuite-registry/
+//
+// but the examples then seem to be a mix and match of everything.
+var vpFormatsSupported = map[string]map[string][]string{
+	"jwt_vp": {"algValuesSupported": algValuesSupported},
```
reinkrul, line 86: do we really support this one?, also, the comment above says jwt_vc_json?

docs/_static/auth/iam.yaml
```diff
@@ -238,6 +275,35 @@ components:
           "token_type": "bearer",
           "expires_in": 3600,
         }
+    OAuthAuthorizationServerMetadata:
```
reinkrul, line 278: if we don't even generate code from this, just make it an `object`

reinkrul (changes requested):

auth/api/iam/api.go
```diff
@@ -153,3 +156,29 @@ func (r Wrapper) HandleAuthorizeRequest(ctx context.Context, request HandleAutho
 		//Description: "missing or invalid parameters",
 	}
 }
+
+// GetOAuthAuthorizationServerMetadata returns the Authorization Server's metadata
+func (r Wrapper) GetOAuthAuthorizationServerMetadata(ctx context.Context, request GetOAuthAuthorizationServerMetadataRequestObject) (GetOAuthAuthorizationServerMetadataResponseObject, error) {
```
reinkrul, line 161: missing unit test

auth/api/iam/types.go
gerardsn: changed to `jwt_vp_json`. I based this on #2357, but could be incorrect?

auth/api/iam/types.go
gerardsn: I can't find the exact details on this. I might be searching for the wrong terms

docs/_static/auth/iam.yaml
gerardsn: I removed the content, but it is still there as a schema since removing it completely makes the codegenerator assign it type `map[string]interface{}` which cannot be aliased to the actual internal struct.

auth/api/iam/types.go
reinkrul: Then it's probably not in there.

reinkrul (approved):

gerardsn:
publicURL panic will be fixed in followup PR

PR #2449, opened 2023-08-29
Openid4vp i18n

beardedfoo:
This adds a localize() function which can be used to localize the template path used in response rendering

### DIFF
--- a/auth/api/iam/assets/authz_wallet_nl.html
+++ b/auth/api/iam/assets/authz_wallet_nl.html
@@ -6,7 +6,7 @@
 </head>
 <body>
     <h1>Credential vereist (Wallet)</h1>
-    <p>{{ .VerifierName }} vraagt om identificatie met behulp van een of meer credentials in uw wallet:</p>
+    <p>{{ .VerifierName }} vraagt om identificatie met één of meer credentials in uw wallet:</p>
 
     <form method="post" class="form" action="./openid4vp_authz_accept">
         <div class="form-group">
--- a/auth/api/iam/assets/authz_wallet_nl.html
+++ b/auth/api/iam/assets/authz_wallet_nl.html
@@ -1,0 +1,60 @@
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <title>Credential vereist</title>
+    <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/bootstrap@5.2.3/dist/css/bootstrap.min.css" integrity="sha384-rbsA2VBKQhggwzxH7pPCaAqO46MgnOM80zW1RWuH61DGLwZJEdK2Kadq2F9CUG65" crossorigin="anonymous">
+</head>
+<body>
+    <h1>Credential vereist (Wallet)</h1>
+    <p>{{ .VerifierName }} vraagt om identificatie met behulp van een of meer credentials in uw wallet:</p>
+
+    <form method="post" class="form" action="./openid4vp_authz_accept">
+        <div class="form-group">
+            {{ range $credential := .Credentials }}
+            <div class="card">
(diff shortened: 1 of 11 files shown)


auth/api/iam/assets/authz_wallet_nl.html
```diff
@@ -0,0 +1,60 @@
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <title>Credential vereist</title>
+    <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/bootstrap@5.2.3/dist/css/bootstrap.min.css" integrity="sha384-rbsA2VBKQhggwzxH7pPCaAqO46MgnOM80zW1RWuH61DGLwZJEdK2Kadq2F9CUG65" crossorigin="anonymous">
+</head>
+<body>
+    <h1>Credential vereist (Wallet)</h1>
+    <p>{{ .VerifierName }} vraagt om identificatie met behulp van een of meer credentials in uw wallet:</p>
```
reinkrul, line 9:
```suggestion
    <p>{{ .VerifierName }} vraagt om identificatie met één of meer credentials in uw wallet:</p>
```

auth/api/iam/localize.go
```diff
@@ -0,0 +1,28 @@
...
+
+// localize selects from a range of localized values based on the accept-language header, returning
+// a default value if no localized value matches the languages specified in the accept-language header.
+func localize(defaultValue string, localizedValues map[string]string, acceptLanguageHeader string) string {
+	// Iterate over the accepted languages specified by the user-agent
+	for _, acceptedLanguage := range strings.Split(acceptLanguageHeader, ",") {
+		// Trim any whitespace from the language specification
+		acceptedLanguage = strings.TrimSpace(acceptedLanguage)
+
+		// Split the language and any associated q-factor value (e.g. "en;q=0.5") where the q-factor specifies
```
reinkrul, line 15: does anyone actually implement that properly... talking about elaborate specs

auth/api/iam/authorized_code.go
```diff
@@ -38,7 +39,7 @@ func (a authorizedCodeFlow) handleAuthzRequest(params map[string]string, session
 
 	// Render HTML
 	buf := new(bytes.Buffer)
-	// TODO: Support multiple languages
+	// TODO: Support multiple languages -- but how with the template path hardcoded prior to the request?
```
reinkrul, line 42: this authorizedCodeFlow type will probablyt go away, it's a reminiscant of the previous, complicated, OAuth2 design I made earlier

auth/api/iam/localize.go
```diff
@@ -0,0 +1,28 @@
+package iam
+
+import (
+	"strings"
+)
+
+// localize selects from a range of localized values based on the accept-language header, returning
+// a default value if no localized value matches the languages specified in the accept-language header.
+func localize(defaultValue string, localizedValues map[string]string, acceptLanguageHeader string) string {
```
reinkrul, line 9: I guess browsers always send their preferred language?

reinkrul (changes requested):

auth/api/iam/localize.go
beardedfoo: Yea this is asking a lot of implementers. It wouldn't be *too* insane to create a map of decimal values and then sort the keys based on them but what would you put as a default value when the q-factor is missing?

auth/api/iam/localize.go
beardedfoo: Never mix the words "browser" and "always" in a sentence. We will just fall back to default if it is blank, which I should create a unit test for.

reinkrul (approved):

PR #2455, opened 2023-08-30
Always write network revocations to backup store

### LINKED ISSUE #2450: backup-revoked-credentials.db contains all revocations after node restart
After the initial sync of a new node (without issuing/revoking any credentials) the `backup-revoked-credentials.db` is empty. On restart the logs contain the message below and the node copies the `revocations` shelf from the `verifier-store.db` to the `backup-revoked-credentials.db`
```
level=info msg="Missing store for shelf, creating from index" module=Storage storeShelf=revocations
```

`backup-revoked-credentials.db` should be filled at the same time as `verifier-store.db`

note that copying the shelf takes ~10sec and is not related to #2448

gerardsn:
fixes #2450

commit 2 is the bug, commit 3 is nice to have but maybe not appropriate for back port?

### DIFF
--- a/e2e-tests/storage/backup-restore/run-test.sh
+++ b/e2e-tests/storage/backup-restore/run-test.sh
@@ -65,15 +65,6 @@ runOnAlpine "$(pwd):/host/" rm -rf /host/node-data
 runOnAlpine "$(pwd):/host/" mv -f /host/node-backup /host/node-data
 BACKUP_INTERVAL=0 docker compose up --wait
 
-echo "Rebuilding data"
-# Sanity check for test; assert that the data is not there before rebuild
-assertDiagnostic "http://localhost:11323" "revocations_count: 0"
-docker compose exec nodeA nuts network reprocess "application/vc+json"
-docker compose exec nodeA nuts network reprocess "application/ld+json;type=revocation"
-
-# Wait for reprocess to finish
-waitForDiagnostic "nodeA" revocations_count 1
-
 assertDiagnostic "http://localhost:11323" "transaction_count: 5"
 assertDiagnostic "http://localhost:11323" "credential_count: 2"
 assertDiagnostic "http://localhost:11323" "issued_credentials_count: 2"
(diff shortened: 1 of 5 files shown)


gerardsn:
e2e fails because `verifier-store.db`'s expected behavior has recently changed to immediately rebuild from `backup-revoked-credentials.db` on startup, and with this fix that now actually works. I'm not sure what changes are needed in the e2e test for it to make sense again

woutslakhorst:
What is the underlying problem that is fixed?

gerardsn:
see commit two. It writes to the Leia store, not the kvBackedStore.

woutslakhorst:
A collection is just a pointer. It's set from the kvBackedStore:
```
kvBackedStore.Collection(leia.JSONCollection, "issuedCredentials")
```

gerardsn:
The `KVBackedLeiaStore` implements the Leia store interface and on `Add` it writes to both stores while the `store.Collection` is only linked to the Leia store, not the kv store.

gerardsn:
Or put differently, `kvBackedLeiaStore.Collection()` returns a `*kvBackedCollection` that also wraps the backup `stoabs.KVStore` while `leia.Store.Collection()` returns a `*leia.Collection`.

woutslakhorst:
Yes, but it already did this, it got a `*kvBackedCollection` by calling `Collection` on the `kvBackedStore`. Nothing has changed in this PR, only that it calls collection every time instead of just once.

gerardsn:
That is what it should have done and is fixed in this PR. It called `Collection` on `store`, which is a `store, err := leia.NewStore(dbPath)`. Note that this is only fixed in the verifier store, the issuer store was/is correct.

gerardsn:
Perhaps this is where the confusion occurs: `NewLeiaVerifierStore` creates `leiaVerifierStore` with field `store` that is a `KVBackedLeiaStore `, but the `store` in `store.Collection` that I'm referring to is a`leia.Store` defined in `NewLeiaVerifierStore` . We have too many stores.

woutslakhorst (approved):

PR #2458, opened 2023-08-31
Auth: make publicurl required and add stricter validation

gerardsn:
The transition from sharing all data via the network to sharing data through public APIs means that the node needs to know what its public URL is in many places. (authorization server and client metadata, and did:web for example). Forcing users to configure the address instead of trying to guess it just makes a lot of sense.

in strict mode the url MUST use https, not be an IP (did:web requirement), and not use a reserved domain name (should still allows names like `https://nodeA` that resolve within docker and what we use in e2e tests)

Q: what should be validated in non-strict mode?


gerardsn:
It looks like the `nuts.yaml` isn't applied after a restart of the container in the failing e2e test. Locally this works just fine...

core/url.go
```diff
@@ -35,3 +41,59 @@ func JoinURLPaths(parts ...string) string {
 	}
 	return result
 }
+
+// ParsePublicURL parses the given input string as URL and asserts that
+// it has a scheme,
+// it is not an IP address, and
+// it is not a reserved address or TLD as described in RFC2606 or https://www.ietf.org/archive/id/draft-chapin-rfc2606bis-00.html.
+func ParsePublicURL(input string) (*url.URL, error) {
```
reinkrul, line 49: can't this just be a private func in the auth module? Not really any reason to put it in core, since no other modules use it?

network/transport/types.go
```diff
@@ -135,71 +133,6 @@ type Contact struct {
 // NutsCommServiceType holds the DID document service type that specifies the Nuts network service address of the Nuts node.
 const NutsCommServiceType = "NutsComm"
 
-// ParseNutsCommAddress parses the given input string to a gRPC target address.
-// The input must include the protocol scheme (e.g. grpc://).
-// The address must NOT be an IP address.
-// The input must not be a reserved address or TLD as described in RFC2606 or https://www.ietf.org/archive/id/draft-chapin-rfc2606bis-00.html.
-func parseNutsCommAddress(input string) (*url.URL, error) {
```
reinkrul, line 142: I would keep this here

reinkrul:
Why change all the test configs if http is still allowed in non-strict mode?

auth/auth.go
```diff
@@ -173,3 +172,37 @@ func (auth *Auth) Start() error {
...
+func parsePublicURL(publicURLStr string, strictmode bool) (*url.URL, error) {
+	if publicURLStr == "" {
+		return nil, errors.New("must provide url")
+	}
+
+	var publicURL *url.URL
+	var err error
+	if strictmode {
+		// PublicURL cannot use a reserved address, IP, or http:// in strictmode
+		publicURL, err = core.ParsePublicURL(publicURLStr)
```
reinkrul, line 186: now the func is split over 2 modules

network/transport/types.go
```diff
@@ -211,10 +144,13 @@ func (s *NutsCommURL) UnmarshalJSON(bytes []byte) error {
 	if err := json.Unmarshal(bytes, &str); err != nil {
 		return errors.New("endpoint not a string")
 	}
-	endpoint, err := parseNutsCommAddress(str)
+	endpoint, err := core.ParsePublicURL(str)
```
reinkrul, line 147: oh now I get it, but kinda confusing though (naming). But maybe a bit of code duplication won't hurt here?

If you're going to have a shared function for checking URLs, I'd make it more useful by adding `allowsSchemes` and `allowReservedHostnames` as parameters, then you can remove most of the new code in auth and in this file.

And the name is probably more like a "validate" function than "parse"?

reinkrul (changes requested):

network/transport/types.go
gerardsn: Added `allowedScemes` and `allowReserved` to the signature. I only expect a boolean or maybe an error with more details from a validate function, not the actual URL, so left at parse.

gerardsn:
> Why change all the test configs if http is still allowed in non-strict mode?

Several of the test specify strictmode=true in the config, or no strictmode which then defaults to true.

reinkrul (approved):

PR #2463, opened 2023-09-06
release notes for v5.4.3

gerardsn:


docs/pages/release_notes.rst
```diff
@@ -3,6 +3,17 @@
...
 #############
 
+************************
+Hazelnut update (v5.4.3)
+************************
+
+Release date: 2023-09-06
+
+- Fixed an issue where revocations received through the network were not written to a backup that was introduced in v5.4.0.
+Nodes upgrading from v5.4.0-v5.4.2 need to make an empty POST call to ``<node-address>/internal/network/v1/reprocess?type=application/ld+json%3Btype=revocation``.
```
reinkrul, line 13: and which performed a backup restore after upgrading to v5.4.x

reinkrul (changes requested):

docs/pages/release_notes.rst
gerardsn: It is also required after the node has been restarted while running one of these versions. 
I prefer to keep the instructions as simple as possible.

woutslakhorst (approved):

PR #2464, opened 2023-09-06
make scope to presentationMapping configurable

### LINKED ISSUE #2390: Add configuration option for scope to PresentationDefinition mapping
For example when an eOverdracht client gets an AuthorizationRequest using the `eOverdrachtV1` scope. The authorization server must translate the `eOverdrachtV1` scope to the correct PresentationDefinition.

It would probably be best to manage this through an API?

woutslakhorst:
closes #2390 

a `auth.presentationexchangemappingfile` configuration option has been added that should point to a json file.
This json file has scopes as fields. The value of these fields should be presentation definition objects. See example in vcr/pe/test


vcr/pe/store.go
```diff
@@ -0,0 +1,72 @@
...
+
+	return nil
+}
+
+// ByScope returns the presentation definition for the given scope.
+// Returns nil if it doesn't exist or if no mappings are loaded.
+func (s *Store) ByScope(scope string) *PresentationDefinition {
+	if s.mapping == nil {
+		return nil
+	}
```
gerardsn, line 66: No need to check, `ok` is just false when `s.mapping == nil`
```suggestion
```

vcr/pe/store.go
```diff
@@ -0,0 +1,72 @@
...
+
+import (
+	"encoding/json"
+	"io"
+	"os"
+)
+
+// Store is a store for presentation definitions
+// It loads a file with the mapping from oauth scope to presentation definition
+type Store struct {
```
reinkrul, line 29: This is now a "presentation exchange store" (since it's in the `pe` package), which is a confusing name. Maybe "DefinitionStore" is better? Or a `DefinitionResolver` with  a

```
type DefinitionResolver interface {
     ByScope(string) *PresentationDefinition
}

type DefinitionMap map[string]PresentationDefinition
```

type which had the `ByScope` function? Then the LoadFromFile and LoadFromBytes can simply we top-level functions, no need to make them methods on the type, I think?

auth/interface.go
```diff
@@ -37,4 +38,7 @@ type AuthenticationServices interface {
 	V2APIEnabled() bool
 	// PublicURL returns the public URL of the node.
 	PublicURL() *url.URL
+
+	// PresentationDefinitions returns the PresentationDefinitions
```
reinkrul, line 42: not really

vcr/pe/store.go
```diff
@@ -0,0 +1,72 @@
...
+	defer reader.Close()
+	bytes, err := io.ReadAll(reader)
+	if err != nil {
+		return err
+	}
+
+	return s.LoadFromBytes(bytes)
+}
+
+func (s *Store) LoadFromBytes(bytes []byte) error {
```
reinkrul, line 50: unused outside of test code in this package, so make it non-exported

reinkrul (changes requested):

auth/auth.go
```diff
@@ -155,6 +162,13 @@ func (auth *Auth) Configure(config core.ServerConfig) error {
 		return err
 	}
 
+	// load presentation definitions
+	if auth.config.PresentationExchangeMappingFile != "" {
```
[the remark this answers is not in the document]
woutslakhorst: which demo?

vcr/pe/store.go
woutslakhorst: I planned on using it somewhere, but wasn't needed so removed.

vcr/pe/types.go
```diff
@@ -69,13 +72,16 @@ type IsHolderItems struct {
...
-	Frame                  *Frame                                         `json:"frame,omitempty"`
-	Id                     string                                         `json:"id"`
-	InputDescriptors       []*InputDescriptor                             `json:"input_descriptors"`
-	Name                   string                                         `json:"name,omitempty"`
-	Purpose                string                                         `json:"purpose,omitempty"`
-	SubmissionRequirements []*SubmissionRequirement                       `json:"submission_requirements,omitempty"`
+	Format *PresentationDefinitionClaimFormatDesignations `json:"format,omitempty"`
+	Frame  *Frame                                         `json:"frame,omitempty"`
+	// Id is the id of the presentation definition, it must be unique within the context.
+	Id               string             `json:"id"`
```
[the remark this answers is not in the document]
woutslakhorst: Not sure how it's used. You have to reference it when creating a presentation submission. In that case it would not matter if it has changed at some point.

vcr/pe/store.go
woutslakhorst: There's no need for an interface

auth/auth.go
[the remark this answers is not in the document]
reinkrul: OpenID4VP server-to-server wallet. Not really documented or thought out and more for testing, so we could also just copy it to all of our dev stations if we want to test it. Are there actually going to be any preloaded presentation exchanges here, or will it be a (yet another) configuration step for setting up your node?

auth/cmd/cmd.go
```diff
@@ -60,6 +62,7 @@ func FlagSet() *pflag.FlagSet {
 	flags.Int(ConfAccessTokenLifeSpan, defs.AccessTokenLifeSpan, "defines how long (in seconds) an access token is valid. Uses default in strict mode.")
 	flags.StringSlice(ConfContractValidators, defs.ContractValidators, "sets the different contract validators to use")
 	flags.Bool(ConfV2APIEnabled, defs.V2APIEnabled, "enables experimental v2 API endpoints")
+	flags.String(ConfPresentationExchangeMappingFile, defs.PresentationExchangeMappingFile, "sets the path to the presentation exchange mapping file")
```
reinkrul, line 65: Do we want to document it (the feature, file format, etc) at this point, for ourselves and others, or is it probable that it'll change in (near) future?

reinkrul (approved):
OK with comments

gerardsn (approved):

auth/cmd/cmd.go
woutslakhorst: marked is as hidden for now. We'll demo first and then decide on the next step

PR #2471, opened 2023-09-07
Chore: rename VDR didservice package to service

reinkrul:
Since we now have DID method implementations in the `didweb`, `didjwk` and `didnuts` packages, `didservice` has become a bad package name. Renamed to `service`.

### DIFF
--- a/auth/api/iam/api.go
+++ b/auth/api/iam/api.go
@@ -30,7 +30,7 @@ import (
 	"github.com/nuts-foundation/nuts-node/core"
 	"github.com/nuts-foundation/nuts-node/vcr"
 	"github.com/nuts-foundation/nuts-node/vcr/openid4vci"
-	"github.com/nuts-foundation/nuts-node/vdr/didservice"
+	"github.com/nuts-foundation/nuts-node/vdr/service"
 	vdr "github.com/nuts-foundation/nuts-node/vdr/types"
 	"html/template"
 	"net/http"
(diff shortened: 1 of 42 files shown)


woutslakhorst (commented):
wouldn't `resolver` be even better? (and nest the specific packages under this dir?)

reinkrul:
> wouldn't `resolver` be even better? (and nest the specific packages under this dir?)

Maybe, at least more descriptive of what functionality it provides (most stuff would fit). Ideally, we'd get rid of the `types` package as well (at least partly) but that's too big of a refactor.

gerardsn (commented):
`service` is a very generic name that IMO just leads to more confusion when used in other packages. 

I think this package mainly exists to fix circular dependencies between `network` and `vdr`. Anything causing s conflict should probably move to `didnuts`, and anything not specific to a particular did method probably belongs in `vdr` (or perhaps `resolvers`).

Now that I think about it, many things related to service references are probably `didnuts` specific since we do not follow the DID spec on dereferencing. (We probably need to look at this in the context of (generating a) did:web too.)

reinkrul:
A `types` package proved to be an anti-pattern (the same goes for `service`, which is a just-dump-it-all package), a new `resolver` (including DIDResolver interface and most of the stuff currently in `didservice`) and maybe `management` (including all `Manipulator` stuff, including types as well) might help clearing up stuff

reinkrul:
Superseeded by https://github.com/nuts-foundation/nuts-node/pull/2513

PR #2473, opened 2023-09-07
Update go-did

reinkrul:
New version has a util function for resolving credential subject DID, eliminating code in the Nuts node.

### DIFF
--- a/vcr/holder/wallet.go
+++ b/vcr/holder/wallet.go
@@ -121,21 +121,16 @@ func (h wallet) BuildPresentation(ctx context.Context, credentials []vc.Verifiab
 }
 
 func (h wallet) resolveSubjectDID(credentials []vc.VerifiableCredential) (*did.DID, error) {
-	type credentialSubject struct {
-		ID did.DID `json:"id"`
-	}
 	var subjectID did.DID
 	for _, credential := range credentials {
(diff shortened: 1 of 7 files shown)


vcr/holder/wallet.go
```diff
@@ -121,21 +121,16 @@ func (h wallet) BuildPresentation(ctx context.Context, credentials []vc.Verifiab
...
 	for _, credential := range credentials {
-		var subjects []credentialSubject
-		err := credential.UnmarshalCredentialSubject(&subjects)
-		if err != nil || len(subjects) != 1 {
-			return nil, errors.New("not all VCs contain credentialSubject.id")
+		sid, err := credential.SubjectDID()
+		if err != nil {
+			return nil, err
 		}
-		subject := subjects[0]
```
gerardsn, line 134: this is changed behavior since it now allows multiple subjects as long as they as have the same Id? (same in issuer) Now less strict so probably fine.

gerardsn (approved):

vcr/holder/wallet.go
reinkrul: Yes, but it will ultimately only use 1 of them or fail in some place else. This is, after all, just storage.

PR #2485, opened 2023-09-13
added auth v2 request access token api for s2s flow

### LINKED ISSUE #2484: internal endpoint for requesting an access token (new version) at your own Nuts node

woutslakhorst:
fixes #2484 

This is a scaffold + input checking for the s2s RequestAccessToken internal API.
Placed under iam package. I could imagine our internal APIs for authv2 to be placed under the auth package instead?

+regenerated APIs using codegen 1.15.0

### DIFF
--- a/auth/api/iam/s2s_vptoken.go
+++ b/auth/api/iam/s2s_vptoken.go
@@ -19,9 +19,12 @@
 package iam
 
 import (
+	"context"
 	"errors"
 	"github.com/labstack/echo/v4"
+	"github.com/nuts-foundation/go-did/did"
 	"github.com/nuts-foundation/nuts-node/core"
+	"github.com/nuts-foundation/nuts-node/vdr/types"
 	"net/http"
 )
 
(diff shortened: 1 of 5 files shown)


reinkrul:
codegen tools update better to have in a separate PR

docs/_static/auth/iam.yaml
```diff
@@ -210,52 +210,52 @@ paths:
...
-#          content:
-#            application/json:
-#              schema:
-#                $ref: '#/components/schemas/TokenResponse'
-#        default:
-#          $ref: '../common/error_response.yaml'
+  /internal/auth/v2/{did}/request-access-token:
+    post:
+      operationId: requestAccessToken
+      summary: Requests an access token using the vp_token-bearer grant.
```
gerardsn, line 216: `make gen-api` for this change to take effect

docs/_static/auth/iam.yaml
```diff
@@ -210,52 +210,52 @@ paths:
...
-#      responses:
-#        '200':
-#          description: Successful request. Responds with an access token as described in rfc6749 section 5.1.
-#          content:
-#            application/json:
-#              schema:
-#                $ref: '#/components/schemas/TokenResponse'
-#        default:
-#          $ref: '../common/error_response.yaml'
+  /internal/auth/v2/{did}/request-access-token:
```
gerardsn, line 213: How many APIs do we expect to add on internal/auth? I don't think this specific API conflicts with the auth/v1 api?

auth/api/iam/s2s_vptoken.go
```diff
@@ -60,3 +63,39 @@ func (s serviceToService) handleAuthzRequest(_ map[string]string, _ *Session) (*
...
 	return nil, nil
 }
+
+func (r Wrapper) RequestAccessToken(ctx context.Context, request RequestAccessTokenRequestObject) (RequestAccessTokenResponseObject, error) {
+	if request.Body == nil {
+		// why did oapi-codegen generate a pointer for the body??
+		return nil, core.InvalidInputError("missing request body")
+	}
+	// resolve wallet
+	requestHolder, err := did.ParseDID(request.Did)
```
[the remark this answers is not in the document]
woutslakhorst: "borrowed" it from the other handlers. They all seem to be wrong.
The APIs are mounted on `/iam/{did}`, this should be `/iam/{id}`. Then we manually add `did:web` to get a DID.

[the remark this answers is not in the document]
woutslakhorst: But this API is an internal API so it should be given the full DID because that is how it is stored in the calling app.

docs/_static/auth/iam.yaml
woutslakhorst: but two different APIs with the same name also doesn't help....

auth/api/iam/s2s_vptoken.go
```diff
@@ -60,3 +63,39 @@ func (s serviceToService) handleAuthzRequest(_ map[string]string, _ *Session) (*
...
+	// resolve verifier metadata
+	requestVerifier, err := did.ParseDID(request.Body.Verifier)
+	if err != nil {
+		return nil, core.InvalidInputError("invalid verifier: %w", err)
+	}
+	_, _, err = r.vdr.Resolver().Resolve(*requestVerifier, nil)
+	if err != nil {
+		if errors.Is(err, types.ErrNotFound) {
+			return nil, core.InvalidInputError("verifier not found: %w", err)
+		}
```
gerardsn, line 94: sounds like a case for `IsFunctionalResolveError`?

gerardsn (approved):
just the one comment

PR #2486, opened 2023-09-13
regenerated APIs with latest codegen

woutslakhorst:
te prevent polluting another PR.


reinkrul (dismissed):
Update Makefile's `install-tools` target.

And optionally the codegen runtime dependency in `go.mod`

woutslakhorst:
Runtime does not have a newer version

gerardsn (approved):

PR #2488, opened 2023-09-13
Add client metadata + API

gerardsn:
I tried to fill out as much as possible, but there are still a bunch of unknowns (tagged with TODOs). I've skipped all the fields related to openid connect / id_token for now. We can add these as needed.

### DIFF
--- a/auth/api/iam/api.go
+++ b/auth/api/iam/api.go
@@ -280,7 +280,7 @@ func (r Wrapper) validateAsNutsFingerprint(ctx context.Context, fingerprint stri
 			return core.NotFoundError(err.Error())
 		}
 		log.Logger().WithField("did", nutsDID.String()).Errorf("oauth metadata: failed to assert ownership of did: %s", err.Error())
-		return err
+		return core.Error(500, err.Error())
 	}
 	if !owned {
 		return core.NotFoundError("did not owned")
--- a/auth/api/iam/api.go
+++ b/auth/api/iam/api.go
@@ -280,7 +280,7 @@ func (r Wrapper) validateAsNutsFingerprint(ctx context.Context, fingerprint stri
 			return core.NotFoundError(err.Error())
 		}
 		log.Logger().WithField("did", nutsDID.String()).Errorf("oauth metadata: failed to assert ownership of did: %s", err.Error())
-		return core.Error(500, err.Error())
+		return err
 	}
 	if !owned {
 		return core.NotFoundError("did not owned")
--- a/auth/api/iam/api.go
+++ b/auth/api/iam/api.go
@@ -22,6 +22,7 @@ import (
 	"context"
 	"embed"
 	"errors"
+	"fmt"
 	"github.com/labstack/echo/v4"
 	"github.com/nuts-foundation/go-did/did"
 	"github.com/nuts-foundation/nuts-node/audit"
@@ -34,6 +35,7 @@ import (
 	vdr "github.com/nuts-foundation/nuts-node/vdr/types"
 	"html/template"
 	"net/http"
+	"strings"
 	"sync"
 )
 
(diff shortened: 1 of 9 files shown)


auth/api/iam/api.go
```diff
@@ -228,6 +230,43 @@ func (r Wrapper) GetOAuthAuthorizationServerMetadata(ctx context.Context, reques
...
+	}
+
+	// assert ownership of did
+	owned, err := r.vdr.IsOwner(ctx, *nutsDID)
+	if err != nil {
+		if didservice.IsFunctionalResolveError(err) {
+			return core.NotFoundError(err.Error())
+		}
+		log.Logger().WithField("did", nutsDID.String()).Errorf("oauth metadata: failed to assert ownership of did: %s", err.Error())
+		return core.Error(500, err.Error())
```
woutslakhorst, line 261: Just return error, it'll map to 500

auth/api/iam/api.go
```diff
@@ -228,6 +230,43 @@ func (r Wrapper) GetOAuthAuthorizationServerMetadata(ctx context.Context, reques
...
+	if strings.HasPrefix(fingerprint, "did:") {
+		return core.InvalidInputError("id contains full did")
+	}
+	nutsDID, err := did.ParseDID("did:nuts:" + fingerprint)
+	if err != nil {
+		return core.InvalidInputError(err.Error())
+	}
+
+	// assert ownership of did
+	owned, err := r.vdr.IsOwner(ctx, *nutsDID)
```
woutslakhorst, line 255: I'm guessing IsOwner doesn't work on web dids since we have did:nuts in filename/vaultKey

Otherwise converting to nuts did is weird.

We should probably fix IsOwner to work with web dids

auth/api/iam/metadata.go
```diff
@@ -34,3 +38,23 @@ func authorizationServerMetadata(identity url.URL) OAuthAuthorizationServerMetad
 		ClientIdSchemesSupported: clientIdSchemesSupported,
 	}
 }
+
+// clientMetadata should only be used for dids managed by the node. It assumes the provided identity URL is correct.
+func clientMetadata(identity url.URL) OAuthClientMetadata {
+	softwareID, softwareVersion, _ := strings.Cut(core.UserAgent(), "/")
+	return OAuthClientMetadata{
+		//RedirectURIs:            nil,
+		TokenEndpointAuthMethod: "client_secret_basic", // TODO: this is the default value, what do we support
```
woutslakhorst, line 47: I'm guessing it would become `vp_token-bearer`?

docs/_static/auth/iam.yaml
```diff
@@ -136,6 +136,39 @@ paths:
                 "$ref": "#/components/schemas/OAuthAuthorizationServerMetadata"
         default:
           $ref: '../common/error_response.yaml'
+#  TODO: what should this path be? It is not a /.well-known/. It will be provided in an authorization request as the "client_metadata_uri"
```
woutslakhorst, line 139: will be temporary right?

auth/api/iam/types.go
```diff
@@ -203,3 +203,72 @@ type OAuthAuthorizationServerMetadata struct {
 	// If omitted, the default value is `pre-registered` (referring to the client), which is currently not supported.
 	ClientIdSchemesSupported []string `json:"client_id_schemes_supported,omitempty"`
 }
+
+type OAuthClientMetadata struct {
```
woutslakhorst, line 207: godoc

woutslakhorst: Any optional string values? SHould be pointers with `omitempty`

auth/api/iam/metadata_test.go
```diff
@@ -19,20 +19,22 @@
 package iam
 
 import (
+	"github.com/nuts-foundation/nuts-node/core"
 	"github.com/stretchr/testify/assert"
 	"net/url"
 	"testing"
 )
 
+var vpFormats = map[string]map[string][]string{
```
woutslakhorst, line 28: let's just refer to the global var

auth/api/iam/metadata.go
```diff
@@ -34,3 +38,23 @@ func authorizationServerMetadata(identity url.URL) OAuthAuthorizationServerMetad
...
+		GrantTypes:              grantTypesSupported,
+		ResponseTypes:           responseTypesSupported,
+		//Scope:                   "",
```
woutslakhorst, line 50: remove comments

woutslakhorst (changes requested):

auth/api/iam/api.go
gerardsn: correct. currently a did:web is just a derivative of did:nuts.

auth/api/iam/metadata.go
gerardsn: no that would require defining a client assertion using the vp_token next to the grant_type similar to RFC7523 describing a
`urn:ietf:params:oauth:client-assertion-type:jwt-bearer` and 
`urn:ietf:params:oauth:grant-type:jwt-bearer`

docs/_static/auth/iam.yaml
gerardsn: what do you mean? OpenID4VP declares `client_metadata` and `client_metadata_uri` parameters for the authorization request. Since the amount of characters in the metadata is too much to add to the query params it should be provided as `client_metadata_uri`. Having to request the client metadata at the client itself also prevents tampering with the `redirect_uris`. So I don't think this is temporary

auth/api/iam/metadata.go
gerardsn: I left them as comments to see what fields still need to added, but I will remove them.

auth/api/iam/metadata_test.go
gerardsn: what is a good way to prevent an accidental typo changing the global var? Besides code review.

docs/_static/auth/iam.yaml
woutslakhorst: right, but it's defined at runtime. So we can decide whatever we want.

auth/api/iam/metadata_test.go
woutslakhorst: not much. A wrong constant would show up in the development network at runtime.

docs/_static/auth/iam.yaml
gerardsn: Ahh.., then yes. I will remove the comment

auth/api/iam/metadata.go
gerardsn: Left the remaining values commented. They either need to be set by the time these flows are implmented or removed.

woutslakhorst (approved):

PR #2489, opened 2023-09-13
Create well-known path from oauth issuer

gerardsn:


auth/api/iam/metadata_test.go
```diff
@@ -20,10 +20,50 @@ package iam
...
 import (
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	"net/url"
 	"testing"
 )
 
+func TestIssuerIdToWellKnown(t *testing.T) {
+	t.Run("ok", func(t *testing.T) {
+		issuer := "https://nuts.nl/iam/id"
```
woutslakhorst, line 30: example.com would be better for test.

auth/api/iam/metadata_test.go
```diff
@@ -20,10 +20,50 @@ package iam
...
+	t.Run("no IP allowed", func(t *testing.T) {
+		issuer := "http://127.0.0.1/iam/id"
+		u, err := IssuerIdToWellKnown(issuer, authzServerWellKnown, false)
+		require.ErrorContains(t, err, "hostname is IP")
+		assert.Nil(t, u)
+	})
+	t.Run("invalid URL", func(t *testing.T) {
+		issuer := "http:// /iam/id"
+		u, err := IssuerIdToWellKnown(issuer, authzServerWellKnown, true)
+		require.ErrorContains(t, err, "invalid character \" \" in host name")
```
woutslakhorst, line 62: I would do a 
```
require.Error(t, err)
assert.ErrorContains(...)
assert.Nil()
```
Then the ErrorContains doesn't block the assert.Nil.

Or do assert.Nil first....

woutslakhorst (approved):

auth/api/iam/metadata_test.go
gerardsn: copy paste error from require.NoError(). updated to assert.ErrorContains()

auth/api/iam/metadata_test.go
woutslakhorst: then ignore my comment.

auth/api/iam/metadata_test.go
```diff
@@ -47,19 +47,19 @@ func TestIssuerIdToWellKnown(t *testing.T) {
 	t.Run("https in strictmode", func(t *testing.T) {
 		issuer := "http://nuts.nl/iam/id"
 		u, err := IssuerIdToWellKnown(issuer, authzServerWellKnown, true)
-		require.ErrorContains(t, err, "scheme must be https")
+		assert.ErrorContains(t, err, "scheme must be https")
```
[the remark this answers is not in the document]
gerardsn: it checks for err == nil

PR #2490, opened 2023-09-14
Feature/2487/oauth authorization metadata client

### LINKED ISSUE #2487: Resolve DID and find authorization server metadata
Part of #2418 

The authorization server endpoint for `did:web:domain/path/id` MUST be located at: `https://domain/.well-known/oauth-authorization-server/path/id/`

woutslakhorst:
waiting for #2489 
waiting for #2485 

closes #2487

Added API client for fetching OAuth Authorization Server metadata.

The api client is generated from the OAS. I changed some names in the spec because they were named from a server handle perspective, from a client perspective they don't make sense.


auth/api/iam/client.go
```diff
@@ -0,0 +1,88 @@
...
+	// TODO: ignoring root web did for now. We can't use the generated client for that type of web:did :(
+	// TODO: use IssuerIdToWellKnown func when merged
+	serverURL := strings.ReplaceAll(serverDID.ID, ":", "/")
+	// remove % encodings
+	serverURL, err := url.QueryUnescape(serverURL)
+	if err != nil {
+		return nil, err
+	}
+
+	response, err := hb.clientWithBase(fmt.Sprintf("https://%s", serverURL)).OAuthAuthorizationServerMetadata(ctx, serverDID.String())
```
gerardsn, line 68: serverURL should be the OAuth issuer identifier (before adding well-known).

from [RFC8414](https://www.rfc-editor.org/rfc/rfc8414.html#section-2):
> issuer
      REQUIRED.  The authorization server's issuer identifier, which is
      a URL that uses the "https" scheme and has no query or fragment
      components.  Authorization server metadata is published at a
      location that is ".well-known" according to [RFC 5785](https://www.rfc-editor.org/rfc/rfc5785) [[RFC5785](https://www.rfc-editor.org/rfc/rfc5785)]
      derived from this issuer identifier, as described in [Section 3](https://www.rfc-editor.org/rfc/rfc8414.html#section-3).
      The issuer identifier is used to prevent authorization server mix-
      up attacks, as described in "OAuth 2.0 Mix-Up Mitigation"
      [[MIX-UP](https://www.rfc-editor.org/rfc/rfc8414.html#ref-MIX-UP)].

`https` is already checked in `IssuerIdToWellKnown`. 
Query/frament are not allowed. Add to `IssuerIdToWellKnown`?

auth/api/iam/client.go
```diff
@@ -0,0 +1,88 @@
...
+func (hb HTTPClient) clientWithBase(baseURL string) ClientInterface {
+	// can only be used for public APIs
+	response, err := NewClientWithResponses(baseURL, WithHTTPClient(hb.httpClient))
+	if err != nil {
+		panic(err)
+	}
+	return response
+}
+
+func (hb HTTPClient) OAuthAuthorizationServerMetadata(ctx context.Context, serverDID did.DID) (*OAuthAuthorizationServerMetadata, error) {
```
gerardsn, line 55: godoc explaining serverDID

auth/api/iam/client.go
```diff
@@ -0,0 +1,88 @@
...
+	if serverDID.Method != "web" {
+		return nil, fmt.Errorf("unsupported DID method: %s", serverDID.Method)
+	}
+	// TODO: ignoring root web did for now. We can't use the generated client for that type of web:did :(
+	// TODO: use IssuerIdToWellKnown func when merged
+	serverURL := strings.ReplaceAll(serverDID.ID, ":", "/")
+	// remove % encodings
+	serverURL, err := url.QueryUnescape(serverURL)
```
gerardsn, line 63: add a DIDToURL to match the inverse in didweb/utils.go? did:web apparently has some extra requirements on url encoding/decoding and a possible port number should be unencoded.

auth/api/iam/client.go
```diff
@@ -0,0 +1,88 @@
...
+		httpClient: core.MustCreateHTTPClient(config, nil),
+	}
+}
+
+func (hb HTTPClient) clientWithBase(baseURL string) ClientInterface {
+	// can only be used for public APIs
+	response, err := NewClientWithResponses(baseURL, WithHTTPClient(hb.httpClient))
+	if err != nil {
+		panic(err)
+	}
```
gerardsn, line 51: 2x panic, is this acceptable(/impossible)? I think the second can be triggered when `baseURL` is not properly validated?

gerardsn (changes requested):

auth/api/iam/client.go
woutslakhorst: An error can only occur from the pass options to `NewClientWithResponses`. We don't pass any that can cause an error. I'll add a comment that when adding options, it should be reassed.

auth/api/iam/client.go
woutslakhorst: Replaced with DIDToURL

gerardsn (changes requested):

vdr/didweb/web.go
```diff
@@ -68,43 +66,19 @@ func (w Resolver) Resolve(id did.DID, _ *types.ResolveMetadata) (*did.Document,
...
-		return nil, nil, fmt.Errorf("invalid did:web: ID must be a domain name, not IP address")
+	splitted := strings.Split(id.ID, ":")
+	if len(splitted) == 1 {
+		// if the id doesn't contain a path we set '/.well-known/did.json' s path
+		baseURL.Path = "/.well-known/did.json"
+	} else {
+		// if there's a path we prepend `./well-known/` to the path
+		baseURL.Path = strings.Join(splitted[1:], "/") + "/did.json"
 	}
+	targetURL := baseURL.String()
```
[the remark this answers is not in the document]
woutslakhorst: I don't think I follow, Example?

gerardsn (approved):
one suggestion

PR #2492, opened 2023-09-15
added presentation_definition endpoint

### LINKED ISSUE #2491: use presentation definition endpoint to get presentation definition from other party
The presentation definition endpoint is not defined in any standard.
We place it under `/iam/:did/presentation_definition?scope=`

woutslakhorst:
closes #2491 

This RP adds the presentation definition endpoint and client API. A new variable has been added to the OAuth Authorization Server Metadata: `presentation_definition_endpoint`


auth/api/iam/api.go
```diff
@@ -252,6 +251,24 @@ func (r Wrapper) OAuthClientMetadata(ctx context.Context, request OAuthClientMet
...
+func (r Wrapper) PresentationDefinition(_ context.Context, request PresentationDefinitionRequestObject) (PresentationDefinitionResponseObject, error) {
+	if len(request.Params.Scope) == 0 {
+		return PresentationDefinition200JSONResponse([]PresentationDefinition{}), nil
+	}
+
+	// todo: only const scopes supported, scopes with variable arguments not supported yet
+	// map all scopes to a presentation definition
+	presentationDefinitions := make([]PresentationDefinition, 0, len(request.Params.Scope))
+	for _, scope := range request.Params.Scope {
+		presentationDefinition := r.auth.PresentationDefinitions().ByScope(scope)
```
gerardsn, line 263: `scope` can contain multiple space separated scopes. This needs to be checked.

docs/_static/auth/iam.yaml
```diff
@@ -123,6 +123,45 @@ paths:
...
+          required: true
+          schema:
+            type: string
+            example: did:nuts:123
+        - name: scope
+          in: query
+          required: true
+          schema:
+            type: array
+            description: The scope for which a presentation definition is requested. Multiple scopes can be specified by separating them with a space.
```
gerardsn, line 150: OAuth defines `scope` as a single space delimited string. We have the freedom to do what we want here, but in all the OAuth/OpenID defined flows the `scope` parameter is always going to be a single string. For consistency I'd do the same here.

docs/_static/auth/iam.yaml
```diff
@@ -123,6 +123,45 @@ paths:
...
+            items:
+              type: string
+              example: patient:x:read
+      responses:
+        "200":
+          description: Authorization request accepted, user is asked for consent.
```
gerardsn, line 156: copy-pasta description?

auth/api/iam/api.go
gerardsn, line 263: This generates a PD per scope while we need a PD per consent backend?

gerardsn: Or are we providing access per scope? (I always thought it was accept/reject entire access token request)

docs/_static/auth/iam.yaml
```diff
@@ -123,6 +123,45 @@ paths:
...
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  "$ref": "#/components/schemas/PresentationDefinition"
```
gerardsn, line 162: OpenID4VP flow requires a single PD for a similar request. Perhaps we should combine the different PDs in to single PD and add a `group` label per PD source. This allows selection of a single group to comply with within the combined PD using `Submission Requirements`.

gerardsn (changes requested):
dealing with some of comments should probably deferred

auth/api/iam/api.go
```diff
@@ -252,6 +251,24 @@ func (r Wrapper) OAuthClientMetadata(ctx context.Context, request OAuthClientMet
 
 	return OAuthClientMetadata200JSONResponse(clientMetadata(*identity)), nil
 }
+func (r Wrapper) PresentationDefinition(_ context.Context, request PresentationDefinitionRequestObject) (PresentationDefinitionResponseObject, error) {
```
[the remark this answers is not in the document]
reinkrul: Since the s2s-flow is custom, I don't see many other implementations and users of the endpoint. So validation could involve getting the definition again given the scope, and comparing it against the submitted PE?

auth/api/iam/client.go
```diff
@@ -80,3 +81,39 @@ func (hb HTTPClient) OAuthAuthorizationServerMetadata(ctx context.Context, webDI
 
 	return &metadata, nil
 }
+
+// PresentationDefinition retrieves the presentation definition for the given web DID and scope(s).
+// We pass the endpoint url for the presentation definition endpoint because we already retrieved the metadata in a previous step.
+// The scopes are evaluated as raw query params and encoded if needed.
```
reinkrul, line 87: This assumes very tight coupling with the calling code ("previous step"). And clients always handle input encoding, so no need to explicitly that that.

```suggestion
// PresentationDefinition retrieves the presentation definition from the presentation definition endpoint (as specified by RFC021) for the given scope.
```

auth/api/iam/client.go
```diff
@@ -80,3 +81,39 @@ func (hb HTTPClient) OAuthAuthorizationServerMetadata(ctx context.Context, webDI
...
+func (hb HTTPClient) PresentationDefinition(ctx context.Context, definitionEndpoint string, scopes []string) ([]PresentationDefinition, error) {
+	presentationDefinitionURL, err := url.Parse(definitionEndpoint)
+	if err != nil {
+		return nil, err
+	}
+	presentationDefinitionURL.RawQuery = url.Values{"scope": scopes}.Encode()
```
reinkrul, line 93: or should they be space-separated? An OAuth2 access token (or request) always has a single scope (but it can actually consist of multiple space-separated scopes)

auth/api/iam/client.go
```diff
@@ -80,3 +81,39 @@ func (hb HTTPClient) OAuthAuthorizationServerMetadata(ctx context.Context, webDI
 
 	return &metadata, nil
 }
+
+// PresentationDefinition retrieves the presentation definition for the given web DID and scope(s).
+// We pass the endpoint url for the presentation definition endpoint because we already retrieved the metadata in a previous step.
+// The scopes are evaluated as raw query params and encoded if needed.
+func (hb HTTPClient) PresentationDefinition(ctx context.Context, definitionEndpoint string, scopes []string) ([]PresentationDefinition, error) {
+	presentationDefinitionURL, err := url.Parse(definitionEndpoint)
```
reinkrul, line 89: if you want a `url.URL`, you can just ask for it (string -> url.URL)?

auth/api/iam/client.go
```diff
@@ -80,3 +81,39 @@ func (hb HTTPClient) OAuthAuthorizationServerMetadata(ctx context.Context, webDI
...
+	}
+
+	definitions := make([]PresentationDefinition, 0)
+	var data []byte
+
+	if data, err = io.ReadAll(response.Body); err != nil {
+		return nil, fmt.Errorf("unable to read response: %w", err)
+	}
+	if err = json.Unmarshal(data, &definitions); err != nil {
+		return nil, fmt.Errorf("unable to unmarshal response: %w, %s", err, string(data))
```
reinkrul, line 115: better not include the data, could be binary (or very large).

auth/api/iam/client.go
```diff
@@ -80,3 +81,39 @@ func (hb HTTPClient) OAuthAuthorizationServerMetadata(ctx context.Context, webDI
...
+	// create a GET request with scope query param
+	request, err := http.NewRequest(http.MethodGet, presentationDefinitionURL.String(), nil)
+	if err != nil {
+		return nil, err
+	}
+	response, err := hb.httpClient.Do(request.WithContext(ctx))
+	if err != nil {
+		return nil, fmt.Errorf("failed to call endpoint: %w", err)
+	}
+	if err = core.TestResponseCode(http.StatusOK, response); err != nil {
```
reinkrul, line 104: unsupported scope might deserve its own error flow path?

docs/_static/auth/iam.yaml
```diff
@@ -123,6 +123,45 @@ paths:
               schema:
                 type: string
                 format: uri
+  "/iam/{did}/presentation_definition":
```
reinkrul, line 126: Maybe reference RFC021?

docs/_static/auth/iam.yaml
```diff
@@ -123,6 +123,45 @@ paths:
...
+    get:
+      summary: Used by relying parties to obtain a presentation definition for desired scopes.
+      description: |
+        Specified by https://identity.foundation/presentation-exchange/spec/v2.0.0/
+        The presentation definition is a JSON object that describes the desired verifiable credentials and presentation formats.
+        A presentation definition is matched against a wallet. If verifiable credentials matching the definition are found, 
+        a presentation can created and submitted together with a presentation submission.
```
reinkrul, line 133:
```suggestion
        a presentation can created together with a presentation submission.
```

docs/_static/auth/iam.yaml
```diff
@@ -332,6 +371,10 @@ components:
         OAuth2 Client Metadata
         Contain properties from several specifications and may grow over time
       type: object
+    PresentationDefinition:
+      description: |
+        A presentation definition is a JSON object that describes the desired verifiable credentials and presentation formats.
```
reinkrul, line 376: add link to spec?

reinkrul (changes requested):

auth/api/iam/api.go
woutslakhorst: The API gets them as slice.

auth/api/iam/client.go
woutslakhorst: You get it as a string from the metadata, so it would move the parsing to the caller. That would duplicate parsing code for every call.

reinkrul (changes requested):

auth/api/iam/api.go
```diff
@@ -245,6 +244,24 @@ func (r Wrapper) OAuthClientMetadata(ctx context.Context, request OAuthClientMet
...
+	}
+
+	// todo: only const scopes supported, scopes with variable arguments not supported yet
+	// todo: we only take the first scope as main scope, when backends are introduced we need to use all scopes and send them as one to the backend.
+	scopes := strings.Split(request.Params.Scope, " ")
+	presentationDefinition := r.auth.PresentationDefinitions().ByScope(scopes[0])
+	if presentationDefinition == nil {
+		return PresentationDefinition400JSONResponse{
+			Error: "invalid_scope",
+		}, nil
```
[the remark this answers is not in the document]
woutslakhorst: Compile error due to missing methods on that type (oapi-codegen specific)

docs/_static/auth/iam.yaml
```diff
@@ -123,6 +123,51 @@ paths:
...
+            type: string
+            example: did:nuts:123
+        - name: scope
+          in: query
+          required: true
+          schema:
+            type: string
+            description: | 
+              The scope for which a presentation definition is requested. Multiple scopes can be specified by separating them with a space.
+              The first scope is the main scope that determines the backend, additional scopes can be used to format the presentation definition.
```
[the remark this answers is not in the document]
woutslakhorst: Removed specifics, they should be specified by use-cases

woutslakhorst:
The PE suggestion won't work since you don't know which one of the backends you pick will actually give you access. The patient specific scope will only be checked after a client submitted a VP. If a client picks the wrong one, it'll fail and does not retry. There's no complexity on the CIS side since it's all handled by the node.

auth/api/iam/client.go
woutslakhorst: Putting ur.URL in the metadata will require custom marshalling/unmarshalling for the metadata struct.

auth/api/iam/api.go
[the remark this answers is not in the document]
woutslakhorst: backends do not have to check this, they only need to check if given VCs are enough for the given scope.

woutslakhorst:
#2540 describes additional features required for Submission Requirement Feature.

auth/api/iam/api.go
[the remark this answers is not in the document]
gerardsn: I just realized that the PD.id generated by the backends will be thrown away when all PDs are merged

gerardsn (changes requested):
just the one comment

auth/api/iam/client.go
```diff
@@ -80,3 +82,41 @@ func (hb HTTPClient) OAuthAuthorizationServerMetadata(ctx context.Context, webDI
 
 	return &metadata, nil
 }
+
+// PresentationDefinition retrieves the presentation definition from the presentation definition endpoint (as specified by RFC021) for the given scope.
+func (hb HTTPClient) PresentationDefinition(ctx context.Context, definitionEndpoint string, scopes []string) (*PresentationDefinition, error) {
+	presentationDefinitionURL, err := url.Parse(definitionEndpoint)
```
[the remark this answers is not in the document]
woutslakhorst: I also changed the parsing so it will allow IPs in non-strictmode

auth/api/iam/client.go
```diff
@@ -85,7 +85,7 @@ func (hb HTTPClient) OAuthAuthorizationServerMetadata(ctx context.Context, webDI
 
 // PresentationDefinition retrieves the presentation definition from the presentation definition endpoint (as specified by RFC021) for the given scope.
 func (hb HTTPClient) PresentationDefinition(ctx context.Context, definitionEndpoint string, scopes []string) (*PresentationDefinition, error) {
-	presentationDefinitionURL, err := url.Parse(definitionEndpoint)
+	presentationDefinitionURL, err := core.ParsePublicURL(definitionEndpoint, !hb.config.Strictmode)
```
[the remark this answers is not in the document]
gerardsn: Merging #2560 should make this easier

gerardsn (approved):
I don't think allowing IP in non-strictmode is going to be possible since parsing of did:web fails for IP addresses. But it might make unit tests easier if this is allowed?

woutslakhorst:
@reinkrul waiting for your review

auth/api/iam/client.go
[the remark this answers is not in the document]
woutslakhorst: Hmmmm #2560 flipped the 2nd argument, although the PR build succeeds, after merging it'll fail....

auth/api/iam/error.go
```diff
@@ -125,3 +126,16 @@ func (p oauth2ErrorWriter) Write(echoContext echo.Context, _ int, _ string, err
 	redirectURI.RawQuery = query.Encode()
 	return echoContext.Redirect(http.StatusFound, redirectURI.String())
 }
+
+const InvalidScope = ErrorCode("invalid_scope")
```
reinkrul, line 130: consts are on line 35

auth/api/iam/error.go
```diff
@@ -125,3 +126,16 @@ func (p oauth2ErrorWriter) Write(echoContext echo.Context, _ int, _ string, err
...
+
+var ErrInvalidScope = errors.New("invalid scope")
```
reinkrul, line 132: could just return `OAuth2Error`? You're already unmarshalling it
Then you also have the description and stuff

reinkrul (approved):
Small remarks

PR #2493, opened 2023-09-15
changed OAS to reflect path component is the ID part of a DID

woutslakhorst:
The APIs under `iam/{id}` no longer expect a DID as param but only the ID part.

### DIFF
--- a/auth/api/iam/api.go
+++ b/auth/api/iam/api.go
@@ -216,7 +216,7 @@ func (r Wrapper) GetOAuthAuthorizationServerMetadata(ctx context.Context, reques
 		return nil, core.NotFoundError("authz server metadata: did not owned")
 	}
 
-	identity := r.auth.PublicURL().JoinPath(ownDID.WithoutURL().String())
+	identity := r.auth.PublicURL().JoinPath("iam", request.Id)
 
 	return GetOAuthAuthorizationServerMetadata200JSONResponse(authorizationServerMetadata(*identity)), nil
 }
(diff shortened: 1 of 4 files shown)


gerardsn (commented):
I added api.go#validateAsNutsFingerprint in #2488 to do just this. I did not apply it to the other APIs in that PR yet for the sake of scope creep. Which one should we keep?

auth/api/iam/api.go
```diff
@@ -208,85 +202,55 @@ func (r Wrapper) HandleAuthorizeRequest(ctx context.Context, request HandleAutho
...
-		log.Logger().WithField("did", id.String()).Errorf("authz server metadata: failed to assert ownership of did: %s", err.Error())
+		log.Logger().WithField("did", ownDID.String()).Errorf("authz server metadata: failed to assert ownership of did: %s", err.Error())
 		return nil, core.Error(500, "authz server metadata: %w", err)
 	}
 	if !owned {
 		return nil, core.NotFoundError("authz server metadata: did not owned")
 	}
 
-	identity := r.auth.PublicURL().JoinPath(apiPath, id.WithoutURL().String())
+	identity := r.auth.PublicURL().JoinPath(ownDID.WithoutURL().String())
```
gerardsn, line 219: same as ClientMetadata
```suggestion
	identity := r.auth.PublicURL().JoinPath("iam", request.Id)
```

gerardsn (changes requested):
the content of `GetOAuthAuthorizationServerMetadata` and `GetOAuthClientMetadata` should match 1:1

gerardsn (approved):

PR #2513, opened 2023-09-26
VDR: Refactor types and didservice packages

reinkrul:
Replaces https://github.com/nuts-foundation/nuts-node/pull/2471

Goals of this PR are:
- Getting rid of the `types` package, since packages should imply function, not technical contents
- Getting rid of the `didservice` package, since it's a dumping place for every "service"-y type that doesn't fit anywhere else. And the same argument as for `types` applies here as well.
- Making a clear separation of management (create, update, lower-level stuff) of DIDs owned by the current node and resolving any DID/key/service (regardless whether it's owned by the node or not).

Changes:

- Removed `vdr/didservice/common.go#CompoundService` (type alias for `map[string]string`), was just used in `didman.go`
- Removed `vdr/didservice/common.go#ErrDIDAlreadyExists` (unused)
- Removed `vdr/didservice/test.go` (unused)
- Moved all did-resolving-related types (constants, interfaces, implementations) from `vdr/types` and `vdr/didservice`  to `vdr/resolvers/did.go`. The same for keys and services.
- Renamed `vdr/didservice#KeyResolver` implementation to `DIDKeyResolver` (otherwise it clashes with the interface name).
- Renamed `vdr/didservice#ServiceResolver` implementation to `DIDServiceResolver` (otherwise it clashes with the interface name).
- Moved all DID creation/updating types in `vdr/didservice` to `vdr/management/management.go`
- Moved all DID "finding" types (iterating over Nuts DID store) to `vdr/management/finder.go`
- Renamed VDR main struct from `VDR` to `Module`, to avoid clash with the interface name (later on, the `VDR` interface should be renamed as well, since packages shouldn't contain types which starts with the same word or are named the same). We could do this for all module implementations, for consistency.

No functionality or behavior has been changed/removed.


woutslakhorst (commented):
Is there a reason to have a separate `management` and `resolver` package?

What happens if you place contents of those packages in the VDR root?

reinkrul:
> Is there a reason to have a separate `management` and `resolver` package?

Separation of concerns; `management` is for everything you want to do with your own DIDs, `resolver` is for resolving any DID (or key, or service) for any DID method.

> What happens if you place contents of those packages in the VDR root?

Cyclic dependencies with at least `network` module.

woutslakhorst (approved):

PR #2515, opened 2023-09-26
IAM: Implement correct OAuth2 error handling

reinkrul:
This PR introduces an OAuth2 error handler that:

- Redirects the browser back to the provided `redirect_uri` (with `error` and `error_description` parameters)
- If there's no `redirect_uri`, it either prints the error as plain text (for a browser) or JSON (for an API client)

It also fixes HTTP request logging status codes for the IAM API package.

### DIFF
--- a/auth/api/iam/api.go
+++ b/auth/api/iam/api.go
@@ -29,7 +29,6 @@ import (
 	"github.com/nuts-foundation/nuts-node/auth/log"
 	"github.com/nuts-foundation/nuts-node/core"
 	"github.com/nuts-foundation/nuts-node/vcr"
-	"github.com/nuts-foundation/nuts-node/vcr/openid4vci"
 	"github.com/nuts-foundation/nuts-node/vdr"
 	"github.com/nuts-foundation/nuts-node/vdr/resolver"
 	"html/template"
@@ -80,8 +79,7 @@ func (r Wrapper) Routes(router core.EchoRouter) {
 				// Add http.Request to context, to allow reading URL query parameters
 				requestCtx := context.WithValue(ctx.Request().Context(), httpRequestContextKey, ctx.Request())
 				ctx.SetRequest(ctx.Request().WithContext(requestCtx))
-				// TODO: Do we need a generic error handler?
-				// ctx.Set(core.ErrorWriterContextKey, &protocolErrorWriter{})
+				ctx.Set(core.ErrorWriterContextKey, &oauth2ErrorWriter{})
 				return f(ctx, request)
 			}
 		},
@@ -118,30 +116,21 @@ func (r Wrapper) HandleTokenRequest(ctx context.Context, request HandleTokenRequ
 		// Options:
 		// - OpenID4VCI
 		// - OpenID4VP, vp_token is sent in Token Response
+		panic("not implemented")
 	case "vp_token":
 		// Options:
 		// - service-to-service vp_token flow
+		panic("not implemented")
 	case "urn:ietf:params:oauth:grant-type:pre-authorized_code":
 		// Options:
 		// - OpenID4VCI
+		panic("not implemented")
 	default:
-		// TODO: Don't use openid4vci package for errors
-		return nil, openid4vci.Error{
-			Code:       openid4vci.InvalidRequest,
-			StatusCode: http.StatusBadRequest,
-			//Description: "invalid grant type",
+		return nil, OAuth2Error{
+			Code:        InvalidRequest,
+			Description: "unsupported grant_type",
 		}
 	}
-
-	// TODO: Handle?
-	//scope, err := handler(request.Body.AdditionalProperties)
-	//if err != nil {
-	//	return nil, err
-	//}
-	// TODO: Generate access token with scope
-	return HandleTokenRequest200JSONResponse(TokenResponse{
-		AccessToken: "",
-	}), nil
 }
 
 // HandleAuthorizeRequest handles calls to the authorization endpoint for starting an authorization code flow.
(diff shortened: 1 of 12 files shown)


auth/api/iam/api.go
```diff
@@ -118,30 +116,21 @@ func (r Wrapper) HandleTokenRequest(ctx context.Context, request HandleTokenRequ
 		// Options:
 		// - OpenID4VCI
 		// - OpenID4VP, vp_token is sent in Token Response
+		panic("not implemented")
```
woutslakhorst, line 119: now we can crash any live node? There's an `unsupported_grant_type` error.

auth/api/iam/api.go
```diff
@@ -118,30 +116,21 @@ func (r Wrapper) HandleTokenRequest(ctx context.Context, request HandleTokenRequ
...
 	default:
-		// TODO: Don't use openid4vci package for errors
-		return nil, openid4vci.Error{
-			Code:       openid4vci.InvalidRequest,
-			StatusCode: http.StatusBadRequest,
-			//Description: "invalid grant type",
+		return nil, OAuth2Error{
+			Code:        InvalidRequest,
+			Description: "unsupported grant_type",
```
woutslakhorst, line 131: `invalid_request` or `unsupported_grant_type`? (not both)

auth/api/iam/error.go
```diff
@@ -0,0 +1,125 @@
...
+		}
+	}
+	if oauthErr.Code == "" {
+		// Somebody forgot to set a code
+		oauthErr.Code = ServerError
+	}
+	redirectURI, _ := url.Parse(oauthErr.RedirectURI)
+	if oauthErr.RedirectURI == "" || redirectURI == nil {
+		// Can't redirect the user-agent back, render error as JSON or plain text (depending on content-type)
+		contentType := echoContext.Request().Header.Get("Content-Type")
```
woutslakhorst, line 104: `Content-Type` or `Accept` header?

auth/api/iam/error.go
```diff
@@ -0,0 +1,125 @@
...
+	// ServerError is returned when the Authorization Server encounters an unexpected condition that prevents it from fulfilling the request.
+	ServerError ErrorCode = "server_error"
+)
+
+// Make sure the error implements core.HTTPStatusCodeError, so the HTTP request logger can log the correct status code.
+var _ core.HTTPStatusCodeError = OAuth2Error{}
+
+// OAuth2Error is an OAuth2 error that signals the error was (probably) caused by the client (e.g. bad request),
+// or that the client can recover from the error (e.g. retry).
+type OAuth2Error struct {
```
woutslakhorst, line 49: There are also some generated error structs, those can be alliased?

auth/api/iam/api.go
```diff
@@ -80,8 +79,7 @@ func (r Wrapper) Routes(router core.EchoRouter) {
 				// Add http.Request to context, to allow reading URL query parameters
 				requestCtx := context.WithValue(ctx.Request().Context(), httpRequestContextKey, ctx.Request())
 				ctx.SetRequest(ctx.Request().WithContext(requestCtx))
-				// TODO: Do we need a generic error handler?
-				// ctx.Set(core.ErrorWriterContextKey, &protocolErrorWriter{})
+				ctx.Set(core.ErrorWriterContextKey, &oauth2ErrorWriter{})
```
woutslakhorst, line 82: Must all APIs under `/aim` return oauth errors?

woutslakhorst (changes requested):

auth/api/iam/api.go
reinkrul: Ah, it's a TODO, but we can return `unsupported_grant_type` indeed.

(node won't crash, the HTTP handler (goroutine) will)

auth/api/iam/api.go
reinkrul: `unsupported_grant_type`

auth/api/iam/error.go
reinkrul: At first, I had both. But that introduced unused code paths (`Accept` header being `application/json`), which adds complexity and no value. So given the assumption that a browser will never send `application/json` as content-type (maybe with a form post with enctype set to that value, but that would be very, very weird?), and API clients will use JSON, I just settles on `Content-Type = application/json` for now.

When proven otherwise in future, we can easily adapt this.

auth/api/iam/api.go
reinkrul: For now I'd say yes, and it doesn't hurt. When new non-OAuth2 APIs are added to IAM, we could/should introduce a skipper func?

auth/api/iam/error.go
reinkrul: Defined as:

```
// ErrorResponse defines model for ErrorResponse.
type ErrorResponse struct {
	// Error Code identifying the error that occurred.
	Error string `json:"error"`
}
```

Our error type also has some helper fields (RedirectURI, InternalError) and additional functions, so I don't see much value in using the generated type.

auth/api/iam/api.go
reinkrul: I see we also have `/internal/auth/v2/{did}/request-access-token`. I'll only apply it for request URLs starting with `/iam/`

reinkrul:
Force-pushed after fixing conflicts a bit too early

auth/api/iam/error.go
woutslakhorst: Not using the generated type, but reusing this new type and omit the generated type from generation.

woutslakhorst (approved):

PR #2523, opened 2023-10-02
Resolving did:key DIDs

reinkrul:
We might need this for supporting demos with Sphereon wallet.

This adds a dependency on `go-multicodec`, which we could go without if we just take the constants (it's a limited set).

TODO:
- [x] Test for other key types
- [x] Register the resolver

### DIFF
--- a/vdr/didkey/resolver.go
+++ b/vdr/didkey/resolver.go
@@ -109,6 +109,7 @@ func (r Resolver) Resolve(id did.DID, _ *resolver.ResolveMetadata) (*did.Documen
 		if rsaKey.Size() < 256 {
 			return nil, nil, errors.New("did:key: RSA public key is too small (must be at least 2048 bits)")
 		}
+		key = rsaKey
 	default:
 		return nil, nil, fmt.Errorf("did:key: unsupported public key type: 0x%x", keyType)
 	}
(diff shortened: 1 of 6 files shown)


woutslakhorst:
Given the number of LOC, it wouldn't hurt to add the method

reinkrul:
I struggle decoding the public keys, so I'm first going ask at Sphereon about the `did:key` DID; it might only be used for the demo credential (and I suspect they'll remove it in future iterations).

vdr/didkey/resolver.go
```diff
@@ -0,0 +1,115 @@
+package didkey
```
woutslakhorst, line 1: copyright

vdr/didkey/resolver_test.go
```diff
@@ -0,0 +1,133 @@
...
+func TestResolver_Resolve(t *testing.T) {
+	resolver := Resolver{}
+	type testCase struct {
+		name string
+		did  string
+		jwk  map[string]interface{}
+	}
+
+	// Fixtures were taken from https://github.com/digitalbazaar/did-method-key/blob/main/test/driver.spec.js
+	testCases := []testCase{
```
woutslakhorst, line 21: missing testcases for other key types

woutslakhorst (changes requested):

reinkrul:
Added lots of tests

woutslakhorst (approved):

PR #2525, opened 2023-10-03
IAM: Use SessionDatabase in OpenID4VP

reinkrul:
Also removed unused authorized code flow.


auth/api/iam/openid4vp.go
```diff
@@ -162,12 +171,16 @@ func (r *Wrapper) handlePresentationRequest(params map[string]string, session *S
...
-		session = r.sessions.Get(sessionID)
+	sessionID := c.FormValue("sessionID")
+	if sessionID == "" {
+		return errors.New("missing sessionID parameter")
 	}
-	if session == nil {
-		return errors.New("invalid session")
+
+	var session Session
+	sessionStore := r.storageEngine.GetSessionDatabase().GetStore(sessionExpiry, session.OwnDID.String(), "session")
```
woutslakhorst, line 180: `session` is too ambiguous. `openid4vp`,`session` would be better when you debug and list keys in session storage.

woutslakhorst (approved):
Won't block the PR for the requested change.

auth/api/iam/openid4vp.go
reinkrul: For the authorized code flow/OpenID4VP, this is more in POC state than actually production-grade, so I expect it to change in future. But I'll change the keys. Idea for keys:

- `openid.(did).session` (can be OpenID4VP/SIOPv2 or OpenID4VCI)
- `s2s.(did).code`

PR #2526, opened 2023-10-04
Verify VC from Sphereon Wallet

reinkrul:


reinkrul:
Very stale now, will probably be reworked after DIIP v2 is out.

PR #2555, opened 2023-10-20
Switch to mrtron/base58 lib, maintainer looks more reputable 

reinkrul:
(more active OSS contributor)

Both libraries were already a transitive dependency, just replaced the direct usage.


gerardsn (changes requested):
relevant tests are failing

reinkrul:
> relevant tests are failing

I didn't even care running them locally, because I assumed it would be a drop-in replacement. As proven again, assumption is the mother of all ******* (censored).

Some error messages obviously changed.

gerardsn (approved):

PR #2560, opened 2023-10-25
Refactor ParsePublicURL

gerardsn:
All URLs obtained from external sources (metadata endpoints) need to be validated for OAuth2/OpenID4VC meaning that all methods using these URLs will require the following codeblok:
```
if strictmode {
	issuerURL, err = core.ParsePublicURL(issuer, false, "https")
} else {
	issuerURL, err = core.ParsePublicURL(issuer, true, "https", "http")
}
```

This PR makes that a bit easier


woutslakhorst (approved):

gerardsn:
Because this is effectively a validation of the `auth.publicUrl` config value configured in the other nuts node. Maybe a bit redundant, but I like that it is specific.

PR #2561, opened 2023-10-25
Support ed25591 keys and signatures

reinkrul:
We might need it for some demos. Maybe we need to introduce configuration for restricting what algorithms are allowed?

Like a "Dutch FIPS mode"


reinkrul:
Will be included in secp256k1 PR.

Issue #2553, opened 2023-10-20
Support path_nested in Presentation Exchanges


reinkrul:
We should support `path_nested` when processing presentation submissions (https://identity.foundation/presentation-exchange/#presentation-submission)

gerardsn:
[OpenID4VC](https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#name-nested-verifiable-presentat) restricts this feature to 1 level of nesting.

`PresentationDefinition.Match` should also apply the nesting in the created `PresentationSubmission`

PR #2563, opened 2023-10-25
generate presentation submission with path_nested

woutslakhorst:
Part of #2553 

This PR covers the presentation submission creation with `path_nested`.

Requires #2559


gerardsn:
can you rebase this so its easier to review

gerardsn (changes requested):

vcr/pe/presentation_definition.go
```diff
@@ -47,16 +53,39 @@ type Candidate struct {
...
 // It doesn't do the credential search, this should be done before calling this function.
-// The PresentationDefinition.Format should be altered/set if an envelope defines the supported format before calling.
-// The resulting PresentationSubmission has paths that are relative to the matching VCs.
-// The PresentationSubmission needs to be altered so the paths use "path_nested"s that are relative to the created VP.
+// The given PresentationContext is used to set the correct vp index in the nested paths and to alter the given PresentationSubmission.
+// It assumes this method is used for OpenID4VP since other envelopes require different nesting.
 // ErrUnsupportedFilter is returned when a filter uses unsupported features.
 // Other errors can be returned for faulty JSON paths or regex patterns.
-func (presentationDefinition PresentationDefinition) Match(vcs []vc.VerifiableCredential) (PresentationSubmission, []vc.VerifiableCredential, error) {
+func (presentationDefinition PresentationDefinition) Match(context PresentationContext, vcs []vc.VerifiableCredential) ([]vc.VerifiableCredential, error) {
```
reinkrul, line 60: weird public-facing API; what to does the caller need to do with "Index"? This should probably be an internal shelper struct. And this function should create the submission, in my opinion (or there should be a  separate `PresentationDefinition.CreateSubmission` which does this, potato/potato). Now callers have to set the right `DefinitionId`, while this is part of creating the submission.

vcr/pe/presentation_definition.go
```diff
@@ -47,16 +53,39 @@ type Candidate struct {
...
+	} else if descriptorMaps, selectedVCs, err = presentationDefinition.matchBasic(vcs); err != nil {
+		return nil, err
 	}
-	return presentationDefinition.matchBasic(vcs)
+
+	// wrap each InputDescriptorMappingObject for the outer VP
+	for _, descriptorMap := range descriptorMaps {
+		nestedDescriptorMap := InputDescriptorMappingObject{
+			Id:     descriptorMap.Id,
+			Format: "ldp_vp",                            // todo support jwt_vp
```
reinkrul, line 79: easy, just get the format from `selectedVCs[i].Format()` (if the slices pair up)

reinkrul (changes requested):

vcr/pe/presentation_definition.go
woutslakhorst: I'll create an `AddMatch` to a PresentationSubmission. That way the submission can also handle the index and single vs plural paths.

vcr/pe/presentation_submission.go
```diff
@@ -0,0 +1,132 @@
...
+// A PresentationSubmissionBuilder can be used to create a PresentationSubmission with multiple wallets as input.
+func (presentationDefinition PresentationDefinition) Builder() PresentationSubmissionBuilder {
+	return PresentationSubmissionBuilder{
+		presentationDefinition: presentationDefinition,
+	}
+}
+
+// AddWallet adds credentials from a wallet that may be used to create the PresentationSubmission.
+func (b *PresentationSubmissionBuilder) AddWallet(holder did.DID, vcs []vc.VerifiableCredential) {
+	b.holders = append(b.holders, holder)
```
reinkrul, line 63: note: we might have to support multiple holders later (match submission from 2 wallets of different holders).

vcr/pe/presentation_submission.go
```diff
@@ -0,0 +1,132 @@
...
+// Builder returns a new PresentationSubmissionBuilder.
+// A PresentationSubmissionBuilder can be used to create a PresentationSubmission with multiple wallets as input.
+func (presentationDefinition PresentationDefinition) Builder() PresentationSubmissionBuilder {
+	return PresentationSubmissionBuilder{
+		presentationDefinition: presentationDefinition,
+	}
+}
+
+// AddWallet adds credentials from a wallet that may be used to create the PresentationSubmission.
+func (b *PresentationSubmissionBuilder) AddWallet(holder did.DID, vcs []vc.VerifiableCredential) {
```
reinkrul, line 62: a builder normally returns itself for call chaining

vcr/pe/presentation_submission.go
```diff
@@ -0,0 +1,132 @@
...
+// Multiple presentation definitions can be added to the builder.
+type PresentationSubmissionBuilder struct {
+	holders                []did.DID
+	presentationDefinition PresentationDefinition
+	wallets                [][]vc.VerifiableCredential
+}
+
+// Builder returns a new PresentationSubmissionBuilder.
+// A PresentationSubmissionBuilder can be used to create a PresentationSubmission with multiple wallets as input.
+func (presentationDefinition PresentationDefinition) Builder() PresentationSubmissionBuilder {
```
reinkrul, line 55: `Builder()` of type X normally builds X (although Go's different since it has no static functions), so maybe rename this to `Submission()`?

reinkrul (approved):
Better, small remarks left

vcr/pe/presentation_submission.go
woutslakhorst: Which already works for the submission part. Only the VP creation part is missing, therefore the SignInstructions are added.

vcr/pe/presentation_submission.go
woutslakhorst: Named it `PresentationSubmissionBuilder`

PR #2564, opened 2023-10-26
Support secp256k1 and ed25519 keys

reinkrul:
TODO:
- [x] JWT signing test for ed25519


woutslakhorst (changes requested):

woutslakhorst:
Can use changes from #2686

reinkrul:
Too outdated, wrong and we now support ES256K in some form.

PR #2567, opened 2023-10-26
Add introspection endpoint for s2s flow

### LINKED ISSUE #2541: support access token introspection (api and internals)

gerardsn:
closes #2541

I've copied the v1 introspection OAPI and edited it. There are still a couple of fields unused for the user authentication part.

I've also removed the issuer key from the s2s access token store since its not available during introspection

### DIFF
--- a/auth/api/iam/api.go
+++ b/auth/api/iam/api.go
@@ -147,13 +147,6 @@ func (r Wrapper) HandleTokenRequest(ctx context.Context, request HandleTokenRequ
 
 // IntrospectAccessToken allows the resource server (XIS/EHR) to introspect details of an access token issued by this node
 func (r Wrapper) IntrospectAccessToken(ctx context.Context, request IntrospectAccessTokenRequestObject) (IntrospectAccessTokenResponseObject, error) {
-	// Client authentication if required, return 401 if fails.
-	// TODO: Is this relevant? Should be handled by API authentication middleware
-	clientAuthorized := true
-	if !clientAuthorized {
-		return IntrospectAccessToken401Response{}, nil
-	}
-
 	// Validate token
 	if request.Body.Token == "" {
 		// Return 200 + 'Active = false' when token is invalid or malformed
--- a/auth/api/iam/api.go
+++ b/auth/api/iam/api.go
@@ -21,6 +21,7 @@ package iam
 import (
 	"context"
 	"embed"
+	"encoding/json"
 	"errors"
 	"github.com/labstack/echo/v4"
 	"github.com/nuts-foundation/go-did/did"
@@ -169,16 +170,17 @@ func (r Wrapper) IntrospectAccessToken(ctx context.Context, request IntrospectAc
 	iat := int(token.IssuedAt.Unix())
 	exp := int(token.Expiration.Unix())
 	response := IntrospectAccessToken200JSONResponse{
-		Active:   true,
-		Iat:      &iat,
-		Exp:      &exp,
-		Iss:      &token.Issuer,
-		Sub:      &token.Issuer,
-		ClientId: &token.ClientId,
-		// TODO: include what resources the token is valid for
-		Aud:   nil,
-		Scope: &token.Scope,
-		Vcs:   &token.Presentation.VerifiableCredential,
+		Active:                 true,
+		Iat:                    &iat,
+		Exp:                    &exp,
+		Iss:                    &token.Issuer,
+		Sub:                    &token.Issuer,
+		ClientId:               &token.ClientId,
+		Scope:                  &token.Scope,
+		PdpMap:                 &token.PDPMap,
+		PresentationDefinition: nil,
+		PresentationSubmission: nil,
+		Vps:                    &token.VPToken,
 
 		// TODO: user authentication, used in OpenID4VP flow
 		FamilyName:     nil,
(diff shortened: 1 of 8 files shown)


auth/api/iam/api.go
```diff
@@ -143,6 +144,60 @@ func (r Wrapper) HandleTokenRequest(ctx context.Context, request HandleTokenRequ
 	}
 }
 
+// IntrospectAccessToken allows the resource server (XIS/EHR) to introspect details of an access token issued by this node
+func (r Wrapper) IntrospectAccessToken(ctx context.Context, request IntrospectAccessTokenRequestObject) (IntrospectAccessTokenResponseObject, error) {
+	// Client authentication if required, return 401 if fails.
+	// TODO: Is this relevant? Should be handled by API authentication middleware
+	clientAuthorized := true
```
reinkrul, line 151: leave TODO, add issue, remove dead code.

auth/api/iam/api.go
```diff
@@ -143,6 +144,60 @@ func (r Wrapper) HandleTokenRequest(ctx context.Context, request HandleTokenRequ
...
+		// Return 200 + 'Active = false' when token is invalid or malformed
+		return IntrospectAccessToken200JSONResponse{}, err
+	}
+
+	if token.Expiration.Before(time.Now()) {
+		// Return 200 + 'Active = false' when token is invalid or malformed
+		// can happen between token expiration and pruning of database
+		return IntrospectAccessToken200JSONResponse{}, nil
+	}
+
```
reinkrul, line 173: should we check `not before`?

auth/api/iam/api.go
```diff
@@ -143,6 +144,60 @@ func (r Wrapper) HandleTokenRequest(ctx context.Context, request HandleTokenRequ
...
+		Active:   true,
+		Iat:      &iat,
+		Exp:      &exp,
+		Iss:      &token.Issuer,
+		Sub:      &token.Issuer,
+		ClientId: &token.ClientId,
+		// TODO: include what resources the token is valid for
+		Aud:   nil,
+		Scope: &token.Scope,
+		Vcs:   &token.Presentation.VerifiableCredential,
```
reinkrul, line 187: later, this should probably the VCs that were mapped by the submission. VP might include more VCs.

auth/api/iam/api.go
```diff
@@ -143,6 +144,60 @@ func (r Wrapper) HandleTokenRequest(ctx context.Context, request HandleTokenRequ
...
+
+	// Create and return introspection response
+	iat := int(token.IssuedAt.Unix())
+	exp := int(token.Expiration.Unix())
+	response := IntrospectAccessToken200JSONResponse{
+		Active:   true,
+		Iat:      &iat,
+		Exp:      &exp,
+		Iss:      &token.Issuer,
+		Sub:      &token.Issuer,
```
reinkrul, line 182: should be subject? Or this should probably be derived from the VCs, but having the resource server/AAA Proxy do that could be a bit unpratical. Maybe we should have 2 maps in the response;
- Map for input descriptor ID to verifiable credential
- Map for input descriptor ID to credential subject ID

(might be out of scope for PR)

auth/api/iam/api.go
reinkrul: Edit: I see we don't set it.

auth/api/iam/api.go
woutslakhorst: This is an internal call so should be an authenticated API. remove code, no issue, no todo

auth/api/iam/api.go
woutslakhorst: You\re referring to the client_id. The issuer of the token is not the VC holder, it's the resource owner

auth/api/iam/api.go
woutslakhorst: Better to put the vp_token here. Also add the presentation_submission. The AAA will need those (and scope) to call the PDP => request URL + vp_token + scope + presentation_subission

auth/api/iam/s2s_vptoken.go
```diff
@@ -133,8 +137,8 @@ func (r Wrapper) createAccessToken(issuer did.DID, issueTime time.Time, presenta
 	}, nil
 }
 
-func (r Wrapper) accessTokenStore(issuer did.DID) storage.SessionStore {
-	return r.storageEngine.GetSessionDatabase().GetStore(accessTokenValidity, "s2s", issuer.String(), "accesstoken")
+func (r Wrapper) s2sAccessTokenStore() storage.SessionStore {
```
woutslakhorst, line 140: This removes the tenant. Now an access token would be introspectable across tenant borders and thus usable on all tenants. It would depend on what the AAA does, but best is to return an error.

woutslakhorst (changes requested):

auth/api/iam/api.go
gerardsn: We require `exp - iat <= 5s`, so further reducing that by adding `nbf` in-between doesn't make sense to me.

auth/api/iam/s2s_vptoken.go
gerardsn: The introspection EP currently doesn't know who is making the call, so the tenant is unknown. We need to discus what options we have here.

woutslakhorst: That is indeed a fundamental change. Do we know for sure that the `aud` in the access token request is transferred to the `iss`/`sub`? This makes the access token introspection determine the tenant. That would be OK, but best to check A-Z if we can't leak anything across tenant boundaries.

gerardsn: The presentation definition should extract tenant (did) from the vp_token to the PDPMap as discussed so it can be validated at the AAA.

docs/_static/auth/iam.yaml
```diff
@@ -456,11 +456,24 @@ components:
         scope:
           type: string
           description: granted scopes
-        vcs:
+        pdp_map:
```
reinkrul, line 459: mixing short form and long form now, gotta choose one?

reinkrul, line 459: mixing short form and long form now, gotta choose one?

auth/api/iam/types.go
```diff
@@ -34,6 +34,9 @@ type DIDDocumentMetadata = resolver.DocumentMetadata
 // VerifiableCredential is an alias
 type VerifiableCredential = vc.VerifiableCredential
 
+// VerifiableCredential is an alias
```
woutslakhorst, line 37: wrong comment, since you're no longer referencing a VC from the OAS, the VerifiableCredential alias can be removed?

docs/_static/auth/iam.yaml
woutslakhorst: you could take a look at https://www.iana.org/assignments/jwt/jwt.xhtml for ideas. `verified_claims` might be a good fit?

gerardsn: I don't think `verified_claims` is a good fit (or any other registered claim name). It comes from the OpenID Connect realm and is focussed on user identity. The required `trust_framework` field per claims object also adds unnecessary complexity IMO. I prefer to reserve this for user identity as it also provides a standardized output on the assurance level of the claims about the user.

We'd probably have to customize the claim for our purpose so I prefer to just use a custom claim. This is an internal endpoint so a custom claim does not impact interoperability. I've changed the name to `input_descriptor_constraint_id_map`. Very verbose, but I am open to changes

auth/api/iam/types.go
gerardsn: It will be back after moving OpenID4VCI from the VCR, but will remove for now.

woutslakhorst (approved):

Issue #2548, opened 2023-10-19
Moving away from shared network state


reinkrul:
For a more robust system, we want to move away from shared network state. The network is currently used for the features below, for which we all have a future replacement
- Public transactions:
  - Publishing DIDs: will be replaced by `did:web`
  - Public VCs acting as "registry": will be replaced by "use case list" (requires https://github.com/nuts-foundation/nuts-node/issues/2549)
  - VC revocations: will be replaced by revocation status list, hosted by over HTTP (requires https://github.com/nuts-foundation/nuts-node/issues/2550)
- Private VCs for NutsAuthorizationCredentials: NutsAuthorizationCredentials will be replaced by a Presentation Exchange-driven protocol (requires https://github.com/nuts-foundation/nuts-node/issues/2551)

After these features have been replaced we can drop the gRPC network and nodes don't need to maintain an active peer-to-peer network and state.

woutslakhorst:
More specific: removing shared state makes it easier to maintain and service a node. We noticed that ease of deployment is a much more important non-functional than expected.

woutslakhorst:
all mentioned points are taken care of in v6 release.

PR #2570, opened 2023-10-27
ignore events with jsonld.ContextURLNotAllowedErr at startup

### LINKED ISSUE #2569: Synchronous processing of (failed) events makes node startup slow
#2414 made retrying at startup to synchronous for all failed events. All previously non-retry able errors are retried once at startup since a version upgrade or configuration change may fixed the reason it previously failed.

The stable network has 150+ non-retry able errors, which causes a ~2.5 min delay at startup on my laptop.

gerardsn:
closes #2569

Nearly all non-retry able errors are for contexts that were previously not on the jsonld remote allow list. By ignoring these types of errors Network module startup went from ~2,5 min to <10 sec for a node on stable network.

This hack does not actually fix the underlying issue, and if the context is later put on the allow list it can be `Reprocess`ed, but that will not remove the event from the list. Is this enough with #2548 in mind?

### DIFF
--- a/network/dag/notifier.go
+++ b/network/dag/notifier.go
@@ -23,6 +23,8 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"github.com/nuts-foundation/nuts-node/jsonld"
+	"strings"
 	"time"
 
 	"github.com/avast/retry-go/v4"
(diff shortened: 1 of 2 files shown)


reinkrul (approved):
Will it still be retried when occuring at runtime (TX with payload triggering this error is received while running)?

gerardsn:
It will only be tried once, since it's not retryable. These changes only affect events at startup

PR #2584, opened 2023-11-05
Discovery: OpenAPI spec for server and client

reinkrul:


docs/_static/usecase/v1.yaml
```diff
@@ -0,0 +1,190 @@
...
+          type: string
+    get:
+      summary: Retrieves a use case list
+      description: |
+        An API provided by the use case maintainer to retrieve a use case list, starting at the given lamport timestamp.
+        The client should provide the timestamp it was returned in the last response.
+        If no timestamp is given or it is 0, it will return the full list.
+
+        error returns:
+        * 400 - incorrect input
```
woutslakhorst, line 26: for query param and a 404 for unknown list.

docs/_static/usecase/v1.yaml
```diff
@@ -0,0 +1,190 @@
...
+                  timestamp:
+                    type: integer
+                    minimum: 0
+                  entries:
+                    type: array
+                    items:
+                      $ref: "#/components/schemas/VerifiablePresentation"
+                  tombstones:
+                    description: |
+                      List of presentation IDs that were removed from the list.
```
woutslakhorst, line 57: since the last query and/or non-expired VPs that have been removed?

docs/_static/usecase/v1.yaml
```diff
@@ -0,0 +1,190 @@
...
+      requestBody:
+        description: Presentation identifying the credential subject to remove from the list.
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: "#/components/schemas/VerifiablePresentation"
+      responses:
+        "201":
+          description: Presentation was added to the list
```
woutslakhorst, line 111: removed

docs/_static/usecase/v1.yaml
```diff
@@ -0,0 +1,190 @@
...
+      - in: query
+        name: query
+        schema:
+          type: object
+          additionalProperties:
+            type: string
+        style: form
+        explode: true
+    get:
+      summary: Searches a for presentations on the usecase list.
```
woutslakhorst, line 134: -a

docs/_static/usecase/v1.yaml
```diff
@@ -0,0 +1,190 @@
...
+      description: |
+        An API of the use case client that searches a for presentations on the usecase list,
```
woutslakhorst, line 136: it's a bit weird to have the client and server api on the same path. Before a client can use it as registry some additional work has to be done (did service resolving). I think the result of the list should feed into the same registry as the network for now. An api can be added later.

docs/_static/usecase/v1.yaml
```diff
@@ -0,0 +1,190 @@
...
+                      type: object
+                      description: The Verifiable Credential that matched the query.
+        default:
+          $ref: "../common/error_response.yaml"
+components:
+  schemas:
+    VerifiablePresentation:
+      $ref: "../common/ssi_types.yaml#/components/schemas/VerifiablePresentation"
+  securitySchemes:
+    jwtBearerAuth:
```
woutslakhorst, line 184: security is indeed a question. But'll probably not be a bearer jwt.

the add and delete are "protected" by the VP, what additional protection should we add?

woutslakhorst (changes requested):
Comment from roadmap update: rename it to `DiscoveryService`

docs/_static/usecase/v1.yaml
reinkrul: Replaced with empty presentation (of special type)

docs/_static/usecase/v1.yaml
reinkrul: Replaced with register an empty VP fo special type

docs/_static/usecase/v1.yaml
reinkrul: Not sure we need any additional protection; the "get" is a cheap operation, and to register a VP you need a VC that matches the presentation definition of the service. Risk would be DoS attacks?

docs/_static/usecase/v1.yaml
reinkrul: Then we have to build pruning into that store (it's the "VCR store") as well. Maybe we should just search in both stores. Although the current VCR search does not know about the discovery service IDs, so you don't know in which service to search.

Perhaps search in all of them, and deduplicate results?

docs/_static/discovery/v1.yaml
```diff
@@ -0,0 +1,153 @@
...
+                    type: string
+                  entries:
+                    type: array
+                    items:
+                      $ref: "#/components/schemas/VerifiablePresentation"
+        default:
+          $ref: "../common/error_response.yaml"
+    post:
+      summary: Register a presentation on the discovery service.
+      description: |
```
woutslakhorst, line 56: should also describe that it's used for delete

woutslakhorst (changes requested):

PR #2566, opened 2023-10-26
PEX: Validate presentation submission

reinkrul:

PR #2589, opened 2023-11-06
Discovery: SQLite-based server implementation

reinkrul:
Base maintainer implementation, not including:

- HTTP API
- Checking revocations
- Registration of the module
- Diagnostics

Requires

- https://github.com/nuts-foundation/nuts-node/pull/2595 (SQLite support)
- https://github.com/nuts-foundation/nuts-node/pull/2598 (PEX JWT support)
- https://github.com/nuts-foundation/nuts-node/pull/2566 (PEX validation)


storage/config.go
```diff
@@ -22,9 +22,18 @@ package storage
...
 	Redis RedisConfig `koanf:"redis"`
+	SQL   SQLConfig   `koanf:"sql"`
 }
 
 // DefaultConfig returns the default configuration for the module.
 func DefaultConfig() Config {
-	return Config{}
+	return Config{
+		SQL: SQLConfig{
+			ConnectionString: "file::memory:?cache=shared",
```
woutslakhorst, line 32: What happens if I use a filename? Where is it stored? (not in data is my guess).
It is missing sane defaults, checks and documentation.

woutslakhorst (changes requested):
This PR is about the usecase maintainer impl but introduces SQL on the fly. The migration towards sql storage is a big one and needs a proper setup. We risk in introducing it with add-ons/patches and forget to properly design it.

Issues that come to mind:
- put the migrations in sub dirs per supported DB
- are the migration prefixes with 1, 2, 3 going to break when using 10 ("10"<"2_" stringwise)
- where are we going to put the models for the tabels, all here or in their own packages?
- migration file naming
- connectionstring documentation
- which DBs to support?
- additional config for those DBs and can that be added to later on?

SQL introduction should be done in its own PR

reinkrul:
Extracted SQL functionality into a separate PR: https://github.com/nuts-foundation/nuts-node/pull/2595

reinkrul:
Refactored with server and client having 1 store.

discoveryservice/module.go
```diff
@@ -0,0 +1,232 @@
...
+	if presentation.ID == nil {
+		return errors.New("presentation does not have an ID")
+	}
+	expiration := presentation.JWT().Expiration()
+	// VPs should not be valid for too short; unnecessary overhead at the server and clients.
+	// Also protects against expiration not being set at all. The factor is somewhat arbitrary.
+	minValidity := float64(definition.PresentationMaxValidity / 4)
+	if expiration.Sub(time.Now()).Seconds() < minValidity {
+		return fmt.Errorf("presentation is not valid for long enough (min %s)", time.Duration(minValidity)*time.Second)
+	}
```
woutslakhorst, line 127: undocumented behaviour where the implementation is doing things different than the spec. I would just leave it out.

discoveryservice/module.go
```diff
@@ -0,0 +1,232 @@
...
+	// VPs should not be valid for too long, as that would prevent the server from pruning them.
+	if int(expiration.Sub(time.Now()).Seconds()) > definition.PresentationMaxValidity {
+		return fmt.Errorf("presentation is valid for too long (max %s)", time.Duration(definition.PresentationMaxValidity)*time.Second)
+	}
+
+	if presentation.IsType(retractionPresentationType) {
```
woutslakhorst, line 133: retract is till a weird word, what was wrong with revoke?

discoveryservice/module.go
```diff
@@ -0,0 +1,232 @@
...
+
+var ErrServerModeDisabled = errors.New("node is not a discovery server for this service")
+
+var _ core.Injectable = &Module{}
+var _ core.Runnable = &Module{}
+var _ core.Configurable = &Module{}
+var _ Server = &Module{}
+
+// var _ Client = &Module{}
+var retractionPresentationType = ssi.MustParseURI("RetractedVerifiablePresentation")
```
woutslakhorst, line 45: This isn't in the spec yet?

discoveryservice/module.go
```diff
@@ -0,0 +1,232 @@
...
+	} else {
+		return m.addPresentation(definition, presentation)
+	}
+}
+
+func (m *Module) addPresentation(definition Definition, presentation vc.VerifiablePresentation) error {
+	// Must contain credentials
+	if len(presentation.VerifiableCredential) == 0 {
+		return errors.New("presentation must contain at least one credentialRecord")
+	}
```
woutslakhorst, line 144: No need to check this, will be done by PEX logic.

discoveryservice/module.go
```diff
@@ -0,0 +1,232 @@
...
+	// VP can't be valid longer than the credentialRecord it contains
```
woutslakhorst, line 145: is this in the spec?

discoveryservice/store.go
```diff
@@ -0,0 +1,367 @@
...
+	credential "github.com/nuts-foundation/nuts-node/vcr/credential"
+	"gorm.io/gorm"
+	"gorm.io/gorm/clause"
+	"gorm.io/gorm/schema"
+	"strconv"
+	"strings"
+	"time"
+)
+
+var ErrServiceNotFound = errors.New("discovery service not found")
```
woutslakhorst, line 36: godoc

discoveryservice/store.go
```diff
@@ -0,0 +1,367 @@
...
+	Timestamp uint64
+}
+
+func (s serviceRecord) TableName() string {
+	return "discoveryservices"
+}
+
+var _ schema.Tabler = (*presentationRecord)(nil)
+
+type presentationRecord struct {
```
woutslakhorst, line 50: what does gorm do with respect to snakecasing and case-sensitivity?

discoveryservice/store.go
```diff
@@ -0,0 +1,367 @@
...
+		Where("service_id = ?", serviceID).
+		Joins("inner join discoveryservice_credentials cred ON cred.presentation_id = discoveryservice_presentations.id")
+	numProps := 0
+	for jsonPath, value := range query {
+		if value == "*" {
+			continue
+		}
+		// sort out wildcard mode
+		var eq = "="
+		if strings.HasPrefix(value, "*") {
```
woutslakhorst, line 256: Prefix wildcards are dangerous. Some DBs require this to be enabled. You could also have a fancy search by inserting `.` wildcards between every character.

woutslakhorst (changes requested):

README.rst
```diff
@@ -252,12 +252,12 @@ The following options can be configured on the server:
...
-    storage.sql.connection                                                                                                                                                                                                                                                                                                                                         Connection string for the SQL database. If not set, it defaults to a SQLite database stored inside the configured data directory
+    storage.sql.connection                                                                                                                                                                                                                                                                                                                                         Connection string for the SQL database. If not set, it defaults to a SQLite database stored inside the configured data directory. If specifying a SQLite database, make sure to enable foreign keys with the '_foreign_keys=on' query parameter.
```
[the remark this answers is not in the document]
reinkrul: shared cache is discouraged (https://www.sqlite.org/sharedcache.html), but appearantly the only way to make sqlite3 behave in Go (https://github.com/mattn/go-sqlite3/issues/632), since `database/sql` tries to open multiple connections anyway (and GORM uses the stdlib package). Otherwise, we'll end up with `database busy` errors when load increases (and multiple connections are opened). I think we've seen that error before.

This then touches isolation semantics:
- I used "SELECT FOR UPDATE" to avoid concurrent updates to the discovery service (messing up the timestamping), but this isn't supported by SQLite.
- It does, however, support table-level transactions. There can only be 1 write-TX for each table active.
- A TX, however, does not become a write-TX until the first actual write operation, so the current `updateTimestamp()` that uses `SELECT FOR UDPATE`, `UPDATE` (latter using the result from the first) does not prevent current modification of the timestamps.
- Solution:
  1.  use a `SELECT FOR UPDATE` to make sure there's a row-level lock for databases that support it. Throw away the result. 
  2. rewrite the timestamp-update query to increment in SQL, rather than using the previous `result + 1`
  3. then query the timestamp to return it from the function

discoveryservice/config.go
```diff
@@ -0,0 +1,50 @@
...
+type DefinitionsConfig struct {
+	Directory string `koanf:"directory"`
+}
+
+// ServerConfig holds the config for the server
+type ServerConfig struct {
+	// DefinitionIDs specifies which use case lists the server serves.
+	DefinitionIDs []string `koanf:"definition_ids"`
+	// Directory is the directory where the server stores the lists.
+	Directory string `koanf:"directory"`
```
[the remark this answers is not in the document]
reinkrul: Leftover from pre-SQL PR, removed

vcr/pe/schema/v2/schema.go
```diff
@@ -51,19 +51,21 @@ var PresentationDefinition *jsonschema.Schema
 // PresentationSubmission is the JSON schema for a presentation submission.
 var PresentationSubmission *jsonschema.Schema
 
+// Compiler is the JSON schema compiler.
+var Compiler = jsonschema.NewCompiler()
```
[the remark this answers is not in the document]
reinkrul: I assumed it to be fine, as resources are only added to it in `init()` functions. But at the same time, this is a dangerous assumption (are those init funcs even executed serially?) and it alters the global in another package. Changed it to a function, with the default (PEX) schemas loaded. So each other user will have its own copy of the JSON schemas.

storage/sql_migrations/2_discoveryservice.up.sql
```diff
@@ -0,0 +1,49 @@
+-- discoveryservices contains the known discovery services and the highest timestamp
+create table discoveryservices
+(
+    id        text    not null primary key,
+    timestamp integer not null
```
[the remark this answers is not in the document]
reinkrul:
```
create table foo
(
    timestamp timestamp
)
```

storage/sql_migrations/2_discoveryservice.up.sql
```diff
@@ -0,0 +1,49 @@
+-- discoveryservices contains the known discovery services and the highest timestamp
+create table discoveryservices
+(
+    id        text    not null primary key,
```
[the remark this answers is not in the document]
reinkrul: Lovely. We probably want to test the implementation with other RDMS's before releasing it into the wild.

discoveryservice/definition.go
```diff
@@ -0,0 +1,68 @@
...
+)
+
+//go:embed *.json
+var jsonSchemaFiles embed.FS
+var definitionJsonSchema *jsonschema.Schema
+
+func init() {
+	serviceDefinitionSchemaData, err := jsonSchemaFiles.ReadFile("service-definition-schema.json")
+	if err != nil {
+		panic(err)
```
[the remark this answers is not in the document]
reinkrul: If this fails, it means an embedded file is corrupt. So it will already fail in unit testing. No need to return errors here.

discoveryservice/module.go
reinkrul: I had "delete", but you don't really "delete" a presentation, since new a presentation just replaces the previous one.
"Revoke" is tightly associated with revoking credentials, a feature which we also have to implement (checking whether the presentation contains VC(s) that are revoked), so that becomes very confusing.

Prefer "delete" or "remove"?

discoveryservice/store.go
reinkrul: snake casing (becomes `presentation_records`), implemented `Tabler` to customize table names

vcr/credential/resolver.go
```diff
@@ -52,3 +55,48 @@ func ExtractTypes(credential vc.VerifiableCredential) []string {
...
+		if issuer == "" {
+			return nil, errors.New("JWT presentation does not have 'iss' claim")
+		}
+		return did.ParseDID(issuer)
+	default:
+		return nil, errors.New("unsupported presentation format")
+	}
+}
+
+func PresentationSigningKeyID(presentation vc.VerifiablePresentation) (*did.DIDURL, error) {
```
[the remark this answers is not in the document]
reinkrul: I think I saw that in a VP from MS Authenticator. Probably won't hurt here, as long as we scope it to the issuer?

reinkrul:
note to self: check that presentation.ID is scoped to the signer DID

reinkrul:
First feedback processed

discoveryservice/store.go
reinkrul: > Prefix wildcards are dangerous. Some DBs require this to be enabled. 
Source? And I don't expect there to be millions of rows, so should be OK? Although I'm also fine with disabling prefix wildcard search.
> You could also have a fancy search by inserting . wildcards between every character.

?

discoveryservice/module.go
reinkrul: Yes;

> - `exp` (expiration) of the presentation MUST NOT be after the expiration date of the credentials.

discoveryservice/module.go
reinkrul: It is;

> - it MUST specify `RetractedVerifiablePresentation` as type, in addition to the `VerifiablePresentation`.

discoveryservice/store.go
woutslakhorst: Actually using `%` instead of `.`. You type `zrg` as user. You expand it to `z%r%g%`. This will match `zorg`, `zerg` and `zaterdag` (and others). Benefit is that you can use a binary index on the first `z`. Depending on the letter this reduces search space 10 to 50 fold.

discoveryservice/module.go
woutslakhorst: or "unsubscribe", "deregister". All still not very catchy.... dunno 🤷

discoveryservice/module.go
woutslakhorst: or add minimum validity to the spec.
Once every minute won't trigger a rate limit but is still way too much

README.rst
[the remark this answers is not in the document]
woutslakhorst: What about a note: don't use in-memory unless for test/debug. Don't use SQLite as discovery server. Using SQLite in production is discouraged.

discoveryservice/store.go
reinkrul: Still don't understand the `.`?


I think we should disallow wildcards-in-the-middle, I believe it's not supported by at least some of them (MySQL). 
So just postfix (and maybe prefix)?

vcr/pe/presentation_definition.go
```diff
@@ -292,8 +292,9 @@ func matchConstraint(constraint *Constraints, credential vc.VerifiableCredential
 // matchField matches the field against the VC.
 // All fields need to match unless optional is set to true and no values are found for all the paths.
 func matchField(field Field, credential vc.VerifiableCredential) (bool, error) {
+	type Alias vc.VerifiableCredential
 	// jsonpath works on interfaces, so convert the VC to an interface
-	asJSON, _ := json.Marshal(credential)
+	asJSON, _ := json.Marshal(Alias(credential))
```
gerardsn, line 297: What does this do? Or why is it needed

storage/sql_migrations/2_discoveryservice.up.sql
```diff
@@ -0,0 +1,49 @@
...
+    constraint fk_discovery_presentation_service_id foreign key (service_id) references discovery_service (id) on delete cascade
+);
+
+-- discovery_credential is a credential in a presentation of the discovery service.
+-- We could do without the table, but having it allows to have a normalized index for credential properties that appear on every credential.
+-- Then we don't need rows in the properties table for them (having a column for those is faster than having a row in the properties table which needs to be joined).
+create table discovery_credential
+(
+    id                    varchar(36) not null primary key,
+    presentation_id       varchar(36) not null,
```
gerardsn, line 28: Easy to incorrectly relate to `presentation_id` in the `discovery_presentation` table? (note by Rein)

discovery/store.go
```diff
@@ -0,0 +1,379 @@
...
+
+	stmt := s.db.Model(&presentationRecord{}).
+		Where("service_id = ?", serviceID).
+		Joins("inner join discovery_credential cred ON cred.presentation_id = discovery_presentation.id")
+	numProps := 0
+	for jsonPath, value := range query {
+		if value == "*" {
+			continue
+		}
+		// sort out wildcard mode
```
gerardsn, line 264: todo?

discovery/store.go
```diff
@@ -0,0 +1,379 @@
...
+		if err != nil {
+			return nil, nil, fmt.Errorf("parse presentation '%s' of service '%s': %w", row.PresentationID, serviceID, err)
+		}
+		presentations = append(presentations, *presentation)
+		timestamp = Timestamp(row.LamportTimestamp)
+	}
+	return presentations, &timestamp, nil
+}
+
+func (s *sqlStore) search(serviceID string, query map[string]string) ([]vc.VerifiablePresentation, error) {
```
gerardsn, line 248: ~50 line method. I'd appreciate a godoc that summarizes behavior.

gerardsn (changes requested):
A PR like this takes me a full day to review to cover about 80% of the content. Adding more comment along the way or summarizing behavior of longer functions/methods (also private ones) would help me a lot.

README.rst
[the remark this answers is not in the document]
reinkrul: I'll add a note SQLite is discouraged in production environments.

discovery/store.go
reinkrul: no, short explanation of the code below. Will elaborate a bit on it.

discovery/module.go
```diff
@@ -0,0 +1,229 @@
...
+
+func (m *Module) addPresentation(definition Definition, presentation vc.VerifiablePresentation) error {
+	// VP can't be valid longer than the credentialRecord it contains
+	expiration := presentation.JWT().Expiration()
+	for _, cred := range presentation.VerifiableCredential {
+		exp := cred.JWT().Expiration()
+		if !exp.IsZero() && expiration.After(exp) {
+			return fmt.Errorf("presentation is valid longer than the credential(s) it contains")
+		}
+	}
```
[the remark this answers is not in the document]
reinkrul: I believe it's optional. But I should use `VerifiableCredential.ExpirationDate` instead

discovery/config.go
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *
+ */
+
+package discovery
+
+// Config holds the config of the module
+type Config struct {
```
woutslakhorst, line 22: Not a pflag yet.

discovery/config.go
```diff
@@ -0,0 +1,48 @@
...
+	Server      ServerConfig             `koanf:"server"`
```
woutslakhorst, line 23: usage of definitions and files should be explained in docs.

discovery/module.go
```diff
@@ -114,51 +120,64 @@ func (m *Module) Add(serviceID string, presentation vc.VerifiablePresentation) e
...
-		exp := cred.JWT().Expiration()
-		if !exp.IsZero() && expiration.After(exp) {
-			return fmt.Errorf("presentation is valid longer than the credentialRecord(s) it contains")
+		if cred.ExpirationDate != nil && expiration.After(*cred.ExpirationDate) {
+			return fmt.Errorf("presentation is valid longer than the credential(s) it contains")
 		}
 	}
-	// VP must fulfill the PEX Presentation Definition
+	// VP must fulfill the PEX Presentation ServiceDefinition
+	// TODO: Use Validate() once PR is merged
```
woutslakhorst, line 172: Do you want to wait for that?

vcr/pe/presentation_definition.go
```diff
@@ -292,9 +292,15 @@ func matchConstraint(constraint *Constraints, credential vc.VerifiableCredential
 // matchField matches the field against the VC.
 // All fields need to match unless optional is set to true and no values are found for all the paths.
 func matchField(field Field, credential vc.VerifiableCredential) (bool, error) {
-	type Alias vc.VerifiableCredential
 	// jsonpath works on interfaces, so convert the VC to an interface
-	asJSON, _ := json.Marshal(Alias(credential))
+	var asJSON []byte
+	if credential.Format() == vc.JWTCredentialProofFormat {
+		// json.Marshal on a JWT VC leads to the JWT string, not a map with the VC properties
+		type Alias vc.VerifiableCredential
```
woutslakhorst, line 299: Don't call it an Alias (it's a type definition)

vcr/pe/presentation_submission.go
```diff
@@ -115,43 +115,57 @@ func (b *PresentationSubmissionBuilder) Build(format string) (PresentationSubmis
 
 	// next we need to map the selected VCs to the correct wallet
 	// loop over all selected VCs and find the wallet that contains the VC
+	signInstructions := make([]SignInstruction, len(b.wallets))
```
woutslakhorst, line 118: 118-132 this was part of a different PR, rebase?

vcr/pe/schema/v2/schema.go
```diff
@@ -51,21 +51,27 @@ var PresentationDefinition *jsonschema.Schema
...
 		return nil, fmt.Errorf("refusing to load unknown schema: %s", url)
 	}
-	Compiler.Draft = jsonschema.Draft7
-	if err := loadSchemas(schemaFiles, Compiler); err != nil {
-		panic(err)
-	}
-	PresentationDefinition = Compiler.MustCompile(presentationDefinition)
-	PresentationSubmission = Compiler.MustCompile(presentationSubmission)
+	compiler := Compiler()
+	PresentationDefinition = compiler.MustCompile(presentationDefinition)
```
woutslakhorst, line 73: already done by `Compiler()`

woutslakhorst (changes requested):

discovery/module.go
reinkrul: Probably not. I'll make an issue.

reinkrul: We can't use it, since we don't have a submission. So we'll have to use match.

vcr/pe/presentation_definition.go
reinkrul: added comment, otherwise `json.Marshal` just returns the compacted JWT as JSON string

discovery/interface.go
```diff
@@ -0,0 +1,52 @@
...
+	"errors"
+	"github.com/nuts-foundation/go-did/vc"
+)
+
+// Timestamp is value that references a point in the list.
+// It is used by clients to request new entries since their last query.
+// It's implemented as lamport timestamp (https://en.wikipedia.org/wiki/Lamport_timestamp);
+// it is incremented when a new entry is added to the list.
+// Pass 0 to start at the beginning of the list.
+type Timestamp uint64
```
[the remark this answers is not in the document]
reinkrul: I wrote an implementation with strings instead, where it is prefixed with a random value generated by the server on first initialization of the particular service. Is an improvement, but makes this PR even larger. I'd prefer to get this one merged first.

discovery/store.go
```diff
@@ -0,0 +1,379 @@
...
+	if err != nil {
+		return err
+	}
+	if num > 0 {
+		log.Logger().Debugf("Pruned %d expired presentations", num)
+	}
+	return nil
+}
+
+func (s *sqlStore) removeExpired() (int, error) {
```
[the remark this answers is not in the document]
reinkrul: > Or will clients only be able to search for use cases that have a configured definition?

I'd say this. Only let registered services be retrieved.

discovery/store.go
```diff
@@ -0,0 +1,379 @@
...
+		// This is in contrast to most other SQL-databases, which let the 2nd thread wait for some time to acquire the lock.
+		// The general advice for SQLite is to retry the operation, which is just poor-man's scheduling.
+		// So to keep behavior consistent across databases, we'll just lock the entire store for the duration of the transaction.
+		// See https://github.com/nuts-foundation/nuts-node/pull/2589#discussion_r1399130608
+		s.writeLock.Lock()
+		defer s.writeLock.Unlock()
+	}
+	if err := s.prune(); err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
reinkrul: I thought I could spare the complexity, and as long as the column is indexed it should be really, really fast.

storage/sql_migrations/2_discoveryservice.up.sql
reinkrul: comment added

discovery/module.go
```diff
@@ -115,47 +120,67 @@ func (m *Module) Add(serviceID string, presentation vc.VerifiablePresentation) e
...
 	if int(expiration.Sub(time.Now()).Seconds()) > definition.PresentationMaxValidity {
 		return fmt.Errorf("presentation is valid for too long (max %s)", time.Duration(definition.PresentationMaxValidity)*time.Second)
 	}
-
+	// Check if the presentation already exists
+	credentialSubjectID, err := credential.PresentationSigner(presentation)
+	if err != nil {
+		return err
+	}
+	exists, err := m.store.exists(serviceID, credentialSubjectID.String(), presentation.ID.String())
```
gerardsn, line 139: mixed use of fields. above we use the `exp` from the JWT, here the ID field in the json object. why not the `jti`? Probably doesn't matter, but being consistent might save some issues in the future?

discovery/store.go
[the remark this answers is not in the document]
gerardsn: it depends how care orgs will be allowed to access discovery services. As long as care orgs can't use use-cases that aren't configured in a service definition I suppose there is no issue.

rationale: If a node only acts as client (applies to most) and all service definitions are gone after a restart (config error), the db will only contain outdated info. The client will also no longer publish its clients to the discovery services. This could result in weird situations where orgs can find orgs managed by other vendors but not their own, etc..

gerardsn (changes requested):
comments are suggestions. Requesting changes because `retracted_jti` is not DID

discovery/module.go
```diff
@@ -0,0 +1,229 @@
...
+		return fmt.Errorf("retraction presentation issuer is not a valid DID: %w", err)
+	}
+	retractJTI, err := did.ParseDIDURL(retractJTIString)
+	if err != nil {
+		return fmt.Errorf("retraction presentation 'retract_jti' claim is not a valid DID URL: %w", err)
+	}
+	if !signerDID.Equals(retractJTI.DID) {
+		return errors.New("retraction presentation 'retract_jti' claim does not match JWT issuer")
+	}
+	exists, err := m.store.exists(serviceID, signer, retractJTIString)
```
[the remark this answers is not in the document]
gerardsn: comment still stands, `retracted_jti` is not a DID

discovery/module.go
reinkrul: The VP spec does not specify expiration date, so there's no property for it. Only in the JSON-LD and JWT proofs.

I think properties from the VC model should be preferred (so presentation.ID over JWT jti, although they map to the same underlying JWT field), unless there of course isn't one. Therefore it's mixed.

reinkrul:
> it depends how care orgs will be allowed to access discovery services. As long as care orgs can't use use-cases that aren't configured in a service definition I suppose there is no issue. Rationale: If a node only acts as client (applies to most) and all service definitions are gone after a restart (config error), the db will only contain outdated info. The client will also no longer publish its clients to the discovery services. This could result in weird situations where orgs can find orgs managed by other vendors but not their own, etc..

If the service definition is wrong, which causes an error on startup. And clients should only be able to search loaded services (validation check in still-missing `Search()` function in `module.go`).
If it can't post presentations to the server, healthchecks/logging should show this error.

discovery/module.go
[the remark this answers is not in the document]
reinkrul: I don't understand the comment, the 3rd argument of `exists` is the presentation ID, thus `retracted_jti` (DID URL) is correct?

If the comment is about the `signerDID.Equals(retractJTI.DID)`, it compared the DID part of `retractJTI`, not the whole DID URL.

discovery/store.go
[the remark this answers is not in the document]
reinkrul: If the service definition is wrong, which causes an error on startup. And clients should only be able to search loaded services (validation check in still-missing Search() function in module.go).
If it can't post presentations to the server, healthchecks/logging should show this error.

gerardsn (approved):

Issue #2340, opened 2023-07-13
SQL storage support


reinkrul:
This issue is meant to identify various parts of supporting storage in SQL databases of what we currently store in KV-stores.

woutslakhorst:
Some of our dependencies already require (parts of) specific sql libs.

Any change in storage should not be seen separately from other functional/architectural changes. For example a web:did method could mean that no document history is required anymore and the removal of distributing them via the Nuts network also removes the requirement for consistency. So instead of storing a did document as json and using json search capabilities, a did document can just be stored by a bunch of tables and a json variant is the result of marshalling the relatinal data....

reinkrul:
> Some of our dependencies already require (parts of) specific sql libs.

And its fine to have _some_ SQL drivers in there, just not all of them (it's not feasible).

> Any change in storage should not be seen separately from other functional/architectural changes. For example a web:did method could mean that no document history is required anymore and the removal of distributing them via the Nuts network also removes the requirement for consistency. So instead of storing a did document as json and using json search capabilities, a did document can just be stored by a bunch of tables and a json variant is the result of marshalling the relatinal data....

As the analysis above points out: JSON search is not required for DID documents in the given the current featureset, only for searching VCs. So moving to did:web does not influence that requirement.

Could be good to further normalize the DID document storage model indeed, having services and maybe keys as separate tables.

reinkrul:
Minimum viable schema:
![sql-schema-v1](https://github.com/nuts-foundation/nuts-node/assets/1481228/3f149f87-6312-4068-a798-960de607da3e)
(source: https://github.com/nuts-foundation/nuts-node/blob/sql-support/docs/diagrams/sql-schema.puml)

woutslakhorst:
would:
- remove TXs
- add verificationMethods and services as tables
- add dateTimes on credential
- split VCR into wallet/issuer

reinkrul:
Idea from today:
- when using did:web, verifiers would still like to verify the cryptographic validness of a issuer's DID document. Especially if it's a high-profile issuer (e.g. CIBG/LrZA). This could extend did:web with a signature signed by a key from the DID document, like we do with did:nuts. This would, however lead to the requirement that the DID producer can serve all historic versions. Having to store multiple versions of the same DID document leads to duplication of its contents in the database, which has consequences when its contents (services, verification methods) are normalized:
   - data is duplicated for each version of the document, even if it was just a small change
[...]

PR #2595, opened 2023-11-09
SQL: introduce SQLite storage

reinkrul:
Part of https://github.com/nuts-foundation/nuts-node/issues/2340


storage/engine.go
```diff
@@ -129,6 +152,49 @@ func (e *engine) GetSessionDatabase() SessionDatabase {
...
+	e.sqlDB, err = gorm.Open(sqlite.Open(connectionString), &gorm.Config{})
+	if err != nil {
+		return err
+	}
+	log.Logger().Debug("Running database migrations...")
+	underlyingDB, err := e.sqlDB.DB()
+	if err != nil {
+		return err
+	}
+	sourceDriver, err := iofs.New(sqlMigrationsFS, "sql_migrations")
```
woutslakhorst, line 177: How will the correct migrations be selected in the future when there's only a connection string? eg: mysql vs postgresql

storage/interface.go
```diff
@@ -37,6 +38,8 @@ type Engine interface {
 	GetProvider(moduleName string) Provider
 	// GetSessionDatabase returns the SessionDatabase
 	GetSessionDatabase() SessionDatabase
+
+	SQLDatabase() *gorm.DB
```
woutslakhorst, line 42: godoc, other methods are prepended with `Get`. All `Get` or no `Get`?

storage/sql_migrations/README.md
```diff
@@ -0,0 +1,8 @@
+This directory contains SQL schema migrations, run at startup of the node.
+
+Refer to https://github.com/golang-migrate/migrate/blob/master/MIGRATIONS.md on how to write migrations.
+
+Files should be named according to the following: `<number>_<short engine name>_<table/feature name>.<up|down>.sql`.
+For instance: `2_usecase_list.up.sql`.
+
+Remember NOT to alter migrations that were included in released versions of the Nuts node: it might break vendor deployments or cause data corruption.
```
woutslakhorst, line 8: important for instances running master. It'll break.

storage/engine.go
reinkrul: probably connection string driver name to folder

storage/sql_migrations/README.md
reinkrul: Maybe alter to "avoid changing migrations in master, DO NOT alter migrations in a released version". E.g., to fix a broken migration (not just one that is revised)?

storage/interface.go
reinkrul: In Go you generally leave out "Get" for "GetXXX", but we should probably do that in one "go". Will change to "GetSQLDatabase()" for the sake of consistency

woutslakhorst (approved):

PR #2598, opened 2023-11-10
PEX: Support matching JWT VCs

reinkrul:


vcr/pe/presentation_definition.go
```diff
@@ -221,32 +222,41 @@ func (presentationDefinition PresentationDefinition) groups() []groupCandidates
...
+					if matchProofType(proofType, credential) {
+						return true
+					}
+				}
+			}
+		}
+	case vc.JWTCredentialProofFormat:
+		// Get signing algorithm used to sign the JWT
+		message, _ := jws.ParseString(credential.Raw()) // can't really fail, JWT has been parsed before.
+		signingAlgorithm, _ := message.Signatures()[0].ProtectedHeaders().Get("alg")
```
woutslakhorst, line 248: `jws.AlgorithmKey` for constant

vcr/pe/test/pd_jsonld_jwt.json
```diff
@@ -0,0 +1,99 @@
+{
+  "id": "Definition requesting NutsOrganizationCredential",
```
woutslakhorst, line 2: All examples online define multiple paths to handle the differente between jsonld and jwt. Would be best to add a test for that as well.

vcr/pe/presentation_definition_test.go
```diff
@@ -92,38 +44,109 @@ var testCredentialString = `
...
-	id4 := ssi.MustParseURI("4")
-	vc1 := vc.VerifiableCredential{ID: &id1}
-	vc2 := vc.VerifiableCredential{ID: &id2}
-	vc3 := vc.VerifiableCredential{ID: &id3}
-	vc4 := vc.VerifiableCredential{ID: &id4}
-	verifiableCredential := credential.ValidNutsOrganizationCredential(t)
+	// Create a JSON-LD VC
+	jsonldVC := credential.ValidNutsOrganizationCredential(t)
+
+	// Create a JWT VC
```
woutslakhorst, line 82: JWT vc creation could be placed next to jsonld vc creation in the credential package.

vcr/pe/presentation_definition_test.go
```diff
@@ -270,51 +307,87 @@ func TestMatch(t *testing.T) {
...
 
 		assert.True(t, match)
 	})
 
-	t.Run("format with jwt_vc always returns false", func(t *testing.T) {
-		asMap := map[string]map[string][]string{"jwt_vc": {"alg": {"ES256K", "ES384"}}}
-		asFormat := PresentationDefinitionClaimFormatDesignations(asMap)
-		match := matchFormat(&asFormat, vc.VerifiableCredential{})
+	t.Run("JWT", func(t *testing.T) {
+		const jwtCredential = `eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCIsImtpZCI6ImRpZDpleGFtcGxlOmFiZmUxM2Y3MTIxMjA0
```
woutslakhorst, line 323: You might want to prevent people from using this example at other points because it's no longer valid (expired). Or generate a valid credential for each test run?

woutslakhorst (changes requested):

gerardsn (approved):
no additional comments

vcr/pe/test/pd_jsonld_jwt.json
reinkrul: Couldn't find them, and pick sounded weird to me (but works), will do

vcr/pe/presentation_definition_test.go
reinkrul: I used it because it came from the VC data model, so we at least have some tests that use data from the specification. I'll add a comment that it's expired and where it came from.

woutslakhorst (approved):

PR #2607, opened 2023-11-15
Policy backend API spec

### LINKED ISSUE #2606: Design scope to presentation definition and request authorization API interface
In both the service-to-service access token request flow and OpenID4VP authorization request flow, a request with scopes needs to be mapped to the correct Presentation Definition. The responsibility for this lies outside the Nuts node. Policy Administration/Decision Points should make the decision. To aid vendor build these 'points' a standardised API contract could help.

The API should contain 2 calls:
- GET /presentation_definition?scope=X Y Z&authorizer=did:web:example.com
- POST /authorized

The did path parameter is used for tenant selection. It uses standard URLEncoding. The first API call is rerouted from a public call and must therefore not give different answers in the case of patient specific scopes.

The POST call should at least contain:
- the request URL
- most fields from the access token introspections (VPs, presentation_submission, scope, client_id, mapped input descriptor fields)
- 200 JSON return with `access` boolean. 403 is reserved for getting no access to the authorized endpoint.

woutslakhorst:
closes #2606

The specification is required to generate a client for the Nuts node and for policy backends to generate the server API.

- [x] Client code generation


docs/_static/policy/v1.yaml
```diff
@@ -0,0 +1,133 @@
...
+        It might be the case that the caller mapped credential fields to additional params.
+      type: object
+      required:
+        - client_id
+        - scope
+        - request_url
+        - request_method
+        - presentation_submission
+        - vps
+      properties:
```
gerardsn, line 96: missing the `input_desciptor_constraint_id_map` from the introspection endpoint? Could probably use a simpler name here.

docs/_static/policy/v1.yaml
```diff
@@ -0,0 +1,133 @@
...
+        The request contains all params involved with the request.
+        It might be the case that the caller mapped credential fields to additional params.
+      type: object
+      required:
+        - client_id
+        - scope
+        - request_url
+        - request_method
+        - presentation_submission
+        - vps
```
gerardsn, line 95: These are optional? Most backends probably can't handle PEX/VC/VP

gerardsn (changes requested):
Also see comments in issue.

---
Will user information also be passed to the backend? If so, what field parameter do they belong in?

docs/_static/policy/v1.yaml
```diff
@@ -0,0 +1,133 @@
...
+        It could be the case that the presentation definition is not found.
+        In that case the response will be 201 with an empty body.
+      operationId: "presentationDefinition"
+      tags:
+        - policy
+      responses:
+        "200":
+          description: |
+            DID has been found and the scope is supported.
+            If the scope is supported but no presentation definition is required, the response will be 200 with a presentation definition without any input descriptors.
```
reinkrul, line 39: what would the use case for this be?

docs/_static/policy/v1.yaml
```diff
@@ -0,0 +1,133 @@
...
+          description: The DID is known but the presented scope is not supported.
+        "404":
+          description: DID is not known to the policy backend.
+  /{did}/authorize:
+    parameters:
+      - name: did
+        in: path
+        description: URLEncoded DID. The DID is used for tenant selection.
+        required: true
+        example: did:web:example.com:1
```
reinkrul, line 54: invalid DID (`:1`)

docs/_static/policy/v1.yaml
```diff
@@ -0,0 +1,133 @@
...
+servers:
+  - url: "http://localhost:1323"
+paths:
+  /{did}/presentation_definition:
+    parameters:
+      - name: did
+        in: path
+        description: URLEncoded DID.
+        required: true
+        example: did:web:example.com:1
```
reinkrul, line 14:
```suggestion
        example: did:web:example.com
```

docs/_static/policy/v1.yaml
```diff
@@ -0,0 +1,133 @@
...
+        schema:
+          type: string
+    post:
+      summary: Authorize a resource request.
+      description: |
+        When an access token is used to request a resource, the resource server needs to know if the access token grants access to the requested resource.
+        The resource server will send a request to the policy backend to check if the access token grants access to the requested resource.
+
```
reinkrul, line 62: maybe add that the policy backend does not need to check the presentation submission itself? Since that's fairly complex, it should be done by the Nuts node.

docs/_static/policy/v1.yaml
woutslakhorst: public service, like a directory service

docs/_static/policy/v1.yaml
```diff
@@ -0,0 +1,133 @@
+openapi: 3.1.0
```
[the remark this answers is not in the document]
woutslakhorst: no

docs/_static/policy/v1.yaml
woutslakhorst: Then the entire id to value mapping would have to be standardised. I think it's too early for a decision on that. With the VPs included, a backend can make a decision.

docs/_static/policy/v1.yaml
woutslakhorst: The first versions might just have to, we'll have to experiment a bit to see what's useful.

policy/api/v1/client/client_test.go
```diff
@@ -175,10 +175,10 @@ func TestHTTPClient_Authorized(t *testing.T) {
 		handler := http2.Handler{StatusCode: http.StatusOK}
 		_, client := testServerAndClient(t, &handler)
 
-		response, err := client.Authorized(ctx, "http://::1:1", request)
+		response, err := client.Authorized(ctx, "http://test.test", request)
```
reinkrul, line 178: bad test, uses external resources and assumes test.test won't be available?

Better use the test server and an invalid (local) port?

woutslakhorst: from wiki

> .test is a reserved top-level domain intended for usage in software testing. It is guaranteed to never be registered into the Internet.

an ipv6 address won't work on our CI. A `.local` takes a second to (not) resolve.

policy/api/v1/client/client.go
```diff
@@ -0,0 +1,115 @@
...
+	var data []byte
+
+	if data, err = io.ReadAll(response.Body); err != nil {
+		return nil, fmt.Errorf("unable to read response: %w", err)
+	}
+	if err = json.Unmarshal(data, &presentationDefinition); err != nil {
+		return nil, fmt.Errorf("unable to unmarshal response: %w", err)
+	}
+
+	return &presentationDefinition, nil
```
gerardsn, line 82: just use generated `ParsePresentationDefinitionResponse()`?

policy/api/v1/client/client.go
```diff
@@ -0,0 +1,115 @@
...
+	var data []byte
+
+	if data, err = io.ReadAll(response.Body); err != nil {
+		return false, fmt.Errorf("unable to read response: %w", err)
+	}
+	if err = json.Unmarshal(data, &authorized); err != nil {
+		return false, fmt.Errorf("unable to unmarshal response: %w", err)
+	}
+
+	return authorized, nil
```
gerardsn, line 114: replace with generated `ParseCheckAuthorizedResponse()`

gerardsn (changes requested):

docs/_static/policy/v1.yaml
woutslakhorst: That'll be an incremental change. We can then also process feedback in the same change.

gerardsn (approved):

PR #2609, opened 2023-11-17
PEX: Changed path_nested to be object instead of slice

reinkrul:
Simplified logic a bit; `1 InputDescriptorMapping = 1 Credential` (instead of `1..n` credentials), nesting only if there's multiple VPs.

### DIFF
--- a/vcr/pe/presentation_submission.go
+++ b/vcr/pe/presentation_submission.go
@@ -121,7 +121,7 @@ func (b *PresentationSubmissionBuilder) Build(format string) (PresentationSubmis
 		for i, walletVCs := range b.wallets {
 			for _, walletVC := range walletVCs {
 				// do a JSON equality check
-				if vcEqual(selectedVCs[j], walletVC) {
+				if selectedVCs[j].Raw() == walletVC.Raw() {
 					signInstructions[i].Holder = b.holders[i]
 					signInstructions[i].VerifiableCredentials = append(signInstructions[i].VerifiableCredentials, selectedVCs[j])
 					// remap the path to the correct wallet index
--- a/vcr/pe/presentation_submission.go
+++ b/vcr/pe/presentation_submission.go
@@ -143,7 +143,7 @@ func (b *PresentationSubmissionBuilder) Build(format string) (PresentationSubmis
 		}
 	}
 
-	presentationIndex := 0
+	index := 0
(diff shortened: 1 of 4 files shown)


vcr/pe/presentation_submission.go
```diff
@@ -115,43 +115,69 @@ func (b *PresentationSubmissionBuilder) Build(format string) (PresentationSubmis
...
-					inputDescriptorMappingObjectForVC.Path = fmt.Sprintf("$.verifiableCredential[%d]", index)
-					signInstructions[i].inputDescriptorMappingObjects = append(signInstructions[i].inputDescriptorMappingObjects, inputDescriptorMappingObjectForVC)
-					index++
+	// there will be at most len(b.wallets) sign instructions
+	var signInstructions []SignInstruction
+	for _, holderID := range b.holders {
+		var signInstruction *SignInstruction
+		// find VCs that need to be in this sign instruction/verifiable presentation
+		for i, selectedVC := range selectedVCs {
+			if b.isHolder(holderID, selectedVC) {
```
woutslakhorst, line 124: This changes logic. Before any selected VC from a wallet would be added regardless of the holder. Currently we said that all VCs need to have the same subject/holder, but if that changes this selection method won't work. But if `isHolder` is changed back to just checking the Raw()/Equal then it would be possible again. Summary: the holderID == subject check does not prevent false positives but does block future combination of VCs.

vcr/pe/presentation_submission.go
```diff
@@ -115,43 +115,69 @@ func (b *PresentationSubmissionBuilder) Build(format string) (PresentationSubmis
...
 		}
 	}
 
 	return presentationSubmission, signInstructions, nil
 }
+
+// isHolder returns true if the wallet of the specified subject contains the given VC.
+func (b *PresentationSubmissionBuilder) isHolder(subjectID did.DID, credential vc.VerifiableCredential) bool {
+	for i, holder := range b.holders {
+		if holder == subjectID {
```
woutslakhorst, line 173: not needed, just checking Raw is sufficient.

woutslakhorst (changes requested):

vcr/pe/presentation_submission.go
reinkrul: reverted and fixed bugs in original code

woutslakhorst (approved):

PR #2613, opened 2023-11-17
ignore

gerardsn:
testing if https://github.com/nuts-foundation/nuts-node/security/code-scanning/44 van also be dismissed for 32-bit systems.


gerardsn:
alert is valid for 32-bit systems, but still dismissed

PR #2616, opened 2023-11-22
PEX: Allow VP format per submission wallet

reinkrul:
Each wallet potentially maps to a sign instruction, which each could have their own VP format.

Found this requirement when validating submissions with a list of VPs (requires rebuilding the presentation submission).

### DIFF
--- a/auth/api/iam/openid4vp.go
+++ b/auth/api/iam/openid4vp.go
@@ -138,8 +138,8 @@ func (r *Wrapper) handlePresentationRequest(params map[string]string, session *S
 	}
 
 	submissionBuilder := presentationDefinition.PresentationSubmissionBuilder()
-	submissionBuilder.AddWallet(session.OwnDID, ownCredentials)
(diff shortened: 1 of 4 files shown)


woutslakhorst:
The format is passed to fill a field in the mapping object. It should be determined by the VP format. This is defined by the client_metadata (verifier metadata) and must match the actual signing.

Does this mean that setting the format is actually some sort of post processing step? Or do we need to align client metadata with wallet capabilities before calling `AddWallet`?

reinkrul:
> The format is passed to fill a field in the mapping object. It should be determined by the VP format. This is defined by the client_metadata (verifier metadata) and must match the actual signing.
> 
> Does this mean that setting the format is actually some sort of post processing step? Or do we need to align client metadata with wallet capabilities before calling `AddWallet`?

No, it is just use as instruction how to create the VP (jsonld/jwt). So the presentation signing function should take it into account.

But it doesn't solve the problem I stated; `Resolve()` and `Validate()` need to work on the PEX envelope (`interface{}`) instead

PR #2619, opened 2023-11-24
Add access token request call for the OpenID4VP user flow

### LINKED ISSUE #2604: Create the internal access token request API on the Nuts node
The internal access token request is the request a xis/ecd makes to the Nuts node to get an access token. The endpoint should:

- be internally focused and secured (JWT, firewall)
- ~return active access tokens that match an earlier request (access token management)~
- return a redirect link for the user-agent to start the OAuth flow for getting a new access token
- ~support requesting multiple tokens at once~
- do some (client) state management which will probably be required to link multiple steps from the OAuth flow together
- accept userID, scope, redirect_url and target DID~s~ as arguments (maybe more)

It will (probably) be a POST call with 200, 302 and 400 returns: 200 with access token, 302 if a new OAuth flow needs to be started and 400 on failure.

woutslakhorst:
closes #2604 

it extends the access token request with a user ID. If a user ID is given, the user flow is started. The API then returns a 302 with the location for the user. The user is redirected to `baseURL/iam/did/user`. If an active redirect token is present, the rest of the flow is started by creating a redirectURL to the remote authorization server.

There are 3 different error states;
- unauthenticated, the flow can forward to a login page
- technical error (inc missing token), forward to ???
- functional error return from verifier, invalid_request, display reason?

The last two would require i18n. #2621 is created for the user facing design which includes error codes.

- [x] some untested code
- [x] user facing error returns


auth/api/iam/api.go
```diff
@@ -343,8 +345,32 @@ func (r Wrapper) PresentationDefinition(_ context.Context, request PresentationD
...
+	if request.Body == nil {
+		// why did oapi-codegen generate a pointer for the body??
+		return nil, core.InvalidInputError("missing request body")
+	}
+	// resolve wallet
+	requestHolder, err := did.ParseDID(request.Did)
+	if err != nil {
+		return nil, core.NotFoundError("did not found: %w", err)
+	}
+	isWallet, err := r.vdr.IsOwner(ctx, *requestHolder)
```
reinkrul, line 358: use `idToOwnedDID` (or something)

auth/api/iam/session.go
```diff
@@ -33,17 +34,20 @@ type Session struct {
...
-	return u
+type UserSession struct {
+	ClientState string
+	SessionID   string
+	UserID      string
+	OwnDID      did.DID
 }
 
-func (s Session) CreateRedirectURI(params map[string]string) string {
+type RedirectSession struct {
```
reinkrul, line 44: godoc

reinkrul (changes requested):
first part

auth/api/iam/user.go
```diff
@@ -0,0 +1,122 @@
...
+		"token": token,
+	})
+	return RequestAccessToken302Response{
+		Headers: RequestAccessToken302ResponseHeaders{
+			Location: redirectURL.String(),
+		},
+	}, nil
+}
+
+// handleUserLanding is the handler for the landing page of the user.
```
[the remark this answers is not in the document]
woutslakhorst: yes it should switch when more different types are added (overview, addDevice, listDevices, whatever). But currently that's not needed yet.

auth/api/iam/user.go
```diff
@@ -0,0 +1,122 @@
...
+		return echoCtx.String(http.StatusForbidden, "")
+	}
+
+	// extract request from store
+	store := r.storageEngine.GetSessionDatabase().GetStore(time.Second, "user", "redirect")
+	redirectSession := RedirectSession{}
+	err := store.Get(token, &redirectSession)
+	if err != nil {
+		log.Logger().Debug("token not found in store")
+		return echoCtx.String(http.StatusForbidden, "")
```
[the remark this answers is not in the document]
woutslakhorst: See PR description: #2621

auth/services/oauth/relying_party.go
```diff
@@ -110,6 +111,31 @@ func (s *relyingParty) CreateJwtGrant(ctx context.Context, request services.Crea
 	return &services.JwtBearerTokenResult{BearerToken: signingString, AuthorizationServerEndpoint: endpointURL}, nil
 }
 
+func (s *relyingParty) AuthorizationRequest(ctx context.Context, requestHolder did.DID, verifier did.DID, scopes string, clientState string) (*url.URL, error) {
```
reinkrul, line 114: no state/session needed?

reinkrul (changes requested):

docs/_static/auth/iam.yaml
```diff
@@ -302,13 +302,31 @@ paths:
...
+                  description: | 
+                    The ID of the user for which this access token is requested.
+                    It's handled as opaque ID and is scoped to the requester DID.
+                redirectURL:
+                  type: string
+                  description: |
+                    The URL to which the user-agent will be redirected after the authorization request.
       responses:
         '200':
           description: Successful request. Responds with an access token as described in rfc6749 section 5.1.
```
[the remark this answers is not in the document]
woutslakhorst: it does both as described in the description

auth/services/oauth/relying_party.go
woutslakhorst: clientState is passed, tokens are (currently) stored on API layer

reinkrul (approved):
Let's see in it action

PR #2624, opened 2023-11-28
PEX: ldp_vp submission with 1 VC should not index credential property

reinkrul:
go-did marshals `verifiableCredential` to a JSON object instead of an array in that case.

### DIFF
--- a/vcr/pe/presentation_submission_test.go
+++ b/vcr/pe/presentation_submission_test.go
@@ -53,7 +53,38 @@ func TestPresentationSubmissionBuilder_Build(t *testing.T) {
 	vc2 := credentialToJSONLD(vc.VerifiableCredential{ID: &id2})
 	vc3 := credentialToJSONLD(vc.VerifiableCredential{ID: &id3})
 
-	t.Run("1 presentation", func(t *testing.T) {
+	t.Run("1 presentation with 1 credential", func(t *testing.T) {
+		expectedJSON := `
+		{
+		 "id": "for-test",
+		 "definition_id": "",
+		 "descriptor_map": [
+		   {
+		     "format": "ldp_vc",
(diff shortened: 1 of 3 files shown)


gerardsn:
The VC v1.1 spec is not clear if this can be an object, but all examples are an array. The v2 spec states that it MUST be an array if the field is present. `VerifiablePresentation.VerifiableCredential` is also always an array in go-did. Am I misinterpreting what you want to fix here?

gerardsn:
ahh... I think the `Unplural` in go-did for VCs in a VP is the bug. Have we used VPs in production yet? If not we can probably just fix that

gerardsn (commented):
is this still relevant given https://github.com/nuts-foundation/go-did/pull/98?

woutslakhorst:
> is this still relevant given [nuts-foundation/go-did#98](https://github.com/nuts-foundation/go-did/pull/98)?

PR should update to latest go-did and make changes if needed.

vcr/pe/presentation_submission_test.go
```diff
@@ -53,7 +53,38 @@ func TestPresentationSubmissionBuilder_Build(t *testing.T) {
...
+
+		require.NoError(t, err)
+		require.NotNil(t, signInstructions)
+		assert.Len(t, signInstructions, 1)
+		require.Len(t, submission.DescriptorMap, 1)
+		assert.Equal(t, "$.verifiableCredential", submission.DescriptorMap[0].Path)
+
+		submission.Id = "for-test" // easier assertion
+		actualJSON, _ := json.MarshalIndent(submission, "", "  ")
+		println(string(actualJSON))
```
woutslakhorst, line 84: leftover println

woutslakhorst (approved):

PR #2630, opened 2023-11-29
add VDR v2 spec, generated code and empty wrapper

### LINKED ISSUE #2629: Create API spec for VDRv2
see #2347 

`/did`: create/delete DID (pass type did:nuts?). For did:web it can use a GUID or any other random string. In future we could allow a named DID instead of a random identifier.
`/did/{did}`: delete/resolve DID
`/did/{did}/service` create service
`/did/{did}/service/{id}` update/delete service
`/did/{did}/verificationmethod` create key (verification method), by default all relationship types
`/did/{did}/verificationmethod/{id}` delete key

Some clarifications:
- the random identifier in the ID was actually the pubkey fingerprint. For did:web this is not needed. An URL friendly random identifier would suffice. UUIDv4 for example
- the relationship types for verificationmethods: we do not need `capabilityInvocation` and `capabilityDelegation`, but since it's no effort to add them, we just add them.

woutslakhorst:
closes #2629 

I described that removing a service does not have any influence on services referencing that service. Other than that everything is just basic CRUD.

One of those PRs where it takes more time to review than to make ;)


docs/_static/vdr/v2.yaml
```diff
@@ -0,0 +1,268 @@
...
+      responses:
+        "200":
+          description: "New DID has been created successfully. Returns the DID document."
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/DIDDocument'
+        default:
+          $ref: '../common/error_response.yaml'
+  /internal/vdr/v2/did/{did}:
```
reinkrul, line 34: this only resolves did:web DIDs owned by the local node? I think we already have `/iam/{did}` which essentially does the same?

reinkrul:
Companion PR: https://github.com/nuts-foundation/nuts-node/pull/2631

docs/_static/vdr/v2.yaml
woutslakhorst: We could limit it to own docs and not do a resolve but a simple GET?

woutslakhorst:
Companion PR can be done in parallel. When implementing the Create on the v2 api, the method specifics on the v1 api can be removed.

e2e-tests/oauth-flow/rfc021/docker-compose.yml
```diff
@@ -1,7 +1,7 @@
 version: "3.7"
 services:
   nodeA-backend:
-    image: "${IMAGE_NODE_A:-nutsfoundation/nuts-node:master}"
+    image: "${IMAGE_NODE_A:-nutsfoundation/nuts-node:latest}"
```
reinkrul, line 4:
```suggestion
    image: "${IMAGE_NODE_A:-nutsfoundation/nuts-node:master}"
```

reinkrul (approved):
almost, small suggestions

reinkrul:
Add to OpenAPI web epxlorer?

PR #2633, opened 2023-11-30
Root URL server config property to replace auth.publicurl

reinkrul:
Multiple modules now require it, time to move it to root config.


woutslakhorst (changes requested):
`AuthenticationServices` has a `PublicURL` method. It must be deprecated (or made private) and all callers can use the ServerConfig. I bet it would also remove dependencies on auth at some places.

docs/pages/deployment/configuration.rst
```diff
@@ -79,7 +79,7 @@ See :ref:`getting started <configure-node>` on how to set this up correctly.
 
 The incorporated `IRMA server <https://irma.app/docs/irma-server/#production-mode>`_ is automatically changed to production mode.
 In fact, running in strict mode is the only way to enable IRMA's production mode.
-In addition, it requires ``auth.irma.schememanager=pbdf`` and the ``auth.publicurl`` where the IRMA client can reach the server must be set.
+In addition, it requires ``auth.irma.schememanager=pbdf`` and the ``url`` where the IRMA client can reach the server must be set.
```
gerardsn, line 82: no longer relevant since the `url` is not optional anymore
```suggestion
In addition, it requires ``auth.irma.schememanager=pbdf``.
```

core/server_config.go
```diff
@@ -278,12 +290,14 @@ func FlagSet() *pflag.FlagSet {
 		"Required when 'network.enabletls' is 'true'.")
 	flagSet.String("network.truststorefile", "", "Deprecated: use 'tls.truststorefile'. PEM file containing the trusted CA certificates for authenticating remote gRPC servers.")
 	flagSet.Int("network.maxcrlvaliditydays", 0, "Deprecated: use 'tls.crl.maxvaliditydays'. The number of days a CRL can be outdated, after that it will hard-fail.")
+	flagSet.String("auth.publicurl", "", "Public URL which can be reached by a users IRMA client, this should include the scheme and domain: https://example.com. Additional paths should only be added if some sort of url-rewriting is done in a reverse-proxy.")
```
gerardsn, line 293: This is one of the key changes in the next (major) release. I think we should no longer accept `auth.publicurl`. If users do not understand this then they should probably go back to reading the release notes.

core/server_config.go
```diff
@@ -261,6 +272,7 @@ func FlagSet() *pflag.FlagSet {
 	flagSet.Bool("strictmode", true, "When set, insecure settings are forbidden.")
 	flagSet.Bool("internalratelimiter", true, "When set, expensive internal calls are rate-limited to protect the network. Always enabled in strict mode.")
 	flagSet.String("datadir", "./data", "Directory where the node stores its files.")
+	flagSet.String("url", "", "Public facing URL of the server (required). Must be HTTPS when strictmode is set. Superseeds 'auth.publicurl', which is deprecated.")
```
gerardsn, line 275: IMO deprecation messages should only be on the deprecated field so they can easily be removed later on? Feel free to ignore

gerardsn (changes requested):

woutslakhorst (approved):

gerardsn (approved):

PR #2634, opened 2023-12-01
Update to go-did v0.11

reinkrul:
Changes:
- `VerifiableCredential.IssuanceDate` is now a required `*time.Time`. I had to add them where it wasn't present, and change it to a pointer where it was present.

Earlier, it included these changes, which we decided to revert since it leads to too many changes in the Nuts node:
- The REST API and `go-leia` credential store assume JSON-LD arrays are compacted. Changing this, although still valid JSON-LD and semantically the same credential, would break expected behavior of API users. So I introduced a `CompactingVerifiableCredential` and `CompactingVerifiablePresentation` which essentially do the same as `go-did` earlier did for JSON-LD types; compacting arrays on JSON marshalling. This is the majority of the change. This can be removed when the VCR v2 API and `go-leia` credential store is replaced with something else.
- `SignedDocument.UnmarshalProofValue` assumed there was only 1 proof which is never the case any more, since it's now always marshalled into an array (even if it has just one value). I extended the function to not burden the caller with checking the number of proofs, since I don't expect there multiple proofs at any time (although still supported).
- PEX paths now always index array properties (mostly `verifiableCredential[0]`).


vcr/api/vcr/v2/types.go
```diff
@@ -21,37 +21,37 @@ package v2
...
 // CredentialSubject is an alias to use from within the API
 type CredentialSubject = interface{}
 
 // Revocation is an alias to use from within the API
 type Revocation = credential.Revocation
 
-// VerifiablePresentation is an alias to use from within the API
-type VerifiablePresentation = vc.VerifiablePresentation
+var _ json.Unmarshaler = &types.CompactingVerifiablePresentation{}
+var _ json.Marshaler = types.CompactingVerifiablePresentation{}
```
gerardsn, line 38: these should be in the file declaring `CompactingVerifiableCredential` and `CompactingVerifiablePresentation`

vcr/issuer/issuer.go
```diff
@@ -215,8 +215,9 @@ func (i issuer) buildVC(ctx context.Context, template vc.VerifiableCredential, o
 		ExpirationDate:    template.ExpirationDate,
 		IssuanceDate:      template.IssuanceDate,
 	}
-	if unsignedCredential.IssuanceDate.IsZero() {
-		unsignedCredential.IssuanceDate = TimeFunc()
+	if unsignedCredential.IssuanceDate == nil || unsignedCredential.IssuanceDate.IsZero() {
```
gerardsn, line 218: the `IsZero` was needed to check that no value was provided. If this true now, the zero value was provided by the user and it should not be changed but return an error? (from the validator later on)

vcr/issuer/leia_store.go
```diff
@@ -89,7 +89,7 @@ func NewLeiaIssuerStore(dbPath string, backupStore stoabs.KVStore) (Store, error
 }
 
 func (s leiaIssuerStore) StoreCredential(vc vc.VerifiableCredential) error {
-	vcAsBytes, _ := json.Marshal(vc)
+	vcAsBytes, _ := json.Marshal(types.CompactingVerifiableCredential(vc))
```
gerardsn, line 92: add comment on why it is needed. (To prevent copy pasta when migrating to a new DB where I assume it is no longer needed?)

vcr/issuer/network_publisher.go
```diff
@@ -86,7 +86,7 @@ func (p networkPublisher) PublishCredential(ctx context.Context, verifiableCrede
 
 	payload, _ := json.Marshal(verifiableCredential)
 	tx := network.TransactionTemplate(types.VcDocumentType, payload, key).
-		WithTimestamp(verifiableCredential.IssuanceDate).
+		WithTimestamp(*verifiableCredential.IssuanceDate).
```
gerardsn, line 89: feels like a panic waiting to happen (although not possible in current flows)

gerardsn (changes requested):
I've added a commit to fix a VC validator issue

woutslakhorst (commented):
It seems this only happens for the JSON stores, we can just change them to JSONLD stores and the problem is solved.

- change all JSON collection types to JSONLD collection types
- change to the correct index and search queries
- rename the index, then the index will be rebuild from the kv kackup

vcr/types/jsonld.go
```diff
@@ -0,0 +1,95 @@
...
+import (
+	"encoding/json"
+	"github.com/nuts-foundation/go-did/vc"
+)
+
+// CompactingVerifiableCredential is a type to use from within the API, which compacts JSON-LD arrays.
+// It retains backwards compatibility for JSON-LD array compacting, which was removed in the go-did v0.10.0 upgrade.
+// This essentially retains the behavior that was removed in https://github.com/nuts-foundation/go-did/pull/98
+// It only applies to JSON-LD; in case of JWT, the credential is returned as-is.
+type CompactingVerifiableCredential vc.VerifiableCredential
```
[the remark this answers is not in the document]
reinkrul: It's also needed by the go-leia store, that's why I put it here

vcr/issuer/issuer.go
reinkrul: True. Would be a functionally invalid value, but user-provided anyway (or due to another bug, but then this "fix" obscures that bug), will revert.

vcr/issuer/leia_store.go
reinkrul: Added comment. Your assumption is right.

vcr/issuer/network_publisher.go
reinkrul: True, although this is the network publisher, which ideally is removed at some foreseeable point in future?

vcr/api/vcr/v2/types.go
reinkrul: Left them there for now until we resolve the go-leia discussion

woutslakhorst:
The need for this change started with simplifying the PEX submission validation. In order to do it without a lot of code, we could change the go-did library so it always marshals to arrays instead of objects when n=1. But this had some more consequences which are now being fixed by this PR. So in short: simplification lead to lib change lead to monkey patch...

Maybe it's better to implement the PEX stuff using the complex approach?

gerardsn (changes requested):
over complete review, but requesting changes mainly on last comment

didman/didman.go
```diff
@@ -599,6 +599,11 @@ func (d *didman) updateService(ctx context.Context, id did.DID, serviceType stri
 
 func generateIDForService(id did.DID, service did.Service) ssi.URI {
 	bytes, _ := json.Marshal(service)
+	// go-did earlier unmarshaled/marshaled the service endpoint to a map[string]interface{} ("NormalizeDocument()"), which changes the order of the keys.
```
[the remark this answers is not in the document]
reinkrul: I didn't revert removal of compaction for service marshalling, so this still applies

vcr/verifier/verifier_test.go
```diff
@@ -443,7 +444,7 @@ func Test_verifier_validateAtTime(t *testing.T) {
 		t.Run("credential is valid", func(t *testing.T) {
 			sut := verifier{}
 			credentialToTest := testCredential(t)
-			valid := sut.validateAtTime(credentialToTest.IssuanceDate, credentialToTest.ExpirationDate, timeToCheck)
+			valid := sut.validateAtTime(*credentialToTest.IssuanceDate, credentialToTest.ExpirationDate, timeToCheck)
```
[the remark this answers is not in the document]
reinkrul: although this test uses a credential, validateAtTime is used to verify validity of the LDProof, not of the VC. I moved it to ProofOptions.

gerardsn (approved):

PR #2638, opened 2023-12-01
Postgres support

reinkrul:
At least have an e2e test to make sure the DDL statements are compatible. To be extended with tests for `did:web`.

We probably don't want unit tests for postgres, since spinning up Docker containers during `go test` will make stuff even slower than it already is.


woutslakhorst:
What do you think about the number of dependencies imported by dbmigrate? An alternative like https://github.com/pressly/goose does not have this problem....

woutslakhorst:
or do we want #2648 (also adds mysql support)

reinkrul:
We probably want to rework this to have it rerun a set of tests (probably RFC021 test) on different databases

PR #2641, opened 2023-12-01
allow local issuance for did:web

### LINKED ISSUE #2640: support self issuance for wallets
Some use cases require an self-issued credential or one that is managed by the same node. In that case there's no need to issue over the web.

- only for did:web
- part of wallet

woutslakhorst:
closes #2640 

this allows an issuer to issue to a wallet within the same node. This only happend when the VC is not public, not published, the receiver is a did:web and when it's owned by the node.

### DIFF
--- a/vcr/issuer/issuer.go
+++ b/vcr/issuer/issuer.go
@@ -23,7 +23,10 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"github.com/nuts-foundation/nuts-node/vcr/holder"
 	"github.com/nuts-foundation/nuts-node/vcr/openid4vci"
+	"github.com/nuts-foundation/nuts-node/vdr/didweb"
+	"github.com/nuts-foundation/nuts-node/vdr/management"
 	"github.com/nuts-foundation/nuts-node/vdr/resolver"
 	"time"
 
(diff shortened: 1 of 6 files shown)


vcr/issuer/issuer.go
```diff
@@ -155,9 +163,47 @@ func (i issuer) Issue(ctx context.Context, template vc.VerifiableCredential, opt
...
 	}
+	// local to local wallet for did:web
+	i.tryLocalIssuance(ctx, options, createdVC)
+
 	return createdVC, nil
 }
 
+func (i issuer) tryLocalIssuance(ctx context.Context, options CredentialOptions, createdVC *vc.VerifiableCredential) {
+	if options.Publish {
+		// not supported, return silently
```
reinkrul, line 174: but JWTs can't be loaded into the local wallet this way?

reinkrul (changes requested):
I don't like this monkey patch too much. The wallet owner now hasn't anything to decide whether it actually wants the credential. And it's a bit of magic with obscure limitations (no JWT).

Wouldn't it be better to provide a REST API to load credentials into the wallet? Then the existing issue API (no publish) can be used to create the VC, then the new API (something like `POST internal/vcr/v1/holder/{did}/vc` for putting it in any wallet.

woutslakhorst:
> I don't like this monkey patch too much. The wallet owner now hasn't anything to decide whether it actually wants the credential. And it's a bit of magic with obscure limitations (no JWT).
> 
> Wouldn't it be better to provide a REST API to load credentials into the wallet? Then the existing issue API (no publish) can be used to create the VC, then the new API (something like `POST internal/vcr/v1/holder/{did}/vc` for putting it in any wallet.

That would be a nice intermediate solution. It would require 2 API calls for vendors to create their credentials. It would help us to continue building the rest.

PR #2643, opened 2023-12-04
load a VC into a wallet via the internal API

### LINKED ISSUE #2642: Add API to load VC into wallet
Add a convenience API to add an unpublished VC into a wallet.

in favour of #2640

woutslakhorst:
closes #2642 

primary use case is to load a VC into the wallet when issued to a web:did.

### DIFF
--- a/docs/_static/vcr/vcr_v2.yaml
+++ b/docs/_static/vcr/vcr_v2.yaml
@@ -420,7 +420,7 @@ paths:
     post:
       summary: Load a VerifiableCredential into the holders wallet.
       description: |
-        If a VerifiableCredential is obtained in another way or when it's created without publishing, this API allows to add it to a wallet.
+        If a VerifiableCredential is not directly issued to the wallet through e.g. OpenID4VCI, this API allows to add it to a wallet.
         The DID of the holder has to be provided in the path.
         It's assumed that the credentialSubject.id equals the holder DID.
 
--- a/docs/_static/vcr/vcr_v2.yaml
+++ b/docs/_static/vcr/vcr_v2.yaml
@@ -416,7 +416,39 @@ paths:
                 $ref: "#/components/schemas/VerifiablePresentation"
         default:
           $ref: '../common/error_response.yaml'
+  /internal/vcr/v2/holder/{did}/vc:
+    post:
+      summary: Load a VerifiableCredential into the holders wallet.
+      description: |
+        If a VerifiableCredential is obtained in another way or when it's created without publishing, this API allows to add it to a wallet.
+        The DID of the holder has to be provided in the path.
+        It's assumed that the credentialSubject.id equals the holder DID.
 
+        error returns:
+        * 400 - Invalid credential
+        * 500 - An error occurred while processing the request
+      operationId: loadVC
+      tags:
+        - credential
+      parameters:
+        - name: did
+          in: path
+          description: URL encoded DID.
+          required: true
+          example: "did:web:example.com:iam:123"
+          schema:
+            type: string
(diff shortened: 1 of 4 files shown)


docs/_static/vcr/vcr_v2.yaml
```diff
@@ -416,7 +416,39 @@ paths:
                 $ref: "#/components/schemas/VerifiablePresentation"
         default:
           $ref: '../common/error_response.yaml'
+  /internal/vcr/v2/holder/{did}/vc:
+    post:
+      summary: Load a VerifiableCredential into the holders wallet.
+      description: |
+        If a VerifiableCredential is obtained in another way or when it's created without publishing, this API allows to add it to a wallet.
```
reinkrul, line 423:
```suggestion
        If a VerifiableCredential is not directly issued to the wallet through e.g. OpenID4VCI, this API allows to add it to a wallet.
```

docs/_static/vcr/vcr_v2.yaml
```diff
@@ -416,7 +416,39 @@ paths:
...
+        The DID of the holder has to be provided in the path.
+        It's assumed that the credentialSubject.id equals the holder DID.
```
reinkrul, line 425: what if it doesn't? Do we still load it?

docs/_static/vcr/vcr_v2.yaml
```diff
@@ -416,7 +416,39 @@ paths:
...
 
+        error returns:
+        * 400 - Invalid credential
```
reinkrul, line 428: or holder DID not managed by this node?

reinkrul (changes requested):
see comments

docs/_static/vcr/vcr_v2.yaml
woutslakhorst: Right now, yes. I think the internal wallet API needs to change. Instead of calling `Wallet()` on the VCR, it needs to be `Wallet(holderDID)` because not all VC.credentialSubject.ID needs to equal the holder DID theoretically.

docs/_static/vcr/vcr_v2.yaml
woutslakhorst: The current wallet does not group credentials by wallet. If you add a VC that is not owned, then you'll never be able to use it.

vcr/api/vcr/v2/api.go
```diff
@@ -321,6 +321,23 @@ func (w *Wrapper) VerifyVP(ctx context.Context, request VerifyVPRequestObject) (
 	return VerifyVP200JSONResponse(result), nil
 }
 
+func (w *Wrapper) LoadVC(ctx context.Context, request LoadVCRequestObject) (LoadVCResponseObject, error) {
+	// the actual holder is ignored for now, since we only support a single wallet...
```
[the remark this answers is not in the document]
woutslakhorst: Ownership via credentialSubject.ID is too strict.

docs/_static/vcr/vcr_v2.yaml
reinkrul: It does group them (on a shelf, identified by the credential subject DID), but without ownership check. So probably OK (we can always add the check later)

docs/_static/vcr/vcr_v2.yaml
reinkrul: or add a wallet DID to the funcs (`Put(ctx, walletDID, credential)`). Our dependency injection style makes it cumbersome to have non-multi-tenant interfaces (see OpenID handlers)

PR #2648, opened 2023-12-05
change gomigrate for dbmate

### LINKED ISSUE #2646: Investigate and migrate dbmigrate to pressly/goose
apparently https://github.com/pressly/goose has way less dependencies. If everything works the same way, it would be better to reduce deps.

woutslakhorst:
closes #2646 

there's one thingy: it requires a connection string so it opens a new connection to do migrations....


storage/engine.go
```diff
@@ -171,29 +175,14 @@ func (e *engine) initSQLDatabase() error {
...
-	}
-	migrations.Log = sqlMigrationLogger{}
-	err = migrations.Up()
-	if errors.Is(err, migrate.ErrNoChange) {
-		// There was nothing to migrate
-		return nil
-	}
-	return err
+
+	url, _ := url2.Parse(fmt.Sprintf("sqlite:%s", connectionString))
```
reinkrul, line 179: well, it looks simpler

reinkrul, line 179: url2 is quite ugly

storage/sql_migrations/001_initial.sql
```diff
@@ -0,0 +1,5 @@
+-- migrate:up
```
reinkrul, line 1: I think we an remove this first migration, now we have more (and we're changing tools anyways)

reinkrul (changes requested):

storage/engine_test.go
```diff
@@ -144,6 +144,6 @@ func Test_engine_sqlDatabase(t *testing.T) {
 		require.NoError(t, row.Err())
 		var count int
 		assert.NoError(t, row.Scan(&count))
-		assert.Equal(t, 1, count)
+		assert.Equal(t, 2, count)
```
reinkrul, line 147: how this did change? Was the test bugged earlier?

reinkrul (approved):

vdr/didweb/store_test.go
```diff
@@ -52,8 +52,6 @@ func Test_sqlStore_create(t *testing.T) {
 		verificationMethods, err := store.get(testDID)
 		require.NoError(t, err)
 		require.Len(t, verificationMethods, 2)
-		require.JSONEq(t, toJSON(vm1), toJSON(verificationMethods[0]))
```
reinkrul, line 55: maybe we should sort them, otherwise DID documents might change if order is not guaranteerd>

PR #2652, opened 2023-12-06
RFC021: VP format should be derived from PD and what is supported by OAuth2 RP and AS

reinkrul:
Earlier, it took a random (since it was a map) format from whatever was in the verifier's metadata.

Introduced proper matching on format parameters (which can be used later to also filter on cryptographic algorithms, currently not used yet), then return a matching format after stable sorting (so JWT will have precedence over JSON-LD).

### DIFF
--- a/auth/api/iam/metadata.go
+++ b/auth/api/iam/metadata.go
@@ -21,6 +21,7 @@ package iam
 import (
 	"github.com/nuts-foundation/nuts-node/auth/oauth"
 	"github.com/nuts-foundation/nuts-node/core"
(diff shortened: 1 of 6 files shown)


reinkrul:
OpenID4VC metadata specifies `alg` as `alg_values_supported` (and the same for `proof_type`), v.s. DIF's claim formats that just specify `alg` and `proof_type`. So I introduced an abstraction which translates OpenID supported formats to DIF's.

woutslakhorst (approved):

PR #2653, opened 2023-12-07
Discovery: opaque timestamp

reinkrul:
Timestamp returned from server changed to string, to allow tags from multiple several to be discerned from each other.

Also changed the name to "tag" since timestamp indicates a clock and/or number. This makes having a lamport clock code-wise easier for function/type naming.

### DIFF
--- a/discovery/interface.go
+++ b/discovery/interface.go
@@ -21,14 +21,55 @@ package discovery
 import (
 	"errors"
 	"github.com/nuts-foundation/go-did/vc"
+	"math"
+	"strconv"
+	"strings"
 )
 
-// Timestamp is value that references a point in the list.
+// Tag is value that references a point in the list.
 // It is used by clients to request new entries since their last query.
+// It is opaque for clients: they should not try to interpret it.
+// The server who issued the tag can interpret it as Lamport timestamp.
+type Tag string
+
+// Timestamp decodes the Tag into a Timestamp, which is a monotonically increasing integer value (Lamport timestamp).
+// Tags should only be decoded by the server who issued it, so the server should provide the stored tag prefix.
+// The tag prefix is a random value that is generated when the service is created.
+// It is not a secret; it only makes sure clients receive the complete presentation list when they switch servers for a specific Discovery Service:
+// servers return the complete list when the client passes a timestamp the server can't decode.
+func (t Tag) Timestamp(tagPrefix string) *Timestamp {
+	trimmed := strings.TrimPrefix(string(t), tagPrefix)
+	if len(trimmed) == len(string(t)) {
+		// Invalid tag prefix
+		return nil
+	}
+	result, err := strconv.ParseUint(trimmed, 10, 64)
+	if err != nil {
+		// Not a number
+		return nil
+	}
+	if result < 0 || result > math.MaxUint64 {
+		// Invalid uint64
+		return nil
+	}
+	lamport := Timestamp(result)
+	return &lamport
+}
+
+// Timestamp is the interpreted Tag.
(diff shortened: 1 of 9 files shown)


discovery/interface.go
```diff
@@ -21,15 +21,61 @@ package discovery
...
+}
+
+// Timestamp is the interpreted Tag.
 // It's implemented as lamport timestamp (https://en.wikipedia.org/wiki/Lamport_timestamp);
 // it is incremented when a new entry is added to the list.
 // Pass 0 to start at the beginning of the list.
 type Timestamp uint64
 
+// Tag returns the Timestamp as Tag.
+func (l Timestamp) Tag(serviceSeed string) Tag {
```
woutslakhorst, line 71: `serviceSeed` vs `tagPrefix`

discovery/store.go
```diff
@@ -221,26 +231,42 @@ func createPresentationRecord(serviceID string, timestamp Timestamp, presentatio
...
 }
 
-// get returns all presentations, registered on the given service, starting after the given timestamp.
-// It also returns the latest timestamp of the returned presentations.
-// This timestamp can then be used next time to only retrieve presentations that were added after that timestamp.
-func (s *sqlStore) get(serviceID string, startAt Timestamp) ([]vc.VerifiablePresentation, *Timestamp, error) {
+// get returns all presentations, registered on the given service, starting after the given tag.
+// It also returns the latest tag of the returned presentations.
+// This tag can then be used next time to only retrieve presentations that were added after that tag.
+func (s *sqlStore) get(serviceID string, tag *Tag) ([]vc.VerifiablePresentation, *Tag, error) {
```
woutslakhorst, line 237: just noticed that the `get` does not check if the service is still in the configured service list. So if control migrated from A to B, but clients still query A, they will always get an outdated list?

discovery/store.go
```diff
@@ -303,29 +329,44 @@ func (s *sqlStore) search(serviceID string, query map[string]string) ([]vc.Verif
...
-		// Increment timestamp
-		result.LamportTimestamp++
+		// Update tag: decode current timestamp, increment it, encode it again.
+		currTimestamp := Timestamp(0)
+		if service.LastTag != "" {
+			// If LastTag is empty, it means the service was just created and no presentations were added yet.
+			ts := service.LastTag.Timestamp(service.TagPrefix)
+			if ts == nil {
+				// would be very weird
+				return nil, fmt.Errorf("invalid tag '%s'", service.LastTag)
```
woutslakhorst, line 355: Better then to make the columns `not null`?

woutslakhorst (changes requested):

discovery/store.go
reinkrul: failure on service layer: `Module.Get` checks this. SQL store doesn't do any checks the service layer can do.

I found out that check used the wrong map (all service definitions, instead of those of the server), so I'll fix that and add a test.

discovery/store.go
reinkrul: It's not that the tag is null, it can't decode its own tag. Indicates invalid data in SQL (maybe when becoming server, after being client first for a particular service).

I didn't want to think about supporting that at first, since it seemed quite complex. But now tags are server-specific,  it has become managable but requires some SQL-fu. I'll add an issue to add the feature; https://github.com/nuts-foundation/nuts-node/issues/2656

woutslakhorst: I meant that someone could change the values to null in the DB and then that error will be triggered.

reinkrul: Or changed the value of one of `service.last_tag` or `service.tag_prefix`, making them not correlate anymore. Making the column non-null won't change that, but I'll make the error message more descriptive.

woutslakhorst (approved):

PR #2659, opened 2023-12-08
Discovery: implement server API

reinkrul:
Implements APIs specified by RFC;
- Register presentation
- Delete presentation
- Get presentations given tag (delta)

Out of scope: search (separate PR to keep changes small)


woutslakhorst (changes requested):
All ok, but I think the error/status return for a wrong audience is incorrect. My guess is that it'll respond with a 500.

reinkrul:
> All ok, but I think the error/status return for a wrong audience is incorrect. My guess is that it'll respond with a 500.

Internal server error? bad request qwould be more applicable?

discovery/api/v1/wapper.go
```diff
@@ -0,0 +1,84 @@
...
+		Entries: presentations,
+		Tag:     string(*newTag),
+	}, nil
+}
+
+func (w *Wrapper) RegisterPresentation(_ context.Context, request RegisterPresentationRequestObject) (RegisterPresentationResponseObject, error) {
+	err := w.Server.Add(request.ServiceID, *request.Body)
+	if err != nil {
+		return nil, err
+	}
```
gerardsn, line 82: should this return a 404 when errors.Is(err, errRetractionReferencesUnknownPresentation)? Also see comment on https://github.com/nuts-foundation/nuts-specification/pull/267

gerardsn (changes requested):

discovery/module.go
```diff
@@ -184,29 +202,28 @@ func (m *Module) validateRetraction(serviceID string, presentation vc.Verifiable
...
-	var retractJTIString string
-	if retractJTIRaw, ok := presentation.JWT().Get("retract_jti"); !ok {
-		return errors.New("retraction presentation does not contain 'retract_jti' claim")
-	} else {
-		if retractJTIString, ok = retractJTIRaw.(string); !ok {
-			return errors.New("retraction presentation 'retract_jti' claim is not a string")
-		}
+	retractJTIRaw, _ := presentation.JWT().Get("retract_jti")
+	retractJTI, ok := retractJTIRaw.(string)
+	if !ok {
```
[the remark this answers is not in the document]
reinkrul: Should return `errInvalidRetractionJTIClaim`, added check and test

discovery/api/v1/wapper.go
reinkrul: Nobody noticed this file is named `wapper.go`?

reinkrul: Specified it as HTTP 400 with JSON error response

gerardsn (approved):

PR #2661, opened 2023-12-08
Chore: remove superseded OpenID4VCI docs

reinkrul:
Old PoC stuff that's now implemented or replaced by better diagrams. It mainly details the VCR server-to-server OpenID4VCI flow which is incorrect and will be removed in (near?) future.


gerardsn (approved):
This documents part of the thought process when designing this. Might be useful to keep this around until we have reimplemented OpenID4VCI?

PR #2664, opened 2023-12-08
Auth: create session and validate signatures perform the same checks

reinkrul:
Signature verification performed more elaborate checks (employee details non-empty strings) than on session creation, which allows EmployeeID sessions to fail on input that was deemed valid earlier. That causes confusion, so should be the same.

Also, validation failure reason isn't logged or returned. Ideally it's returned to the client, but then we have to alter (extend) the OpenAPI spec. So for now, only log it.

Or should we still return it?


woutslakhorst:
Enough for know, on creation it'll give an error. Backport to v5 and create a bugfix release.

woutslakhorst (approved):
Test failed?

PR #2667, opened 2023-12-09
PEX: Resolve values mapped by Input Descriptor constraint fields

reinkrul:
Required for token introspection and Discovery Service search results.

### DIFF
--- a/vcr/pe/presentation_definition.go
+++ b/vcr/pe/presentation_definition.go
@@ -77,7 +77,7 @@ func (presentationDefinition PresentationDefinition) Match(vcs []vc.VerifiableCr
 // to the corresponding value from the Verifiable Credentials that map to the InputDescriptor.
 // The credentialMap is a map with the InputDescriptor.Id as key and the VerifiableCredential as value.
 // Constraints that contain no ID are ignored.
-func (presentationDefinition PresentationDefinition) ResolveConstraintsFields(credentialMap map[string]vc.VerifiableCredential) map[string]interface{} {
+func (presentationDefinition PresentationDefinition) ResolveConstraintsFields(credentialMap map[string]vc.VerifiableCredential) (map[string]interface{}, error) {
 	result := make(map[string]interface{})
 	for inputDescriptorID, cred := range credentialMap {
 		// Find the input descriptor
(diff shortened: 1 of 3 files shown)


vcr/pe/presentation_definition.go
```diff
@@ -73,6 +73,32 @@ func (presentationDefinition PresentationDefinition) Match(vcs []vc.VerifiableCr
...
+		var inputDescriptor InputDescriptor
+		for _, curr := range presentationDefinition.InputDescriptors {
+			if curr.Id == inputDescriptorID {
+				inputDescriptor = *curr
+				break
+			}
+		}
+		if inputDescriptor.Constraints == nil {
+			continue
+		}
```
gerardsn, line 93: maybe this belongs at the start of `matchConstraint`

vcr/pe/presentation_definition.go
```diff
@@ -73,6 +73,32 @@ func (presentationDefinition PresentationDefinition) Match(vcs []vc.VerifiableCr
...
+		_, values, _ := matchConstraint(inputDescriptor.Constraints, cred)
```
gerardsn, line 94: err check

gerardsn (changes requested):

vcr/pe/presentation_definition.go
reinkrul: `nil` check on `Constraints` is there in 1 other place, which seems OK to me now. Having functions do checks like "can I actually do my job?" and otherwise returns `nil` leads to fuzzy code in my experience. When there's a third spot we do the same, we should consider moving the check to a single place i.m.o.

vcr/pe/presentation_definition_test.go
```diff
@@ -710,7 +710,7 @@ func TestPresentationDefinition_ResolveConstraintsFields(t *testing.T) {
 			"any_credential": jwtCredential,
 		}
 
-		fieldValues := definition.ResolveConstraintsFields(credentialMap)
+		fieldValues, _ := definition.ResolveConstraintsFields(credentialMap)
```
gerardsn, line 713: this one probably needs an `assert.NoError` check to make sure`fieldValues` is not empty due to an error

gerardsn (approved):

PR #2678, opened 2023-12-12
Derive signer of JWT-VP from the kid header, not the 'iss' field

### LINKED ISSUE #2679: PresentationSigner should derive JWT signer from 'kid' header
`PresentationSigner` returns the JWT 'iss' field as the signer while this should be derived from the 'kid' header

gerardsn:
closes #2679

### DIFF
--- a/vcr/credential/resolver.go
+++ b/vcr/credential/resolver.go
@@ -20,12 +20,12 @@
 package credential
 
 import (
+	"errors"
 	"fmt"
 	"github.com/nuts-foundation/go-did/did"
 	"github.com/nuts-foundation/go-did/vc"
 	"github.com/nuts-foundation/nuts-node/crypto"
 	"github.com/nuts-foundation/nuts-node/vcr/signature/proof"
-	"strings"
 )
 
 // FindValidator finds the Validator the provided credential based on its Type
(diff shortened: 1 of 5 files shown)


vcr/credential/resolver.go
```diff
@@ -64,12 +65,12 @@ func ExtractTypes(credential vc.VerifiableCredential) []string {
...
-		issuer := token.Issuer()
-		if issuer == "" {
-			return nil, errors.New("JWT presentation does not have 'iss' claim")
+		kid, _, err := crypto.JWTKidAlg(presentation.Raw())
+		if err != nil {
+			return nil, err
 		}
-		return did.ParseDID(issuer)
+		signer := strings.Split(kid, "#")[0]
+		return did.ParseDID(signer)
```
reinkrul, line 73: `ParseDIDURL` then take result.DID

reinkrul (changes requested):

vcr/credential/resolver.go
```diff
@@ -59,17 +60,23 @@ func ExtractTypes(credential vc.VerifiableCredential) []string {
...
-		issuer := token.Issuer()
-		if issuer == "" {
-			return nil, errors.New("JWT presentation does not have 'iss' claim")
+		kid, _, err := crypto.JWTKidAlg(presentation.Raw())
+		if err != nil {
+			return nil, err
+		}
+		if kid == "" {
+			return nil, errors.New("no kid header in JWT")
+		}
```
gerardsn, line 74: `verifier.resolveSigningKey` tries to apply some magic in case the kid header is missing or only contains the fragment. Should this also be applied here?

reinkrul: I'd say not if missing, but yes if it only contains the fragment?

reinkrul (approved):

vcr/credential/resolver.go
gerardsn: deferred to #2684

PR #2688, opened 2023-12-14
StatusList2021Credential Read operations

### LINKED ISSUE #2689: StatusList2021 read operations (accept credentials with status)

gerardsn:
This PR allows the node to receive credentials that contain a StatusList2021 credentialStatus according to https://www.w3.org/TR/2023/WD-vc-status-list-20230427/

currently it only supports `statusPurpose == revocation`

Verifier.Verify only fails when the revocation can be confirmed. other errors are logged

StatusList2021Credentials are fetched on every verification. caching will be added later

closes #2689


vcr/credential/statuslist2021/bitstring.go
```diff
@@ -0,0 +1,108 @@
...
+		return "", err
+	}
+	err = gz.Close()
+	if err != nil {
+		return "", err
+	}
+
+	// encode to base64 string
+	// the examples contain '-' from which I took that is uses URLEncoding over StdEncoding
+	return base64.URLEncoding.EncodeToString(buf.Bytes()), nil
```
gerardsn, line 84: [this](https://github.com/TBD54566975/ssi-sdk/blob/main/credential/status/statuslist2021.go) uses STDEncoding
[this](https://github.com/trustbloc/vc-go/blob/v1.1.0/status/internal/bitstring/bitstring.go) one uses RAWURLEncoding

vcr/credential/validator.go
```diff
@@ -110,6 +114,74 @@ func (d defaultCredentialValidator) Validate(credential vc.VerifiableCredential)
...
+		if credentialStatus.ID.String() == "" {
+			return errors.New("credentialStatus.id is required")
+		}
+		if credentialStatus.Type == "" {
+			return errors.New("credentialStatus.type is required")
+		}
+
+		// only accept StatusList2021EntryType for now
+		if credentialStatus.Type != StatusList2021EntryType {
+			continue
```
reinkrul, line 150: Maybe return an error if the status type is not supported?

vcr/credential/validator.go
```diff
@@ -110,6 +114,74 @@ func (d defaultCredentialValidator) Validate(credential vc.VerifiableCredential)
...
+		}
+
+		if !credential.ContainsContext(statusList2021ContextURI) {
+			return errors.New("StatusList2021 context is required")
+		}
+
+		// marshal as StatusList2021Entry
```
reinkrul, line 157:
```suggestion
		// unmarshal as StatusList2021Entry
```

vcr/verifier/credential_status.go
```diff
@@ -0,0 +1,221 @@
...
+	}
+	// TODO: cache updated credential so it does not have to be downloaded everytime
+	//  	 also cache if statusPurposes != 'revocation' to prevent unnecessary downloads
+	return &sl, nil
+}
+
+// download the StatusList2021Credential found at statusList2021Entry.statusListCredential
+func (cs *credentialStatus) download(statusListCredential string) (*vc.VerifiableCredential, error) {
+	var cred vc.VerifiableCredential // VC containing CredentialStatus of the credentialToVerify
+	req, err := http.NewRequest(http.MethodGet, statusListCredential, new(bytes.Buffer))
```
reinkrul, line 162:
```suggestion
	req, err := http.NewRequest(http.MethodGet, statusListCredential, nil)
```

reinkrul: No need for body in GET request

vcr/verifier/credential_status.go
```diff
@@ -0,0 +1,221 @@
...
+		return nil, err
+	}
+	body, err := io.ReadAll(res.Body)
+	if err = res.Body.Close(); err != nil {
+		// log, don't fail
+		log.Logger().
+			WithError(err).
+			WithField("method", "credentialStatus.download").
+			Debug("failed to close response body")
+	}
```
reinkrul, line 177:
```suggestion
	defer func() {
		if err = res.Body.Close(); err != nil {
			// log, don't fail
			log.Logger().
				WithError(err).
				WithField("method", "credentialStatus.download").
				Debug("failed to close response body")
		}
	}()
	body, err := io.ReadAll(res.Body)
```

reinkrul: Close() before reading in defer, after `Do`'s error check

vcr/verifier/credential_status.go
```diff
@@ -0,0 +1,221 @@
...
+	if res.StatusCode > 299 || err != nil {
+		return nil, fmt.Errorf("fetching StatusList2021Credential from '%s' failed", statusListCredential)
```
reinkrul, line 179: might not want to throw err away. You could use `errors.Join`.

vcr/verifier/credential_status.go
```diff
@@ -0,0 +1,221 @@
...
+
+// verifyStatusList2021Credential checks that the StatusList2021Credential is currently valid
+func (cs *credentialStatus) verifyStatusList2021Credential(cred vc.VerifiableCredential) (*credential.StatusList2021CredentialSubject, error) {
+	// make sure we have the correct credential.
+	if len(cred.Type) > 2 || !cred.IsType(ssi.MustParseURI(credential.StatusList2021CredentialType)) {
+		return nil, errors.New("incorrect credential types")
+	}
+
+	// returns statusList2021CredentialValidator, or Validate() fails because base type is missing
+	if err := credential.FindValidator(cred).Validate(cred); err != nil {
```
reinkrul, line 195: probably should just directly create the validator, since we exactly know which one we need...

vcr/vcr.go
```diff
@@ -262,7 +263,8 @@ func (c *vcr) Configure(config core.ServerConfig) error {
 		c.openidSessionStore = c.storageClient.GetSessionDatabase()
 	}
 	c.issuer = issuer.NewIssuer(c.issuerStore, c, networkPublisher, openidHandlerFn, didResolver, c.keyStore, c.jsonldManager, c.trustConfig)
-	c.verifier = verifier.NewVerifier(c.verifierStore, didResolver, c.keyResolver, c.jsonldManager, c.trustConfig)
+	c.verifierHttpClient = core.NewStrictHTTPClient(config.Strictmode, c.config.OpenID4VCI.Timeout, tlsConfig)
```
reinkrul, line 266: I don't think strict HTTP client is required, since the credential is signed?

Side question: do we check that the status credential is signed by the credential (being checked)'s issuer?

reinkrul (changes requested):

vcr/vcr.go
gerardsn: https is probably not required, same for the tlsConfig? and it also abuses the OpenID4VCI timeout. I wasn't sure so went with the safest option for now. CRLs are often served on http since they are signed and it solves a chicken-egg problem, but here that problem does not exist. Applying the node's strictmode setting made sense to me, but I am open to a broader solution to what is needed here.

Side question: no, will add to credentialStatus.Verify (the spec does not list this as requirement, but it is probably implied)

vcr/verifier/credential_status.go
gerardsn: added to note on epic. dependency cycles vs consistency are a big issue in this feature

vcr/vcr.go
gerardsn: Side question: no, ~~will add to credentialStatus.Verify (the spec does not list this as requirement, but it is probably implied)~~ the validate algorithm does not include a check that the `statusListCredential` and the `credentialToVerify` have the same issuer, only that they have the same `statusPurpose`.

Should we include this check? Without this check a vendor could have 1 statuslist for all the dids it manages (makes our lives a bit easier), but how should the trust in the issuer of the statuslist be derived? If we trust the credential issuer containing a `credentialStatus`, just trust the revocation mechanism it uses?

vcr/credential/statuslist2021/bitstring_test.go
```diff
@@ -0,0 +1,111 @@
...
+	t.Run("error - index OOB", func(t *testing.T) {
+		bs := Bitstring{0} // single byte
+		assert.ErrorIs(t, bs.SetBit(-1, true), ErrIndexNotInBitstring)
+		assert.NoError(t, bs.SetBit(0, true))
+		assert.ErrorIs(t, bs.SetBit(8, true), ErrIndexNotInBitstring)
+	})
+}
+
+func Test_CompressExpand(t *testing.T) {
+	t.Run("ok - empty bitstring from example", func(t *testing.T) {
```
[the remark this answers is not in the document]
gerardsn: sounds useful but I will need some help getting it to work.

A quick scan of the tests suggests that it might not be that useful since we only implement `revocation`, not `suspension`, and we apply soft-fail approach to anything that is not a confirmed `revocation`

vcr/credential/statuslist2021/bitstring.go
gerardsn: https://www.github.com/w3c/vc-bitstring-status-list/issues/58 says URLEncoding, but usage of padding is still unclear. Go's DecodeString methods on RawURLEncoding (no pad) and URLEncoding (with pad) return an error if the wrong one is used. Maybe try 1, and if it fails try the other?

gerardsn: https://www.rfc-editor.org/rfc/rfc4648.html#section-3.2
> Implementations MUST include appropriate pad characters at the end of
   encoded data unless the specification referring to this document
   explicitly states otherwise.

so follow all the specs and be interoperable with no-one

reinkrul: >Maybe try 1, and if it fails try the other?
Sounds like the most interoperable option?

vcr/credential/types.go
```diff
@@ -84,3 +85,46 @@ type Resource struct {
...
+
+var statusList2021ContextURI = ssi.MustParseURI(jsonld.W3cStatusList2021Context)
+var statusList2021CredentialTypeURI = ssi.MustParseURI(StatusList2021CredentialType)
+
+// StatusList2021Entry is the "credentialStatus" property used by issuers to enable VerifiableCredential status information.
+type StatusList2021Entry struct {
+	// ID is expected to be a URL that identifies the status information associated with the verifiable credential.
+	// It MUST NOT be the URL for the status list, which is in StatusListCredential.
+	ID string `json:"id,omitempty"`
+	// Type MUST be "StatusList2021Entry"
```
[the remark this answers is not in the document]
reinkrul: Oh I thought we were implementing BitstringStatusList

vcr/credential/types.go
```diff
@@ -84,3 +85,46 @@ type Resource struct {
...
+	// StatusPurpose indicates what it means if the VerifiableCredential is on the list.
+	// The value is arbitrary, with predefined values `revocation` and `suspension`.
+	// This value must match credentialSubject.statusPurpose value in the VerifiableCredential.
+	StatusPurpose string `json:"statusPurpose,omitempty"`
+	// StatusListIndex is an arbitrary size integer greater than or equal to 0, expressed as a string.
+	// The value identifies the bit position of the status of the verifiable credential.
+	StatusListIndex string `json:"statusListIndex,omitempty"`
+	// The statusListCredential property MUST be a URL to a verifiable credential.
+	// When the URL is dereferenced, the resulting verifiable credential MUST have type property that includes the "StatusList2021Credential" value.
+	StatusListCredential string `json:"statusListCredential,omitempty"`
```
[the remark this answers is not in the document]
reinkrul: then, nothing

vcr/credential/validator.go
```diff
@@ -110,6 +114,74 @@ func (d defaultCredentialValidator) Validate(credential vc.VerifiableCredential)
...
+		if credentialStatus.Type == "" {
+			return errors.New("credentialStatus.type is required")
+		}
+
+		// only accept StatusList2021EntryType for now
+		if credentialStatus.Type != StatusList2021EntryType {
+			continue
+		}
+
+		if !credential.ContainsContext(statusList2021ContextURI) {
```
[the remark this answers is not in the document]
reinkrul: I saw it in de VC data integrity processing guidelines as well, so let's keep it like this

vcr/vcr.go
reinkrul: IDeally you'd have multiple layers of protection, if issuer isn't checked and we have no HTTPS, you could spoof the endpoint (MITM) and you're in. If there's HTTPS, spoofing becomes a lot harder. If we have no HTTPS but the status credential issuer is checked, you need to steal a credential issuer private key and spoof DNS.

Enforcing HTTPS is probably the easiest fix here. Although I don't expect other DIDs than the credential issuer to sign the status credential? (maybe EBSI, but out of scope..)

reinkrul (approved):

PR #2702, opened 2023-12-21
Handle direct_post return from wallet to verifier

### LINKED ISSUE #2694: Handle authorization response from holder by verifier

woutslakhorst:
closes #2694 

the /token API is also implemented but not yet used in the e2e test. The next PR should combine the two. 

some method signatures are getting quite long. We might need to refactor these after completion of the entire flow.

the constructed callback url for the holder will be implemented by the next PR.

`nonce` and `challenge` are still a bit of a mess, but types will change in the future anyway...


vcr/pe/presentation_submission.go
```diff
@@ -149,13 +149,16 @@ func (b *PresentationSubmissionBuilder) Build(format string) (PresentationSubmis
 
 	// the verifiableCredential property in Verifiable Presentations can be a single VC or an array of VCs when represented in JSON.
 	// go-did always marshals a single VC as a single VC for JSON-LD VPs. So we might need to fix the mapping paths.
-	if format == vc.JSONLDPresentationProofFormat {
-		for _, signInstruction := range nonEmptySignInstructions {
-			if len(signInstruction.Mappings) == 1 {
-				signInstruction.Mappings[0].Path = "$.verifiableCredential"
-			}
+
+	// todo the check below actually depends on the format of the credential and not the format of the VP
```
reinkrul, line 153: I think this isn't right?

vcr/pe/presentation_submission.go
```diff
@@ -149,13 +149,16 @@ func (b *PresentationSubmissionBuilder) Build(format string) (PresentationSubmis
...
+	// commented for now because it's needed for VPs in JWT format as well.
+	//if format == vc.JSONLDPresentationProofFormat {
```
reinkrul, line 155: just remove comment and commented out code?

e2e-tests/oauth-flow/openid4vp/docker-compose.yml
```diff
@@ -0,0 +1,56 @@
+version: "3.7"
+services:
+  nodeA-backend:
```
reinkrul, line 3: why do we need NGINX in front, we can terminate TLS on the Nuts node in this case, I think?

As the Nuts network isn't needed.

e2e-tests/oauth-flow/openid4vp/run-test.sh
```diff
@@ -0,0 +1,125 @@
...
+echo "Starting Docker containers..."
+echo "------------------------------------"
+docker compose up -d
+docker compose up --wait nodeA nodeA-backend nodeB nodeB-backend
+
+echo "------------------------------------"
+echo "Registering vendors..."
+echo "------------------------------------"
+# Register Vendor A
+VENDOR_A_DIDDOC=$(docker compose exec nodeA-backend nuts vdr create-did --v2)
```
reinkrul, line 22: actually not a vendor-DID document, but just a DID document or care organization DID doc?

e2e-tests/oauth-flow/openid4vp/run-test.sh
```diff
@@ -0,0 +1,125 @@
...
+  echo $RESPONSE
+  echo "FAILED: Could not get redirectURL from node-B" 1>&2
+  exitWithDockerLogs 1
+fi
+
+echo "--------------------------------------"
+echo "Redirect user to local OAuth server..."
+echo "--------------------------------------"
+
+LOCATION=$(echo $LOCATION | sed -E 's/nodeB/localhost:20443/')
```
reinkrul, line 70: we might want to make this an actual browser test soon?

docs/_static/auth/iam.yaml
```diff
@@ -149,6 +153,52 @@ paths:
                 "$ref": "#/components/schemas/PresentationDefinition"
         "default":
           $ref: '../common/error_response.yaml'
+  "/iam/{id}/response":
```
reinkrul, line 156: is `/response` specific enough? Or should it be something like `/response_direct_post` or `/authz_response`

docs/_static/auth/iam.yaml
```diff
@@ -149,6 +153,52 @@ paths:
...
+    post:
+      summary: Used by wallets to post the authorization response or error to.
+      description: | 
+        Specified by https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#name-response-mode-direct_postjw
+        The response is either an error response with error, error_description and state filled or a submission with vp_token and presentation_submission filled.
```
reinkrul, line 161: Why no state in success flow?

auth/services/oauth/relying_party.go
```diff
@@ -126,12 +127,18 @@ func (s *relyingParty) CreateAuthorizationRequest(ctx context.Context, requestHo
 	if err != nil {
 		return nil, fmt.Errorf("failed to parse authorization endpoint URL: %w", err)
 	}
-	// todo: redirect_uri
+	// construct callback URL for wallet
+	callbackURL, err := didweb.DIDToURL(requestHolder)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create callback URL: %w", err)
+	}
+	callbackURL = callbackURL.JoinPath("callback")
```
reinkrul, line 135: not specified on the OpenAPI spec?

auth/services/oauth/interface.go
```diff
@@ -63,11 +62,11 @@ type Verifier interface {
 // Holder implements the OpenID4VP Holder role which acts as Authorization server in the OpenID4VP flow.
 type Holder interface {
 	// BuildPresentation builds a Verifiable Presentation based on the given presentation definition.
-	BuildPresentation(ctx context.Context, walletDID did.DID, presentationDefinition pe.PresentationDefinition, acceptedFormats map[string]map[string][]string, nonce string) (*vc.VerifiablePresentation, *pe.PresentationSubmission, error)
+	BuildPresentation(ctx context.Context, walletDID did.DID, presentationDefinition pe.PresentationDefinition, acceptedFormats map[string]map[string][]string, nonce string, audience did.DID) (*vc.VerifiablePresentation, *pe.PresentationSubmission, error)
```
reinkrul, line 65: audience could be any URI? (DID is also a URI)

auth/services/oauth/holder.go
```diff
@@ -102,14 +106,20 @@ func (v *HolderService) ClientMetadata(ctx context.Context, endpoint string) (*o
...
 	}
-	redirectURL, err := iamClient.PostError(ctx, auth2Error, *responseURL)
+	validURL := *responseURL
+	if verifierClientState != "" {
+		validURL = http.AddQueryParams(*responseURL, map[string]string{
+			oauth.StateParam: verifierClientState,
+		})
+	}
+	redirectURL, err := iamClient.PostError(ctx, auth2Error, validURL)
 	if err != nil {
```
reinkrul, line 123: do we need to check the redirect URL, whether it's ours? (do we trust the AS)

auth/services/oauth/holder.go
```diff
@@ -102,14 +106,20 @@ func (v *HolderService) ClientMetadata(ctx context.Context, endpoint string) (*o
 	return metadata, nil
 }
 
-func (v *HolderService) PostError(ctx context.Context, auth2Error oauth.OAuth2Error, verifierResponseURI string) (string, error) {
+func (v *HolderService) PostError(ctx context.Context, auth2Error oauth.OAuth2Error, verifierResponseURI string, verifierClientState string) (string, error) {
```
reinkrul, line 109: `verifier` is superfluous in `verifierClientState` here; in this flow the wallet is the AS and verifier the RP (thus client), thus `clientState` can only be the other party's.

auth/api/iam/session.go
```diff
@@ -30,11 +32,57 @@ type OAuthSession struct {
...
+		if pd, ok := val.(pe.PresentationSubmission); ok {
+			return pd
+		}
+	}
+	return pe.PresentationSubmission{}
+}
+
+// CredentialMap returns the credential map from the server state.
+func (s ServerState) CredentialMap() map[string]vc.VerifiableCredential {
+	if val, ok := s[credentialMapStateKey]; ok {
```
reinkrul, line 78: If I remember correctly, you can shorten this to:

```go
if mapped, ok := s[credentialMapStateKey].(map[string]vc.VerifiableCredential); ok {
```

`nil` will return `!ok`

reinkrul (changes requested):

vcr/pe/presentation_submission.go
woutslakhorst: It did create an error when running the e2e-test. But "fixing" it now is a bit pro-active since VC formats are going to be changed anyway.

e2e-tests/oauth-flow/openid4vp/run-test.sh
woutslakhorst: made it a PARTY

docs/_static/auth/iam.yaml
woutslakhorst: In this PR the nonce is used to get the state. In the next PR it might just have the state for redirecting the error result

auth/services/oauth/relying_party.go
woutslakhorst: next PR

e2e-tests/oauth-flow/openid4vp/run-test.sh
woutslakhorst: Using Demo-EHR and Admin. Since only 302s are used, the current test is fine.

docs/_static/auth/iam.yaml
woutslakhorst: There's a description on it. The other oauth2/openid4vp specific endpoints are: `token` and `authorize`, so a simple name is in line with the others.

auth/services/oauth/holder.go
woutslakhorst: Can't find it in the security considerations of the spec(s). But won't hurt to add.

auth/services/oauth/holder.go
woutslakhorst: that's why it's called the `verifierClientState`, as in: the client state of the verifier who has the RP role in this part.

reinkrul: You're right, it's defined as just `state` (https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.1). But a parameter `state` still always refers to the client/RP. But it's so local here, it doesn't really matter.

auth/api/iam/openid4vp.go
```diff
@@ -269,14 +269,22 @@ func (r Wrapper) sendAndHandleDirectPost(ctx context.Context, vp vc.VerifiablePr
...
+			return nil, err
+		}
+		if strings.HasPrefix(redirectURI, registeredURL.String()) {
+			return HandleAuthorizeRequest302Response{
+				HandleAuthorizeRequest302ResponseHeaders{
+					Location: redirectURI,
+				},
+			}, nil
+		}
+		log.Logger().Errorf("verifier responded with incorrect callbackURI: \"%s\" for wallet: %s", redirectURI, walletDID.String())
```
reinkrul, line 287:
```suggestion
		log.Logger().Errorf("Verifier responded with incorrect callbackURI: \"%s\" for wallet: %s", redirectURI, walletDID)
```

reinkrul (approved):

auth/api/iam/openid4vp.go
```diff
@@ -286,6 +309,253 @@ func (r Wrapper) sendAndHandleDirectPostError(ctx context.Context, auth2Error oa
...
+
+	// check if the authorization code is valid
+	var oauthSession OAuthSession
+	err = r.oauthCodeStore().Get(*authorizationCode, &oauthSession)
+	if err != nil {
+		return nil, withCallbackURI(oauthError(oauth.InvalidRequest, "invalid authorization code"), callbackURI)
+	}
+
+	// check if the redirectURI matches the one from the authorization request
+	if oauthSession.redirectURI() != nil && oauthSession.redirectURI().String() != *redirectURI {
```
gerardsn, line 534: only accept registered callback URIs?
```suggestion
	if oauthSession.redirectURI() != nil || oauthSession.redirectURI().String() != *redirectURI {
```

auth/api/iam/openid4vp.go
```diff
@@ -286,6 +309,253 @@ func (r Wrapper) sendAndHandleDirectPostError(ctx context.Context, auth2Error oa
...
+	callbackURI, err := url.Parse(*redirectURI)
+	if err != nil {
+		return nil, oauthError(oauth.InvalidRequest, "invalid redirect_uri parameter")
+	}
+
+	// check if the authorization code is valid
+	var oauthSession OAuthSession
+	err = r.oauthCodeStore().Get(*authorizationCode, &oauthSession)
+	if err != nil {
+		return nil, withCallbackURI(oauthError(oauth.InvalidRequest, "invalid authorization code"), callbackURI)
```
gerardsn, line 530: callbackURI should not be used until confirmed
```suggestion
		return nil, oauthError(oauth.InvalidRequest, "invalid authorization code")
```

auth/api/iam/openid4vp.go
```diff
@@ -286,6 +309,253 @@ func (r Wrapper) sendAndHandleDirectPostError(ctx context.Context, auth2Error oa
...
+	// check if the client_id matches the one from the authorization request
+	if oauthSession.ClientID != *clientId {
+		return nil, withCallbackURI(oauthError(oauth.InvalidRequest, "client_id does not match"), callbackURI)
+	}
+
+	presentations := oauthSession.ServerState.VerifiablePresentations()
+	submission := oauthSession.ServerState.PresentationSubmission()
+	definition, err := r.policyBackend.PresentationDefinition(ctx, verifier, oauthSession.Scope)
+	if err != nil {
+		return nil, withCallbackURI(oauthError(oauth.ServerError, fmt.Sprintf("failed to fetch presentation definition: %s", err.Error())), callbackURI)
```
gerardsn, line 547: this is going to be very annoying with external policy backends

auth/api/iam/openid4vp.go
```diff
@@ -286,6 +309,253 @@ func (r Wrapper) sendAndHandleDirectPostError(ctx context.Context, auth2Error oa
...
+	}
+	credentialMap := oauthSession.ServerState.CredentialMap()
+	subject, _ := did.ParseDID(oauthSession.ClientID)
+
+	response, err := r.createS2SAccessToken(verifier, time.Now(), presentations, submission, *definition, oauthSession.Scope, *subject, credentialMap)
```
gerardsn, line 552: rename method if not specific to S2S flow

auth/api/iam/openid4vp.go
```diff
@@ -286,6 +309,253 @@ func (r Wrapper) sendAndHandleDirectPostError(ctx context.Context, auth2Error oa
...
+	pexEnvelope, err := pe.ParseEnvelope([]byte(*request.Body.VpToken))
+	if err != nil || len(pexEnvelope.Presentations) == 0 {
+		return nil, oauthError(oauth.InvalidRequest, "invalid vp_token")
+	}
+
+	// note: instead of using the challenge to lookup the oauth session, we could also add a client state from the verifier.
+	// this would allow us to lookup the redirectURI without checking the VP first.
+
+	// extract the nonce from the vp(s)
+	nonce, err := extractChallenge(pexEnvelope.Presentations[0])
```
gerardsn, line 370: should check all presentations. this will have 2 VPs (user and care org)

auth/api/iam/validation.go
```diff
@@ -0,0 +1,100 @@
...
+// Errors are returned as OAuth2 errors.
+func (r Wrapper) validatePresentationSubmission(ctx context.Context, authorizer did.DID, scope string, submission *pe.PresentationSubmission, pexEnvelope *pe.Envelope) (map[string]vc.VerifiableCredential, *PresentationDefinition, error) {
+	definition, err := r.policyBackend.PresentationDefinition(ctx, authorizer, scope)
+	if err != nil {
+		return nil, nil, oauth.OAuth2Error{
+			Code:          oauth.InvalidScope,
+			InternalError: err,
+			Description:   fmt.Sprintf("unsupported scope (%s) for presentation exchange: %s", scope, err.Error()),
+		}
+	}
```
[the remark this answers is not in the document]
woutslakhorst: added check for `policy.ErrNotFound`

auth/api/iam/openid4vp.go
```diff
@@ -254,14 +269,22 @@ func (r Wrapper) sendAndHandleDirectPost(ctx context.Context, vp vc.VerifiablePr
...
+			return nil, err
+		}
+		if strings.HasPrefix(redirectURI, registeredURL.String()) {
+			return HandleAuthorizeRequest302Response{
+				HandleAuthorizeRequest302ResponseHeaders{
+					Location: redirectURI,
+				},
+			}, nil
+		}
+		log.Logger().Errorf("verifier responded with incorrect callbackURI: \"%s\" for wallet: %s", redirectURI, walletDID.String())
```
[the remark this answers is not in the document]
woutslakhorst: it was added after a comment from @reinkrul with the comment "wouldn't hurt to check, but now we have a reason for not checking"

auth/api/iam/openid4vp.go
```diff
@@ -286,6 +309,253 @@ func (r Wrapper) sendAndHandleDirectPostError(ctx context.Context, auth2Error oa
...
+// errors are returned as OAuth2 errors.
+func (r Wrapper) validatePresentationNonce(presentations []vc.VerifiablePresentation) error {
+	var nonce string
+	var returnErr error
+	for _, presentation := range presentations {
+		nextNonce, err := extractChallenge(presentation)
+		_ = r.oauthNonceStore().Delete(nextNonce)
+		if nextNonce == "" {
+			// fallback on nonce instead of challenge, todo: should be uniform, check vc data model specs for JWT/JSON-LD
+			nextNonce, err = extractNonce(presentation)
```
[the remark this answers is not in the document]
woutslakhorst: placed lower

auth/api/iam/openid4vp.go
woutslakhorst: We don't have any registered callbacks, suggestion would generate nilpointer

auth/api/iam/openid4vp.go
woutslakhorst: this is done 20 lines later

auth/api/iam/openid4vp.go
gerardsn: first part should have checked redirectURI == nil. Now that I think about it again, the flow have failed before if there is no redirect_uri, so probably doesn't make sense.

I think the session's redirect_uri is considered a registered callback, but I can't find it anymore.

auth/api/iam/openid4vp.go
```diff
@@ -286,6 +309,253 @@ func (r Wrapper) sendAndHandleDirectPostError(ctx context.Context, auth2Error oa
...
+				}, nil
+			}
+		}
+	}
+	// we don't have a client state, so we can't redirect to the holder redirectURI
+	// return an error page instead
+	return nil, oauthError(oauth.ErrorCode(code), description)
+}
+
+func (r Wrapper) handleAuthorizeResponseSubmission(ctx context.Context, request HandleAuthorizeResponseRequestObject) (HandleAuthorizeResponseResponseObject, error) {
```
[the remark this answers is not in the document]
gerardsn: I think at this point the inner flow is complete. It's either successful and the return is a redirect with authorization code, or it is unsuccessful and the return is a redirect with the error. (Ignoring the missing callbackURI case)

The second half of the errors uses `withCallbackURI`. Can you explain how these are different than the first half?

Issue #2065, opened 2023-04-19
Auth: remove auth.http.timeout in favor of http-client.timeout


reinkrul:
Global `http-client.timeout` was introduced, `auth.http.timeout` may be removed in v6.

woutslakhorst:
Global is now called `httpclient.timeout`

PR #2725, opened 2024-01-16
Config: Add HTTP-client timeout

reinkrul:
Also moved defaults to `NewServerConfig()`, which caused the need to alter `TestServerConfig()` to now take a visitor instead of a struct


woutslakhorst (approved):
closes an issue?

reinkrul:
No, introduces the timeout mentioned in https://github.com/nuts-foundation/nuts-node/issues/2065

PR #2730, opened 2024-01-17
Discovery: fix API of Client.Search()

reinkrul:
Found these issues while fixing the e2e test.

Fixes:

- Endpoint was intended as an internal endpoint, moved it to `/internal/discovery/v1`
- Use of dynamic query parameters isn't supported by codegen, workaround

### DIFF
--- a/discovery/api/v1/api.go
+++ b/discovery/api/v1/api.go
@@ -27,10 +27,13 @@ import (
 	"github.com/nuts-foundation/nuts-node/core"
 	"github.com/nuts-foundation/nuts-node/discovery"
 	"net/http"
+	"net/url"
 )
 
 var _ StrictServerInterface = (*Wrapper)(nil)
 var _ core.ErrorStatusCodeResolver = (*Wrapper)(nil)
+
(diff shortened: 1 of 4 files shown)


discovery/api/v1/api.go
```diff
@@ -55,6 +58,12 @@ func (w *Wrapper) Routes(router core.EchoRouter) {
 				ctx.Set(core.OperationIDContextKey, operationID)
 				ctx.Set(core.ModuleNameContextKey, discovery.ModuleName)
 				ctx.Set(core.StatusCodeResolverContextKey, w)
+				// deepmap/openapi codegen does not support dynamic query parameters ("exploded form parameters"),
```
gerardsn, line 61: is this a feature of the strict-server option?

gerardsn (approved):

discovery/api/v1/api.go
reinkrul: Not sure, but it's a rather obscure/new feature which not all code generators support properly

PR #2734, opened 2024-01-19
CRUD for did:web services

reinkrul:


vdr/api/v2/api.go
```diff
@@ -104,19 +105,48 @@ func (a *Wrapper) ListDIDs(ctx context.Context, _ ListDIDsRequestObject) (ListDI
...
-	panic("implement me")
+	targetDID, err := did.ParseDID(request.Did)
+	if err != nil {
+		return nil, err
+	}
+	serviceID, err := ssi.ParseURI(request.ServiceId)
+	if err != nil {
+		return nil, err
+	}
+	err = w.VDR.DeleteService(ctx, *targetDID, *serviceID)
```
woutslakhorst, line 129: Can result in an `didweb.errServiceNotFound` which is not wrapped by the VDR/management which then will result in a 500 error, should be in the 400 range. Same goes for other errors and other services. (including unique key violations, etc)

vdr/didweb/store.go
```diff
@@ -120,3 +151,43 @@ func (s *sqlStore) list() ([]did.DID, error) {
 	}
 	return result, nil
 }
+
+// createService creates a new service in the DID document identified by subjectDID.
+// It does not validate the service.
+func (s *sqlStore) createService(subjectDID did.DID, service did.Service) error {
```
woutslakhorst, line 157: For all DB related operations, make sure to deal with:
- `ErrRecordNotFound`
- `ErrDuplicatedKey`
- (`ErrForeignKeyViolated`)

Otherwise 500 errors will be generated where a 400 should be generated.

storage/sql_migrations/002_didweb.sql
```diff
@@ -21,6 +21,20 @@ create table vdr_didweb_verificationmethod
     foreign key (did) references vdr_didweb (did) on delete cascade
 );
 
+-- this table is used to store the services for a did:web
+create table vdr_didweb_service
+(
+    -- id is the unique id of the service as it appears in the DID document.
+    id   varchar(500) not null,
```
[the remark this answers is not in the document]
reinkrul: Yes, we store the complete DID

vdr/api/v2/api.go
reinkrul: Now mapping these errors to `management.ErrInvalidService`, which is then mapped to HTTP 400

PR #2737, opened 2024-01-19
VDR: Group all DID Document management funcs in a new interface

reinkrul:
To produce a DID document of a certain type, you need to implement the `DocumentManager` interface. This is in preparation of extending the functionality of updating `did:web` documents; after this PR, https://github.com/nuts-foundation/nuts-node/pull/2734 will add CRUD for services. Later on, Verification Method CRUD might be added as well.

The goal is to have VDR only delegate calls to supported DID methods.

When `did:nuts` support is removed, many of the "old" (current) interfaces can be removed, this one will remain.


woutslakhorst (approved):
Not sure if this is going to make things clearer/leaner, but let's see where it goes.

Issue #2727, opened 2024-01-16
Discovery: API for Discovery Client to return the service/DID activation status


reinkrul:
Administrative applications will want to know whether a service/DID combination has been registered.

PR #2740, opened 2024-01-22
Discovery: get DID activation status on service

reinkrul:
Most probably required by future "Nuts Admin"-applications.

Fixes https://github.com/nuts-foundation/nuts-node/issues/2727

### DIFF
--- a/discovery/module.go
+++ b/discovery/module.go
@@ -307,6 +307,29 @@ func (m *Module) Services() []ServiceDefinition {
 	return result
 }
 
+// GetServiceActivation is a Discovery Client function that retrieves the activation status of a service for a DID.
+// See interface.go for more information.
+func (m *Module) GetServiceActivation(_ context.Context, serviceID string, subjectDID did.DID) (bool, *vc.VerifiablePresentation, error) {
+	refreshTime, err := m.store.getPresentationRefreshTime(serviceID, subjectDID)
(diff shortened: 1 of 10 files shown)


docs/_static/discovery/v1.yaml
```diff
@@ -139,6 +139,31 @@ paths:
         required: true
         schema:
           type: string
+    get:
```
gerardsn, line 161: feels like a duplication of search

discovery/module.go
```diff
@@ -307,6 +307,29 @@ func (m *Module) Services() []ServiceDefinition {
...
+// GetServiceActivation is a Discovery Client function that retrieves the activation status of a service for a DID.
+// See interface.go for more information.
+func (m *Module) GetServiceActivation(_ context.Context, serviceID string, subjectDID did.DID) (bool, *vc.VerifiablePresentation, error) {
+	refreshTime, err := m.store.getPresentationRefreshTime(serviceID, subjectDID)
+	if err != nil {
+		return false, nil, err
+	}
+	if refreshTime == nil {
+		return false, nil, nil
+	}
```
gerardsn, line 319: this tests if the client is managing the subjectDID's status on the discovery server, which is not in the API description

gerardsn (approved):

PR #2741, opened 2024-01-22
Discovery: more documentation on client usage

reinkrul:


docs/pages/deployment/discovery.rst
```diff
@@ -24,6 +24,50 @@ Service definitions are JSON files loaded from the ``discovery.definitions.direc
...
+
+Registration
+============
+
+To make an organization discoverable through a Discovery Service, need to activate the organization's DID on that particular service.
+This can be done by performing a HTTP POST to the ``/internal/discovery/v1/{serviceID}/{did}`` endpoint of the Nuts node,
+where the service ID must match the ID of the service definition.
+The Nuts node will try to create a Verifiable Presentation according to the Service Definition and register it on the Discovery Server.
+
+Registration on the Discovery Service will only succeed if then the DID's wallet contains the required credentials.
```
gerardsn, line 53:
```suggestion
Registration on the Discovery Service will only succeed if the DID's wallet contains the required credentials.
```

docs/pages/deployment/discovery.rst
```diff
@@ -24,6 +24,50 @@ Service definitions are JSON files loaded from the ``discovery.definitions.direc
...
+If registration fails it will be retried after a set interval, which can be configured with ``discovery.client.registration_refresh_interval``.
```
gerardsn, line 54:
```suggestion
If registration fails, it will be retried after a set interval, which can be configured with ``discovery.client.registration_refresh_interval``.
```

docs/pages/deployment/discovery.rst
```diff
@@ -24,6 +24,50 @@ Service definitions are JSON files loaded from the ``discovery.definitions.direc
...
+
+The Nuts node will automatically refresh Verifiable Presentations that are about to expire.
```
gerardsn, line 56: Not clear that this is also managed by the `registration_refresh_interval`

gerardsn (changes requested):
If we put a number on the discovery service RFC it can be referenced here

It may help readers to describe the link between `serviceID` and use-cases, or just something that helps them connect the 2

PR #2744, opened 2024-01-23
VDR: Introduce DID method-specific creation options

reinkrul:
DID methods have their own creation options, SelfControl and Controllers are `did:nuts` specific. `did:web` has (user) "id" (last path part in the DID).

This moves `did:nuts` options to the `didnuts` package and introduces an option for `did:web`.

### DIFF
--- a/vdr/api/v1/api.go
+++ b/vdr/api/v1/api.go
@@ -122,12 +122,12 @@ func (a *Wrapper) CreateDID(ctx context.Context, request CreateDIDRequestObject)
 			if err != nil {
 				return nil, core.InvalidInputError("controller entry (%s) could not be parsed: %w", c, err)
 			}
-			options.Controllers = append(options.Controllers, *id)
+			options.MethodSpecificOptions = append(options.MethodSpecificOptions, didnuts.Controllers(*id))
 		}
 	}
 	options.KeyFlags = request.Body.VerificationMethodRelationship.ToFlags(options.KeyFlags)
 	if request.Body.SelfControl != nil {
-		options.SelfControl = *request.Body.SelfControl
+		options.MethodSpecificOptions = append(options.MethodSpecificOptions, didnuts.SelfControl(*request.Body.SelfControl))
 	}
 
 	doc, _, err := a.VDR.Create(ctx, options)
--- a/vdr/api/v1/api.go
+++ b/vdr/api/v1/api.go
@@ -90,7 +90,7 @@ func (a *Wrapper) AddNewVerificationMethod(ctx context.Context, request AddNewVe
 		opts = &VerificationMethodRelationship{}
 	}
 
-	vm, err := a.DocManipulator.AddVerificationMethod(ctx, *d, opts.ToFlags(didnuts.DefaultCreationOptions().KeyFlags))
+	vm, err := a.DocManipulator.AddVerificationMethod(ctx, *d, opts.ToFlags(didnuts.DefaultKeyFlags()))
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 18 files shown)


vdr/cmd/cmd.go
```diff
@@ -117,7 +117,7 @@ func createCmd() *cobra.Command {
 	result.Flags().BoolVar(createRequest.CapabilityDelegation, "capabilityDelegation", defs.KeyFlags.Is(management.CapabilityDelegationUsage), setUsage(defs.KeyFlags.Is(management.CapabilityDelegationUsage), "Pass '%t' to %s capabilityDelegation capabilities."))
 	result.Flags().BoolVar(createRequest.CapabilityInvocation, "capabilityInvocation", defs.KeyFlags.Is(management.CapabilityInvocationUsage), setUsage(defs.KeyFlags.Is(management.CapabilityInvocationUsage), "Pass '%t' to %s capabilityInvocation capabilities."))
 	result.Flags().BoolVar(createRequest.KeyAgreement, "keyAgreement", defs.KeyFlags.Is(management.KeyAgreementUsage), setUsage(defs.KeyFlags.Is(management.KeyAgreementUsage), "Pass '%t' to %s keyAgreement capabilities."))
-	result.Flags().BoolVar(createRequest.SelfControl, "selfControl", defs.SelfControl, setUsage(defs.SelfControl, "Pass '%t' to %s DID Document control."))
+	result.Flags().BoolVar(createRequest.SelfControl, "selfControl", true, setUsage(true, "Pass '%t' to %s DID Document control."))
```
gerardsn, line 120: not sure if this is correct, but requires regen docs if it is?

vdr/api/v1/api.go
```diff
@@ -122,12 +122,12 @@ func (a *Wrapper) CreateDID(ctx context.Context, request CreateDIDRequestObject)
 			if err != nil {
 				return nil, core.InvalidInputError("controller entry (%s) could not be parsed: %w", c, err)
 			}
-			options.Controllers = append(options.Controllers, *id)
+			options.AddOption(didnuts.Controllers(*id))
```
gerardsn, line 125: this adds a controllers option per controller instead of one option for all controllers. It works, but cleaner to add the option outside the loop imo

vdr/didnuts/creator.go
```diff
@@ -77,13 +77,31 @@ type Creator struct {
 // DefaultCreationOptions returns the default DIDCreationOptions when creating DID Documents.
 func DefaultCreationOptions() management.DIDCreationOptions {
 	return management.DIDCreationOptions{
-		Method:      MethodName,
-		Controllers: []did.DID{},
-		KeyFlags:    management.AssertionMethodUsage | management.CapabilityInvocationUsage | management.KeyAgreementUsage,
-		SelfControl: true,
+		Method:                MethodName,
+		KeyFlags:              management.AssertionMethodUsage | management.CapabilityInvocationUsage | management.KeyAgreementUsage,
+		MethodSpecificOptions: []management.DIDCreationOption{SelfControl(true)},
```
gerardsn, line 82: if the API receives `SelfControl`=false, the `DIDCreateOptions` will contain 2 conflicting values. It works because it applies the last one, but it is complicated to understand. An options map is clearer, or the documentation on the usage of `DIDCreateOptions`/`MethodSpecificOptions` should be very clear.

gerardsn (changes requested):

vdr/management/management.go
```diff
@@ -82,16 +82,20 @@ type DIDCreationOptions struct {
...
-	// SelfControl indicates whether the generated DID Document can be altered with its own capabilityInvocation key.
-	// Defaults to true when not given.
-	SelfControl bool
+	// MethodSpecificOptions contains creation options specific to the DID method.
+	MethodSpecificOptions []DIDCreationOption
+}
+
+// AddOption adds a method-specific DID creation option.
+func (o *DIDCreationOptions) AddOption(opt DIDCreationOption) *DIDCreationOptions {
+	o.MethodSpecificOptions = append(o.MethodSpecificOptions, opt)
```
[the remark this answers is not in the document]
reinkrul: I don't think "self control" is an option/concept from the DID core spec? "Controllers" yes, but we don't support it in did:web. So no need to provide it generically.

vdr/didnuts/creator.go
```diff
@@ -77,13 +77,31 @@ type Creator struct {
...
+
+type controllersCreationOption struct {
+	controllers []did.DID
+}
+
+// Controllers is a DID document creation option that indicates which DIDs can control the generated DID Document.
+// If selfControl = true and controllers is not empty, the newly generated DID will be added to the list of controllers.
+func Controllers(controllers ...did.DID) management.DIDCreationOption {
+	return controllersCreationOption{controllers: controllers}
+}
```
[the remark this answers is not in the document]
reinkrul: in our implementation, they are

vdr/cmd/cmd.go
reinkrul: defaults haven't changed, but the field isn't a struct property anymore

gerardsn (approved):

vdr/didnuts/creator.go
reinkrul: Fixed this, last option prevales

Issue #2755, opened 2024-01-30
An access_token request does not work for VCs with jwt_vc_json because https://identity.foundation/claim-format-registry/ expects jwt_vc.


rolandgroen:
The method RequestRFC021AccessToken filters the formatCandidates on both the metadata settings from the presentationDefinition.Format and the oauth server metadata. 

The VP metadata returns this:

```
func DefaultOpenIDSupportedFormats() map[string]map[string][]string {
	return map[string]map[string][]string{
		"jwt_vp_json": {"alg_values_supported": algValuesSupported},
		"jwt_vc_json": {"alg_values_supported": algValuesSupported},
		"ldp_vc":      {"proof_type_values_supported": proofTypeValuesSupported},
		"ldp_vp":      {"proof_type_values_supported": proofTypeValuesSupported},
	}
}
```

The presentationDefinition (note that https://identity.foundation/claim-format-registry/ does not allow `jwt_vp_json` and `jwt_vc_json`):

```   "format": {
      "ldp_vc": {
        "proof_type": [
          "JsonWebSignature2020"
        ]
      },
      "ldp_vp": {
        "proof_type": [
          "JsonWebSignature2020"
        ]
      },
      ...
    }
```

The VC that sits in the wallet is of type `jwt_vp_json`.

Problem: the VC of type `jwt_vc_json` does not get evaluated as candidate.
Expected: the VC of type `jwt_vc_json`  should be evaluated as candidate.

rolandgroen:
Created PR: https://github.com/nuts-foundation/nuts-node/pull/2756

gerardsn:
fixed by #2892

PR #2756, opened 2024-01-30
Issue #2755 An access_token request does not work for VCs with jwt_vc_json because https://identity.foundation/claim-format-registry/ expects jwt_vc.

rolandgroen:
Issue #2755 An access_token request does not work for VCs with jwt_vc_json because https://identity.foundation/claim-format-registry/ expects jwt_vc.

- Add format aliases to credential formats. Updated the Formats struct and associated methods in the vcr/credential/formats.go file to include the handling of format aliases. This provides additional flexibility in mapping equivalent format names. Also created a new method "normalizeFormat" to facilitate the normalization of these format aliases.

### DIFF
--- a/vcr/credential/formats.go
+++ b/vcr/credential/formats.go
@@ -26,6 +26,10 @@ func DIFClaimFormats(formats map[string]map[string][]string) Formats {
 		ParamAliases: map[string]string{
 			// no aliases for this type
 		},
+		FormatAliases: map[string]string{
+			"jwt_vp_json": "jwt_vp",
+			"jwt_vc_json": "jwt_vc",
+		},
 	}
 }
 
@@ -43,8 +47,9 @@ func OpenIDSupportedFormats(formats map[string]map[string][]string) Formats {
 // Formats is a map of supported formats and their parameters according to https://identity.foundation/claim-format-registry/
 // E.g., ldp_vp: {proof_type: [Ed25519Signature2018, JsonWebSignature2020]}
 type Formats struct {
-	Map          map[string]map[string][]string
-	ParamAliases map[string]string
+	Map           map[string]map[string][]string
+	ParamAliases  map[string]string
+	FormatAliases map[string]string
 }
 
 // Match takes the other supports formats and returns the formats that are supported by both sets.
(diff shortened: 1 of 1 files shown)


reinkrul:
When did you encounter this issue?

rolandgroen:
To clarify, issue https://github.com/nuts-foundation/nuts-node/issues/2755 happens when the presentationDefinition in the RFC021 procedure contains type definitions of type `ldp_vc` or `ldp_vp` for the VPs or VCs that are being requested. This is enforced by the json-ld at https://identity.foundation/claim-format-registry/schemas/presentation-definition-claim-format-designations.json. In this case the nuts-node fails to select the correct VCs from the wallet. The reason is that ldp_[vc|vp]\_json and ldp_[vc|vp]\_vc are synonyms, and the match code does not handle that correctly.

woutslakhorst:
@rolandgroen this is one of those things that will most likely be solved with future updates of the involved specs. Is it still valuable to implement this bypass or should we wait for updated specs?

vcr/credential/formats.go
```diff
@@ -43,22 +47,29 @@ func OpenIDSupportedFormats(formats map[string]map[string][]string) Formats {
 // Formats is a map of supported formats and their parameters according to https://identity.foundation/claim-format-registry/
 // E.g., ldp_vp: {proof_type: [Ed25519Signature2018, JsonWebSignature2020]}
 type Formats struct {
-	Map          map[string]map[string][]string
-	ParamAliases map[string]string
+	Map           map[string]map[string][]string
+	ParamAliases  map[string]string
+	FormatAliases map[string]string
```
reinkrul, line 52:
```suggestion
	// FormatAliases allows aliasing the VP and VC formats. This feature can be removed when https://identity.foundation/claim-format-registry/ and the OpenID4VC specifications have agreed on the format designators.
	FormatAliases map[string]string
```

reinkrul (approved):
We discussed that we can add this for now, we can remove it again when the standards have converged.

rolandgroen:
Closing in favor of PR: https://github.com/nuts-foundation/nuts-node/pull/2892

PR #2759, opened 2024-01-31
Add Delete function to vcr wallet

rolandgroen:
A new Delete function is added to the wallet in the vcr/holder/wallet.go file. This function allows a credential to be removed from the wallet. Additionally, an interface in the vcr/holder/interface.go file has been updated to include this Delete function. Proper error handling and stats consistency are ensured in this implementation.

### DIFF
--- a/vcr/holder/wallet.go
+++ b/vcr/holder/wallet.go
@@ -275,6 +275,28 @@ func (h wallet) IsEmpty() (bool, error) {
 	return count == 0, err
 }
 
+func (h wallet) Delete(ctx context.Context, subjectDID did.DID, id ssi.URI) error {
+	err := h.walletStore.Write(ctx, func(tx stoabs.WriteTx) error {
+		stats := tx.GetShelfWriter(statsShelf)
+		walletKey := stoabs.BytesKey(id.String())
+		walletShelf := tx.GetShelfWriter(subjectDID.String())
+		err := walletShelf.Delete(walletKey)
+		if err != nil {
+			return err
+		}
+		// Update stats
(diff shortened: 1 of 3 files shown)


vcr/holder/wallet.go
```diff
@@ -275,6 +275,28 @@ func (h wallet) IsEmpty() (bool, error) {
...
+		err := walletShelf.Delete(walletKey)
+		if err != nil {
+			return err
+		}
+		// Update stats
+		currentCount, err := h.readCredentialCount(stats)
+		if err != nil {
+			return fmt.Errorf("unable to read wallet credential count: %w", err)
+		}
+		return stats.Put(credentialCountStatsKey, binary.BigEndian.AppendUint32([]byte{}, currentCount+1))
```
reinkrul, line 292: I think this should be -1?

vcr/holder/wallet.go
```diff
@@ -275,6 +275,28 @@ func (h wallet) IsEmpty() (bool, error) {
 	return count == 0, err
 }
 
+func (h wallet) Delete(ctx context.Context, subjectDID did.DID, id ssi.URI) error {
```
reinkrul, line 278: missing test

reinkrul (changes requested):
Who needs this function? Is it correct that the REST API does not support this operation?

rolandgroen:
I'm not sure, how do we delete VCs from the wallet? Am I missing something?

reinkrul:
I've recently changed the implementation of the wallet from BBolt to SQL, so I'll reimplement this feature one of these days.

https://github.com/nuts-foundation/nuts-node/issues/2820

PR #2762, opened 2024-02-01
Discovery: return no search results as empty array

reinkrul:
Instead of `null`


gerardsn (approved):
Q: Why is this a 200 with empty array over a 404? (offline is fine too)

reinkrul:
It's about expectations: when searching, you expect zero more results (find all entries that match this query). 404 not found is an error indication, when there is none.

When it was a GET/read request for a specific resource, a 404 would be in order ("give me this credential")

PR #2776, opened 2024-02-06
VDR: Allow deactivation for did:web, added support to v2 API

reinkrul:
Added VDRv2 API support for `Deactivate()`, changed v1 API to use the same function to reduce complexity (and as side effect `did:web` DIDs can be deactivated using v1 API).

### DIFF
--- a/vdr/didweb/manager.go
+++ b/vdr/didweb/manager.go
@@ -67,8 +67,9 @@ type Manager struct {
 	keyStore crypto.KeyStore
 }
 
-func (m Manager) Deactivate(_ context.Context, _ did.DID) error {
-	return errors.New("Deactivate() is not yet supported for did:web")
+func (m Manager) Deactivate(_ context.Context, subjectDID did.DID) error {
+	// TODO: Should we remove private keys as well?
+	return m.store.delete(subjectDID)
 }
 
 func (m Manager) RemoveVerificationMethod(ctx context.Context, id did.DID, keyID did.DIDURL) error {
--- a/vdr/management/management.go
+++ b/vdr/management/management.go
@@ -39,6 +39,9 @@ type DocumentManager interface {
 	DocumentOwner
 	resolver.DIDResolver
 
+	// Deactivate deactivates a DID document, making it unusable for future interactions.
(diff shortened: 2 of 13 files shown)


vdr/management/management.go
```diff
@@ -39,6 +39,9 @@ type DocumentManager interface {
 	DocumentOwner
 	resolver.DIDResolver
 
+	// Deactivate deactivates a DID document, making it unusable for future interactions.
+	Deactivate(ctx context.Context, id did.DID) error
+
```
gerardsn, line 44: `DocManager` only needs `AddVerificationMethod` and `RemoveVerificationMethod` before it completely implements  `DocManipulator`. Both sound like they should be added for did:web?

vdr/didweb/manager.go
```diff
@@ -67,8 +67,9 @@ type Manager struct {
 	keyStore crypto.KeyStore
 }
 
-func (m Manager) Deactivate(_ context.Context, _ did.DID) error {
-	return errors.New("Deactivate() is not yet supported for did:web")
+func (m Manager) Deactivate(_ context.Context, subjectDID did.DID) error {
+	// TODO: Should we remove private keys as well?
```
gerardsn, line 71: Are there any risks if we keep them?

woutslakhorst: it depends....
if keys are used for encryption then it becomes a choice between preventing stolen keys to be used to decrypt old data and allowing the owner to recover encrypted data at a later point in time.

Since we're only using keys for signing, I would say sanitise and delete, it's up to the secure storage to handle things like backups and recovery.

vdr/management/management.go
woutslakhorst: I wonder if 2 interfaces are needed?

vdr/didweb/manager.go
reinkrul: Actually had to implement this throughout the crypto layer...

woutslakhorst (approved):
LGTM,

it's a bit symptomatic that 30 files need to be altered to add 1 (simple) piece of functionality.

PR #2777, opened 2024-02-06
move ami client api related things to client package

woutslakhorst:
this is a refactor, no functionality has been harmed.

I moved OpenID4VP "service/client" related stuff from the auth/services package to the client/iam package. I removed the role based split of functionality to a component based split (no split) which reflects how the server API is structured. I also moved one of the BuildVP methods to the wallet (and thus removed part of it). The remaining auth/services contents is for the older auth/v1/api

I think it's a bit clearer with a bit less code.


auth/services/oauth/relying_party.go
```diff
@@ -188,115 +128,6 @@ func (s *relyingParty) RequestRFC003AccessToken(ctx context.Context, jwtGrantTok
...
-		formatCandidates = formatCandidates.Match(credential.OpenIDSupportedFormats(metadata.VPFormats))
-	}
-	if presentationDefinition.Format != nil {
-		formatCandidates = formatCandidates.Match(credential.DIFClaimFormats(*presentationDefinition.Format))
-	}
-	format := pe.ChooseVPFormat(formatCandidates.Map)
-	if format == "" {
-		return nil, errors.New("requester, verifier (authorization server metadata) and presentation definition don't share a supported VP format")
-	}
-	// TODO: format parameters (alg, proof_type, etc.) are ignored, but should be used in the actual signing
```
gerardsn, line 231: disappeared from `BuildSubmission`, has this been fixed?

vcr/holder/wallet.go
```diff
@@ -69,6 +74,65 @@ func New(
...
+
+	// todo: support multiple wallets
+	vp, err := h.BuildPresentation(ctx, signInstructions[0].VerifiableCredentials, PresentationOptions{
+		Format: format,
+		ProofOptions: proof.ProofOptions{
+			Created:   time.Now(),
+			Challenge: &params.Nonce,
+			Domain:    &params.Audience,
+			Expires:   &params.Expires,
+			Nonce:     &params.Nonce,
```
gerardsn, line 127: Add `params.Challenge` so you can set whatever is appropriate? Adding params.Nonce to both sounds like something that could break with other implementations.

vcr/holder/interface.go
```diff
@@ -47,6 +48,9 @@ type Wallet interface {
 	// If signerDID is not provided, it will be derived from the credentials credentialSubject.id fields. But only if all provided credentials have the same (singular) credentialSubject.id field.
 	BuildPresentation(ctx context.Context, credentials []vc.VerifiableCredential, options PresentationOptions, signerDID *did.DID, validateVC bool) (*vc.VerifiablePresentation, error)
 
+	// BuildSubmission builds a Verifiable Presentation based on the given presentation definition.
+	BuildSubmission(ctx context.Context, walletDID did.DID, presentationDefinition pe.PresentationDefinition, acceptedFormats map[string]map[string][]string, params BuildParams) (*vc.VerifiablePresentation, *pe.PresentationSubmission, error)
```
gerardsn, line 52: should we add a `type Formats map[string]map[string][]string` in the `pe` package? This is an incomprehensible data format that is used 30+ times of which half in non-test code.

gerardsn (changes requested):

auth/client/iam/openid4vp.go
```diff
@@ -0,0 +1,228 @@
...
+
+// NewClient returns an implementation of Holder
+func NewClient(wallet holder.Wallet, strictMode bool, httpClientTimeout time.Duration, httpClientTLS *tls.Config) *OpenID4VPClient {
+	return &OpenID4VPClient{
+		httpClientTimeout: httpClientTimeout,
+		httpClientTLS:     httpClientTLS,
+		strictMode:        strictMode,
+		wallet:            wallet,
+	}
+}
```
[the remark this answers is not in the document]
woutslakhorst: to prevent direct access to the members, also to signal it's a pointer return.

vcr/holder/interface.go
woutslakhorst: correct, but let's do that separately to prevent merge conflicts.

auth/client/iam/openid4vp.go
```diff
@@ -0,0 +1,228 @@
...
+
+	// get the presentation definition from the verifier
+	parsedURL, err := core.ParsePublicURL(metadata.PresentationDefinitionEndpoint, c.strictMode)
+	if err != nil {
+		return nil, err
+	}
+	presentationDefinitionURL := http.AddQueryParams(*parsedURL, map[string]string{
+		"scope": scopes,
+	})
+	presentationDefinition, err := c.PresentationDefinition(ctx, presentationDefinitionURL)
```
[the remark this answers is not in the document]
woutslakhorst: by calling the local and not the on eon the http client, the error is wrapped.

auth/api/iam/openid4vp.go
```diff
@@ -210,42 +209,52 @@ func (r Wrapper) handleAuthorizeRequestFromVerifier(ctx context.Context, walletD
...
-	metadata, err := r.auth.Holder().ClientMetadata(ctx, clientMetadataURI)
+	metadata, err := r.auth.IAMClient().ClientMetadata(ctx, clientMetadataURI)
 	if err != nil {
-		return r.sendAndHandleDirectPostError(ctx, oauth.OAuth2Error{Code: oauth.ServerError, Description: "failed to get client metadata (verifier)"}, walletDID, responseURI, state)
+		return r.sendAndHandleDirectPostError(ctx, oauth.OAuth2Error{Code: oauth.ServerError, Description: "failed to get client metadata (verifier)"}, responseURI, state)
 	}
 	// get presentation_definition from presentation_definition_uri
 	presentationDefinitionURI := params[presentationDefUriParam]
-	presentationDefinition, err := r.auth.Holder().PresentationDefinition(ctx, presentationDefinitionURI)
+	parsedURL, err := url.Parse(presentationDefinitionURI)
```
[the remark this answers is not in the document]
woutslakhorst: Reverted signature back to string from url.URL, with the additional URL parsing there's little need to pass a url.URL.

auth/services/oauth/relying_party.go
woutslakhorst: no added todo at the right location.

vcr/holder/wallet.go
woutslakhorst: I'd rather keep it the same for now and reevaluate with the updated specs for proofs later this year

auth/client/iam/openid4vp.go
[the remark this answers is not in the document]
gerardsn: nvm, fields are unexported

gerardsn (approved):

PR #2780, opened 2024-02-07
VDR: Consistent capital usage in API

reinkrul:


auth/api/iam/api.go
```diff
@@ -446,14 +446,14 @@ func (r Wrapper) getWalletDID(ctx context.Context, didString string) (*did.DID,
...
-		return nil, core.NotFoundError("did not found: %w", err)
+		return nil, core.NotFoundError("DID not found: %w", err)
 	}
 	isWallet, err := r.vdr.IsOwner(ctx, *requestHolder)
 	if err != nil {
 		return nil, err
 	}
 	if !isWallet {
-		return nil, core.InvalidInputError("did not owned by this node")
+		return nil, core.InvalidInputError("DID not owned by this node")
```
gerardsn, line 456: no change in tests required means this is untested?

gerardsn (approved):

auth/api/iam/api.go
reinkrul: There is, but the assertion was weird. Fixed it.

PR #2783, opened 2024-02-07
Add JAR/RFC9101 to authorization requests

woutslakhorst:
This adds auth request signing (JAR, RFC9101).
Both the OAuth2 and OpenID4VP request will use JAR if the server metadata requires it. Some logic has been altered to make sure the handling can use both form params as JWT claims. Making the request happens from the iam/client package. This is a bit weird for the openid4vp auth request, but extracting logic and encapsulating it in another place is a bit overkill.

A non signed request is currently allowed although our server metadata always sets the `RequestAuthorizationRequestSigning` value to true. This is to prevent braking stuff for some of the other flows. I'll make an issue after 👍 s.

- [x] waiting for #2777
- [x] fix e2e test
- [x] coverage


auth/api/iam/api.go
```diff
@@ -297,10 +299,50 @@ func (r Wrapper) HandleAuthorizeRequest(ctx context.Context, request HandleAutho
 	// Workaround: deepmap codegen doesn't support dynamic query parameters.
 	//             See https://github.com/deepmap/oapi-codegen/issues/1129
 	httpRequest := ctx.Value(httpRequestContextKey).(*http.Request)
-	params := make(map[string]string)
+	params := make(map[string]interface{})
 	for key, value := range httpRequest.URL.Query() {
-		params[key] = value[0]
+		if len(value) == 1 {
```
reinkrul, line 304: why the change? can parameters occur multiple times? Note that `createSession` casts everything to string again

auth/api/iam/openid4vp.go
```diff
@@ -81,8 +87,33 @@ func (r Wrapper) handleAuthorizeRequestFromHolder(ctx context.Context, verifier
...
+	for _, param := range []string{oauth.ClientIDParam, oauth.ScopeParam, oauth.StateParam} {
+		if _, ok := params[param].(string); !ok {
+			return nil, withCallbackURI(oauthError(oauth.InvalidRequest, fmt.Sprintf("missing %s parameter", param)), redirectURL)
+		}
+	}
+
+	// additional JAR checks
+	// check if the audience is the verifier
+	audience, ok := params[jwt.AudienceKey].([]string)
+	if !ok || len(audience) == 0 || audience[0] != verifier.String() {
```
reinkrul, line 100: if we only check the first entry, maybe assure that the length is exactly 1, instead of not empty? Then at least it breaks whne behavior becomes unspecified.

auth/api/iam/api.go
```diff
@@ -297,10 +299,50 @@ func (r Wrapper) HandleAuthorizeRequest(ctx context.Context, request HandleAutho
...
+		} else {
+			params[key] = value
+		}
 	}
+
+	// if the request param is present, JAR (RFC9101, JWT Authorization Request) is used
+	// we parse the request and check the client_id param
+	var signerKid string
+	if rawToken, ok := params[oauth.RequestParam].(string); ok {
+		// Parse and validate the JWT
```
reinkrul, line 315: function becomes quite long and does too much (request parsing, switching to the right handler)

auth/api/iam/openid4vp.go
```diff
@@ -81,8 +87,33 @@ func (r Wrapper) handleAuthorizeRequestFromHolder(ctx context.Context, verifier
...
+		return nil, withCallbackURI(oauthError(oauth.InvalidRequest, fmt.Sprintf("invalid audience, verifier = %s, audience = %s", verifier.String(), audience)), redirectURL)
+	}
+	// check if issuer of the JWT equals client_id
+	clientIDClaim := params[oauth.ClientIDParam].(string)
+	issuerClaim, ok := params[jwt.IssuerKey].(string)
+	if !ok || issuerClaim != clientIDClaim {
+		return nil, withCallbackURI(oauthError(oauth.InvalidRequest, "invalid issuer claim"), redirectURL)
```
reinkrul, line 107: claim/parameter, confusing

auth/api/iam/openid4vp.go
```diff
@@ -193,44 +217,53 @@ func (r Wrapper) handleAuthorizeRequestFromHolder(ctx context.Context, verifier
...
 		return nil, oauth.OAuth2Error{Code: oauth.InvalidRequest, Description: "missing response_uri parameter"}
 	}
 	// we now have a valid responseURI, if we also have a clientState then the verifier can also redirect back to the original caller using its client state
-	state := params[oauth.StateParam]
+	state, ok := params[oauth.StateParam].(string)
+	if !ok {
+		return nil, oauth.OAuth2Error{Code: oauth.InvalidRequest, Description: "missing state parameter"}
+	}
+	// do a quick check for any missing string values where a conversion would blow up
+	for _, param := range []string{clientMetadataURIParam, presentationDefUriParam, oauth.ClientIDParam} {
```
reinkrul, line 237: repeated function, make a util func

auth/api/iam/openid4vp.go
```diff
@@ -193,44 +217,53 @@ func (r Wrapper) handleAuthorizeRequestFromHolder(ctx context.Context, verifier
...
-	verifierID := params[oauth.ClientIDParam]
+	verifierID := params[oauth.ClientIDParam].(string)
 	// the verifier must be a did:web
 	verifierDID, err := did.ParseDID(verifierID)
 	if err != nil || verifierDID.Method != "web" {
 		return r.sendAndHandleDirectPostError(ctx, oauth.OAuth2Error{Code: oauth.InvalidRequest, Description: "invalid client_id parameter (only did:web is supported)"}, responseURI, state)
 	}
 
-	nonce, ok := params[nonceParam]
+	nonce, ok := params[nonceParam].(string)
```
reinkrul, line 255: with so many string casts, better take another approach. Maybe make params a type with a `String(key string) string` function that returns the param as string, if we really need it to be `map[string]interface{}`

auth/api/iam/openid4vp.go
```diff
@@ -832,18 +865,20 @@ func (r Wrapper) oauthNonceStore() storage.SessionStore {
 	return r.storageEngine.GetSessionDatabase().GetStore(oAuthFlowTimeout, oauthNonceKey...)
 }
 
-func assertParamPresent(params map[string]string, param ...string) error {
+func assertParamPresent(params map[string]interface{}, param ...string) error {
```
reinkrul, line 868: oh here we have it, so just use this one instead of the new `for` range loops?

auth/client/iam/openid4vp.go
```diff
@@ -149,34 +154,63 @@ func (c *OpenID4VPClient) AccessToken(ctx context.Context, code string, verifier
...
+			oauth.ClientIDParam: client.String(),
+			oauth.RequestParam:  token,
+		})
+		return &redirectURL, nil
+	}
+	// else return an unsigned regular authorization request
+	// left here for completeness, node 2 node interaction always uses JAR since the AS metadata has it hardcoded
+
+	// additional claims can be added by the caller
+	modifier(params)
```
reinkrul, line 208: modifier must be there, or can it be nil?

What I tend to do it make modifier varargs slice, then loop over it. Then the caller can either provide none, 1 or multipl.

auth/client/iam/openid4vp.go
```diff
@@ -149,34 +154,63 @@ func (c *OpenID4VPClient) AccessToken(ctx context.Context, code string, verifier
...
 	endpoint, err := url.Parse(metadata.AuthorizationEndpoint)
 	if err != nil {
 		return nil, fmt.Errorf("failed to parse authorization endpoint URL: %w", err)
 	}
-	// construct callback URL for wallet
-	callbackURL, err := didweb.DIDToURL(requestHolder)
-	if err != nil {
-		return nil, fmt.Errorf("failed to create callback URL: %w", err)
+	// one default param for both signed and unsigned
+	params := map[string]interface{}{
```
reinkrul, line 173: why not `map[string]string`? OAuth2 parameters are always strings (as they are encoded in URL query parameters), so no need to support anything else?

auth/oauth/types.go
```diff
@@ -149,6 +157,9 @@ type AuthorizationServerMetadata struct {
 	// If omitted, the default value is true. (hence pointer, or add custom unmarshalling)
 	PresentationDefinitionUriSupported *bool `json:"presentation_definition_uri_supported,omitempty"`
 
+	// RequireSignedRequestObject specifies if the authorization server requires the use of signed request objects.
+	RequireSignedRequestObject bool `json:"require_signed_request_object,omitempty"`
```
reinkrul, line 161: I believe Sphereon does not use this (https://github.com/search?q=repo%3ASphereon-Opensource%2FSSI-SDK%20require_signed_request_object&type=code).

Are there instances in which we know for sure we have to use it?

reinkrul (changes requested):

auth/client/iam/openid4vp.go
woutslakhorst: JWT claims are not exclusively strings. A JAR request is encoded as JWT and does include string arrays as values.

auth/client/iam/openid4vp.go
woutslakhorst: The instances where it is used always require a minimum of 4 options or so, never 0. A single modifier function is less code than options through var args.

auth/oauth/types.go
woutslakhorst: When using DIDs, it's required. (at least to get an acceptable level of security)

auth/api/iam/api.go
```diff
@@ -297,10 +299,50 @@ func (r Wrapper) HandleAuthorizeRequest(ctx context.Context, request HandleAutho
...
+		token, err := crypto.ParseJWT(rawToken, func(kid string) (crypto2.PublicKey, error) {
+			signerKid = kid
+			return resolver.DIDKeyResolver{Resolver: r.vdr}.ResolveKeyByID(kid, nil, resolver.AssertionMethod)
+		}, jwt.WithValidate(true))
+		if err != nil {
+			return nil, oauth.OAuth2Error{Code: oauth.InvalidRequest, Description: "invalid request parameter", InternalError: err}
```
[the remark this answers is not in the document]
woutslakhorst: >  Let's make sure we have a clear error message also pointing to the fact we use JAR/it must be a JWT signed request.
The fact that it says `invalid request parameter` points to that fact. `request` is the param for JAR. Also our specs should say that JAR is required. I don't think we can expose any details about the error due to security reasons.

auth/api/iam/api.go
woutslakhorst: obsolete due to addition of `singleStringParam`

auth/api/iam/openid4vp.go
woutslakhorst: fixed with  `singleStringParam`

auth/api/iam/openid4vp.go
woutslakhorst: fixed with  `singleStringParam`

auth/api/iam/openid4vp.go
woutslakhorst: fixed with  `singleStringParam`

auth/api/iam/openid4vp.go
woutslakhorst: fixed with  `singleStringParam`

auth/client/iam/openid4vp.go
woutslakhorst: fixed with  `singleStringParam`

auth/api/iam/openid4vp.go
woutslakhorst: has been removed since check is done one level higher up

auth/oauth/types.go
reinkrul: Ah and if it's not specified in the metadata, it defaults to false, which is fine then

reinkrul (approved):

PR #2738, opened 2024-01-19
StatusList: issuance

gerardsn:
closes #2690 

this PR adds
- option to add credentialStatus entry to a VC during issuance
- revoking credentials with a credentialStatus
- fetching StatusList2021Credentials for issuers that have issued credentials with a credentialStatus
`example.com/iam/123/statuslist/1`
- in memory store that will be replaced by a sql store in a follow-up PR

PR #2784, opened 2024-02-07
StatusList: replace memstore with sqlstore

### LINKED ISSUE #2785: Add storage for status list credentials

gerardsn:
builds on #2738 that needs to be merged first
closes #2785 

should we add a timestamp when a statuslist entry is set (credential is revoked)?


vcr/statuslist/store.go
```diff
@@ -0,0 +1,287 @@
...
+	return "status_list_statuses"
+}
+
+// revocationRecord is created when a
+type revocationRecord struct {
+	StatusListCredential string `gorm:"primaryKey"`                     // https://example.com/iam/id/status/page
+	StatusListIndex      int    `gorm:"primaryKey;autoIncrement:false"` // 0 <= statusListIndex <= statuslist2021.MaxBitstringIndex
+	CredentialID         string // did:web:example.com:iam:id#unique-identifier
+	// TODO: do we want to store the timestamp the revocation was added?
+	//CreatedAt time.Time
```
reinkrul, line 72: I'd say yes, maybe for informational purposes later on

vcr/statuslist/store.go
```diff
@@ -0,0 +1,287 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	// confirm that status list credential exists
+	// SELECT * FROM status_list_credentials WHERE status_list_credential = 'statusListCredential'
+	err = s.DB(ctx).First(new(statusListCredentialRecord), "status_list_credential = ?", statusListCredential).Error
+	if err != nil {
+		if errors.Is(err, gorm.ErrRecordNotFound) {
+			// status list credential does not exist or is not managed byt his node
```
reinkrul, line 101:
```suggestion
			// status list credential does not exist or is not managed by this node
```

vcr/statuslist/store.go
```diff
@@ -0,0 +1,287 @@
...
+			return nil, errNotFound
+		}
+		return nil, err
+	}
+
+	// get revocations
+	var revocations []revocationRecord
+	// SELECT status_list_index FROM status_list_statuses WHERE status_list_credential = 'statusListCredential'
+	err = s.DB(ctx).Select("status_list_index").Find(&revocations, "status_list_credential = ?", statusListCredential).Error
+	if err != nil && errors.Is(err, gorm.ErrRecordNotFound) {
```
reinkrul, line 111: I think this should be `!errors.Is(not found)` instead?

vcr/statuslist/store.go
```diff
@@ -0,0 +1,287 @@
...
+			// status list credential does not exist or is not managed byt his node
+			return nil, errNotFound
+		}
+		return nil, err
+	}
+
+	// get revocations
+	var revocations []revocationRecord
+	// SELECT status_list_index FROM status_list_statuses WHERE status_list_credential = 'statusListCredential'
+	err = s.DB(ctx).Select("status_list_index").Find(&revocations, "status_list_credential = ?", statusListCredential).Error
```
reinkrul, line 110: If you model the relationship in the `statusListCredentialRecord` and `revocationRecord ` structs with `gorm` tags, you can qurey them in 1 go with `Preload()`.

vcr/statuslist/store.go
```diff
@@ -0,0 +1,287 @@
...
+func toStatusListCredential(issuer did.DID, page int) (string, error) {
+	switch issuer.Method {
+	case "web":
+		issuerAsURL, err := didweb.DIDToURL(issuer) // https://example.com/iam/id
+		if err != nil {
+			return "", err
+		}
+		return issuerAsURL.JoinPath("statuslist", strconv.Itoa(page)).String(), nil // https://example.com/iam/id/statuslist/page
+	}
+	return "", fmt.Errorf("status list: unsupported DID method: %s", issuer.Method)
```
reinkrul, line 286: `DIDToURL` also returns this error, so you can skip the switch statement altogether

vcr/statuslist/store.go
```diff
@@ -0,0 +1,287 @@
...
+
+			// first time issuer; prepare to create a new Page / StatusListCredential
+			credentialRecord = statusListCredentialRecord{
+				Issuer:          issuer.String(),
+				LastIssuedIndex: statuslist2021.MaxBitstringIndex,
+			}
+		}
+
+		// next index
+		credentialRecord.LastIssuedIndex++
```
reinkrul, line 163: base 1 or 0? Because now the first index of the first page is 1 instead of 0?

Edit: oh now I see, you initialize `LastIssuedIndex: statuslist2021.MaxBitstringIndex,`, so it gets recognized as "full" page, so a new one gets created. THat's quite confusing?

vcr/statuslist/store_test.go
```diff
@@ -0,0 +1,256 @@
...
+	})
+	t.Run("optimistic locking", func(t *testing.T) {
+		t.Run("gorm translates errors", func(t *testing.T) {
+			// without this the lock conditions are not handled correctly
+			assert.True(t, s.db.Config.TranslateError)
+		})
+		t.Run("trigger gorm.ErrRecordNotFound on UPDATE of LastIssuedIndex", func(t *testing.T) {
+			t.SkipNow()
+			// This test is skipped since in needs manual evaluation.
+			// Run this test if the logic around optimistic locking has been changed.
```
reinkrul, line 116: You can evaluate this test automatically, if you query the database for `LastIssuedIndex`? The index should be `len(jobs) - 1`? If not, a race condition occured and the idnex wasn't updated properly.

For bonus points, lower the `MaxBitStringIndex` in the test so you can also automatically test a full page (without mkaing the test super long-running).

vcr/statuslist/store_test.go
```diff
@@ -0,0 +1,256 @@
...
+
+			worker := func(jobs <-chan int, results chan<- bool) {
+				for range jobs {
+					_, _ = s.Create(nil, aliceDID, StatusPurposeRevocation)
+					results <- true
+				}
+			}
+
+			const numWorkers = 10
+			const numJobs = 4000
```
reinkrul, line 137: this is a complicated way to implement a wait group?

reinkrul (changes requested):

vcr/statuslist/store_test.go
gerardsn: The test is aiming to confirm there were 2 parallel updates and that it's handled correctly. The test you propose does not guarantee the race condition occurs. (CI only has 1 core and might never trigger this situation)

storage/engine_test.go
```diff
@@ -143,6 +143,6 @@ func Test_engine_sqlDatabase(t *testing.T) {
 		require.NoError(t, row.Err())
 		var count int
 		assert.NoError(t, row.Scan(&count))
-		assert.Equal(t, 3, count)
+		assert.Equal(t, 4, count)
```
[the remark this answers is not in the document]
gerardsn: this builds on #2738 that needs to be rebased

storage/sql_migrations/004_statuslist.sql
```diff
@@ -0,0 +1,43 @@
+-- migrate:up
+
+-- status_list_credentials: keeps track of all status list credentials issued by an issuer, and the highest status list index issued for each credential.
+create table status_list_credentials
```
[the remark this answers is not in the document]
gerardsn: I used plural because its the convention used in GORM. I don't really care either way, but singular would create column `status_list_credential.status_list_credential` if that is fine?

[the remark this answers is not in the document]
reinkrul: I'd name the column `url`?

vcr/issuer/issuer_test.go
```diff
@@ -1016,3 +1018,11 @@ func TestIssuer_StatusList(t *testing.T) {
 		assert.Nil(t, result)
 	})
 }
+
+func NewTestStatusListStore(t testing.TB) statuslist.StatusList2021Issuer {
```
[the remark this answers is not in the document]
gerardsn: I only create one per test function and have no need to reset the store. A TEST util function to clear a SQL db in de storage module could be useful though

vcr/statuslist/store.go
gerardsn: `gorm.ErrRecordNotFound` is never returned for `Find`, removed code

storage/sql_migrations/004_statuslist.sql
```diff
@@ -0,0 +1,43 @@
...
+    status_list_credential varchar(500) not null primary key,
+    -- issuer: the DID of the issuer of this statusListCredential.
+    issuer varchar(500) not null,
+    -- page: the n-th StatusList2021Credential issued by this issuer.
+    --      Used to find the credential statusList2021Entries are being issued to. Could require an index in the future.
+    page integer not null,
+    -- last_issued_index: the highest status_list_index issued for this page. Should not be incremented above the max statusListIndex.
+    last_issued_index integer not null
+    -- Ties status list credentials to DID management.
+    --!! There should be a single table for all managed DIDs, not per DID method.
```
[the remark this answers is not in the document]
gerardsn: added. but now the test DIDs need to be written to `vdr_didweb` table to prevent foreign key constraint errors

storage/sql_migrations/004_statuslist.sql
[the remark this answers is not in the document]
gerardsn: I think the only usage of a non-index value is preceded by selection on primary key.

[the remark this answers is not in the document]
gerardsn: I was naming from the statusListEntry point of view, but viewing from the credential makes much more sense

vcr/statuslist/store.go
gerardsn: `DIDToURL` joins two errors resulting in `URL does not represent a Web DID\nunsupported DID method: nuts`. This doesn't make much sense to me; even less when I wrap it in `status list: %w`

We could probably use generic `unsupported DID method` and `not supported for did method` errors to prevent this

vcr/statuslist/store.go
gerardsn: other options I though of all resulted in more code / separate code path for a new issuer. I added a comment to explain he behavior.

storage/sql_migrations/004_statuslist.sql
```diff
@@ -0,0 +1,43 @@
...
+    constraint status_list_entry_id primary key (status_list_credential, status_list_index),
+    -- credential_id: references a credential in the issued credentials table, once it exists.
+    --!! Should this cascade? The credential revocation status is unchanged.
+    -- foreign key (credential_id) references issuer_store (id),
+    -- Ties the status_list_credential to an issuer (did) via the status_list_issuer table
+    constraint fk_status_list_credential foreign key (status_list_credential) references status_list_credentials (status_list_credential) on delete cascade
+);
+
+-- migrate:down
+drop table status_list_issuer;
```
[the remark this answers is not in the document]
gerardsn: hmm... sloppy

storage/sql_migrations/004_statuslist.sql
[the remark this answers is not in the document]
gerardsn: The issuer now produces errors because of this. I'm assuming we are going to have issues like this more often, what would be a good way to deal with this without turning everything into an integration test?

vcr/statuslist/store_test.go
gerardsn: Wait groups are used to ensure all routines are closed. 
Changing this to a wait group would fire a new go routine for every job, which is presumably slower than letting active routines pick new jobs as long as there are any?

storage/sql_migrations/004_statuslist.sql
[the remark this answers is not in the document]
reinkrul: When testing a SQL store implementation, it only makes sense to build it as an integration test i.m.o, otherwise you need to mock the SQL/GORM driver which overcomplicates things.

From tests point of view, we could load some default data into it when creating a "test" storage engine?

vcr/statuslist/store_test.go
gerardsn: The suggested test revealed that optimistic locking does not work. Simultaneous reads of the old value during update of last_issued_index results in the next index value being issue twice. Optimistic locking is replaced with SELECT FOR UPDATE and a mutex for cases where the db is SQLite.

storage/sql_migrations/004_statuslist.sql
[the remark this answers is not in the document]
gerardsn: added function `storage.AddDIDtoSQLDB` for now.

gerardsn:
@reinkrul I've added Testcontainers to run confirm the code is working properly with both Postgres and SQLite.

We should consider if we want this, because it adds a lot of dependencies and it looks like CI requirements may increase too. (failing test requires https://golang.testcontainers.org/system_requirements/ci/circle_ci/)

gerardsn:
testcontainers are removed (adds 100+ lines in go.mod/go.sum) and postgres will be tested in the e2e test.

vcr/issuer/issuer.go
```diff
@@ -223,7 +223,7 @@ func (i issuer) buildVC(ctx context.Context, template vc.VerifiableCredential, o
 		Issuer:            template.Issuer,
 		ExpirationDate:    template.ExpirationDate,
 		IssuanceDate:      template.IssuanceDate,
-		CredentialStatus:  template.CredentialStatus, // TODO: needs api update for this to work. do we allow external issuers to revoke this credential?
+		//CredentialStatus:  template.CredentialStatus, // not allowed for now since it requires API changes to be able to determine what status to revoke.
```
reinkrul, line 226: remove

vcr/statuslist/store.go
```diff
@@ -48,34 +50,35 @@ type StatusList2021Issuer interface {
...
 // statusListCredentialRecord keeps track of the StatusListCredential issued by an issuer, and what the LastIssuedIndex is for each credential.
 type statusListCredentialRecord struct {
-	StatusListCredential string `gorm:"primaryKey"`
-	Issuer               string // did:web:example.com:iam:id
-	Page                 int    // >= 1
-	LastIssuedIndex      int    // 0 <= statusListIndex < statuslist2021.MaxBitstringIndex
+	SubjectID       string             `gorm:"primaryKey"`
+	Issuer          string             // did:web:example.com:iam:id
+	Page            int                // >= 1
+	LastIssuedIndex int                // 0 <= statusListIndex < statuslist2021.MaxBitstringIndex
```
reinkrul, line 61: comments could be godoc

vcr/statuslist/store.go
```diff
@@ -48,34 +50,35 @@ type StatusList2021Issuer interface {
...
-// revocationRecord is created when a
+// revocationRecord is created when a statusList entry has been revoked.
 type revocationRecord struct {
 	StatusListCredential string `gorm:"primaryKey"`                     // https://example.com/iam/id/status/page
 	StatusListIndex      int    `gorm:"primaryKey;autoIncrement:false"` // 0 <= statusListIndex <= statuslist2021.MaxBitstringIndex
 	CredentialID         string // did:web:example.com:iam:id#unique-identifier
-	// TODO: do we want to store the timestamp the revocation was added?
-	//CreatedAt time.Time
+	RevokedAt            int64  `gorm:"autoCreateTime;column:created_at"`
 }
```
reinkrul, line 75: godoc

reinkrul (approved):

vcr/issuer/issuer.go
gerardsn: left there intentionally so we don't activate before making sure it can be used

PR #2795, opened 2024-02-14
Discovery: improve test stability

reinkrul:
Service Definitions is a map, which caused the call order of the mocked HTTP client to be undefined.

(I wonder why this didn't fail more often).


gerardsn (approved):
It was failing about 50% of the time for me. Was actually just looking into it.

PR #2804, opened 2024-02-16
Statuslist: move all code to statuslist package

gerardsn:
This PR just moves all code to one package for easy reviewing. Actual changes will be made in the next PR

### DIFF
--- a/storage/gorm_logger.go
+++ b/storage/gorm_logger.go
@@ -1,3 +1,21 @@
+/*
+ * Copyright (C) 2024 Nuts community
+ *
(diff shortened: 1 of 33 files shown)


reinkrul (commented):
Definitely an improvement I think?

Did you get a lot of "Type name shouldn't start with package name" warnings? Would it be better to call the package `statuslist2021` so you can shorten all type names? So;

- ContextURI
- Entry
- Credential
- etc

reinkrul (approved):

Issue #2798, opened 2024-02-14
Migrate credential wallet to SQL storage


reinkrul:
Starting v6 we support SQL storage. One of the stores we still need to migrate is the credential wallet (`VCR/Holder/wallet.go`).

On the `master` branch there is functionality to migrate credentials from VCR store to the current BBolt wallet, but since this feature isn't released yet we can simply change the store implementation to SQL.

gerardsn:
Since the wallets only apply to did:web, and no credentials have been issued for this yet, we can also remove the migration. So just change store implementation

reinkrul:
Requires https://github.com/nuts-foundation/nuts-node/issues/2799

PR #2806, opened 2024-02-19
VCR: Implement holder's wallet as SQL store instead of KV

reinkrul:
Fixes https://github.com/nuts-foundation/nuts-node/issues/2798

### DIFF
--- a/storage/sql_migrations/007_wallet.sql
+++ b/storage/sql_migrations/007_wallet.sql
@@ -1,0 +1,10 @@
+-- migrate:up
+create table wallet_credential
+(
+    holder_did    varchar(500) not null primary key,
+    credential_id varchar(500) not null primary key,
+    constraint fk_wallet_credential foreign key (credential_id) references credential (id)
+);
+
+-- migrate:down
+drop table wallet_credential;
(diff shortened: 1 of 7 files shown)


storage/sql_migrations/007_wallet.sql
```diff
@@ -0,0 +1,11 @@
+-- migrate:up
+create table wallet_credential
+(
+    holder_did    varchar(500) not null,
+    credential_id varchar(500) not null,
+    primary key (holder_did, credential_id),
+    constraint fk_wallet_credential foreign key (credential_id) references credential (id)
```
gerardsn, line 7: holder_did is a did that should be managed by this node. Should there be a foreign key on this too? Or why not?

vcr/holder/wallet.go
```diff
@@ -330,23 +275,64 @@ func (h wallet) Diagnostics() []core.DiagnosticResult {
...
+
+func (s walletStore) count() (int64, error) {
+	var count int64
+	err := s.db.Model(walletRecord{}).Count(&count).Error
+	return count, err
+}
+
+func (s walletStore) list(holderDID did.DID) ([]vc.VerifiableCredential, error) {
+	var records []walletRecord
+	err := s.db.Model(walletRecord{}).Preload("Credential").Where("holder_did = ?", holderDID.String()).Find(&records).Error
```
gerardsn, line 306: return error immediately

gerardsn (approved):

storage/sql_migrations/007_wallet.sql
reinkrul: Probably, yes. Unless we start issuing from did:jwk's in the future, but then we'll change it again.

PR #2807, opened 2024-02-19
Remove credential.validationError

gerardsn:
It's a weird thing and its only purpose apparently is to prepend `validation failed: ` to the error string you give it. Since it only takes a string, you also lose the other error types being wrapped.


reinkrul (approved):
It worked that way since `fmt.Errorf` only allowed 1 error to be wrapped, but it's nicer now

PR #2809, opened 2024-02-19
Move AllFieldsDefinedValidator to jsonld package

gerardsn:
needed in statuslist2021 package, but is a better fit in jsonld package anyway

### DIFF
--- a/jsonld/jsonld.go
+++ b/jsonld/jsonld.go
@@ -23,7 +23,6 @@ import (
 	"encoding/json"
 	"fmt"
 
-	"github.com/nuts-foundation/go-did/vc"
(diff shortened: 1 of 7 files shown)


reinkrul (changes requested):
I don't think it belongs here; when it takes a VerifiableCredential. VC functionality uses JSON-LD, not the other way around. You can probably fix it by simply changing the type of `input` to `interface{}`

gerardsn:
on second though. I added `string` because test data are strings, but probably not relevant in production code?

reinkrul (approved):
Probably fine, as long as it's tested.

PR #2814, opened 2024-02-19
Remove dead code

gerardsn:


reinkrul (approved):

reinkrul:
Nitpicking; "Dead code" is not the right term here (dead code is never hit), but it's still useless code.

PR #2815, opened 2024-02-20
Enable ES256K support with build flag

rolandgroen:
As part of the mitzXnuts PoC. For completeness.

This feature is toggled with the jwx_es256k flag.

The JWS signed by MS Authenticator uses/requires ES256K signed VCs. This patch enables the support for ES256K signed JWTs & VCs. Reject if undesired / unwanted.

Test included.


reinkrul (approved):

reinkrul:
Before I can merge your PR, your commits need to be signed. You can do this either by signing locally, or let GitHub sign for you (kinda weird, but it works).

rolandgroen:
Commits are signed now.

gerardsn (changes requested):
We first need to decide under what conditions we allow importing low-level crypto implementations and if this qualifies. Perhaps this should stay under a build flag or config option.

reinkrul:
> We first need to decide under what conditions we allow importing low-level crypto implementations and if this qualifies. Perhaps this should stay under a build flag or config option.

I think we can include it, but at the same time we want to allow limitation of supported algorithms/key types for 2 reasons:
- The ES256K implementation isn't from the Golang maintainers. We can't judge the quality and don't know who'll maintain it in the future.
- We decided on a governing-body list of approved ciphers/key types (Dutch NCSC)

So we should make it configurable i.m.o. I see 2 options:
- Define profiles that enable/disable certain ciphers and key types (e.g. `all` or `ncsc`). Like FIPS modes.
- Allow to explicitly enable/disable ciphers and key types like in `mod_ssl` (or is it NGINX?);
  - `+ES256K` enable es256k
  - `-rsa2045` disable
  - `all` enable everything

Regarding what should be enabled by default; the intersection of what is specified by NCSC and supported by Golang crypto std lib.

woutslakhorst:
In this particular case I would suggest putting it under a build flag and not include it in the default build. If you would do a (crypto) code review, it would be listed as an issue. It's not needed for production use cases, only experimental ones. Configuration is already possible through PEX.

gerardsn:
Supported types are listed in the oauth-server-metadata. The build flag should also add the type to the metadata.

reinkrul (approved):
Looks OK now, but we decided to not enable it by default, since it's not recommended by NCSC and our implementation is an unaudited, third party library. So the default Docker container now doesn't include it.

reinkrul:
@rolandgroen I merged the PR, but the default Docker file now does not include ES256K support. Are you building your own image, or do we need to provide an additional image with support "unsanctioned" (for lack of better term) cryptographic algorithms?

PR #2832, opened 2024-02-27
change did OAS param to use content instead of schema

### LINKED ISSUE #2831: OAPICodeGen runtime double unescapes path params
This relates to all paths where a did is used.

Workaround is to use the content tag in the spec instead of the schema type.

issue at codegen: oapi-codegen/runtime#35

woutslakhorst:
fixes #2831 

when using schema oapi codegen uses `BindStyledParamWithOptions` which does an url unescape. This is not needed since go already does this. When using port numbers in a DID this causes the `:` to be decoded (should remain percentage encoded)

note: oapi codegen also updated to 2.1.0


gerardsn (commented):
LGTM.
If I search for `{did}` in the `docs/_static` I find several other APIs. Should the same change be applied to (some of) these? (vdr/v2 and vcr/v2/holder are did:web specific)

gerardsn (approved):

gerardsn (changes requested):
`/internal/vcr/v2/holder/{did}/vc/{id}` has been added to vcr_v2.yaml on master since you started this PR.

gerardsn (approved):

PR #2838, opened 2024-02-27
Backport VCR JSON-LD context not allowed fix, release notes for v5.4.6

reinkrul:


gerardsn:
we have different branches for V5.{0,1,2,3,4} and a V5. 
I don't know how V5 is supposed to relate to the others, and I'm sure we will forget to keep V5 in sync with the relevant branch (V5 and V5.4 are currently already not in sync, hence the missing release note). I prefer to merge V5 into V5.4, merge this into V5.4, and delete V5.

reinkrul:
V5 and V5.4 branches are in sync, so no need to merge there. We have a major version branch (V5) as "working branch" for new releases (e.g. V5.5), but as we don't need it for V5 it might be better to remove it, to avoid confusion.

woutslakhorst:
> V5 and V5.4 branches are in sync, so no need to merge there. We have a major version branch (V5) as "working branch" for new releases (e.g. V5.5), but as we don't need it for V5 it might be better to remove it, to avoid confusion.

better to branch of v5.4 when development is done for v5.5

woutslakhorst:
@reinkrul since this PR would be a trigger for a new release, let's also check for dependency updates (patches)

woutslakhorst (approved):

gerardsn (approved):

PR #2839, opened 2024-02-27
Restrict SQLite driver to 1 DB connection

gerardsn:
by limiting SQLite to 1 connection there is no need to have write locks in code to simulate SELECT FOR UPDATE.

this results in a deadlock when a new transaction is started from within a transaction, but this will fail during development.


vcr/credential/store/sql_test.go
```diff
@@ -276,11 +276,11 @@ func TestCredentialStore_BuildSearchStatement(t *testing.T) {
 			setupStore(t, storageEngine.GetSQLDatabase())
 			for _, credential := range tc.inputVCs {
 				err := db.Transaction(func(tx *gorm.DB) error {
-					credentialRecord, err := store.Store(db, credential)
+					credentialRecord, err := store.Store(tx, credential)
```
reinkrul, line 279: oops

reinkrul (approved):
Sehr gut

PR #2840, opened 2024-02-28
allow for empty VPs to be build when no credentials are requested

### LINKED ISSUE #2800: Allow VP creation with empty set of credentials
finding from OpenID4VP implementation into demo.

When an empty presentation definition is given by the verifier, the wallet will currently answer with "no matching credentials". It could just create an empty vp.

requires changes in :
- auth/api/iam/validation.go (# 33)
- vcr/holder/wallet.go (# 117)
- vcr/pe/presentation_submission.go (# 20)

woutslakhorst:
closes #2800

needed changes to allow for empty VPs to be built when requested so by an empty presentation definition.
This makes testing without any credentials easier when still conforming to specs.

### DIFF
--- a/auth/api/iam/validation.go
+++ b/auth/api/iam/validation.go
@@ -33,6 +33,9 @@ import (
 // validatePresentationSigner checks if the presenter of the VP is the same as the subject of the VCs being presented.
 // All returned errors can be used as description in an OAuth2 error.
 func validatePresentationSigner(presentation vc.VerifiablePresentation, expectedCredentialSubjectDID did.DID) (*did.DID, error) {
+	if len(presentation.VerifiableCredential) == 0 {
+		return credential.PresentationSigner(presentation)
+	}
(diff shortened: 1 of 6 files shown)


gerardsn (approved):
This can now issue VPs that contain no credentials, audience, or challenge. The only restriction is expiration time, which can be set to 1000 year during development and then forgot about. This could result in a golden replayable VP, should this be allowed? Possible a security consideration worth tracking in an issue for now.

reinkrul:
> This can now issue VPs that contain no credentials, audience, or challenge. The only restriction is expiration time, which can be set to 1000 year during development and then forgot about. This could result in a golden replayable VP, should this be allowed? Possible a security consideration worth tracking in an issue for now.

VP proofs can contain a domain and/or challenge (JSON-LD), or audience and/or nonce (JWT). So that can still be used?

I don't think we need to limit the expiration time (not up to us to decide on that i.m.o.).

Issue #2868, opened 2024-03-06
VDR: Add APIv2 operation to resolve service endpoint


reinkrul:
Didman provides functions to resolve API endpoints (as is and an endpoint from a "compound service"). As Didman is largely `did:nuts` specific and was introduced to do heavy-lifting with updating complicated documents, it will be deprecated and removed when `did:nuts` is dropped.

But for integrators, it would be beneficial to have an API endpoint to resolve a service endpoint from a DID document. Otherwise, they'd have to parse JSON-LD, which we want to avoid due to its complexity.

API:

- `/internal/vdr/v2/{did}/service/{id}/endpoint` (resolves by service ID)
- `/internal/vdr/v2/{did}/service/endpoint?type={type}` (resolves by type)

It returns the endpoint as-is; no reference following or compound service parsing. New use cases using `did:web` don't need to support reference resolving, since document management is much simpler (contrary to `did:nuts`).

PR #2873, opened 2024-03-06
VDR: v2 API to resolve service endpoints

reinkrul:
Fixes https://github.com/nuts-foundation/nuts-node/issues/2868

Changed the solution to less convenience-function returning the endpoint, to a more flexible filter-style function that return a list of matching services. To aid the caller in parsing the service endpoint, I added the `endpointType` filter to control which service endpoint types (string, object, array) should be returned.


vdr/api/v2/api.go
```diff
@@ -166,12 +168,58 @@ func (w Wrapper) UpdateService(ctx context.Context, request UpdateServiceRequest
 	return UpdateService200JSONResponse(*newService), nil
 }
 
-func (w Wrapper) AddVerificationMethod(ctx context.Context, request AddVerificationMethodRequestObject) (AddVerificationMethodResponseObject, error) {
+func (w *Wrapper) ResolveServiceEndpointByType(_ context.Context, request ResolveServiceEndpointByTypeRequestObject) (ResolveServiceEndpointByTypeResponseObject, error) {
```
gerardsn, line 171: did spec allows multiple services with the same type, is this handled correctly everywhere (else)?

gerardsn (changes requested):
double checking: this still allows compound services, the node just doesn't unpack it for you anymore?

vdr/api/v2/api.go
reinkrul: We now have a search/filter-style operation that can return multiple services.

gerardsn (approved):
LGTM. This needs some examples for users to understand possible endpointType formats. I'm fine adding them now, or when improving the API docs in general.

PR #2878, opened 2024-03-06
set column length for did and id

### LINKED ISSUE #2858: wallet_credential primary key too long for mysql
it uses 2 columns of 500 chars each. UTFmb4 uses 4 bytes per character which results in 4000 bytes for the primary key, max is 3072.
`holder_did` is also specified as length 500 while other migrations use 255 for the did.

- holder_did length to 255?
- Drop the primary key

woutslakhorst:
fixes #2858 

fixed varchar length of 370 for did and 415 for id (did+identifier)

@reinkrul and @gerardsn could you check if this is ok for the tables you added?

### DIFF
--- a/storage/sql_migrations/001_credential.sql
+++ b/storage/sql_migrations/001_credential.sql
@@ -1,8 +1,8 @@
 -- migrate:up
 create table credential
 (
-    id         varchar(370) not null primary key,
-    issuer     varchar(415) not null,
+    id         varchar(415) not null primary key,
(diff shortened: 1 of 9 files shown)


reinkrul:
Can the document the "why" behind those magic numbers? Maybe in the SQL migration files, or a readme in the migrations directory?

reinkrul (approved):

gerardsn (approved):

Issue #2861, opened 2024-03-05
SQL: Rename `vdr_didweb` table to `did`


reinkrul:
We don't prefix with module names, and we might want to register other in it than `did:web`. At this point we can still change the name easily, with v6 it becomes somewhat harder (migration required).

PR #2882, opened 2024-03-07
SQL: Rename table vdr_didweb to did

reinkrul:
Fixes https://github.com/nuts-foundation/nuts-node/issues/2861


storage/test.go
```diff
@@ -25,7 +25,6 @@ import (
 	"testing"
 
 	"github.com/nuts-foundation/go-did/did"
-	"github.com/nuts-foundation/go-stoabs"
```
gerardsn, line 28: oops

gerardsn (changes requested):
the migration file names still contain `didweb`, and `did:web` occurs several times in the migration files comments

IMO merging migrations files 003 and 005 makes sense, but probably fine as is

PR #2886, opened 2024-03-07
Docs: update API Reference

### LINKED ISSUE #2863: Add VDR v2 OAS to API reference docs
only v1 is in the dropdown box.

gerardsn:
closes #2863 

https://nuts-node.readthedocs.io/en/api-update-rtd-2863/pages/integrating/api.html

specs are now order alphabetically with deprecated specs at the end. This conveniently places the new auth/v2 at the top of the list. 

I've removed the download links since I always click on the spec I am interested in before swagger ui is loaded. The files can still be downloaded from the individual spec references.

### DIFF
--- a/docs/pages/integrating/api.rst
+++ b/docs/pages/integrating/api.rst
@@ -3,22 +3,6 @@
 API Reference
 =============
 
-Below you can discover the Nuts Node APIs and download their OpenAPI specifications:
-
-- Common: `SSI types <../../_static/common/ssi_types.yaml>`_, `Default Error <../../_static/common/error_response.yaml>`_
-- `DID Manager <../../_static/didman/v1.yaml>`_
-- `Discovery Service <../../_static/discovery/v1.yaml>`_
-- `Crypto <../../_static/crypto/v1.yaml>`_
-- `Verifiable Credential Registry (v2) <../../_static/vcr/vcr_v2.yaml>`_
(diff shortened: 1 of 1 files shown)


docs/pages/integrating/api.rst
```diff
@@ -14,16 +14,16 @@ API Reference
             const ui = SwaggerUIBundle({
                 "dom_id": "#swagger-ui",
                 urls: [
-                    {url: "../../_static/auth/iam.yaml", name: "Auth (iam)"},
-                    {url: "../../_static/auth/v1.yaml", name: "Auth (v1) - DEPRECATED"},
+                    {url: "../../_static/auth/iam.yaml", name: "Auth (v2)"},
                     {url: "../../_static/crypto/v1.yaml", name: "Crypto"},
```
reinkrul, line 18: Maybe we can move Crypto and Monitoring to the bottom of the non-deprecated list? Since there are little to no users of those APIs

reinkrul (approved):

woutslakhorst (changes requested):
There's no longer a link to download the common OAS types and error definitions

docs/pages/integrating/api.rst
gerardsn: Alphabetic makes more sense to me. If we want to make it easier for users to find the APIs they need we're better off using more descriptive names in stead of the module names.

woutslakhorst (approved):

Issue #2866, opened 2024-03-06
Introduce dev Docker image


reinkrul:
This Docker image is intended for developers to quickly spin up a node for local/dev environments.

- Builds off the latest `master` image
- Disables strict mode
- Stores private keys on disk
- Stores SQL data in SQLite
- Disables Nuts network
- Disables IRMA
- Embeds `ngrok` to super-easy `did:web` usage (which requires a TLS certificate)

woutslakhorst:
since ngrok is going commercial all the way, maybe checkout https://github.com/localtunnel/localtunnel?

reinkrul:
> since ngrok is going commercial all the way, maybe checkout https://github.com/localtunnel/localtunnel?

Worth a try.

gerardsn:
now available

PR #2884, opened 2024-03-07
HTTP: Rework interfaces to fixed 8080 and 8081

reinkrul:
Fixes https://github.com/nuts-foundation/nuts-node/issues/2860

Changes:
- Kept `MultiEcho` type to easily map specific endpoints (`/internal`, `/status`, `/metrics`) to a specific Echo server (`8081`). But instead of allowing mapping endpoints to an Echo server, you can only configure the port/address of the 2 HTTP interfaces (internal and external).
  - had to  
- Removed support for HTTPS on the server-side; parties must deploy a reverse proxy that handles TLS
- Moved HTTP logging configuration to the module-level config, now applies all HTTP interfaces
- API authentication now only applies to `/internal`
- Removed support for CORS: only used by IRMA/EmployeeID which will be removed at some point. Parties that need it can set the headers in their reverse proxy.

Most changes are in the e2e tests, because:
- `/internal` is bound to `127.0.0.1:8081` by default, to make it available from outside the container it has to be remapped to `:8081`
- some e2e tests require HTTPS (in particular the new OpenID4VC tests), which isn't supported by the Nuts node any more, so I had to put an NGINX proxy before it
- the old default port `1323` is now split into `8081` and `8080`, so if a test used both external and internal endpoints test actions needed to be rewritten.

TODO:
- [x] Fix e2e tests
- [x] Rewrite documentation

PR #2888, opened 2024-03-07
Provide dev Docker image for easier testing

reinkrul:
Fixes https://github.com/nuts-foundation/nuts-node/issues/2866

TODO:
- [x] Handle SIGKILL, ^C does not seem to work any more now
- [x] Re-use same localtunnel domain, if possible
- [ ] Documentation


reinkrul:
Use of `exec` in entrypoint script causes signals to be captured by Nuts node, causing `^C` to be handled. But now it looks like the process is killed before the process completes its shutdown;

```
time="2024-03-11T08:35:29Z" level=info msg="Shutting down..."
time="2024-03-11T08:35:29Z" level=info msg="Stopping HTTP..." module=core
```

reinkrul:
Documentation should be in "Running on Docker", but that's partly rewritten by https://github.com/nuts-foundation/nuts-node/pull/2884

Waiting for that PR.

development/dev-image/entrypoint.sh
```diff
@@ -0,0 +1,32 @@
...
+  cat /localtunnel.log
+  exit 1
+fi
+# Extract subdomain (TUNNEL_URL after https:// up to the first .)
+SUBDOMAIN=$(echo $TUNNEL_URL | awk -F/ '{print $3}' | awk -F. '{print $1}')
+# Store subdomain in localtunnel.cfg so that the container retains its subdomain after restart
+echo "${SUBDOMAIN}" > /localtunnel.cfg
+
+echo Your Nuts node URL is: ${TUNNEL_URL}
+NUTS_URL="${TUNNEL_URL}" NUTS_STRICTMODE=false NUTS_AUTH_CONTRACTVALIDATORS=dummy exec /usr/bin/nuts server
```
woutslakhorst, line 32: The persistent hostname is not working for me, here is an example that uses the `TUNNEL_NAME` env var. It also kills both processes.

```suggestion
#!/bin/sh

# Try to use TUNNEL_NAME as subdomain
SUBDOMAIN=${TUNNEL_NAME}
echo Trying $SUBDOMAIN as subdomain

# Run lt to create the local tunnel with --port 8080 as background process.
# It writes the URL to stdout which we need to extract and feed to the nuts command
# capture the pid in a file so we can kill the process when the container stops
lt --port 8080 --subdomain="${SUBDOMAIN}" > /localtunnel.log 2>&1 &
echo $! > /localtunnel.pid

# Try 30 times to read the tunnel URL from the log file.
# The format is "your url is:" followed by the URL.
echo "Waiting for localtunnel URL (time-out in 30s)..."
for i in $(seq 1 30); do
  TUNNEL_URL=$(grep "your url is:" /localtunnel.log | awk '{print $4}')
  if [ -n "$TUNNEL_URL" ]; then
    break
  fi
  sleep 1
done
# Check whether we retrieved the URL and if not, exit with an error.
if [ -z "$TUNNEL_URL" ]; then
  echo "Failed to retrieve the localtunnel URL"
  cat /localtunnel.log
  exit 1
fi
cat /localtunnel.log
# Extract subdomain (TUNNEL_URL after https:// up to the first .)
SUBDOMAIN=$(echo $TUNNEL_URL | awk -F/ '{print $3}' | awk -F. '{print $1}')

echo Your Nuts node URL is: ${TUNNEL_URL}
NUTS_URL="${TUNNEL_URL}" NUTS_STRICTMODE=false NUTS_AUTH_CONTRACTVALIDATORS=dummy exec /usr/bin/nuts server

# Kill the localtunnel process when the container stops
kill $(cat /localtunnel.pid)
```

reinkrul:
Replaced by https://github.com/nuts-foundation/nuts-node/pull/2915

PR #2891, opened 2024-03-08
set primary key for did service and vm to id

woutslakhorst:
according to DID spec, the id always contains the DID, no need to have a primary key consist of both the did and id

also it would be too long (3140 bytes)

### DIFF
--- a/storage/sql_migrations/003_did.sql
+++ b/storage/sql_migrations/003_did.sql
@@ -17,7 +17,7 @@ create table did_verificationmethod
     -- data is a JSON object containing the verification method data, e.g. the public key.
     -- When producing the verificationMethod, data is used as JSON base object and the id and type are added.
     data text         not null,
-    primary key (did, id),
+    primary key (id),
     foreign key (did) references did (did) on delete cascade
 );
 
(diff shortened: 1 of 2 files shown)


reinkrul (changes requested):
But IDs can be relative (e.g. `#0`)

woutslakhorst:
> But IDs can be relative (e.g. `#0`)

§3.2 of the spec:
ID's MUST be DID URLs, references to resources within the same document may be relative DID URLs. The ID of a service and verificationMethod do not reference a resource. The content of assertionMethod could be a relative URL since it references the verificationMethod.

reinkrul:
> > But IDs can be relative (e.g. `#0`)
> 
> §3.2 of the spec: ID's MUST be DID URLs, references to resources within the same document may be relative DID URLs. The ID of a service and verificationMethod do not reference a resource. The content of assertionMethod could be a relative URL since it references the verificationMethod.

We might run into issues with DID documents using `@base` in which, I think, you can omit the base URI. But that's for later concern, as this only applies to the DID documents we produce from the Nuts node. So probably OK for now.

reinkrul (approved):

woutslakhorst:
> > > But IDs can be relative (e.g. `#0`)
> > 
> > 
> > §3.2 of the spec: ID's MUST be DID URLs, references to resources within the same document may be relative DID URLs. The ID of a service and verificationMethod do not reference a resource. The content of assertionMethod could be a relative URL since it references the verificationMethod.
> 
> We might run into issues with DID documents using `@base` in which, I think, you can omit the base URI. MS Authenticator does this.
> 
> But that's for later concern, as this only applies to the DID documents we produce from the Nuts node. So probably OK for now.

Wouldn't that be a parsing concern? (If you want to store other docs in our DB setup that is...)

Issue #2854, opened 2024-03-04
Move non node related documentation to Nuts wiki


woutslakhorst:
Everything on proposed integration, bolts and getting started that are not directly software related will have to move to the Nuts wiki.

## Getting Started

* [x] https://nuts-node.readthedocs.io/en/stable/pages/getting-started/0-hello-world-docker.html uses the Nuts network. Page could go away entirely.
* [x] https://nuts-node.readthedocs.io/en/stable/pages/getting-started/1-running-docker.html remove TLS configuration
* [x] https://nuts-node.readthedocs.io/en/stable/pages/getting-started/2-running-native.html nobody uses this, just use Makefile?
* [x] https://nuts-node.readthedocs.io/en/stable/pages/getting-started/3-configure-your-node.html can go entirely
* [x] https://nuts-node.readthedocs.io/en/stable/pages/getting-started/4-connecting-crm.html rewrite given `did:web` and Discovery Services. See https://github.com/nuts-foundation/nuts-node/issues/2870
* [x] https://nuts-node.readthedocs.io/en/stable/pages/getting-started/5-authentication.html is IRMA and login contract functionality, which must be removed
* [x] https://nuts-node.readthedocs.io/en/stable/pages/getting-started/6-adding-authorizations.html is `NutsAuthorizationCredential`-related, which is replaced by PDPs, documented in the wiki. See https://github.com/nuts-foundation/nuts-node/issues/2871
* [x] https://nuts-node.readthedocs.io/en/latest/pages/getting-started/6-connecting-ehr.html is `did:nuts` Registry and `NutsAuthorizationCredential`-specific, can be removed

## Development:

- [x] https://nuts-node.readthedocs.io/en/stable/pages/development/2-releasing.html references Registry and Demo EHR.

## Deployment&configuration

- [x] https://nuts-node.readthedocs.io/en/stable/pages/deployment/recommended-deployment.html diagram could use some minor changes. Production checklist needs a rewrite to v6 situation
- [x] https://nuts-node.readthedocs.io/en/stable/pages/deployment/resource-requirements.html requires addition for SQL.
- [x] https://nuts-node.readthedocs.io/en/stable/pages/deployment/monitoring.html remove dag visualization and conflicted did documents
- [x] https://nuts-node.readthedocs.io/en/stable/pages/deployment/cli-reference.html only works for did:nuts, did:web is not represented here. Deprecate?
[...]

PR #2893, opened 2024-03-11
remove XIS and EHR mentions from docs

woutslakhorst:
part of #2854 

changed to `client application`.

Some other small parts have been removed as well.


reinkrul (changes requested):
What about "care organization"?

reinkrul (approved):

Issue #2894, opened 2024-03-11
Extend issuer API to find credentials that have been issued in an interval with a specific key


woutslakhorst:
In order to do key rotation, you'll need to remove keys at a certain point in time. This is only possible if there are no valid credentials for a specific key. The search API needs to be extended to find these.

gerardsn:
Key rotation could be a means to invalidate credentials without any other revocation/expiry mechanism. The search API should only be used to make an informed decision, not be a condition for key rotation.

PR #2895, opened 2024-03-11
rewrite key rotation

woutslakhorst:
part of #2854 

key rotation is rewritten to a add/delete procedure. A new issue has been created for finding issued VCs based on period/kid (#2894)


woutslakhorst:
> Note that the API does not support the required operations yet..

Yes, it was added as note (and an issue)

reinkrul (approved):

PR #2904, opened 2024-03-12
Tweaked strict mode checks to allow running without TLS

reinkrul:
When deploying my Discovery Server, I came across the issue that I must configure TLS (truststore, certificate and key) when strict mode is enabled.

This Nuts node only acts as Discovery Server; it only uses did:web and does not participate in a did:nuts network, meaning it does not need a client certificate (it does not connect to other node over gRPC or HTTP endpoints that require a client certificate).

This PR fixes the following:
- fixes the existing lenient check in the `network` module to allow TLS to be disabled when no bootstrap nodes are configured and there are no discovered nodes. The intended leniency is broken because `assumeNewNode` is initialized after the said check
- remove TLS enabled check in strict mode in the auth module; the strict HTTP client already checks whether the endpoints it connects to are HTTPS in strict mode, so this check does not add extra security.

### DIFF
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -142,11 +142,6 @@ func (auth *Auth) Configure(config core.ServerConfig) error {
 		StrictMode:            config.Strictmode,
 	}, auth.vcr, resolver.DIDKeyResolver{Resolver: auth.vdrInstance.Resolver()}, auth.keyStore, auth.jsonldManager, auth.pkiProvider)
 
-	tlsEnabled := config.TLS.Enabled()
-	if config.Strictmode && !tlsEnabled {
-		return errors.New("in strictmode TLS must be enabled")
-	}
-
 	auth.tlsConfig, err = auth.pkiProvider.CreateTLSConfig(config.TLS) // returns nil if TLS is disabled
 	if err != nil {
 		return err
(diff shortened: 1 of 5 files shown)


auth/auth.go
```diff
@@ -142,11 +142,6 @@ func (auth *Auth) Configure(config core.ServerConfig) error {
 		StrictMode:            config.Strictmode,
 	}, auth.vcr, resolver.DIDKeyResolver{Resolver: auth.vdrInstance.Resolver()}, auth.keyStore, auth.jsonldManager, auth.pkiProvider)
 
-	tlsEnabled := config.TLS.Enabled()
-	if config.Strictmode && !tlsEnabled {
-		return errors.New("in strictmode TLS must be enabled")
-	}
```
gerardsn, line 148: does this check still exists in the n2n flow somewhere?

network/network.go
gerardsn, line 1: what is the point of the network related changes?

reinkrul: Before, the intention was to let an "empty" node (from network point of view = no bootstrap node, no gRPC services in DID documents discovered) start without TLS configured (for the gRPC network), but this didn't work. This PR fixes that leniency check.

auth/auth.go
reinkrul: It's performed by the strict HTTP client, which checks the request URL scheme (HTTP/HTTPS)

gerardsn (approved):

network/network.go
gerardsn: or didn't read the description at all because I started reviewing to see if this also fixes 2909, sorry

Issue #2903, opened 2024-03-12
IAM: specify full DID in OAuth2/OpenID endpoints


reinkrul:
Instead of just the did:web "user ID" (last part after `:iam:`). This is, it works for rooted `did:web` DIDs, `did:web` DIDs with a user path other than `:iam:<id>` and even for non-web DIDs.

See discussion in https://github.com/nuts-foundation/nuts-node/issues/2833#issuecomment-1980955648

PR #2906, opened 2024-03-12
IAM: Specify full DID instead of did:web user ID in OAuth2 endpoints

reinkrul:
Moves all OAuth2/OpenID endpoints (`/authorize, /token, /callback, /statuslist, /response, /presentation_definition`) to from `/iam/{id}/<endpoint>` to `/oauth2/{did}/<endpoint>`, changing the base path to `oauth2` and the`id` parameter to the full DID. This has the following advantages:

- It works with DIDs other than did:web
- It works for rooted did:web DIDs
- `/iam` now only serves DID documents, allowing more flexibility for deeper subpaths in future (e.g. `:iam:care:1234`)

Fixes https://github.com/nuts-foundation/nuts-node/issues/2903


docs/_static/auth/iam.yaml
```diff
@@ -535,15 +571,17 @@ paths:
           description: |
             This is returned when an OAuth2 Client is unauthorized to talk to the introspection endpoint.
             Note: introspection of an invalid or malformed token returns a 200 where with field 'active'=false
-  /iam/{id}/statuslist/{page}:
+  /oauth2/{did}/statuslist/{page}:
```
woutslakhorst, line 574: This is not an oauth related endpoint. Change base path and maybe also put it in a different yaml

auth/api/iam/api.go
```diff
@@ -607,3 +619,16 @@ func (r Wrapper) accessTokenClientStore() storage.SessionStore {
...
 }
+
+// createOAuth2EndpointURL creates an OAuth2 endpoint URL for an owned did:web DID, for the given endpoint.
+// It creates a URL in the following format: https://<did:web host>/oauth2/<did>/<endpoint>
+func createOAuth2EndpointURL(webDID did.DID, endpointPath ...string) (*url.URL, error) {
+	didURL, err := didweb.DIDToURL(webDID)
+	if err != nil {
+		return nil, fmt.Errorf("failed to convert DID to URL: %w", err)
+	}
+	didURL.RawPath = ""
```
woutslakhorst, line 630: this deserves a comment

auth/api/iam/api_test.go
```diff
@@ -847,13 +848,56 @@ func TestWrapper_StatusList(t *testing.T) {
...
+	})
+	t.Run("did:web with port", func(t *testing.T) {
+		const didAsString = "did:web:example.com%3A8080:iam:holder"
+		webDID := did.MustParseDID(didAsString)
+
+		actual, err := createOAuth2EndpointURL(webDID, "something")
+
+		require.NoError(t, err)
+		require.NotNil(t, actual)
+		assert.Equal(t, "https://example.com:8080/oauth2/did:web:example.com%253A8080:iam:holder/something", actual.String())
```
woutslakhorst, line 891: Since you fiddled with rawpath, what happens if I put this return path in a Location header. Is the framework going to escape it again or not?

auth/oauth/types_test.go
```diff
@@ -0,0 +1,48 @@
+package oauth
```
woutslakhorst, line 1: license

woutslakhorst (changes requested):

gerardsn (changes requested):
I expected the `e2e-tests/oauth-flow/statuslis2021` test to fail. The path where the statuslist can be resolved has changed, but `statuslist2021_issuer.go` still sets the old path. (nodeA should don't be able to resolve the statuslist from nodeB, last check)

auth/api/iam/api.go
reinkrul: Made it less fault prone by just creating a new `url.URL`.

docs/_static/auth/iam.yaml
reinkrul: Moved to `/statuslist/<did>/<page>`

reinkrul:
> I expected the `e2e-tests/oauth-flow/statuslis2021` test to fail. The path where the statuslist can be resolved has changed, but `statuslist2021_issuer.go` still sets the old path. (nodeA should don't be able to resolve the statuslist from nodeB, last check)

There also was no unit test to assert the constructed URL, so I added that. Fixing statuslist e2e test is not in scope for this PR. Made an issue: https://github.com/nuts-foundation/nuts-node/issues/2922

woutslakhorst (approved):

gerardsn (changes requested):

auth/api/iam/api.go
```diff
@@ -607,3 +619,18 @@ func (r Wrapper) accessTokenClientStore() storage.SessionStore {
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to convert DID to URL: %w", err)
+	}
+	result := new(url.URL)
+	result.Scheme = didURL.Scheme
+	result.Host = didURL.Host
+	result.Path = path.Join("oauth2", webDID.String())
+	result = result.JoinPath(endpointPath...)
+	return result, nil
+}
```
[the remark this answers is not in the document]
gerardsn: Not sure what the didURL.Parse does to URL escaped did's though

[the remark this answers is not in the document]
woutslakhorst: if the path contains escaped characters, Parse will fill u.path with the unescaped value and u.rawPath with the unescaped value. When calling String(), the rawPath is used when set.

docs/_static/auth/iam.yaml
```diff
@@ -261,11 +271,11 @@ paths:
     get:
       tags:
         - well-known
-      summary: Get the OAuth2 Authorization Server metadata
+      summary: Get the OAuth2 Authorization Server metadata for a did:web with a :iam:<id> user path.
```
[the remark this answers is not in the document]
reinkrul: In our case, it's did:web DID specific. We can always support other DIDs in future, but it'll probably have another path?

gerardsn (approved):

PR #2913, opened 2024-03-13
Fail to start when deprecated network params are used

### LINKED ISSUE #2909: Network: fail to start if deprecated TLS config is present
#2877 removed the deprecated TLS config under `network` (moved to `tls`) i.e.,
```
strictmode: false
network:
  truststorefile: /ssl/truststore.pem
  certfile: /ssl/certificate.pem
  certkeyfile: /ssl/certificate.pem
  grpcaddr:  :5555
```

As this is no longer parsed, nodes using this config now try to connect to peers without TLS. This obviously fails, but the message is extremely unhelpful:
```
2024-03-13 11:22:56 time="2024-03-13T10:22:56Z" level=debug msg="failed to open a grpc ClientConn" error="context deadline exceeded: connection error: desc = \"error reading server preface: http2: frame too large\"" module=Network peerAddr=<rein> peerAuthenticated=false peerDID= peerID=
2024-03-13 11:22:56 time="2024-03-13T10:22:56Z" level=debug msg="failed to open a grpc ClientConn" error="context deadline exceeded: connection error: desc = \"error reading server preface: EOF\"" module=Network peerAddr=<wout> peerAuthenticated=false peerDID= peerID=
2024-03-13 11:22:56 time="2024-03-13T10:22:56Z" level=debug msg="failed to open a grpc ClientConn" error="context deadline exceeded: connection error: desc = \"error reading server preface: EOF\"" module=Network peerAddr=<gerard> peerAuthenticated=false peerDID= peerID=
2024-03-13 11:22:57 time="2024-03-13T10:22:57Z" level=debug msg="failed to open a grpc ClientConn" error="context deadline exceeded: connection error: desc = \"error reading server preface: http2: frame too large\"" module=Network peerAddr=<e-care> peerAuthenticated=false peerDID= peerID=
```

Solution: A node should fail to start if the deprecated config settings are present

(Nodes not wanting to use TLS still works fine as demonstrated by e2e tests)

gerardsn:
fixes #2909

### DIFF
--- a/core/server_config.go
+++ b/core/server_config.go
@@ -53,17 +53,19 @@ var redactedConfigKeys = []string{
 
 // ServerConfig has global server settings.
 type ServerConfig struct {
-	Verbosity           string            `koanf:"verbosity"`
-	LoggerFormat        string            `koanf:"loggerformat"`
-	CPUProfile          string            `koanf:"cpuprofile"`
-	Strictmode          bool              `koanf:"strictmode"`
-	InternalRateLimiter bool              `koanf:"internalratelimiter"`
-	Datadir             string            `koanf:"datadir"`
-	HTTPClient          HTTPClientConfig  `koanf:"httpclient"`
-	TLS                 TLSConfig         `koanf:"tls"`
-	LegacyTLS           *NetworkTLSConfig `koanf:"network"`
 	// URL contains the base URL for public-facing HTTP services.
-	URL       string `koanf:"url"`
+	URL                 string           `koanf:"url"`
+	Verbosity           string           `koanf:"verbosity"`
+	LoggerFormat        string           `koanf:"loggerformat"`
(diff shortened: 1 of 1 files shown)


reinkrul (commented):
Do we do this just for this case, or for all deprecated config we remove? Since then, we always have to spread deprecation/removal over at least 3 versions.

This only happens in non-strict mode (not posible in strict mode).

gerardsn:
I was thinking only for this case since it took me a couple of hours to figure out what caused this to fail. User trying to upgrade are going to be very frustrated if they have to go through the same proces.

A better option imo would be to fail (maybe just in production) if the config contains invalid parameters, but I don't know if this is possible/desirable.

reinkrul (approved):
Maybe we can remove e.g. half a year after release of v6

PR #2915, opened 2024-03-13
Create developer image that auto starts a https tunnel

woutslakhorst:
Closes #2888

Use Microsoft devtunnel instead of localtunnel. Localtunnel's stability is not that good.

With devtunnel, you need to login at startup which is very easy with a github account. Latest tunnel id is stored under `/devtunnel/tunnel.id`, if you mount that directory you can reuse tunnels. With docker compose you are still logged in after restart

- [x] documentation


auth/api/iam/user.go
```diff
@@ -38,7 +38,7 @@ const (
 	oAuthFlowTimeout = time.Minute
 	// userRedirectTimeout is the timeout for the user redirect session.
 	// This is the maximum time between the creation of the redirect for the user and the actual GET request to the user/wallet page.
-	userRedirectTimeout = time.Second * 5
+	userRedirectTimeout = time.Second * 10
```
reinkrul, line 41: unrelated change?

development/dev-image/Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM nutsfoundation/nuts-node:latest as nutsnode
```
reinkrul, line 1: I'd say from `master`

reinkrul:
How does this work then, does it log a link to GH?

development/dev-image/Dockerfile
gerardsn: `make docker` uses tag `latest`. So building the dev-image locally would then require an extra step. I'd prefer to change both to `master`

gerardsn:
To me it looks like the login requirement limits you to using this container on your local machine aka laptop. This means no e2e-tests, and not useful for parties that develop everything in the cloud? (This could be considered a feature)

development/dev-image/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+FROM debian:buster-slim
```
gerardsn, line 2: is it possible to use alpine here? This would prevent distro specific issues.

woutslakhorst: Unfortunately the tunnel install script uses `apt-get`

development/dev-image/entrypoint.sh
```diff
@@ -0,0 +1,48 @@
...
+  echo "Failed to retrieve the devtunnel URL"
+  cat /devtunnel/tunnel.log
+  exit 1
+fi
+
+# store the tunnel id in /devtunnel/tunnel.id
+echo $TUNNEL_ID > /devtunnel/tunnel.id
+echo Your Nuts node URL is: ${TUNNEL_URL}
+
+NUTS_URL="${TUNNEL_URL}" NUTS_STRICTMODE=false NUTS_AUTH_CONTRACTVALIDATORS=dummy NUTS_HTTP_INTERNAL_ADDRESS=:8081 NUTS_AUTH_V2APIENABLED=true exec /usr/bin/nuts server
```
gerardsn, line 45: config flag is removed
```suggestion
NUTS_URL="${TUNNEL_URL}" NUTS_STRICTMODE=false NUTS_AUTH_CONTRACTVALIDATORS=dummy NUTS_HTTP_INTERNAL_ADDRESS=:8081 exec /usr/bin/nuts server
```

development/dev-image/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+
+RUN apt-get update && apt-get install -y curl sudo musl
+RUN curl -sL https://aka.ms/DevTunnelCliInstall | bash
+
+COPY entrypoint.sh /
+RUN chmod +x /entrypoint.sh
+
+COPY --from=nutsnode /usr/bin/nuts /usr/bin/nuts
+RUN chmod +x /usr/bin/nuts
```
gerardsn, line 11: The number of layers needed can be reduced by moving stuff around and reducing the number of RUN commands.
Line 10-11 are the only parts that ever change, can they go at the very end?

gerardsn (approved):

woutslakhorst:
Docker action isn't running anymore. Checking...

development/dev-image/Dockerfile
woutslakhorst: I don't think those last 2 layers (config) have a lot of influence on image building performance when cached.

reinkrul:
We still need to build this in GH Actions

Issue #2916, opened 2024-03-13
Can't register for a discovery service that doesn't require VCs in a VP


woutslakhorst:
Registration fails with "DID wallet does not have credentials required for registration on".

This comes from discovery/client.go#140 where len==0 is checked. But if no VCs are required, len==0 is fine.

PR #2923, opened 2024-03-15
allow for empty VP when registering with discovery service

woutslakhorst:
fixes: #2916 

added simple check to see if to VCs are required

### DIFF
--- a/discovery/client.go
+++ b/discovery/client.go
@@ -137,7 +137,7 @@ func (r *defaultClientRegistrationManager) findCredentialsAndBuildPresentation(c
 	if err != nil {
 		return nil, fmt.Errorf("failed to match Discovery Service's Presentation Definition (service=%s, did=%s): %w", service.ID, subjectDID, err)
 	}
-	if len(matchingCredentials) == 0 {
+	if len(matchingCredentials) == 0 && len(service.PresentationDefinition.InputDescriptors) > 0 {
 		return nil, fmt.Errorf("DID wallet does not have credentials required for registration on Discovery Service (service=%s, did=%s)", service.ID, subjectDID)
 	}
 	return r.buildPresentation(ctx, subjectDID, service, matchingCredentials, nil)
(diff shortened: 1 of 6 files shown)


discovery/client.go
```diff
@@ -137,7 +137,7 @@ func (r *defaultClientRegistrationManager) findCredentialsAndBuildPresentation(c
 	if err != nil {
 		return nil, fmt.Errorf("failed to match Discovery Service's Presentation Definition (service=%s, did=%s): %w", service.ID, subjectDID, err)
 	}
-	if len(matchingCredentials) == 0 {
+	if len(matchingCredentials) == 0 && len(service.PresentationDefinition.InputDescriptors) > 0 {
```
reinkrul, line 140: they could be optional? (`min: 0`)

Not sure if we already need to support this, though... Use case?

reinkrul (changes requested):

gerardsn (commented):
Isn't the real issue here that `PresentationDefinition.Match` doesn't return an error if it cannot fulfill the presentations definition? I think we allowed this because we initially thought a PD could be fulfilled by multiple wallets, but this is no longer the case?

discovery/client.go
woutslakhorst: this is the check we do now on multiple locations.

woutslakhorst:
> Isn't the real issue here that `PresentationDefinition.Match` doesn't return an error if it cannot fulfill the presentations definition? I think we allowed this because we initially thought a PD could be fulfilled by multiple wallets, but this is no longer the case?

uncertain for now, the spec allows for multiple wallets to provide the proof. When we return an error, we block that option. In our case we know we can't combine a organisation wallet with a user wallet.

reinkrul:
Could be fixed with a specific error that can be tested? Since from PEX API intention perspective, not being able to fulfill a request (don't have the required credentials) is really different from fulfilling an empty request. It's now not possible to differentiate those 2.

woutslakhorst:
> Could be fixed with a specific error that can be tested? Since from PEX API intention perspective, not being able to fulfill a request (don't have the required credentials) is really different from fulfilling an empty request. It's now not possible to differentiate those 2 (which is the actual problem, i.m.o.)

PD matching would have to rewritten to operate on a set of wallets, a bit too much for now (given future of pex is not certain as well).

reinkrul (approved):

PR #2926, opened 2024-03-15
return empty list on empty wallet

### LINKED ISSUE #2924: Return value for an empty wallet is null
If a holder's wallet is empty and you list it via `GET {{baseUrl}}/internal/vcr/v2/holder/:did/vc` it returns `null` and not `[]`

woutslakhorst:
fixes #2924

### DIFF
--- a/vcr/holder/wallet.go
+++ b/vcr/holder/wallet.go
@@ -325,7 +325,7 @@ func (s walletStore) list(holderDID did.DID) ([]vc.VerifiableCredential, error) 
 	if err != nil {
 		return nil, err
 	}
-	var results []vc.VerifiableCredential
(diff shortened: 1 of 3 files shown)


vcr/holder/wallet.go
```diff
@@ -325,7 +325,7 @@ func (s walletStore) list(holderDID did.DID) ([]vc.VerifiableCredential, error)
 	if err != nil {
 		return nil, err
 	}
-	var results []vc.VerifiableCredential
+	results := make([]vc.VerifiableCredential, 0)
```
reinkrul, line 328: again, and again...

reinkrul (approved):

PR #2939, opened 2024-03-18
Dev image2

woutslakhorst:
image is created correctly but wasn't uploaded to docker hub

now with a push action


development/dev-image/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+RUN apt-get update && apt-get install -y curl sudo musl
+RUN curl -sL https://aka.ms/DevTunnelCliInstall | bash
+
+COPY entrypoint.sh /
+RUN chmod +x /entrypoint.sh
+
+COPY --from=nutsnode /usr/bin/nuts /usr/bin/nuts
+RUN chmod +x /usr/bin/nuts
+
+HEALTHCHECK --start-period=30s --timeout=5s --interval=10s \
```
reinkrul, line 13: do we need a longer starting period because it blocks until you logged in on Github?

Issue #2942, opened 2024-03-18
Rewrite TLS configuration options


reinkrul:
The documentation for TLS configuration is `tls.certfile`, `tls.certheader`, `tls.certkeyfile`, `tls.offload` and `tls.truststorefile` is confusing; TLS offloading is now mandatory for HTTP, but setting `tls.offload` (which is only applicable to gRPC) then requires you to set the other properties as well.

We should mark these properties as deprecated and alter the description to reflect this.

PR #2949, opened 2024-03-19
Mark v5/did:nuts config flags as such

reinkrul:
To make it more clear what's required for did:web use cases.

Also made Golden Hammer/OpenID4VCI server-to-server hidden (shouldn't be used actively?)

Fixes https://github.com/nuts-foundation/nuts-node/issues/2942


reinkrul:
Discussed with @gerardsn: make hidden, add note above the table referring to the v5 documentation

gerardsn (approved):
approved after changes as discussed: make all v5 config flags hidden, and add a note that config flags related to did:nuts / gRPC network can be found in the v5 documentation

Issue #2953, opened 2024-03-20
/internal/vcr/v2/holder/{did}/vc ( getCredentialsInWallet ) returns null instead of empty array on empty wallet


rolandgroen:
As description, the method: ```w.VCR.Wallet().List(ctx, *holderDID)```  returns null and populates the GetCredentialsInWallet200JSONResponse with nil instead of []

gerardsn:
fixed by #2926

PR #2954, opened 2024-03-20
Fix for #2953: Handle nil credentials in GetCredentialsInWallet

rolandgroen:
This PR ensures that in the function GetCredentialsInWallet, a nil slice of credentials is replaced with an empty slice.


reinkrul (changes requested):
@woutslakhorst already fixed this on master recently.

https://github.com/nuts-foundation/nuts-node/commit/b5217ef60c4e98369139245156dd6d9ab997f812

So I.m.o., this one can be closed?

PR #2965, opened 2024-03-21
Don't set default SQLite in strictmode

### LINKED ISSUE #2964: SQL storage should have no default in strictmode
storage configuration is not an afterthought. if this is not set in production setting the node should fail to start

gerardsn:
fixes #2964


reinkrul:
This causes SQL to be mandatory even if people don't use the v6 functionality.

reinkrul (approved):

gerardsn:
strictmode is to prevent unsafe, not stupid configurations

Issue #2966, opened 2024-03-21
Use linters to check our codebase


gerardsn:
Problematic results found by `staticcheck` (use `golangci-lint`):
```
auth/services/x509/uzi_validator.go:134:3: SA4006: this value of `rawCert` is never used (staticcheck)
                rawCert = append(rawCerts, rawCert...)
```

and

```
crypto/jwx.go:437:2: SA4020: unreachable case clause: crypto.PublicKey will always match before crypto/rsa.PublicKey (staticcheck)
        case rsa.PublicKey:
        ^
crypto/jwx.go:439:2: SA4020: unreachable case clause: crypto.PublicKey will always match before crypto/ecdsa.PublicKey (staticcheck)
        case ecdsa.PublicKey:
```

reinkrul:
Can we run these as part of our build and add them to the Makefile?

gerardsn:
Yes, but it takes some time to find the correct configuration. By default it also finds stuff like a function with 61 lines while 60 is permitted etc.

reinkrul:
Not bound to v6, so i removed the milestone.

PR #2967, opened 2024-03-21
go vet and remove dead code

gerardsn:
part of #2966


vcr/revocation/statuslist2021_issuer.go
```diff
@@ -426,7 +426,6 @@ func (cs *StatusList2021) Revoke(ctx context.Context, credentialID ssi.URI, entr
 		}
 
 		// append new revocation and re-issue the StatusList2021Credential.
-		issuerRecord.Revocations = append(issuerRecord.Revocations)
```
reinkrul, line 429: isn't this an actual bug, since there's nothing appended?

reinkrul (approved):
Nice, just 1 question

vcr/revocation/statuslist2021_issuer.go
gerardsn: This used to append the revocation from L397, but the query on L414 now includes new revocation. I just forgot to remove this. The expected behavior is tested.

PR #2972, opened 2024-03-22
storage.sql.connection string redaction

### LINKED ISSUE #2971: storage.sql.connection may contain secrets that are logged at startup
username and password can be in the connection string for MySQL and Postgres

gerardsn:
closes #2971 

I figured we might want to provide some useful info in the error, so this PR tries to redact the secret:

``` yaml
storage:
  sql:
    connection: "fake://user:password@example.com:123/db"
```

fails with

```
"unable to configure Storage: failed to initialize SQL database: unsupported SQL database connection: fake://<redacted>@example.com:123/db"
```

This might still fail, so just not logging the string ever is the safest option.

What do you think? If this is preferred I will add tests etc.

### DIFF
--- a/storage/config.go
+++ b/storage/config.go
@@ -18,6 +18,8 @@
 
 package storage
 
+import "strings"
+
 // Config specifies config for the storage engine.
 type Config struct {
 	BBolt BBoltConfig `koanf:"bbolt"`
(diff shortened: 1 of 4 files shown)


storage/config.go
```diff
@@ -33,5 +35,26 @@ func DefaultConfig() Config {
 // SQLConfig specifies config for the SQL storage engine.
 type SQLConfig struct {
 	// ConnectionString is the connection string for the SQL database.
+	// This string may contain secrets (user:password), so should never be logged. User RedactedConnectionString.
```
reinkrul, line 38: User RedactedConnectionString?

storage/config.go
```diff
@@ -33,5 +35,26 @@ func DefaultConfig() Config {
...
 	ConnectionString string `koanf:"connection"`
 }
+
+// RedactedConnectionString sanitizes the configured connection string so that is can be logged.
+func RedactedConnectionString(connection string) string {
```
reinkrul, line 43: can be unexported/private

storage/config.go
```diff
@@ -33,5 +35,26 @@ func DefaultConfig() Config {
...
+	if strings.HasPrefix(connection, "sqlite:") {
+		// this may still contain userauth, but this is not supported by nuts node
+		return connection
+	}
+	rest, url, found := strings.Cut(connection, "@")
+	if !found {
+		// does not contain user:pw
+		return connection
+	}
+	result := "<redacted>@" + url
```
reinkrul, line 53: probably good enough

reinkrul (changes requested):

storage/config.go
gerardsn: removed now that is unexported

gerardsn:
changed to never log the connection string. Certain typos can cause the redaction to fail, its safer to just never log the connection string

reinkrul (approved):

PR #2980, opened 2024-03-25
VDR: Move FilterServices to /service instead of /service/{id}

reinkrul:
Was in the wrong place.

### DIFF
--- a/docs/_static/vdr/v2.yaml
+++ b/docs/_static/vdr/v2.yaml
@@ -147,6 +147,91 @@ paths:
                 $ref: '#/components/schemas/Service'
         default:
           $ref: '../common/error_response.yaml'
+    get:
+      summary: "Filters services of a resolved DID document"
+      parameters:
+        - name: type
+          in: query
+          description: |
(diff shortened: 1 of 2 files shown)


docs/_static/vdr/v2.yaml
```diff
@@ -147,71 +147,6 @@ paths:
...
-        "200":
-          description: "Service has been updated successfully. Returns the service."
-          content:
-            application/json:
-              schema:
-                $ref: '#/components/schemas/Service'
-        default:
-          $ref: '../common/error_response.yaml'
     get:
       summary: "Filters services of a resolved DID document"
```
gerardsn, line 151: The diff doesn't match the PR, so just confirming: you moved this `get ` to a different path?

gerardsn (approved):

docs/_static/vdr/v2.yaml
reinkrul: a, weird yes. I moved `FilterServices` to `/service`, it was under `/service/{id}` by accident (which doesn't make sense since it's meant to filter the services, not pre-select a specific one).

PR #2984, opened 2024-03-26
Add AllowUntrustedIssuer option to VerifyVP

### LINKED ISSUE #2983: Don't require trusted VC issuers when verifying VP for did:web
Individual credential issuers no longer need to be trusted as this is now part of the presentation definition. VerifyVP still has `allowUntrustedVCs` set to `false` by default with no option to change it

gerardsn:
closes #2983

### DIFF
--- a/docs/_static/vcr/vcr_v2.yaml
+++ b/docs/_static/vcr/vcr_v2.yaml
@@ -725,6 +725,10 @@ components:
           type: boolean
           description: Indicates whether the Verifiable Credentials within the VP must be verified, default true.
           default: true
+        allowUntrustedIssuer:
+          description: If set to true, trust in credential issuer is not verified. Non did:nuts use cases should set this to true.
+          type: boolean
+          default: false
     VPVerificationResult:
       description: Contains the verifiable presentation verification result.
       type: object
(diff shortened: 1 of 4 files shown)


docs/_static/vcr/vcr_v2.yaml
```diff
@@ -725,6 +725,10 @@ components:
           type: boolean
           description: Indicates whether the Verifiable Credentials within the VP must be verified, default true.
           default: true
+        allowUntrustedIssuer:
+          description: If set to true, trust in credential issuer is not verified. Non did:nuts use cases should set this to true.
```
reinkrul, line 729: Or derive it from issuer DID? If `did:web`, allowUntrusted = true?

gerardsn: That works too, but I don't like magic. The API then has different behavior depending on the did method. We could add it in the documentation, but everyone just skims through the documentation except when it comes to config flags.

Would changing the default value if !did:nuts method in addition to the current work? This provides sane defaults, but also allows overriding them.

gerardsn:
superseded by #2992

PR #2989, opened 2024-03-27
Add credentialStatus to common ssi_types

gerardsn:
spec only contains
- `id`: OPTIONAL
- `type`: REQUIRED

all other properties in a `CredentialSubject` are defined by the `type`. Leaving this as a plain object seems to the clearest solution, same `CredentialStatus`


woutslakhorst:
Do we also need to add the JSON-LD context definition to our assets?

woutslakhorst (approved):

PR #2990, opened 2024-03-27
IAM: Allow user details to be passed in requestUserAccessToken

reinkrul:

PR #2991, opened 2024-03-27
IAM: Create session-bound user wallet when starting user OpenID4VP flow

reinkrul:
The previous PR added the user details to the `RequestUserAccessToken` API, this PR builds on that by issuing an EmployeeCredential to the user session. This user session is started when the browser initially lands on the Nuts node and is authenticated with a secure session cookie.

The next PR will use the EmployeeCredential in PEXs.

Builds on https://github.com/nuts-foundation/nuts-node/pull/2990


auth/api/iam/session.go
```diff
@@ -39,10 +39,25 @@ type OAuthSession struct {
...
+// UserSession is a session-bound Verifiable Credential wallet.
+type UserSession struct {
+	// TenantDID is the requesting DID when the user session was created, typically the employer's (of the user) DID.
+	// A session needs to be scoped to the tenant DID, since the session gives access to the tenant's wallet,
+	// and the user session might contain session-bound credentials (e.g. EmployeeCredential) that were issued by the tenant.
+	TenantDID did.DID    `json:"tenantDID"`
+	Wallet    UserWallet `json:"wallet"`
+}
+
+// UserWallet is a session-bound Verifiable Credential wallet.
```
woutslakhorst, line 54: Add some info about the security level of this wallet, since it's session bound. In the case of clustering the private key will be synchronised plain text. This is OK since it doesn't really add any security, just protocol compatibility.

auth/api/iam/user.go
```diff
@@ -79,6 +91,19 @@ func (r Wrapper) handleUserLanding(echoCtx echo.Context) error {
 		return err
 	}
 
+	// TODO: When we support external Identity Providers, we should try to load the existing user session
```
woutslakhorst, line 94: Not a todo, just a note.

auth/api/iam/user.go
```diff
@@ -124,10 +149,146 @@ func (r Wrapper) userRedirectStore() storage.SessionStore {
...
 }
 
 func (r Wrapper) oauthCodeStore() storage.SessionStore {
 	return r.storageEngine.GetSessionDatabase().GetStore(oAuthFlowTimeout, oauthCodeKey...)
 }
+
+// loadUserSession loads the user session given the session ID in the cookie.
+// If there is no session cookie (not yet authenticated, or the session expired), nil is returned.
+// If another, technical error occurs when retrieving the session.
+func (r Wrapper) loadUserSession(ctx echo.Context, tenantDID did.DID) (*UserSession, error) {
```
woutslakhorst, line 167: unused.

auth/api/iam/user.go
woutslakhorst: Load the user session first, otherwise you'll create a new one every call?

woutslakhorst (changes requested):

auth/api/iam/user.go
reinkrul: It's used now.

Also used in follow-up PR.

woutslakhorst (approved):

reinkrul:
@woutslakhorst some additional changes.

Issue #2995, opened 2024-03-28
IAM OAuth2 client uses TLS config and truststore from PKI module


reinkrul:
Which in turns is sourced from `truststore.pem`, which is loaded with the PKIOverheid Private Services trust anchors.

This does not work for OpenID4VC/OpenID4VP and `did:web`, which use publicly trusted CAs, not private ones.

reinkrul:
Probably only an issue for hybrid nodes which have configured `truststore.pem`, otherwise it'll _probably_ fallback to the OS bundle.

PR #2998, opened 2024-03-28
Auth: IAM client should not use PKI TLS config

reinkrul:
Fixes https://github.com/nuts-foundation/nuts-node/issues/2995

### DIFF
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -108,7 +108,7 @@ func (auth *Auth) RelyingParty() oauth.RelyingParty {
 
 func (auth *Auth) IAMClient() iam.Client {
 	keyResolver := resolver.DIDKeyResolver{Resolver: auth.vdrInstance.Resolver()}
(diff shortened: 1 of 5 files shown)


auth/auth.go
```diff
@@ -108,7 +108,7 @@ func (auth *Auth) RelyingParty() oauth.RelyingParty {
 
 func (auth *Auth) IAMClient() iam.Client {
 	keyResolver := resolver.DIDKeyResolver{Resolver: auth.vdrInstance.Resolver()}
-	return iam.NewClient(auth.vcr.Wallet(), keyResolver, auth.keyStore, auth.strictMode, auth.httpClientTimeout, auth.tlsConfig)
```
gerardsn, line 111: it can be removed from the underlying struct completely

gerardsn (changes requested):

woutslakhorst (approved):

Issue #2853, opened 2024-03-04
Clearly define VC format types when issuing


woutslakhorst:
With all the work that is being done for ARFv1.2, VC data model 2.0 and data integrity, there'll be a multitude of VC formats that need to be supported. For our APIs we need a clear set of `format` and media types, etc.

woutslakhorst:
> Change current `format` enum values to `json-ld+jwt` and `json-ld+ld-proofs`

don't do this for any PEX related things. Only for the issuance API

reinkrul:
go-did's `VerifiableCredential.Format()` should also return these values (instead of `jwt_vc`), right @woutslakhorst ?

woutslakhorst:
> go-did's `VerifiableCredential.Format()` should also return these values (instead of `jwt_vc`), right @woutslakhorst ?

Yes, which probably means you have to map the formats to PEX formats somehow....

reinkrul:
> From VC data model v1.1
> JSON‑LD+JWT: The current JWT type credential. (for newer stuff, revocation and discovery)
> JSON‑LD+LD‑Proofs: The default VC type on the network

Where did you read this? I think the type/format designators are in flux currently. Can't we keep it as it currently is for the v6 release, and just add the new formats (e.g. SD), and maybe an alias for the JWT/JSON-LD VCs (and deprecate the current ones)?

woutslakhorst:
for v1.1 they aren't defined. Also PEX formats/types are different from W3C.

gerardsn:
Considering that each spec (group) currently seems to use its own convention it probably comes down to whom we want to be compatible with? I think most parties that implement the openid4vc specs use the PEX formats (and possibly the weird `*_json` types). 

What is the argument to change this now? I agree that it would be nice to have one fixed convention for this, but I don't think there is any consensus on this yet?

PR #3006, opened 2024-04-02
VCR: Correct VC formats in preparation of supporting VC data model v2

reinkrul:
Fixes https://github.com/nuts-foundation/nuts-node/issues/2853


reinkrul:
Not relevant for now

PR #2977, opened 2024-03-22
extended policy mapping for user/organization

woutslakhorst:
closes #2973 

presentation definition mapping files have been extended to allow for multiple PDs. Each PD also has a `audience_type` that defines if the PD is intended for a user or organisation.

Documentation on PEX stuff has also been added.

Issue #2901, opened 2024-03-11
Support EmployeeID authentication means for OpenID4VP


woutslakhorst:
For the 'old' flow a credential was created on the fly after the user clicked on OK.
Any information is passed to the other side for auditing purposes. We need a new (temporary) way to transfer user information to the other side as long as there's no (Q)EAA credential in a user wallet.

woutslakhorst:
The preferred way is to forward the user to a node user page (for example for starting OpenID4VP). The node should support OIDC and custom params via the initial API for development. The OIDC flow will return claims that can be used to generate a VC. When the user first goes to the node user page, the node will create a did:jwk and issue a EmployeeCred from the Organization to the did:jwk using data from OIDC ID_TOKEN. The did:jwk and credential are stored in the session store. In the OpenID4VP flow, this user wallet is also allowed to be used for matching against PEX.

This way the ephemeral did:jwk wallet can be extended to a cloud user wallet using webauthn. Development effort is significant but is most future proof.

woutslakhorst:
- [ ] design user flow, currently there's one entry point 'startOpenID4VP'. Another one will be added for OpenID4VCI. For each user flow:
  - [ ] check if user has a wallet
  - [ ] if not (default for now), create a did:jwk
  - [ ] issue a NutsEmployeeCredential to the did:jwk
  - [ ] store key and VC in session store
- [ ] extend OpenID4VP to support multiple consecutive flows (request and check)
- [ ] extend PEX configuration to allow for multiple flows.

woutslakhorst:
1 PD per scope, RP sends multiple scopes when both org and user wallet need to be used.

woutslakhorst:
An option for the verifier is to send the 2nd authorization request as a `openid4vp://?request_uri=` redirect in response to the post body. The `openid4vp` can be picked up by nuts-node and used to activate the user cloud wallet or render a QR code (because the user cloud wallet doesn't have it or the user selected to use a cross device wallet)

reinkrul:
The current `NutsEmployeeCredential` is wrong:
- We use Person/Initials, which isn't defined in the schema.org context
- The context defines "schema" as `@vocab` which causes anything to be mapped to the schema.org context (so we don't actually use it).

We should probably introduce a new `EmployeeCredential(2024?)` that reads as follows:

```json
...
```

PR #3009, opened 2024-04-02
IAM: match user-targeted OpenID4VP PDs against session-wallet

reinkrul:
This PR adds support for matching Presentation Definitions that target a "user wallet" against the authenticated user's session-bound wallet. In practice, this allows using the session-bound `EmployeeCredential` (created by #2991) in OpenID4VP alongside organization-wallet VCs.

Builds on:
- https://github.com/nuts-foundation/nuts-node/pull/2991
- https://github.com/nuts-foundation/nuts-node/pull/2977

TODO:
- [x] Enable e2e test (fails on headless Chrome Docker container, because the self-signed certificate isn't accepted)
- [x] Fix OpenID4VP e2e test (token introspection result doesn't contain mapped fields)

Part of https://github.com/nuts-foundation/nuts-node/issues/2901


auth/api/iam/api.go
```diff
@@ -362,7 +369,14 @@ func (r Wrapper) HandleAuthorizeRequest(ctx context.Context, request HandleAutho
 	case responseTypeVPToken:
 		// Options:
 		// - OpenID4VP flow, vp_token is sent in Authorization Response
-		return r.handleAuthorizeRequestFromVerifier(ctx, *ownDID, params)
+		// non-spec: wallet_owner_type is an unsigned parameter that hints whether the request targets an organization or user wallet.
+		// Requests to user wallets can be rendered as QR-code.
+		walletOwnerType := pe.WalletOwnerType(params.get("wallet_owner_type"))
```
woutslakhorst, line 374: In the initial design we talked about handling all urls beginning with `openid4vp://` as user target. This switch to a specific param, is it registered somewhere (ticket)?

auth/api/iam/generated.go
```diff
@@ -423,6 +426,2603 @@ func (a TokenIntrospectionResponse) MarshalJSON() ([]byte, error) {
 	return json.Marshal(object)
 }
 
+// RequestEditorFn  is the function signature for the RequestEditor callback function
```
woutslakhorst, line 429: These additional 2600 LoC are used for a single call in a test. Generate code to a test package so we can give that part of the code another status as *production* code

e2e-tests/browser/openid4vp_employeecredential/selfsigned.go
```diff
@@ -0,0 +1,178 @@
...
+
+type SelfSignedSession struct {
+	ID                 string
+	Contract           string
+	EmployeeIdentifier string
+	EmployeeName       string
+	EmployeeRole       string
+}
+
+func (s SelfSigned) Start(organizationDID string, employee EmployeeInfo) (*SelfSignedSession, error) {
```
woutslakhorst, line 55: Why is there auth v1 contract stuff in the openid4vp e2e tests?

woutslakhorst (changes requested):

auth/api/iam/generated.go
reinkrul: Moved to `e2e` folder.

auth/api/iam/api.go
reinkrul: Changed it to internal "dispatch". Feels kinda sketchy just calling the HTTP function, but let's see if it works.

e2e-tests/browser/openid4vp_employeecredential/docker-compose.yml
```diff
@@ -0,0 +1,28 @@
+services:
+  nodeA-backend:
+#    image: "${IMAGE_NODE_A:-nutsfoundation/nuts-node:master}"
+    image: nutsfoundation/nuts-node:local
```
woutslakhorst, line 4: remainder of local tests?

e2e-tests/browser/openid4vp_employeecredential/main_test.go
```diff
@@ -0,0 +1,123 @@
...
+	"github.com/stretchr/testify/require"
+	"os"
+	"testing"
+	"time"
+)
+
+var nodeClientConfig = core.ClientConfig{Address: "http://localhost:8081"}
+
+func init() {
+	//os.Setenv("SHOW_BROWSER", "true")
```
woutslakhorst, line 40: `// uncomment this to get feedback during development`

e2e-tests/browser/openid4vp_employeecredential/main_test.go
```diff
@@ -0,0 +1,123 @@
...
+	//os.Setenv("KEEP_BROWSER_OPEN", "true")
+}
+
+func Test_UserAccessToken_EmployeeCredential(t *testing.T) {
```
woutslakhorst, line 44: Note: Definitely shorter then our other e2e tests, but also less of an example/test on raw http requests. (We can make the same mistake in client/server which still make the test succeed).

woutslakhorst (changes requested):
small stuff left, which todo's are required for the PR?

e2e-tests/browser/openid4vp_employeecredential/main_test.go
reinkrul: We could opt to use raw(ish) HTTP requests in our e2e tests, which would definitely improve effectiveness (then the chances of a bug in the API and client will have much more chance of being found) and it would be more documentation on how the API is supposed to be used. At the cost of maintainability.

But as the number of e2e tests we have is limited, I think it's OK. But we'll probably make a mini-client in the e2e tests because you don't want to have the "create DID" HTTP request and response handling over and over again.

reinkrul:
> small stuff left, which todo's are required for the PR?

I'd say bugfix in token introspection and enabling the e2e test GH Actions/CircleCI (a manual test is a forgotten test)

e2e-tests/browser/openid4vp_employeecredential/main_test.go
reinkrul: I'd say have that refactor separate from this feature change. So merge this first, then clean up?

reinkrul:
Codeclimate is being weird again about terrible code coverage in unchanged files.

woutslakhorst (approved):

PR #3011, opened 2024-04-03
IAM: fix loading of server state

reinkrul:
After loading the session from the store, the fields became typed as `[]interface{}` and `map[string]interface{}`, making them uncastable.

Should we consider making `ServerState` a struct instead of a map, instead?

### DIFF
--- a/auth/api/iam/session.go
+++ b/auth/api/iam/session.go
@@ -19,6 +19,7 @@
 package iam
 
 import (
+	"encoding/json"
 	"github.com/nuts-foundation/go-did/did"
 	"github.com/nuts-foundation/go-did/vc"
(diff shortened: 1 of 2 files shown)


woutslakhorst:
> Should we consider making `ServerState` a struct instead of a map, instead?

We know of all the data that's being stored in the `ServerState`, there's no runtime magic. So let's make it a struct and just use normal marshalling/unmarshalling.

reinkrul:
> > Should we consider making `ServerState` a struct instead of a map, instead?
> 
> We know of all the data that's being stored in the `ServerState`, there's no runtime magic. So let's make it a struct and just use normal marshalling/unmarshalling.

Fixed

woutslakhorst (approved):

PR #3014, opened 2024-04-04
Update golang.org/x/net@v0.23.0

reinkrul:
govuln reported a CVE in 0.22.0


gerardsn (approved):
run `go mod tidy` to cleanup go.sum

gerardsn:
It looks like the docker action still resolves go 1.22.1 as the most recent version. 1.22.2 fixes the vulnerability but is less than a day old so I expect this issue to resolve itself

reinkrul:
But still in this dependency, so merging this won't hurt.

PR #3015, opened 2024-04-04
Discovery: split Discovery OAS in internal API and Server API

reinkrul:
The internal API is what people use to integrate with the Nuts node, the server API is a reflection of what the Nuts RFC specifies and mainly for implementing the Nuts node itself.

No functional changes, only split off `server.yaml` (as in "Discovery Server") API from `discovery/v1.yaml`.

Mainly improves QoL of people integrating the Nuts node.


woutslakhorst (commented):
generation probably forces you to change package, but the server definition should also be versioned....

reinkrul:
> generation probably forces you to change package, but the server definition should also be versioned....

There's no use in versioning it at this point, since the API is specified by the Nuts RFC. So we won't change it, unless the RFC changes (and the RFC currently isn't versioned).

reinkrul:
@woutslakhorst do you want to approve or comment on it?

woutslakhorst (approved):

PR #3019, opened 2024-04-05
Policy: add 'type' query parameter to /presentation_definition endpoint

reinkrul:
Required for specifying the right PD when an Authorization Code Flow results in multiple OpenID4VP requests.

Builds on https://github.com/nuts-foundation/nuts-node/pull/2977


gerardsn (commented):
only reviewed commit https://github.com/nuts-foundation/nuts-node/pull/3019/commits/b48224482e3cb94e53dcebb25a5ff4dfa9156deb

I think this just moves the problem somewhere else. Adding a generic query param for state/session to this endpoint could be used for other purposes in the future too.

`/presentation_definition?ref=123`

docs/_static/auth/iam.yaml
```diff
@@ -819,6 +823,14 @@ components:
...
           type: string
           description: Role of the End-User.
+    WalletOwnerType:
+      type: string
+      description: |
+        Wallet owner type that should fulfill the presentation definition.
+        Can either be an organization wallet or a user (personal) wallet.
+      enum:
+        - organization
+        - user
```
gerardsn, line 833: A machine can also be a user. I would stick to `person` since personal wallet is the common name for this type.

woutslakhorst:
rebase on master for a leaner PR

docs/_static/auth/iam.yaml
reinkrul: I agree; "user" is a technical term for the actual user of the Nuts node, while what was intended is the sort of entity (legal organization v.s. natural person) that holds the wallet. But since it's already used in other places, I'll make an  issue for it.

gerardsn (approved):

PR #3022, opened 2024-04-05
IAM: Remove OpenID4VP PoC code

reinkrul:
Functionality is either very stale or replaced by the "production" version of the OpenID4VP implementation.


gerardsn (approved):
nothing breaks

PR #3023, opened 2024-04-07
VCR: Allow credentials to be passed when creating PEX submission

reinkrul:
Allows matching on wallets other than the SQL wallet in VCR. Required for using user-session wallet with EmployeeCredential.

I tried refactoring `Wallet` into VC storage (`SQLWallet`) and building presentations (`Presenter`), but that just lead to another type that needed to be passed everywhere, where `Wallet` is also currently passed. So no real benefit, other than 2 smaller types instead of 1 larger.


reinkrul:
Refactor is still required: session wallet needs an alternative key resolver and store.

Issue #3031, opened 2024-04-09
IAM: Expand OAuth Token Introspection Response's `input_descriptor_constraint_id_map`


reinkrul:
The `input_descriptor_constraint_id_map` field in the response contains the fields that resulted from the constraints in the PEX PD. It can be used to add fields from VCs in the Token Introspection response, that can be used in PDP authorization decisions.

The intention was to use standard field names here, so we should expand the field onto the response, not having a separate field for it. Make sure not to overwrite existing field (e.g. `client_id`) for security reasons.

PR #3032, opened 2024-04-09
IAM: Expand InputDescriptorConstraintIdMap onto introspection response

reinkrul:
Fixes https://github.com/nuts-foundation/nuts-node/issues/3031

### DIFF
--- a/auth/api/iam/api.go
+++ b/auth/api/iam/api.go
@@ -219,26 +219,14 @@ func (r Wrapper) IntrospectAccessToken(_ context.Context, request IntrospectAcce
 	iat := int(token.IssuedAt.Unix())
 	exp := int(token.Expiration.Unix())
 	response := IntrospectAccessToken200JSONResponse{
-		Active:                         true,
-		Iat:                            &iat,
-		Exp:                            &exp,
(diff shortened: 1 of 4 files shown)


auth/api/iam/api.go
```diff
@@ -255,6 +243,22 @@ func (r Wrapper) IntrospectAccessToken(_ context.Context, request IntrospectAcce
 		log.Logger().WithError(err).Error("IntrospectAccessToken: failed to marshal presentation submission")
 		return IntrospectAccessToken200JSONResponse{}, err
 	}
+
+	if token.InputDescriptorConstraintIdMap != nil {
+		// Make sure InputDescriptorConstraintIdMap does not override existing properties (e.g. client_id)
+		// We could use a hardcoded list, but that risks getting outdated. So just remarshal into a map
+		// and check if any of the fields would be overwritten.
```
gerardsn, line 250: (nearly) all values are optional meaning that they could (somewhere in the future) be missing from the marshaled response map. An omitted claim could then be filled by a IDC_ID. Probably an acceptable edge case?

gerardsn (approved):
This should probably be documented somewhere

PR #3039, opened 2024-04-09
Add NutsURACredential to JSONLD context

reinkrul:
To progress on authorization using logical identifiers (ABAC), it would be nice to have a credential with a well-understood organization number in the JSON-LD contexts that are packaged in the Nuts node. This can then be used for ABAC (e.g. "organization with URA 123456 can access this and that") in authz policies. This would also help Localisation/Authorization PoCs that require such an attribute (without having to fumble with cumbersome JSONLD configuration/mapping).

And as long as there is no official issuer for that number, I propose packaging a credential containing such number in the Nuts node. I know there's currently not 1 number to rule them all, but URA is probably the safest bet? We can always add more credentials (e.g. AGB).

Maybe we should introduce an "experimental" JSON-LD context for such "support" contexts that we don't intend to keep forever? E.g. `https://nuts.nl/experimental/credentials/v1`


reinkrul:
Solved it for now by publishing a context on nuts-services.nl

https://github.com/reinkrul/nuts-services.nl/blob/main/jsonld/credentials/experimental.json

PR #3042, opened 2024-04-11
Refactor nuts-node chart for separate internal and external services

rolandgroen:
This revision refactors nuts-node charts to support separate internal and external services. It provides different ingress classes, annotations, hosts, and rules for external and internal services. It also adjusts `service.yaml` to accommodate these changes and provides different application URLs for internal and external services. Configurations in `values.yaml` have been altered to reflect these changes as well.

### DIFF
--- a/charts/nuts-node/values.yaml
+++ b/charts/nuts-node/values.yaml
@@ -124,9 +124,6 @@ nuts:
         address: :8081
       external:
         address: :8080
-      default:
-        cors:
-          origin: "https://chart-example.local"
     verbosity: debug
     tls:
       truststorefile: /opt/nuts/ssl/ca.pem
--- a/charts/nuts-node/values.yaml
+++ b/charts/nuts-node/values.yaml
@@ -56,7 +56,7 @@ ingress:
       # kubernetes.io/ingress.class: nginx
       # kubernetes.io/tls-acme: "true"
     hosts:
-      - host: chart-example.external
+      - host: chart-example.internal
         paths:
           - path: /
             pathType: ImplementationSpecific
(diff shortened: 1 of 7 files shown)


JorisHeadease (approved):
Kleine comment toegevoegd, niet een heel belangrijke. Ik approve de PR alvast aangezien dit geen functionele impact heeft

rolandgroen:
@gerardsn could you pls review this before we merge.

charts/nuts-node/values.yaml
```diff
@@ -99,8 +120,11 @@ nuts:
...
     datadir: /opt/nuts/data
     http:
+      internal:
+        address: :8081
+      external:
+        address: :8080
       default:
-        address: :1323
         cors:
           origin: "https://chart-example.local"
```
gerardsn, line 129: the http.default section no longer exists

gerardsn (approved):
I'm no Kubernetes expert, so just skimmed through the PR. If tested and works I'm fine merging this

charts/nuts-node/values.yaml
rolandgroen: In the files README.rst komt nog de tekst voor:

> As a general safety precaution ``auth.contractvalidators`` ignores the ``dummy`` option if configured,
requesting an access token from another node on ``/n2n/auth/v1/accesstoken`` does not return any error details,
``auth.accesstokenlifespan`` is always 60 seconds, ``http.default.cors.origin`` does not allow a wildcard (``*``),
json-ld context can only be downloaded from trusted domains configured in ``jsonld.contexts.remoteallowlist``,
and the ``internalratelimiter`` is always on.

rolandgroen: We halen hem weg uit de config.

PR #3051, opened 2024-04-15
fix DID creation with invalid tenant and default DID

### LINKED ISSUE #3044: DID: invalid did:web tenant causes nil deref
The tenant passed in the API isn't validated (e.g. a tenant containing a space causes this); this leads to a nil deref on `vdr/didweb/manager.go:130` (because the resulting DID isn't valid).:

```
parsedNewDID, _ := did.ParseDID(newDID.String()) // make sure internal state is consistent (DecodedID)
return m.create(ctx, *parsedNewDID)
```

Maybe we should limit the allowed character set and/or percent-encode the value.

woutslakhorst:
fixes #3044 

check parsing error on creation and return err.

### DIFF
--- a/vdr/didweb/manager.go
+++ b/vdr/didweb/manager.go
@@ -126,7 +126,10 @@ func (m Manager) Create(ctx context.Context, opts management.CreationOptions) (*
 		newDID = m.rootDID
 		newDID.ID += ":" + m.tenantPath + ":" + uuid.NewString()
 	}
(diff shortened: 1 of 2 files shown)


reinkrul:
Is the error meaningful enough when calling the API?

woutslakhorst:
> Is the error meaningful enough when calling the API?

There's nothing more to say than that it's invalid

reinkrul (approved):

PR #3052, opened 2024-04-16
Extend VCR an VDRv2 API clients

reinkrul:
Required by coming e2e browser test for EmployeeCredential.

Seperate PR to keep other PR smaller.

### DIFF
--- a/vcr/api/vcr/v2/client.go
+++ b/vcr/api/vcr/v2/client.go
@@ -23,6 +23,8 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"github.com/nuts-foundation/go-did/did"
+	"github.com/nuts-foundation/go-did/vc"
 	"github.com/nuts-foundation/nuts-node/core"
 	"io"
 	"net/http"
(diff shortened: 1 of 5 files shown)


woutslakhorst (commented):
Why do we need a client for this? I thought we're no longer investing in the command-line utilities?

reinkrul:
> Why do we need a client for this? I thought we're no longer investing in the command-line utilities?

The HTTP client is used in the end-to-end test, which needs to specify the DID tenant. The CLI uses that HTTP client, and because its signature changed, the CLI needed to change.

woutslakhorst:
> > Why do we need a client for this? I thought we're no longer investing in the command-line utilities?
> 
> The HTTP client is used in the end-to-end test, which needs to specify the DID tenant. The CLI uses that HTTP client, and because its signature changed, the CLI needed to change.

Why the addition of the `Load` method?

reinkrul:
> > > Why do we need a client for this? I thought we're no longer investing in the command-line utilities?
> > 
> > 
> > The HTTP client is used in the end-to-end test, which needs to specify the DID tenant. The CLI uses that HTTP client, and because its signature changed, the CLI needed to change.
> 
> Why the addition of the `Load` method?

e2e test issues then loads a NutsOrganizationCredential.

reinkrul:
> shouldn't the e2e-test be changed to load the credential through the API instead of investing in new CLI methods

I repeat:

>> Why do we need a client for this? I thought we're no longer investing in the command-line utilities?
> The HTTP client is used in the end-to-end test, which needs to specify the DID tenant. The CLI uses that HTTP client, and because its signature changed, the CLI needed to change.

If we don't want to change the CLI if the HTTP clients change, we need to remove the CLI.

woutslakhorst:
@reinkrul it's ok to change the CLI for the changed API signature, but the `Load` method is new to the client and not used from the CLI.

reinkrul:
> @reinkrul it's ok to change the CLI for the changed API signature, but the `Load` method is new to the client and not used from the CLI.

It's used by the browser test introduced by https://github.com/nuts-foundation/nuts-node/pull/3009

woutslakhorst:
Hmmm, we have to discuss what to do with code only used for test but publicly exposed. Removing it would officially require a major update. I think we should consider moving all client API code to `internal`.

reinkrul:
https://github.com/nuts-foundation/nuts-node/pull/3009 is also through, which includes this.

PR #3055, opened 2024-04-16
IAM: Support multiple consecutive OpenID4VP flows

reinkrul:
This adds support on the Authorizer Server's side to perform multiple OpenID4VP flows (to the wallet) before issuing the authz code.

Split off from https://github.com/nuts-foundation/nuts-node/pull/3009 to make that one smaller to digest later on.


auth/api/iam/session.go
```diff
@@ -39,17 +39,76 @@ type OAuthSession struct {
...
 	VerifierDID            *did.DID
+	OpenID4VPVerifier      *OpenID4VPVerifier
+}
 
-	// TODO use these 2 fields to track if all OpenID4VP flows have been concluded
-	// PresentationSubmissions tracks which PresentationSubmissions have been submitted through OpenID4VP
-	PresentationSubmissions map[string]pe.PresentationSubmission
+// OpenID4VPVerifier tracks the verifier's state of multiple OpenID4VP flows that all relate to a single OAuthSession (e.g. Authorization Code flow).
+// It's a sort of state machine with 2 states.
+type OpenID4VPVerifier struct {
```
woutslakhorst, line 51: It would be a good idea to start adding JSON tags to these session storage structs. Non in-memory storage would benefit from it.

auth/api/iam/session.go
```diff
@@ -39,17 +39,76 @@ type OAuthSession struct {
...
+	WalletDID                       did.DID
+	RequiredPresentationDefinitions pe.WalletOwnerMapping
+	// Submissions tracks which Submissions have been submitted through OpenID4VP
+	Submissions map[string]pe.PresentationSubmission
 	// Presentations tracks which VerifiablePresentations have been received through OpenID4VP
-	Presentations map[string]vc.VerifiablePresentation
+	Presentations []vc.VerifiablePresentation
```
woutslakhorst, line 57: Submissions is a map, this is a slice. Shouldn't submissions and presentations be stored in pairs?

auth/api/iam/s2s_vptoken.go
```diff
@@ -97,18 +97,35 @@ func (r Wrapper) handleS2SAccessTokenRequest(ctx context.Context, issuer did.DID
...
+				// Should be prevented by Presentation Definition author,
+				// but still check this for security reasons.
+				return nil, oauth.OAuth2Error{
+					Code:        oauth.ServerError,
+					Description: "duplicate mapped field in Presentation Definitions",
+				}
+			}
+			fieldsMap[k] = v
+		}
 	}
```
woutslakhorst, line 127: benefits from its own function for readability

auth/api/iam/openid4vp.go
```diff
@@ -106,21 +106,86 @@ func (r Wrapper) handleAuthorizeRequestFromHolder(ctx context.Context, verifier
...
+// If there are more Presentation Definitions to fulfill, the next OpenID4VP Authorization Request is sent.
+// If all Presentation Definitions are fulfilled, the OAuth2 session is completed. It then returns nil and no error.
+func (r Wrapper) nextOpenID4VPFlow(ctx context.Context, state string, session OAuthSession) (*url.URL, error) {
+	// Find next Presentation Definition to fulfill.
+	walletOwnerType, _ := session.OpenID4VPVerifier.next()
+
+	if walletOwnerType == nil {
+		// All OpenID4VP flows completed
+		return nil, errors.New("all OpenID4VP flows already completed")
+	}
```
woutslakhorst, line 175: this is a private method, no need to do this check twice (caller and here)

woutslakhorst (changes requested):
small stuff

auth/api/iam/session.go
reinkrul: Probably, yes. And `Credentials` be removed, because it's the output of `resolve(presentation, submission, definition)`

woutslakhorst (approved):

PR #3056, opened 2024-04-17
Fixed sphinx document warnings

### LINKED ISSUE #2952: check documentation for broken cross references
With the cleanup of all documentation we are bound to have broken some links. Some may need to reference the wiki and some may need to be removed.

woutslakhorst:
part of #2952


docs/pages/release_notes.rst
```diff
@@ -9,14 +9,16 @@ Peanut (v6.0.0)
 **Release date:** TBD
 **Full Changelog**: https://github.com/nuts-foundation/nuts-node/compare/v5.0.0...v6.0.0
 
-Breaking
-********
```
gerardsn, line 13: I think (based on the rst2myst conversion results) that this puts it at the same level as
```
*******************
Peanut (v6.0.0)
*******************
```
(too many *** by the way)

At the bottom of this file we sort of specified how add subsections.

gerardsn (approved):

gerardsn:
also not sure if the linked issue should be closed. We still have some documentation to move around and remove before the release.

docs/pages/release_notes.rst
woutslakhorst: no `=` vs `*`

PR #3057, opened 2024-04-17
Add sqlserver support to storage engine

rolandgroen:
This commit introduces support for the SQL Server database in the storage engine. A sqlserver driver has been added and appropriate configuration has been set up. This will allow users to use Microsoft SQL Server as one of the database options.

### DIFF
--- a/storage/engine.go
+++ b/storage/engine.go
@@ -38,6 +38,7 @@ import (
 	"gorm.io/driver/mysql"
 	"gorm.io/driver/postgres"
 	"gorm.io/driver/sqlite"
+	"gorm.io/driver/sqlserver"
 	"gorm.io/gorm"
 
 	"github.com/amacneil/dbmate/v2/pkg/dbmate"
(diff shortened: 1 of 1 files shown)


rolandgroen:
Assigning PR to @JorisHeadease to do some testing, if positive, pls add someone from nuts to review.

reinkrul:
Might be worth adding an end-to-end test, we already do this for Postgres and MySql, see `e2e-tests/oauth-flow/rfc021`.

Should be "just" a matter of creating a Docker Compose setup for MS SQL Server and adding it to the scripts.

reinkrul (changes requested):
Also add documentation on it, see https://nuts-node.readthedocs.io/en/latest/pages/deployment/storage.html

(`storage.rst`)

woutslakhorst:
> This commit introduces support for the SQL Server database in the storage engine. A sqlserver driver has been added and appropriate configuration has been set up. This will allow users to use Microsoft SQL Server as one of the database options.

IS there a specific reason to add it, since Azure allows for MySQL and Postgres hosting options out of the box?

reinkrul:
Now supported; see https://github.com/nuts-foundation/nuts-node/pull/3080

PR #3065, opened 2024-04-19
Storage: support SQL session storage

reinkrul:
To allow clustering without having to deploy another database/storage server (e.g. Redis).

@woutslakhorst had concerns about performance of session storage in SQL, maybe we can reason about that whether we want it.

Renamed `session.go` to `session_inmemory.go`.

TODO:

- [x] Allow it to be configured
- [ ] Documentation
- [x] Cleanup job to remove expired entries


woutslakhorst:
It's all about the propagation time. A dedicated session cache could be setup to return after a `set` after all nodes are synchronised. For a SQL setup this is defined by the cluster setup (master/master, write master/read slaves, etc). You have to make sure that a combination of propagation and connection setup will make sure clients won't experience any problems. By supporting SQL, you won't get this by default.

reinkrul:
Let's not do this for now

bramwesselo:
In a high load scenario, I see your point @woutslakhorst. However, for deployment in a Kubernetes-like environments (for high availability, not high load), this PR could avoid the requirement for Redis (and lower the bar for adoption).

PR #3066, opened 2024-04-22
Docs: fix list styles

reinkrul:
Fixes https://github.com/nuts-foundation/nuts-node/issues/2952 (there were some leftovers)


woutslakhorst (approved):
I believe @gerardsn wanted to keep the issue open until v6 is done.

PR #3071, opened 2024-04-23
Storage: support Redis session storage

rolandgroen:
Rework of the https://github.com/nuts-foundation/nuts-node/pull/3065 PR, now for storing the session in Redis.

 - Added a new Session Type: RedisSessionStoreType 
 - Added configuration logic for Redis and RedisSessionStoreType
 - Implemented SessionDatabase and SessionStore for Redis

Remark / Notes

- Needed to patch go-stoabs to support TTL functions of redis, PR is here: https://github.com/nuts-foundation/go-stoabs/pull/97


reinkrul:
What were the considerations to use `go-stoabs`?

Note that you don't need to use `go-stoabs` per se, since it was meant as abstraction for key-value stores for the Nuts node. It's only used in v5 features (and we might want to scrap it when doing away with did:nuts/gRPC network). By using it, you (might) lose Redis-specific features that could be useful (e.g. specific data types).

I'm not completely against using it if it brings advantages, but if it's just as easy to directly use the Redis driver, it could be preferred.

@woutslakhorst ?

reinkrul (changes requested):

woutslakhorst:
> What were the considerations to use `go-stoabs`?
> 
> Note that you don't need to use `go-stoabs` per se, since it was meant as abstraction for key-value stores for the Nuts node. It's only used in v5 features (and we might want to scrap it when doing away with did:nuts/gRPC network). By using it, you (might) lose Redis-specific features that could be useful (e.g. specific data types).
> 
> I'm not completely against using it if it brings advantages, but if it's just as easy to directly use the Redis driver, it could be preferred.
> 
> Also important to note that, to enable session storage in Redis, you need to configure Redis storage (doh), which causes did:nuts/gRPC network data to be stored in Redis instead of on disk.
> 
> @woutslakhorst ?

I would prefer to have a redis session storage directly use the redis library. This enables you to use the key/value options with automatic expiry. Furthermore go-stoabs was designed for persistent storage.

rolandgroen:
I didn't know the rationale behind go-stoas, so I just aligned to the "way of working" with Redis.

I agree about the config split, you might even want to configure another Redis for sessions than for the network. The config will get a little more complex, and the session.type can be come implicit:

```
storage:
  redis:
    address: 127.0.0.1:6379
  sql:
    connection: postgres://..:..@127.0.0.1:5432/db
  session:
    redis:
      address: 127.0.0.1:6379
```

rolandgroen:
Refactored the code to directly use the redis library.

go.mod
```diff
@@ -27,7 +27,7 @@ require (
 	github.com/nuts-foundation/crypto-ecies v0.0.0-20211207143025-5b84f9efce2b
 	github.com/nuts-foundation/go-did v0.13.0
 	github.com/nuts-foundation/go-leia/v4 v4.0.1
-	github.com/nuts-foundation/go-stoabs v1.9.0
+	github.com/nuts-foundation/go-stoabs v1.9.1-0.20240423163219-4ceec5cef809
```
reinkrul, line 30: I think we can revert this now?

reinkrul (commented):
I'm OK with the general approach (one last thing remaining). There's some duplicity in `storage.redis` and `session.redis` configuration now, but when v5 features are removed, `storage.redis` will be removed as well.

go.mod
rolandgroen: Yes, rolled back

reinkrul:
I took the liberty of fixing some things I found (some simplifications, and use support for expiry from Miniredis). I hope you don't mind.

reinkrul (approved):

Issue #2712, opened 2024-01-08
Support request_uri in OpenID4VP flow


woutslakhorst:
The current happy flow uses a request in the query param.

woutslakhorst:
request_uri changes have been merged, checkout https://github.com/openid/OpenID4VP for details

woutslakhorst:
@gerardsn #3074 added support for the `request_uri`, does it still need to be activated or supported through metadata? Or it's not yet needed until we add support for QR codes?

PR #3074, opened 2024-04-30
RFC9101 request_uri

gerardsn:
part of #2712

this adds the `request_uri` param as specced in [RFC9101](https://www.rfc-editor.org/rfc/rfc9101.html)

implemented separate from the OpenID4VP extension using `request_uri_method` to keep the PRs small

### DIFF
--- a/auth/api/iam/api.go
+++ b/auth/api/iam/api.go
@@ -349,14 +349,14 @@ func (r Wrapper) HandleAuthorizeRequest(ctx context.Context, request HandleAutho
 // GetRequestJWT returns the Request Object referenced as 'request_uri' in an authorization request.
 // RFC9101: The OAuth 2.0 Authorization Framework: JWT-Secured Authorization Request (JAR).
 func (r Wrapper) GetRequestJWT(ctx context.Context, request GetRequestJWTRequestObject) (GetRequestJWTResponseObject, error) {
-	signedRequestObject := new([]byte)
-	err := r.authzRequestObjectStore().Get(request.Id, signedRequestObject)
+	var signedRequestObject string
+	err := r.authzRequestObjectStore().Get(request.Id, &signedRequestObject)
 	if err != nil {
 		return nil, err
 	}
 	return GetRequestJWT200ApplicationoauthAuthzReqJwtResponse{
-		Body:          bytes.NewReader(*signedRequestObject),
-		ContentLength: int64(len(*signedRequestObject)),
+		Body:          bytes.NewReader([]byte(signedRequestObject)),
+		ContentLength: int64(len(signedRequestObject)),
 	}, nil
 }
 
@@ -855,7 +855,6 @@ func (r Wrapper) openidIssuerEndpoints(ctx context.Context, issuerDid did.DID) (
 // - jwt.Audience
 // - nonce
 // any of these params can be overridden by the requestObjectModifier.
-// if requestID can be provided to generate an expected request_uri (/oauth2/request.jwt/{requestID}), a random value is used if not provided.
 func (r Wrapper) CreateAuthorizationRequest(ctx context.Context, client did.DID, server did.DID, modifier requestObjectModifier) (*url.URL, error) {
 	// we want to make a call according to §4.1.1 of RFC6749, https://www.rfc-editor.org/rfc/rfc6749.html#section-4.1.1
 	// The URL should be listed in the verifier metadata under the "authorization_endpoint" key
@@ -889,7 +888,7 @@ func (r Wrapper) CreateAuthorizationRequest(ctx context.Context, client did.DID,
 			return nil, fmt.Errorf("failed to sign authorization request: %w", err)
 		}
 		requestID := cryptoNuts.GenerateNonce()
-		if err := r.authzRequestObjectStore().Put(requestID, []byte(token)); err != nil {
+		if err := r.authzRequestObjectStore().Put(requestID, token); err != nil {
 			return nil, err
 		}
 		didURL, err := didweb.DIDToURL(client)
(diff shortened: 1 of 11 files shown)


auth/api/iam/api.go
```diff
@@ -316,22 +297,12 @@ func (r Wrapper) HandleAuthorizeRequest(ctx context.Context, request HandleAutho
 	// Workaround: deepmap codegen doesn't support dynamic query parameters.
 	//             See https://github.com/deepmap/oapi-codegen/issues/1129
 	httpRequest := ctx.Value(httpRequestContextKey).(*http.Request)
-	params := parseQueryParams(httpRequest.URL.Query())
-	clientId := params.get(oauth.ClientIDParam)
+	qParams := httpRequest.URL.Query()
```
reinkrul, line 300: qParams and params are now somewhat ambiguous, maybe `queryParams` and `requestObject`

auth/api/iam/api.go
```diff
@@ -373,6 +344,50 @@ func (r Wrapper) HandleAuthorizeRequest(ctx context.Context, request HandleAutho
...
+		if err != nil {
+			return nil, oauth.OAuth2Error{Code: oauth.InvalidRequestURI, Description: "failed to get Request Object", InternalError: err}
+		}
+	} else {
+		// require_signed_request_object is true, so we reject anything that isn't
+		return nil, oauth.OAuth2Error{Code: oauth.InvalidRequest, Description: "authorization request are required to use signed request objects (RFC9101)"}
+	}
+
+	// already OAuth errors
+	return r.validateJARRequest(ctx, rawRequestObject, q.Get(oauth.ClientIDParam))
```
reinkrul, line 388: shouldn't there be an existence check on the `oauth.ClientIDParam` parameter?

auth/api/iam/api.go
```diff
@@ -864,28 +880,27 @@ func (r Wrapper) CreateAuthorizationRequest(ctx context.Context, client did.DID,
...
-		}
-
-		token, err := r.jwtSigner.SignJWT(ctx, paramsAny, nil, keyId.String())
+		requestObjectParams := createRequestObject(client, server, modifier)
+		// TODO: signature type produced here must be in metadata.RequestObjectSigningAlgValuesSupported
+		token, err := r.jwtSigner.SignJWT(ctx, requestObjectParams, nil, keyId.String())
 		if err != nil {
 			return nil, fmt.Errorf("failed to sign authorization request: %w", err)
 		}
+		uriID := cryptoNuts.GenerateNonce()
```
reinkrul, line 889: `uriID` -> `requestID` or `requestObjectID`?

auth/api/iam/api.go
```diff
@@ -864,28 +880,27 @@ func (r Wrapper) CreateAuthorizationRequest(ctx context.Context, client did.DID,
...
+		if err := r.authzRequestObjectStore().Put(uriID, []byte(token)); err != nil {
```
reinkrul, line 890: you can just Put() it as string value?

auth/api/iam/api.go
```diff
@@ -953,6 +991,13 @@ func (r Wrapper) accessTokenServerStore() storage.SessionStore {
 	return r.storageEngine.GetSessionDatabase().GetStore(accessTokenValidity, "serveraccesstoken")
 }
 
+var oauthRequestObjectKey = []string{"oauth", "requestobject"}
```
reinkrul, line 994: can we put these together somewhere?

docs/_static/auth/iam.yaml
```diff
@@ -145,7 +145,60 @@ paths:
...
+        in: path
+        required: true
+        description: Identifier of the Request Object
+        schema:
+          type: string
+    get:
+      summary: Get Request Object referenced in an authorization request to the Authorization Server.
+      description: |
+        Get the Request Object containing the OAuth 2.0 authorization request parameters, including extension parameters.
+        The Request Object is a JWT with signature (JWS) and may be encrypted (JWE).
```
reinkrul, line 160: (but we don't support JWE)

auth/client/iam/interface.go
```diff
@@ -50,4 +50,6 @@ type Client interface {
 	AccessTokenOid4vci(ctx context.Context, clientId string, tokenEndpoint string, redirectUri string, code string, pkceCodeVerifier *string) (*oauth.Oid4vciTokenResponse, error)
 
 	VerifiableCredentials(ctx context.Context, credentialEndpoint string, accessToken string, proofJWT string) (*CredentialResponse, error)
+	// RequestObject is returned from the authorization request's 'request_uri' defined in RFC9101.
```
reinkrul, line 53: newline

reinkrul (changes requested):

auth/api/iam/api.go
gerardsn: `validateJARRequest` compares the value to the singerKID, so will fail if missing or empty

auth/api/iam/api.go
gerardsn: now `requestID` to match the API

auth/api/iam/api.go
gerardsn: it is used as []byte after Get, so needs to be cast somewhere.

auth/client/iam/interface.go
gerardsn: ? newlines above the other interface methods should be replaced with a godoc.

auth/api/iam/api.go
gerardsn: moved to the top of the file to match `openid4vp.go` and `user.go`. 

Another option would be to move all session stores to a separate file? We currently have 10.

reinkrul (approved):

PR #3077, opened 2024-05-01
VCR: Deduplicate AccessToken functions in IAM client

reinkrul:
There was an OpenID4VCI-specific `Oidc4vciAccessToken()` function in the IAM client that did the same as the "regular" `AccessToken()` function. I removed the first.

Also made the `TokenResponse` type basic again, specifying only the core OAuth2 parameters. Then, with an additionalProperties-esque construct, OpenID4VCI (and the weird "status" for internal access token request status calls) can use the same TokenResponse type.

### DIFF
--- a/auth/oauth/types.go
+++ b/auth/oauth/types.go
@@ -28,7 +28,7 @@ import (
 // this file contains constants, variables and helper functions for OAuth related code
 
 // TokenResponse is the OAuth access token response.
-// Through WithParam() and GetString() additional parameters (for OpenID4VCI, for instance) can be set and retrieved.
+// Through With() and Get() additional parameters (for OpenID4VCI, for instance) can be set and retrieved.
 type TokenResponse struct {
 	AccessToken string  `json:"access_token"`
 	ExpiresIn   *int    `json:"expires_in,omitempty"`
@@ -75,9 +75,10 @@ func (t TokenResponse) MarshalJSON() ([]byte, error) {
 	return json.Marshal(result)
 }
 
-// Set adds an additional parameter to the token response.
+// With adds a parameter to the token response.
 // It's a builder-style function.
-func (t *TokenResponse) WithParam(key string, value interface{}) *TokenResponse {
+// It should not be used to set any of the base parameters (access_token, expires_in, token_type, scope).
+func (t *TokenResponse) With(key string, value interface{}) *TokenResponse {
 	if t.additionalParams == nil {
 		t.additionalParams = make(map[string]interface{})
 	}
@@ -85,18 +86,19 @@ func (t *TokenResponse) WithParam(key string, value interface{}) *TokenResponse 
 	return t
 }
 
-// GetString returns the value of the additional parameter with the given key as a string.
-// If the key does not exist or the value is not a string, nil is returned.
-func (t TokenResponse) GetString(key string) *string {
+// Get returns the value of the additional parameter with the given key as a string.
+// If the key does not exist or the value is not a string, it returns an empty string.
+// It should not be used to get any of the base parameters (access_token, expires_in, token_type, scope).
+func (t TokenResponse) Get(key string) string {
 	if t.additionalParams == nil {
-		return nil
+		return ""
 	}
 	if val, ok := t.additionalParams[key]; ok {
 		if str, ok := val.(string); ok {
-			return &str
+			return str
 		}
 	}
-	return nil
+	return ""
 }
 
 const (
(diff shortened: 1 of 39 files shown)


auth/oauth/types.go
```diff
@@ -20,20 +20,83 @@
...
+	if t.additionalParams == nil {
+		t.additionalParams = make(map[string]interface{})
+	}
+	t.additionalParams[key] = value
+	return t
+}
+
+// GetString returns the value of the additional parameter with the given key as a string.
+// If the key does not exist or the value is not a string, nil is returned.
+func (t TokenResponse) GetString(key string) *string {
```
gerardsn, line 90: `Get` or `GetParam`, `String` is obvious from the return type.

auth/oauth/types.go
```diff
@@ -20,20 +20,83 @@
...
+	if t.additionalParams == nil {
+		return nil
+	}
+	if val, ok := t.additionalParams[key]; ok {
+		if str, ok := val.(string); ok {
+			return &str
```
gerardsn, line 96: why not return an error? it does't really matter if you check for nil or an error, except a missing nil check causes panics. I think there are several dereferences in this PR that could panic

auth/oauth/types.go
```diff
@@ -20,20 +20,83 @@
...
+	result["expires_in"] = t.ExpiresIn
+	result["token_type"] = t.TokenType
+	result["scope"] = t.Scope
+
+	return json.Marshal(result)
+}
+
+// Set adds an additional parameter to the token response.
+// It's a builder-style function.
+func (t *TokenResponse) WithParam(key string, value interface{}) *TokenResponse {
```
gerardsn, line 80: is using this to set keys `access_token`, `expires_in`, `token_type`, or `scope` incorrect usage (godoc) or should it set the struct param? (same for the getter)

vcr/api/openid4vci/v0/issuer.go
```diff
@@ -131,10 +131,9 @@ func (w Wrapper) RequestAccessToken(ctx context.Context, request RequestAccessTo
 		return nil, err
 	}
 	expiresIn := int(issuer.TokenTTL.Seconds())
-	return RequestAccessToken200JSONResponse(TokenResponse{
+	return RequestAccessToken200JSONResponse(*(&TokenResponse{
```
gerardsn, line 134: code smell? `func NewTokenResponse(...) *TokenResponse {}`

vcr/holder/openid.go
```diff
@@ -200,7 +200,7 @@ func (h *openidHandler) retrieveCredential(ctx context.Context, issuerClient ope
 	claims := map[string]interface{}{
 		"aud":   issuerClient.Metadata().CredentialIssuer,
 		"iat":   nowFunc().Unix(),
-		"nonce": tokenResponse.CNonce,
+		"nonce": *tokenResponse.GetString("c_nonce"),
```
gerardsn, line 203: panic

vcr/api/openid4vci/v0/issuer.go
```diff
@@ -131,10 +131,9 @@ func (w Wrapper) RequestAccessToken(ctx context.Context, request RequestAccessTo
...
 		AccessToken: accessToken,
-		CNonce:      &cNonce,
 		ExpiresIn:   &expiresIn,
 		TokenType:   "bearer",
-	}), nil
+	}).WithParam("c_nonce", cNonce)), nil
```
gerardsn, line 138: is there no `c_nonce` contant?

auth/oauth/types.go
gerardsn: or maybe return an empty string. For the authz request params I know that an empty value should be ignored, so  assuming that the same holds here.

gerardsn (changes requested):
good improvement :+1:

vcr/api/openid4vci/v0/issuer.go
reinkrul: probably better, yes...

vcr/holder/openid.go
reinkrul: checked in the calling func. But i'll change it to an empty string

vcr/api/openid4vci/v0/issuer.go
reinkrul: i'll make one

auth/oauth/types.go
reinkrul: `Param` to make it in sync with `WithParam`

auth/oauth/types.go
reinkrul: Illegal usage. I considered allowing it, but that leads to a lot of extra code and basically changes it to a map with helper functions that validate. It's how the JWX library works (with headers, for instance), but although it's elegant, usage is awkward because of the errors that can be returned.

So `With` and `Get` are only for the parameters not defined in the OAuth2 RFC. I added that as godoc.

mock/mock_echo.go
gerardsn, line 1: replace all error?

auth/oauth/types_test.go
```diff
@@ -64,3 +65,29 @@ func TestIssuerIdToWellKnown(t *testing.T) {
...
+		var actual TokenResponse
+		err := json.Unmarshal(data, &actual)
+		require.NoError(t, err)
+		assert.Equal(t, expected, actual)
+
+		assert.Equal(t, actual.Get("c_nonce"), "hello")
+	})
+}
+
+func TestTokenResponse_Param(t *testing.T) {
```
gerardsn, line 88:
```suggestion
func TestTokenResponse_Get(t *testing.T) {
```

gerardsn (approved):
approval conditional on fixing the comments

mock/mock_echo.go
reinkrul: oops

reinkrul:
But less code paths, so easier to understand?

PR #3078, opened 2024-05-01
Refactor JAR logic

gerardsn:
this change allows mocking and prepares to for adding `request_uri_method`

part of #2712


auth/api/iam/jar.go
```diff
@@ -0,0 +1,172 @@
...
+
+var _ JAR = &jar{}
+
+type jar struct {
+	auth        auth.AuthenticationServices
+	jwtSigner   cryptoNuts.JWTSigner
+	keyResolver resolver.KeyResolver
+}
+
+type JAR interface {
```
reinkrul, line 47: Maybe split it in JARCreater and JARParser? Since I'd expected `JAR` to be the signed request object itself

auth/api/iam/jar.go
```diff
@@ -0,0 +1,172 @@
...
+	"github.com/nuts-foundation/nuts-node/auth"
+	"github.com/nuts-foundation/nuts-node/auth/oauth"
+	cryptoNuts "github.com/nuts-foundation/nuts-node/crypto"
+	"github.com/nuts-foundation/nuts-node/vdr/resolver"
+)
+
+// requestObjectModifier is a function that modifies the Claims/params of an unsigned or signed (JWT) OAuth2 request
+type requestObjectModifier func(claims map[string]string)
+
+type requestObject struct {
```
reinkrul, line 20: why not have a SignedRequestObject (which is always signed)? Then; `jar.Sign(parameters) -> signedRequestObject`

(then less can go wrong when forgetting checks/using the incorrect state)

auth/api/iam/jar.go
```diff
@@ -0,0 +1,172 @@
...
+	// By default, it adds the following parameters:
+	// - client_id
+	// - jwt.Issuer
+	// - jwt.Audience (if server is not nil)
+	// - nonce
+	Create(client did.DID, server *did.DID, modifier requestObjectModifier) *requestObject
+	// Sign the requestObject, which is available on requestObject.Token.
+	// Returns an error if the requestObject already contains a signed JWT.
+	// TODO: check if signature type of client is supported by the AS/wallet.
+	Sign(ctx context.Context, ro *requestObject) error
```
reinkrul, line 58: could return a signed requestobject

auth/api/iam/jar.go
reinkrul: Does this follow the design pattern Wout used in his DPoP PR (can it?)? Since it's sort of the same thing/principle

reinkrul (changes requested):

docs/_static/auth/iam.yaml
```diff
@@ -145,8 +145,17 @@ paths:
               schema:
                 type: string
                 format: uri
-  /oauth2/request.jwt/{id}:
+  /oauth2/{did}/request.jwt/{id}:
```
gerardsn, line 148: technically we don't need the DID in the path, but all other paths have it...

auth/api/iam/jar.go
gerardsn: signing now happens when the request object is fetched, so this now returns the token as sting. This change allows for consistent behavior in both get/post `request_uri_method`s. It also means the signed token is no longer part of the state / in storage. 

Signing now happens on the external call, and not the internal call. This probably means the audit on signing is in the wrong place again?

auth/api/iam/jar.go
gerardsn: I've left this as is since the goal is to logically group some methods together so they can be mocked. Might change when the `request_uri_method=post` is added. (next PR)

As for comparison to the `DPoP` PR. I see the resemblance between the two, but I don't want to mirror it since that requires moving `JAR` to the `crypto` module. The `JWTSigner` interface currently meets the needs we had for `did:nuts`, but not for `did:web` I think that interface should be updated so we don't pollute the `crypto` package.

docs/_static/auth/iam.yaml
reinkrul: The same applies to a OpenID4VCI endpoint, so I don't think it needs to be per se. One more thing to validate...

reinkrul (approved):

PR #3080, opened 2024-05-01
MS SQL Server support

reinkrul:
TODO:
- [x] Revert ID length changes (200 -> 36). They are incorrect: 36 is a GUID, when larger they come from an external source (so 36 is too small). Need to come up with alternative fix.
- [x] Documentation
- [x] Missing test coverage

Test coverage is somewhat lacking, but DB-specific functionality is tested in the e2e test.


woutslakhorst:
All is green, anything specific to look at?

woutslakhorst (approved):
Too bad about the text type difference between DBs

PR #3083, opened 2024-05-03
Code cleanup

reinkrul:
Fixes https://github.com/nuts-foundation/nuts-node/issues/2966

Only thing left is to upgrade the JWT middleware:

```
http/engine.go:224:18: SA1019: middleware.JWTWithConfig is deprecated: Please use https://github.com/labstack/echo-jwt instead (staticcheck)
                echoServer.Use(middleware.JWTWithConfig(middleware.JWTConfig{
```

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,0 +1,8 @@
+run:
+  tests: false
+issues:
+  exclude-files:
+    - ".*/generated.go"
+    - ".*/.*_test.go"
+    - ".*/test.go"
+    - "docs/.*"
--- a/auth/api/iam/types.go
+++ b/auth/api/iam/types.go
@@ -25,7 +25,6 @@ import (
 	"github.com/nuts-foundation/nuts-node/vcr/pe"
 	"github.com/nuts-foundation/nuts-node/vdr/resolver"
 	"net/http"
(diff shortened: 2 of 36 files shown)


auth/api/iam/types.go
gerardsn, line 1: I think unused params in this section are a sign of missing features, not dead code.

We only check for `clientMetadataURIParam` and `presentationDefUriParam` because the nuts node creates these, but we should also check for `clientMetadataParam` and `presentationDefParam` in case we're talking to some other system.

.golangci.yml
gerardsn, line 1: what does this do?

auth/api/iam/types.go
reinkrul: If there's missing features, there should be issues, not dead code? As it seems you have it quite clear in your mind, can you create issues for them?

.golangci.yml
reinkrul: Ignore test code

auth/api/iam/types.go
gerardsn: #3084 
keep `clientMetadataParam` and `presentationDefParam`, the others can be removed

gerardsn (approved):

Issue #2867, opened 2024-03-06
Auth: split IAM OpenAPI spec into OAuth/DID/.well-known and API endpoints


reinkrul:
Only the API endpoints are relevant from integration perspective, so should be in the API reference documentation. The rest isn't relevant, only for Nuts developers. We should probably split them into 2 files.

woutslakhorst:
Split API spec into 2 'partials', create a new root OAS for swagger docs which only imports 1 partial. The current OAS that is used for code generation should import both.

Same for discovery service APIs

reinkrul:
When looking at the Discovery Server:

Splitting it is a bit awkward with pure OAS: `$ref` can only be used for parameters, components specific paths. Not for including complete paths. We could preprocess the OAS files during the `make gen-api` operation, concatenating the 2. But then, generated clients would still contain endpoints from both OAS files.

This would be confusing, since the "external" endpoints aren't in the API reference, but they're still there in the generated code. So, there are 2 "clean" solutions:

- Make 2 separate OpenAPI files, 1 for the internal endpoints (to be integrated with) and 1 with the external endpoints (only used by other Nuts nodes). Leads to more OpenAPI files a separation, so more complexity there.
- Make 1 OpenAPI file with proper tagging of operations: some code generators create a separate class/type for each tag (at least OpenAPI Generator Tools does).

woutslakhorst:
I think separation is better, documentation wise. The server package is basically a n2n API, which should also be versioned.

reinkrul:
> I think separation is better, documentation wise. The server package is basically a n2n API, which should also be versioned.

The versioning should actually follow the RFC versioning (which isn't versioned atm?)

woutslakhorst:
@reinkrul #3015 has been merged, status of issue?

reinkrul:
I want to do this when there's no other WIP in the API layer, as it will result in lots of merge conflicts otherwise

PR #3097, opened 2024-05-07
IAM: Split internal OAS file and OAuth2/OpenID4XX OAS file

reinkrul:
I tried refactoring them into 2 completely separate files, but that became far to complex. This is a simpler approach in which we combine the 2 APIs, but only serve the parties integrating with the Nuts node the OAS with the internal API endpoints.

Fixes https://github.com/nuts-foundation/nuts-node/issues/2867


codegen/oas-merge/main.go
gerardsn, line 1: untested

gerardsn (commented):
LGTM but not sure if this has any unintended side effects, so just commenting

gerardsn (approved):

Issue #3063, opened 2024-04-19
Check nonce re-usage prevention


woutslakhorst:
In lots of OAuth2 protocol parts we need to check a nonce is used once. This is not done everywhere (yet)

gerardsn:
this applies to other `SessionStore`s too, like `userRedirectStore` and `authzRequestObjectStore`. 
Perhaps add a convenience function on the `SessionStore` interface that combines `Get` and `Delete`, or let `Delete` also return the value...

woutslakhorst:
> or let Delete also return the value...

sounds like a valuable addition

gerardsn:
this consists of 2 parts
1. burn single use entries from local session storage after it has been used
2. temporarily store nonce/jti values received from other parties to prevent replay

The second is a bit harder since it requires going through all code/flows to see where these values are received.

PR #3116, opened 2024-05-14
Add SessionStore.GetAndDelete

gerardsn:
part of #3063 

I chose this solution over changing `Delete(key string) (interface{}, error)` since the returned interface is always going to be an `[]byte` that still needs to be marshaled to the target object.

### DIFF
--- a/storage/session_redis.go
+++ b/storage/session_redis.go
@@ -22,10 +22,11 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
-	"github.com/nuts-foundation/nuts-node/storage/log"
-	"github.com/redis/go-redis/v9"
 	"strings"
 	"time"
+
+	"github.com/nuts-foundation/nuts-node/storage/log"
+	"github.com/redis/go-redis/v9"
 )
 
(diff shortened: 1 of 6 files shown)


storage/session_redis.go
```diff
@@ -96,7 +97,18 @@ func (s redisSessionStore) Put(key string, value interface{}) error {
...
+func (s redisSessionStore) GetAndDelete(key string, target interface{}) error {
+	// GetDel requires redis-server version >= 6.2.0.
+	result, err := s.client.GetDel(context.Background(), s.toRedisKey(key)).Result()
+	if err != nil {
+		if errors.Is(redis.Nil, err) {
+			return ErrNotFound
+		}
+		return err
+	}
+	return json.Unmarshal([]byte(result), target)
```
reinkrul, line 111: semantics differ from in-memory store; if unmarshal fails here it's still deleted. But I guess that's okay.

reinkrul (approved):

storage/session_redis.go
gerardsn: if unmarshal fails the data is corrupted and therefore unusable. no harm in deleting it

PR #3120, opened 2024-05-17
Crypto: alter Storage interface to create keys inside key store

reinkrul:
Instead of generating one in application space, and then importing it into the key storage. This allows supporting unexportable keys, meaning keys that never leave the key store.

This is a preparation PR for supporting Azure Key Vault directly.

Kept the `SavePrivateKey` functionality since it's still used for migrating from one store to another.

### DIFF
--- a/crypto/storage/spi/interface.go
+++ b/crypto/storage/spi/interface.go
@@ -104,12 +104,17 @@ func (pke PublicKeyEntry) JWK() jwk.Key {
 	return pke.parsedJWK
 }
 
+// GenerateAndStore generates a new key pair and stores it in the provided storage.
 func GenerateAndStore(ctx context.Context, store Storage, namingFunc func(crypto.PublicKey) (string, error)) (crypto.PublicKey, string, error) {
 	keyPair, kid, err := GenerateKeyPairAndKID(namingFunc)
 	if err != nil {
 		return nil, "", err
 	}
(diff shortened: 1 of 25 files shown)


crypto/test/keys.go
```diff
@@ -39,3 +41,16 @@ func GenerateECKey() *ecdsa.PrivateKey {
...
 }
+
+// StringNamingFunc can be used to give a key a simple string name
+func StringNamingFunc(name string) storage.KIDNamingFunc {
+	return func(key crypto.PublicKey) (string, error) {
+		return name, nil
+	}
+}
+
+func ErrorNamingFunc(err error) storage.KIDNamingFunc {
```
gerardsn, line 52: replacing `KIDNamingFunc` with `func(key crypto.PublicKey) (string, error)` removes cyclical dependencies so `KIDNamingFunc` can be with the rest in `spi` (or `storage`)

gerardsn (approved):

crypto/storage/spi/interface.go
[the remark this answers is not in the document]
reinkrul: I rescoped "interface" to meaning "the thing you interface with" (haha). Jokes aside, I'll see if I can remove this package (not part of this PR though).

crypto/test/keys.go
reinkrul: -60 lines changed 👍

PR #3123, opened 2024-05-17
Burn nonce type SessionStore entries after first use

### LINKED ISSUE #3063: Check nonce re-usage prevention

gerardsn:
closes #3063 

rewrote nonce validation in `handleAuthorizeResponseSubmission` since it assumed the first VP contained the "correct" nonce and later on did a second validation over all VPs

### DIFF
--- a/auth/api/iam/api.go
+++ b/auth/api/iam/api.go
@@ -239,12 +239,10 @@ func (r Wrapper) RetrieveAccessToken(_ context.Context, request RetrieveAccessTo
 	}
 	// access token is active, return to caller and delete access token from store
 	// change this when tokens can be cached
-	defer func() {
-		err = r.accessTokenClientStore().Delete(request.SessionID)
-		if err != nil {
-			log.Logger().WithError(err).Warn("Failed to delete access token")
-		}
-	}()
+	err = r.accessTokenClientStore().Delete(request.SessionID)
+	if err != nil {
+		log.Logger().WithError(err).Warn("Failed to delete access token")
+	}
 	// return access token
 	return RetrieveAccessToken200JSONResponse(token), nil
 }
--- a/auth/api/iam/api.go
+++ b/auth/api/iam/api.go
@@ -237,12 +237,14 @@ func (r Wrapper) RetrieveAccessToken(_ context.Context, request RetrieveAccessTo
 		// return pending status
 		return RetrieveAccessToken200JSONResponse(token), nil
 	}
-	// delete access token from store
+	// access token is active, return to caller and delete access token from store
 	// change this when tokens can be cached
-	err = r.accessTokenClientStore().Delete(request.SessionID)
-	if err != nil {
-		return nil, err
-	}
+	defer func() {
+		err = r.accessTokenClientStore().Delete(request.SessionID)
+		if err != nil {
+			log.Logger().WithError(err).Warn("Failed to delete access token")
+		}
+	}()
 	// return access token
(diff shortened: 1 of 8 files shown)


auth/api/iam/api.go
```diff
@@ -236,12 +237,14 @@ func (r Wrapper) RetrieveAccessToken(_ context.Context, request RetrieveAccessTo
...
 	}
-	// delete access token from store
+	// access token is active, return to caller and delete access token from store
 	// change this when tokens can be cached
-	err = r.accessTokenClientStore().Delete(request.SessionID)
-	if err != nil {
-		return nil, err
-	}
+	defer func() {
+		err = r.accessTokenClientStore().Delete(request.SessionID)
```
reinkrul, line 243: why do this in a deferred func? As you just log the error and return token in any case, there's no difference in doing it  deferred or not?

auth/api/iam/api.go
```diff
@@ -384,8 +387,7 @@ func (r Wrapper) handleAuthorizeRequest(ctx context.Context, ownDID did.DID, req
 // RFC9101: The OAuth 2.0 Authorization Framework: JWT-Secured Authorization Request (JAR).
 func (r Wrapper) RequestJWTByGet(ctx context.Context, request RequestJWTByGetRequestObject) (RequestJWTByGetResponseObject, error) {
 	ro := new(jarRequest)
-	// TODO: burn request object to prevent DoS through signing requests https://github.com/nuts-foundation/nuts-node/issues/3063
-	err := r.authzRequestObjectStore().Get(request.Id, ro)
+	err := r.authzRequestObjectStore().GetAndDelete(request.Id, ro)
```
reinkrul, line 390: is this principle mentioned in the RFC (request object can only be retrieved once)? Since from HTTP perspective, HTTP GET requests are non-mutative, meaning they should be executable multiple times without side effects, and without the response changing. My thought; could there be clients that try to retrieve it multiple times (for whatever reason)?

reinkrul (commented):
I think checking of `wallet_nonce` and OpenID4VCI's `c_nonce` is missing

auth/api/iam/api.go
gerardsn: This is a fallback for when it fail before it gets to the GetAndDelete method, to make sure the session is always burned

auth/api/iam/api.go
gerardsn: the spec says it should exist short-ish, or once if it contains personal identifiable information. Considering this results in a signing operation I chose short-ish==once

gerardsn:
> I think checking of `wallet_nonce` and OpenID4VCI's `c_nonce` is missing

Our wallet/AS does not add a `wallet_nonce`, the client adds it to comply with wallets that do send it.

`c_nonce` is managed by the server/credential issuer which we do not implement.

auth/api/iam/api.go
reinkrul: defer calls the function when the current scope finishes, meaning after return at line 249, so there's nothing that can fail after this defer?

gerardsn: my bad, I thought I had only added 1 defer function. Removed defer, but kept logging instead of failing when delete token fails. Since this is retrieval by the local node this is preferred over failing?

reinkrul (approved):

PR #3125, opened 2024-05-17
Remove nonce from default Request Object params

### LINKED ISSUE #3124: Authorization request should only contain a nonce if clients requires the return to contain it
Currently we add a `nonce` to all (JAR) Request Objects. The nonce should only be present if the client/RP expects the server to incorporate it into the result, such as a `vp_token`. (or openid `id_token`, where this param was first defined).

authorization request already contain a `state` param that is unique for randomness in signatures.

gerardsn:
closes #3124


reinkrul (approved):
I assume nonce checking is part of the next PR?

woutslakhorst (approved):

PR #3127, opened 2024-05-18
Crypto: let Exists() return an error if one occurs

reinkrul:
Could lead to dangerous assumptions (key does not exist, so (over)write it).


gerardsn (approved):
assuming conflicts with #3120 do not result in significant changes

PR #3128, opened 2024-05-18
Crypto: Support Azure Key Vault

reinkrul:
Uses:
- https://github.com/nuts-foundation/nuts-node/pull/3127
- https://github.com/nuts-foundation/nuts-node/pull/3120
- https://github.com/nuts-foundation/nuts-node/pull/3129


reinkrul:
@woutslakhorst can you review this?

My main concern is that we need to provide Azure Key Vault with a key name, for which I chose to use the verification method's key ID. But, Azure Key Vault only allows `[a-z0-9-]` in key names, so we have to "dumb down" the key name, replacing all illegal characters with `-`.
The alternative is having a SQL table which maps verification method IDs (key IDs) to keys in Azure Key Vault, but then mapping keys in Azure Key Vault back to DIDs becomes hard(er) again. So I think this is the lesser evil.

But, if we want to support creating keys in key storage that require the key thumbprint (e.g. `did:nuts` or `did:jwk`), we need that coupling again, since we must determine the key name for Vault before creating the key.

I also checked if we can rename the key in Key Vault, but that's not possible.

crypto/storage/azure/keyvault.go
```diff
@@ -0,0 +1,276 @@
...
+		return nil, "", fmt.Errorf("unable to create key in Azure Key Vault (name=%s): %w", keyName, err)
+	}
+	publicKey, _, err := parseKey(response.Key)
+	if err != nil {
+		return nil, "", err
+	}
+	return publicKey, keyID, nil
+}
+
+func (a keyvault) GetPrivateKey(ctx context.Context, kid string) (crypto.Signer, error) {
```
gerardsn, line 114: this no longer provides direct access to the key. change to `Signer` to follow go convention?

crypto/storage/azure/keyvault.go
```diff
@@ -0,0 +1,276 @@
...
+		return spi.ErrNotFound
+	} else if err != nil {
+		return fmt.Errorf("unable to delete key from Azure Key Vault (name=%s): %w", keyIDToKeyName(kid), err)
+	}
+	return nil
+}
+
+func (a keyvault) getPrivateKey(ctx context.Context, keyName string) (*azkeys.GetKeyResponse, error) {
+	response, err := a.client.GetKey(ctx, keyName, "", nil)
+	responseError := new(azcore.ResponseError)
```
gerardsn, line 157: how much overhead does it add to check this every time before using a key? If this succeeds once it probably will every time after that. Should the results be cached in `keyvault` and let Sign also check for 404 incase it has been deleted?

crypto/storage/azure/keyvault.go
```diff
@@ -0,0 +1,276 @@
...
+	var publicKey crypto.PublicKey
+	if err := keyAsJWK.Raw(&publicKey); err != nil {
+		return nil, "", fmt.Errorf("unable to convert key from Azure Key Vault Key to crypto.PublicKey: %w", err)
+	}
+	if !(*key.Kty == azkeys.KeyTypeEC || *key.Kty == azkeys.KeyTypeECHSM) || *key.Crv != azkeys.CurveNameP256 {
+		return nil, "", errors.New("only ES256 keys are supported")
+	}
+	// Code to support all types, supported by Azure Key Vault (kept here for future support):
+	//var signatureAlgorithm azkeys.SignatureAlgorithm
+	//switch *key.Kty {
```
gerardsn, line 195: If this is uncommented, would that allow users to create other key types directly in azure and use them by updating the did doc? (This provides a workaround for parties that need other keys while the node can only create keys capable of ES256)

crypto/storage/azure/keyvault.go
```diff
@@ -0,0 +1,276 @@
...
+	//case azkeys.KeyTypeEC:
+	//	fallthrough
+	//case azkeys.KeyTypeECHSM:
```
gerardsn, line 198:
```suggestion
	//case azkeys.KeyTypeEC, azkeys.KeyTypeECHSM:
```

I believe comma separating values is preferred over using empty `fallthrough` cases

crypto/storage/azure/keyvault.go
```diff
@@ -0,0 +1,276 @@
...
+	for bytes[0] == 0 {
+		bytes = bytes[1:]
+	}
+	b.AddASN1(asn1.INTEGER, func(c *cryptobyte.Builder) {
+		if bytes[0]&0x80 != 0 {
+			c.AddUint8(0)
+		}
+		c.AddBytes(bytes)
+	})
+}
```
gerardsn, line 276: magic

gerardsn (changes requested):

crypto/storage/azure/keyvault.go
reinkrul: Too complicated (especially with the KID -> Key Vault key name translation), so then I'd rather support it properly throught the API. I removed the commented out code.

crypto/storage/azure/keyvault.go
reinkrul: Removed

crypto/storage/azure/keyvault.go
reinkrul: Copied it literally and made it clear it's copied. I don't understand the code 100%, but this has to do with signed/unsigned requiring a leading zero. I added this as comment.

crypto/storage/azure/keyvault.go
```diff
@@ -0,0 +1,276 @@
...
+	useHSM  bool
+}
+
+func (a keyvault) Name() string {
+	return StorageType
+}
+
+func (a keyvault) CheckHealth() map[string]core.Health {
+	return nil
+}
```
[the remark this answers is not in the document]
reinkrul: The Golang SDK does not support it. I could list the keys (first page), but in practice the user might not have this privilege. So I guess not, for now?

crypto/storage/azure/keyvault.go
```diff
@@ -0,0 +1,276 @@
...
+	}
+
+	keyName := keyIDToKeyName(keyID)
+
+	var keyType azkeys.KeyType
+	if a.useHSM {
+		keyType = azkeys.KeyTypeECHSM
+	} else {
+		keyType = azkeys.KeyTypeEC
+	}
```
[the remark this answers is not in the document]
reinkrul: You have to configure it in Azure Key Vault. But it's stated in the documentation, and we have to assume people configuring this have basic knowledge of Azure Key Vault.

crypto/storage/azure/keyvault.go
```diff
@@ -0,0 +1,276 @@
...
+	// Only used for migrating to a new storage backend, which is not implemented yet for Azure Key Vault
+	return errors.New("SavePrivateKey() is not supported for Azure Key Vault")
+}
+
+func (a keyvault) ListPrivateKeys(ctx context.Context) []string {
+	// Only used for:
+	// - migrating to a new storage backend, which is not implemented yet for Azure Key Vault
+	// - did:nuts DIDs ownership checking, which won't be supported by this Azure Key Vault integration
+	return nil
+}
```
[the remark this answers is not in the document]
reinkrul: Implemented the function.

crypto/storage/azure/keyvault.go
reinkrul: ?

crypto/storage/azure/keyvault.go
reinkrul: Premature optimization, could always be added later (and it should at least be faster than the external storage API that requires 2 hops).

crypto/storage/azure/keyvault.go
reinkrul: We could do this, but out of scope.

crypto/storage/azure/keyvault.go
```diff
@@ -0,0 +1,276 @@
...
+	"regexp"
+	"time"
+)
+
+// New creates a new Azure Key Vault storage backend.
+// If useHSM is true, the key type will be azkeys.KeyTypeECHSM, otherwise azkeys.KeyTypeEC.
+func New(keyVaultUrl string, timeout time.Duration, useHSM bool) (spi.Storage, error) {
+	if keyVaultUrl == "" {
+		return nil, errors.New("missing Azure Key Vault URL")
+	}
```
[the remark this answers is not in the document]
reinkrul: We don't take vault URLs from user input, and this implementation also uses the validated KID wrapper, preventing the use of slashes and such in KIDs

crypto/crypto.go
```diff
@@ -122,6 +125,17 @@ func (client *Crypto) setupVaultBackend(_ core.ServerConfig) error {
...
 }
 
+func (client *Crypto) setupAzureKeyVaultBackend(_ core.ServerConfig) error {
+	log.Logger().Debug("Setting up Azure Key Vault backend for storage of private key material. ")
+	var err error
+	azureBackend, err := azure.New(client.config.AzureKeyVault.URL, client.config.AzureKeyVault.Timeout, client.config.AzureKeyVault.UseHSM)
+	if err != nil {
+		return err
+	}
+	client.storage = spi.NewValidatedKIDBackendWrapper(azureBackend, spi.KidPattern)
```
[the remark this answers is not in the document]
reinkrul: no

crypto/jwx.go
```diff
@@ -424,6 +428,9 @@ func encryptionAlgorithm(key crypto.PublicKey) (jwa.KeyEncryptionAlgorithm, erro
 
 // Thumbprint generates a Nuts compatible thumbprint: Base58(SHA256(rfc7638-json))
 func Thumbprint(key jwk.Key) (string, error) {
+	if key == nil {
+		return "", errors.New("key is nil")
+	}
```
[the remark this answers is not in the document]
reinkrul: made it panic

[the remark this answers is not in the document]
gerardsn: it also panics without this check? which is fine in go

gerardsn (approved):
LGTM
just a comment on the panic comment

crypto/jwx.go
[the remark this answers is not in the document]
reinkrul: but this is a nicer panic ;-)

PR #3137, opened 2024-05-23
return 404 for missing token sessionID

### LINKED ISSUE #3130: Incorrect return status on unknown auth session when requesting access token
```
node-right-1   | time="2024-05-22T08:42:19Z" level=info msg="HTTP request" method=GET module=http remote_ip=172.90.10.8 status=500 uri=/internal/auth/v2/accesstoken/pl5hGvDnlhmh5Iqevfeo-CgKjCqd95Lm-UFsupOGCQw
```

1st call returns access token, 2nd call should return 404

gerardsn:
closes #3130

### DIFF
--- a/auth/api/iam/api.go
+++ b/auth/api/iam/api.go
@@ -231,6 +231,9 @@ func (r Wrapper) RetrieveAccessToken(_ context.Context, request RetrieveAccessTo
 	var token TokenResponse
 	err := r.accessTokenClientStore().Get(request.SessionID, &token)
 	if err != nil {
+		if errors.Is(err, storage.ErrNotFound) {
+			return nil, core.NotFoundError("session not found")
+		}
 		return nil, err
 	}
 	if token.Get("status") == oauth.AccessTokenRequestStatusPending {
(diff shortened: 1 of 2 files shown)


woutslakhorst (approved):

auth/api/iam/api.go
```diff
@@ -231,6 +231,9 @@ func (r Wrapper) RetrieveAccessToken(_ context.Context, request RetrieveAccessTo
 	var token TokenResponse
 	err := r.accessTokenClientStore().Get(request.SessionID, &token)
 	if err != nil {
+		if errors.Is(err, storage.ErrNotFound) {
+			return nil, core.NotFoundError("session not found")
```
reinkrul, line 235: shouldn't this be an oauth2 error?

reinkrul (changes requested):

auth/api/iam/api.go
woutslakhorst: it's an internal API

reinkrul (approved):

PR #3138, opened 2024-05-23
Fix internal external ports

rolandgroen:
The helm chart does not reflect the changes from a single 8080 part to a split up configuration.


reinkrul:
No changes, already fixed on master? Closing.

Issue #3005, opened 2024-04-02
IAM: move user session management to middleware


reinkrul:
User session management is now being hardcoded into the OpenID4VP HTTP request handlers (creation in user landing, loading in authorization handlers), but this complicates integration with external Identity Providers.

From software design perspective: the user session isn't "owned" by the OpenID4VP (or other OAuth2) flows; it's an injected trait/resource used by those flows. 

Thus: user session management (creating, loading) should be moved to middleware.

Can be done after the first beta/RC release of v6

PR #3139, opened 2024-05-24
IAM: Refactor user session management to middleware

reinkrul:
Fixes https://github.com/nuts-foundation/nuts-node/issues/3005

Moves user session management to Echo middleware. Functional changes:

- If a user session already exists and preauthorized user details are provided, earlier it checked whether the user details of the subsequent "user access token flow" were the same as the one that were used to create the session (XIS sanity check). This check is now became unpractical, since the user details aren't available as-is in the session any more. Also, this is the responsibility of the calling XIS.
- Before, only 1 session cookie could exist for a user, so if the user performed a user access token flow for a different `did:web` DID, things would break (session not found/invalid). This is now fixed, by allowing multiple sessions (bound to web DID) to co-exist, allowing multi-tenancy.
- Changed the session from `Host` prefix to `Secure` prefix, as `Host` prefix does not allow a path other than `/`


auth/api/iam/usersession/util.go
```diff
@@ -0,0 +1,28 @@
...
+ *
+ */
+
+package usersession
+
+import "net/http"
+
+// CookieReader is an interface for reading cookies from an HTTP request.
+// It is implemented by echo.Context and http.Request.
+type CookieReader interface {
```
woutslakhorst, line 25: weird to have a util with only an interface

auth/api/iam/usersession/data.go
```diff
@@ -0,0 +1,65 @@
...
+	"errors"
+	"fmt"
+	"github.com/lestrrat-go/jwx/v2/jwk"
+	"github.com/nuts-foundation/go-did/did"
+	"github.com/nuts-foundation/go-did/vc"
+	"time"
+)
+
+// Data is a session-bound Verifiable Credential wallet.
+type Data struct {
```
woutslakhorst, line 31: extremely generic name. Either name the package `session` and this struct `User` or place everything in the `iam` package and name it `usersession`?

woutslakhorst (changes requested):
I think middleware should be located next to other middleware. Maybe in http/session or http/user? This would solve most weird naming. `user.Wallet` vs `usersession.UserWallet`.... and `user.Session` vs `usersession.Data`

auth/api/iam/usersession/util.go
reinkrul: Moved it all to 1 file

auth/api/iam/usersession/data.go
reinkrul: moved `usersession` to `http` module, renamed package to `user` and `Data` -> `Session`

woutslakhorst (approved):

Issue #3142, opened 2024-05-24
Caching did:web documents, OAuth/OpenID metadata and StatusList2021


reinkrul:
An OAuth2 authz code flow with 2 OpenID4VP causes AS metadata to be retrieved 4 times and the DID document to be resolved 8 times. Since these are relatively static resources they could be cached quite easily, improving UX (faster flows) and reducing traffic (and access logs).

Relates to:
- https://github.com/nuts-foundation/nuts-node/issues/2420
- https://github.com/nuts-foundation/nuts-node/issues/3028
- https://github.com/nuts-foundation/nuts-node/issues/2034

## Suggestion
- Use HTTP/1.1 Cache Control headers to communicate max-age to the client reading the resource. If parties want to override this, they can do this in their reverse proxy.
- Don't solve caching in the DID resolver and IAM client, but do this in the HTTP clients used. There are several Go libraries for this (e.g. https://github.com/pquerna/cachecontrol).
  - Since you don't want unbound in-memory caches, you might want a cache per resource type?

Issues:
- Do we want to prevent clustering issues: e.g. node A having a different version of a resource than node B? Although if the resource server gracefully migrates the endpoint, this shouldn't be an issue?

## did:web documents
The did:web method designers somewhat agree that regular HTTP cache control headers are a good direction: https://github.com/w3c-ccg/did-method-web/issues/40

When DID documents are used to store services, long caching times (e.g., a day) could cause unpractical staleness, so caching should be relatively short in that case. That said, parties changing endpoints should always do this by graceful migration: the old endpoint should stay available for a reasonable amount of time. On the other hand, caching DID documents for a minute already reduces the number of resolutions from 8 to 1 for the flow described above. 

Suggestion: max-age of 5 minutes by default

## OAuth2/OpenID Metadata
The same applies to metadata endpoints.

Suggestion: max-age of 5 minutes by default

## StatusList2021
Max. cache time is influenced by the governance, on how fast credentials should be revoked. eIDAS requires CAs to revoke certificates within 4 hours. But as status list is a small resource (since it uses the bit string), it can be refreshed more often than that.

Suggestion: max-age of 5 minutes by default

PR #3148, opened 2024-05-27
IAM: Add caching to HTTP client

reinkrul:
Fixes https://github.com/nuts-foundation/nuts-node/issues/3142

@woutslakhorst @gerardsn is this an approach you agree on?

I think the Nuts node should having client-side caching features on board, since I've not often seen an outbound proxy, let alone one that performs caching. So I suspect most parties won't be considering an outbound proxy for caching capabilities. This could severely downgrade performance/UX, given how often the same resources are requested.

There are a few libraries that do this, but all of them were archived and unmaintained with reasons like:
- This is so easy to implement, no need for a library
- Caching is not needed in modern architectures (?)
- Caching is very application-specific, so a library is too hard/complex/big

This solution just looks at expiry time (typically `max-age`), keeps cached resources around for max. 1 hour (or `max-age` if less). It also maxes out the available cache to 1mb (ofc. should be configurable).


auth/client/iam/caching.go
```diff
@@ -0,0 +1,198 @@
...
+	}
+	if httpRequest.Method == http.MethodGet {
+		reasons, expirationTime, err := cachecontrol.CachableResponse(httpRequest, httpResponse, cachecontrol.Options{PrivateCache: false})
+		if err != nil {
+			log.Logger().WithError(err).Infof("error while checking cacheability of response (url=%s), not caching", httpRequest.URL.String())
+			return httpResponse, nil
+		}
+		// We don't want to cache responses for too long, as that increases the risk of staleness,
+		// and could keep cause very long-lived entries to never be pruned.
+		maxExpirationTime := time.Now().Add(time.Hour)
```
woutslakhorst, line 84: Make constant

auth/client/iam/caching.go
```diff
@@ -0,0 +1,198 @@
...
+		}
+		if len(reasons) > 0 || expirationTime.IsZero() {
+			log.Logger().Debugf("response (url=%s) is not cacheable: %v", httpRequest.URL.String(), reasons)
+			return httpResponse, nil
+		}
+		responseBytes, err := io.ReadAll(httpResponse.Body)
+		if err != nil {
+			return nil, fmt.Errorf("error while reading response body for caching: %w", err)
+		}
+		h.mux.Lock()
```
woutslakhorst, line 96: For readability purposes place a single lock/unlock at the beginning of the only public method.

auth/client/iam/caching.go
```diff
@@ -0,0 +1,198 @@
...
+		}
+	}
+	h.currentSizeBytes -= len(h.head.responseData)
+	h.head = h.head.next
+	return h.head
+}
+
+func cacheHTTPResponses(requestDoer core.HTTPRequestDoer) *CachingHTTPRequestDoer {
+	return &CachingHTTPRequestDoer{
+		maxBytes:     10 * 1024 * 1024,
```
woutslakhorst, line 193: make configurable where 0 means no cache (for development)

auth/client/iam/caching.go
```diff
@@ -0,0 +1,198 @@
...
+		} else {
+			break
+		}
+	}
+}
+
+// insert adds a new entry to the cache.
+func (h *CachingHTTPRequestDoer) insert(entry *cacheEntry) {
+	// See if we need to make room for the new entry
+	for h.currentSizeBytes+len(entry.responseData) >= h.maxBytes {
```
woutslakhorst, line 146: document this strategy at the beginning (evict oldest entry).

woutslakhorst (changes requested):
Also add an issue for adding metrics to the client cache for hits vs misses. (enterprise feature ;))

auth/client/iam/caching.go
reinkrul: That causes IO to be included in the lock scope, so that causes scalability/performance issues. I refactored the function a bit to make it seem less random.

gerardsn:
out of interest / sanity check; how many simultaneous requests (order of magnitude) do you think it takes before the mutex is slower than ignoring cache? A successful authorization_code flow locks around 15-20 for the verifier and for the requester/holder

reinkrul:
> out of interest / sanity check; how many simultaneous requests (order of magnitude) do you think it takes before the mutex is slower than ignoring cache? A successful authorization_code flow locks around 15-20 for the verifier and for the requester/holder

Locking is only done while checking whether there's a cached copy and when inserting an item into the cache (memory/CPU operations only), not during HTTP requests/IO. So caching with mutexes will be much, much faster than performing HTTP requests, even with many parallel connections (I'd say especially with many parallel connections).

woutslakhorst (approved):

reinkrul:
@woutslakhorst I noticed the `did:web` resolver calls weren't cached, since it creates its own HTTP client (through `core.NewStrichtHTTPClient()`. I moved the strict HTTP client to the `http/client` package so we can manage the caching HTTP client centrally.

v5 functionality does not use the caching HTTP client.

PR #3151, opened 2024-05-28
Refactor and merge Callbacks

gerardsn:
cleanup and merge callbacks. This contains some fixes listed in #3152, but the majority is left for when we have access to a compliant issuer so we can do an end to end test of the flow.

### DIFF
--- a/auth/api/iam/api.go
+++ b/auth/api/iam/api.go
@@ -216,6 +216,12 @@ func (r Wrapper) Callback(ctx context.Context, request CallbackRequestObject) (C
 	}
 	// check if state is present and resolves to a client state
 	if request.Params.State == nil || *request.Params.State == "" {
+		// without state it is an invalid request, but try to provide as much useful information as possible
+		if request.Params.Error != nil && *request.Params.Error != "" {
+			callbackError := callbackRequestToError(request, nil)
+			callbackError.InternalError = errors.New("missing state parameter")
+			return nil, callbackError
+		}
 		return nil, oauthError(oauth.InvalidRequest, "missing state parameter")
 	}
 	oauthSession := new(OAuthSession)
@@ -229,14 +235,7 @@ func (r Wrapper) Callback(ctx context.Context, request CallbackRequestObject) (C
 
 	// if error is present, redirect error back to application initiating the flow
 	if request.Params.Error != nil && *request.Params.Error != "" {
-		requestErr := oauth.OAuth2Error{
-			Code:        oauth.ErrorCode(*request.Params.Error),
-			RedirectURI: oauthSession.redirectURI(),
-		}
-		if request.Params.ErrorDescription != nil {
-			requestErr.Description = *request.Params.ErrorDescription
-		}
-		return nil, requestErr
+		return nil, callbackRequestToError(request, oauthSession.redirectURI())
 	}
 
 	// check if code is present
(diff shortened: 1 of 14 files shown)


auth/api/iam/api.go
```diff
@@ -208,19 +208,52 @@ func (r Wrapper) HandleTokenRequest(ctx context.Context, request HandleTokenRequ
...
+	// check did in path
+	ownDID, err := r.toOwnedDID(ctx, request.Did)
 	if err != nil {
-		// this is an OAuthError already, will be rendered as 400 but that's fine (for now) for an illegal id
 		return nil, err
 	}
+	// check if state is present and resolves to a client state
+	if request.Params.State == nil || *request.Params.State == "" {
+		return nil, oauthError(oauth.InvalidRequest, "missing state parameter")
+	}
```
woutslakhorst, line 220: If the Callback has an error but is missing the client state, the actual error will not be shown

auth/api/iam/api.go
```diff
@@ -208,19 +208,52 @@ func (r Wrapper) HandleTokenRequest(ctx context.Context, request HandleTokenRequ
...
 
-	return r.handleCallback(ctx, request)
+	// check if code is present
+	if request.Params.Code == nil || *request.Params.Code == "" {
+		return nil, withCallbackURI(oauthError(oauth.InvalidRequest, "missing code parameter"), oauthSession.redirectURI())
+	}
+
+	// continue flow
+	switch oauthSession.ClientFlow {
+	case "openid4vci_credential_request":
```
woutslakhorst, line 249: make constants

auth/api/iam/openid4vci.go
```diff
@@ -79,24 +83,21 @@ func (r Wrapper) RequestOid4vciCredentialIssuance(ctx context.Context, request R
...
+	redirectUri = redirectUri.JoinPath(oauth.CallbackPath)
 	// Store the session
-	err = r.openid4vciSessionStore().Put(state, &Oid4vciSession{
-		HolderDid:         requestHolder,
-		IssuerDid:         issuerDid,
-		RemoteRedirectUri: request.Body.RedirectUri,
-		RedirectUri:       redirectUri.String(),
-		PKCEParams:        pkceParams,
+	err = r.oauthClientStateStore().Put(state, &OAuthSession{
+		ClientFlow:  "openid4vci_credential_request",
```
woutslakhorst, line 93: const

docs/_static/auth/iam.partial.yaml
```diff
@@ -472,64 +472,6 @@ paths:
                 "$ref": "#/components/schemas/VerifiableCredential"
         default:
           $ref: '../common/error_response.yaml'
-  /iam/oid4vci/callback:
```
woutslakhorst, line 475: also add to calback API description that it's used for both calls.

woutslakhorst (changes requested):

auth/api/iam/api.go
gerardsn: RFC6749 [4.1.2.1](https://www.rfc-editor.org/rfc/rfc6749.html#section-4.1.2.1).  Error Response
> state
         REQUIRED if a "state" parameter was present in the client
         authorization request.  The exact value received from the
         client.

I wasn't sure what to do with this, but I will wrap the error if there is one

woutslakhorst (approved):

PR #3174, opened 2024-06-05
Crypto: use managed identity credential directly to circumvent time-out

reinkrul:
Error:

```
DefaultAzureCredential: failed to acquire a token.\nAttempted credentials:\n\tEnvironmentCredential: incomplete environment variable configuration. Only AZURE_TENANT_ID and AZURE_CLIENT_ID are set\n\tWorkloadIdentityCredential: no token file specified. Check pod configuration or set TokenFilePath in the options\n\tManagedIdentityCredential: managed identity timed out. See https://aka.ms/azsdk/go/identity/troubleshoot#dac for more information\n\tAzureCLICredential: Azure CLI not found on path\n\tAzureDeveloperCLICredential: Azure Developer CLI not found on path"
```

According to https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azidentity/TROUBLESHOOTING.md#dac we should use it directly to avoid time-outs.

Zorg bij jou's Orca deployment needs to be able to explicitly specify Managed Identity as auth credential to Azure Key Vault.

Update: made it configurable, since in dev environments and maybe other deployments don't want to use managed identities.


crypto/storage/azure/interface.go
```diff
@@ -34,13 +34,24 @@ type Config struct {
 	// UseHSM specifies whether	to store the key in a hardware security module (HSM).
 	// If true, the Azure Key Vault must be configured for HSM usage.
 	UseHSM bool `koanf:"hsm"`
+	// Credential specifies the credential to use for authentication to the Azure Key Vault.
+	Credential CredentialConfig `koanf:"credential"`
```
gerardsn, line 38: I also saw a `TokenCredential` interface somewhere, is that a more specific name that could be used to avoid confusion with verifiable credentials?

crypto/storage/azure/keyvault.go
```diff
@@ -40,18 +40,37 @@ import (
 	"time"
 )
 
+const (
+	DefaultChainCredentialType    string = "default"
+	ManagedIdentityCredentialType string = "managed_identity"
+)
```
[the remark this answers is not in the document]
reinkrul: There are constants in the `go-mssqldb` library, but they're kinda cumbersome and seem specific for that library; `ActiveDirectoryManagedIdentity`. So I guess, no.

crypto/storage/azure/interface.go
reinkrul: I believe "credential" is the term they stick to that gives you access to an identity (user/account), through a token. 

```
TokenCredential represents a credential capable of providing an OAuth token.
```

I can change it to "Auth".

PR #3177, opened 2024-06-08
VCR: Remove @context default value from OAS

reinkrul:
It breaks code generation in Java. Can be safely deleted, we don't necerassily need it.


woutslakhorst (approved):

gerardsn (changes requested):
add to description that it is the default to prevent magic behavior

reinkrul:
> add to description that it is the default to prevent magic behavior

It's already there:

> If omitted, the "https://nuts.nl/credentials/v1" context is used.

gerardsn (approved):

PR #3188, opened 2024-06-12
Tenants: OAS for creating/getting tenants

reinkrul:


woutslakhorst:
We need to compare this with the VDR v1/v2 API. Do we allow individual DID Document changes (or can we make it foolproof). Maybe the VDR v2 api is already what we need and we only need to change `did` references to `subject_id`?

PR #3196, opened 2024-06-17
Prepare DID storage for multiple DIDs per DID Subject

### LINKED ISSUE #3193: Add `subject_id` to `did` table. Add `findBySubjectId`
Same subject DIDs should automatically add AlsoKnownAs of other DIDs to documents.

woutslakhorst:
closes #3193  

- removed selfControl & controllers for did:nuts
- removed tenant option for did:web
- moved sql storage from did:web to vdr


vdr/didweb/manager.go
```diff
@@ -107,64 +113,62 @@ func (m Manager) AddVerificationMethod(_ context.Context, _ did.DID, _ managemen
...
+	vmAsJson, err := json.Marshal(verificationMethod)
 	if err != nil {
 		return nil, nil, err
 	}
-	if err := m.store.create(newDID, *verificationMethod); err != nil {
-		return nil, nil, fmt.Errorf("store new DID: %w", err)
+
+	sqlDid := sql.DID{
+		ID:      newDID.String(),
+		Subject: newDID.String(),
```
reinkrul, line 156: shouldn't this be a UUID (subject) and be supplied through the options?

vdr/didweb/manager.go
```diff
@@ -107,64 +113,62 @@ func (m Manager) AddVerificationMethod(_ context.Context, _ did.DID, _ managemen
...
+	}
+	var doc *sql.DIDDocument
+	if doc, err = documentStore.CreateOrUpdate(sqlDid, []sql.SqlVerificationMethod{{
```
reinkrul, line 159: You're not really creating a new DID document here? You could have multiple DIDs (even multiple did:web's?) belonging to the same subject, which should then all return the same services and VMs?

vdr/didweb/manager.go
```diff
@@ -107,64 +113,62 @@ func (m Manager) AddVerificationMethod(_ context.Context, _ did.DID, _ managemen
...
 	}
-	parsedNewDID, err := did.ParseDID(newDID.String()) // make sure internal state is consistent (DecodedID)
 	if err != nil {
-		return nil, nil, fmt.Errorf("invalid new DID: %s: %w", newDID.String(), err)
+		return nil, nil, fmt.Errorf("parse new DID: %w", err)
 	}
-	return m.create(ctx, *parsedNewDID)
-}
+	tx := m.db.Begin()
+	defer tx.Rollback()
```
reinkrul, line 133: better use `db.Transaction(func())`

vdr/sql/did.go
```diff
@@ -0,0 +1,110 @@
...
+	// DeleteAll removes all DIDs for a subject from the database
+	DeleteAll(subject string) error
+	// Find returns a DID by its ID
+	Find(id did.DID) (*DID, error)
+	// FindBySubject returns all DIDs for a subject
+	FindBySubject(subject string) ([]DID, error)
+}
+
+// SqlDIDManager is the implementation of the DIDManager interface
+type SqlDIDManager struct {
```
reinkrul, line 59: already in `sql` package, so no need to prepend types with `Sql`

reinkrul (changes requested):
How are we going to keep services and VMs over multiple DIDs belonging to the same subject in sync?

woutslakhorst:
> How are we going to keep services and VMs over multiple DIDs belonging to the same subject in sync?

Use the DB as primary and (for now) the network update as side-effect (maybe within the sql transaction)

vdr/didweb/manager.go
```diff
@@ -30,39 +31,28 @@ import (
...
 }
 
 var _ management.DocumentManager = (*Manager)(nil)
 
 // NewManager creates a new Manager to create and update did:web DID documents.
 func NewManager(rootDID did.DID, tenantPath string, keyStore crypto.KeyStore, db *gorm.DB) *Manager {
 	return &Manager{
-		store:      &sqlStore{db: db},
+		db: db,
+		//store:      &sqlStore{db: db},
```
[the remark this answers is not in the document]
woutslakhorst: leftover, remove

vdr/didweb/manager.go
```diff
@@ -71,23 +61,39 @@ func NewManager(rootDID did.DID, tenantPath string, keyStore crypto.KeyStore, db
...
+	documentStore := sql.NewDIDDocumentManager(tx)
+	sqlDocument, err := documentStore.Latest(subjectDID)
+	if err != nil {
+		if errors.Is(err, gorm.ErrRecordNotFound) {
+			return resolver.ErrNotFound
+		}
+		return err
+	}
+	if sqlDocument == nil {
+		return nil
```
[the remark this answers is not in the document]
woutslakhorst: can't, remove

vdr/didweb/manager.go
woutslakhorst: correct, but allowing a subject to be passed through the API would be part of another PR. Added a todo as comment

vdr/didweb/manager.go
woutslakhorst: True, but I wanted this PR to just update the DB model. In a later PR we'll move this logic one level higher (so the web manager will no longer have a create just a read/list/resolve)

vdr/sql/did.go
woutslakhorst: true, but `DIDManager` is the interface. When declaring vars from other packages you should use the interface. We won't see `var a sql.SqlDIDManager` anywhere. Also I'm not sure the package will not be renamed after some refactors.

vdr/didweb/manager.go
```diff
@@ -71,23 +60,36 @@ func NewManager(rootDID did.DID, tenantPath string, keyStore crypto.KeyStore, db
...
+	db         *gorm.DB
 	rootDID    did.DID
-	store      store
 	keyStore   crypto.KeyStore
 	tenantPath string
 }
 
 func (m Manager) Deactivate(ctx context.Context, subjectDID did.DID) error {
-	verificationMethods, _, err := m.store.get(subjectDID)
+	tx := m.db.Begin()
```
reinkrul, line 70: db.Transaction(func())

reinkrul (approved):

PR #3210, opened 2024-06-24
Auth v2: change OAuth2 issuer URL from /iam to /oauth2

reinkrul:
To support other DID methods. Changes:

- Changed the issuer URL to contain the complete DID instead of just the did:web ID part of the DID
- Changed the path of the issuer URL to `/oauth2` instead of `/iam`, to have it align with the other OAuth endpoints, and to explicitly decouple it from the did:web DID.

This is also preparation work for decoupling did:web from OAuth2; which will require the caller to provide the AS url in (near) future.

### DIFF
--- a/auth/api/iam/jar.go
+++ b/auth/api/iam/jar.go
@@ -123,11 +123,8 @@ func (j jar) Parse(ctx context.Context, ownDID did.DID, q url.Values) (oauthPara
 				return nil, oauth.OAuth2Error{Code: oauth.InvalidRequestURI, Description: "failed to get Request Object", InternalError: err}
 			}
 		case "post":
-			issuerURL, err := createOAuth2BaseURL(ownDID)
-			if err != nil {
-				return nil, err
-			}
-			md, err := authorizationServerMetadata(ownDID, issuerURL)
+			issuerURL := ownDID.URI().URL
+			md, err := authorizationServerMetadata(ownDID, &issuerURL)
 			if err != nil {
 				// DB error
 				return nil, err
--- a/auth/api/iam/jar.go
+++ b/auth/api/iam/jar.go
@@ -123,7 +123,11 @@ func (j jar) Parse(ctx context.Context, ownDID did.DID, q url.Values) (oauthPara
 				return nil, oauth.OAuth2Error{Code: oauth.InvalidRequestURI, Description: "failed to get Request Object", InternalError: err}
 			}
 		case "post":
-			md, err := authorizationServerMetadata(ownDID)
+			issuerURL, err := createOAuth2BaseURL(ownDID)
+			if err != nil {
+				return nil, err
+			}
+			md, err := authorizationServerMetadata(ownDID, issuerURL)
 			if err != nil {
 				// DB error
 				return nil, err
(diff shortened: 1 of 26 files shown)


core/to/util.go
```diff
@@ -0,0 +1,5 @@
+package to
+
+func Ptr[T any](v T) *T {
```
woutslakhorst, line 3: I think there exists more than one of these:
- api/iam/api_test.go
- somewhere else?

auth/client/iam/interface.go
```diff
@@ -45,7 +46,7 @@ type Client interface {
 	// PresentationDefinition returns the presentation definition from the given endpoint.
 	PresentationDefinition(ctx context.Context, endpoint string) (*pe.PresentationDefinition, error)
 	// RequestRFC021AccessToken is called by the local EHR node to request an access token from a remote Nuts node using Nuts RFC021.
-	RequestRFC021AccessToken(ctx context.Context, requestHolder did.DID, verifier did.DID, scopes string, useDPoP bool,
+	RequestRFC021AccessToken(ctx context.Context, requestHolder did.DID, verifier did.DID, oauthIssuer *url.URL, scopes string, useDPoP bool,
```
woutslakhorst, line 49: You might also want to update the `AuthorizationServerMetadata` method godoc in this interface.

auth/api/iam/jar_test.go
```diff
@@ -110,8 +111,10 @@ func TestJar_Parse(t *testing.T) {
 	})
 	require.NoError(t, err)
 	token := string(bytes)
+	walletIssuerURL := test.MustParseURL(walletDID.String())
```
woutslakhorst, line 114: doesn't this create a DID as URL? Doesn't break the tests but is functionally wrong?

auth/api/iam/jar.go
```diff
@@ -123,7 +123,8 @@ func (j jar) Parse(ctx context.Context, ownDID did.DID, q url.Values) (oauthPara
 				return nil, oauth.OAuth2Error{Code: oauth.InvalidRequestURI, Description: "failed to get Request Object", InternalError: err}
 			}
 		case "post":
-			md, err := authorizationServerMetadata(ownDID)
+			issuerURL := ownDID.URI().URL
```
woutslakhorst, line 126: which scheme does the URL get?

woutslakhorst (changes requested):
small stuff

auth/api/iam/jar.go
reinkrul: `did`...

It's used for the wallet's client metadata, which in case of a mobile phone doesn't really have a (reachable) HTTPS endpoint... I'd think this would be actually SIOP-ish (`https://self-issued.me` or something).

auth/api/iam/jar_test.go
reinkrul: Yes, but we don't know the URL beforehand, since it's to be provided by the calling application. With unsolicited requests (OpenID4VP request form verifier), we simply don't know it? I wonder if this shouldn't be the SIOPv2 self-issued URL.

This is actually how it worked in practice before; for did:web DIDs we derived the issuer URL, for other DID methods (`did:jwk` in our case) it's assumed it's the wallet DID, and we just set the DID as issuer URL.

core/to/util.go
reinkrul: oh, forgot to deduplicate

woutslakhorst (approved):

PR #3213, opened 2024-06-25
Fix darkmode ui for employeeId means

### LINKED ISSUE #2665: Disable dark-mode for EmployeeID screen
The CSS of the EmployeeID screen detects the OS dark mode settings, which makes the whole form a bit funky. We probably should disable this feature.

Relevant code: https://github.com/nuts-foundation/nuts-node/tree/master/auth/services/selfsigned

Example of the form in dark-mode, embedded in an iframe of the ECD.
image

stevenvegt:
Fixes #2665 

Note: amount of new lines is from formatting css from one big line to multiple lines.

Changes:
- Remove the css which checks for os dark mode setting. Always a light layout.
- Use embedded templates instead of duplicating layout and css over 2 pages and 2 languages.
- Remove the done endpoint and render the form or done page based on the session state.
- Remove unused query params from api spec


auth/services/selfsigned/web/handler.go
```diff
@@ -22,65 +22,76 @@ import (
...
+func (h Handler) RenderEmployeeIDPage(ctx echo.Context, sessionID SessionID) error {
 	session, ok := h.store.Load(sessionID)
 	if !ok {
 		return echo.NewHTTPError(http.StatusNotFound, "session not found")
 	}
 
+	var templateToRender string
+	// the form for a created session should only be loaded once
+	if h.store.CheckAndSetStatus(sessionID, types.SessionCreated, types.SessionInProgress) {
+		templateToRender = "employee_identity"
```
woutslakhorst, line 72: Some `const`s for string values wouldn't hurt.

woutslakhorst (approved):
Needs backport to v5 branch as well

PR #3215, opened 2024-07-01
VCR: StatusList2021 forward compatibility for other DID methods

reinkrul:
To support other DID methods in future (`did:tdw`), the StatusList2021 issuer can't use `didweb.DIDtoURL` to determine the URL the StatusCredential lives on. This PR changes it to more explicit configuration.

Note: should we lookup the statuslist through `(DID, page)` combination instead? Since that also supports migration scenarios?

### DIFF
--- a/vcr/vcr.go
+++ b/vcr/vcr.go
@@ -225,7 +225,7 @@ func (c *vcr) Configure(config core.ServerConfig) error {
 		c.openidSessionStore = c.storageClient.GetSessionDatabase()
 	}
 
-	status := revocation.NewStatusList2021(c.storageClient.GetSQLDatabase(), client.NewWithCache(config.HTTPClient.Timeout))
+	status := revocation.NewStatusList2021(c.storageClient.GetSQLDatabase(), client.NewWithCache(config.HTTPClient.Timeout), config.URL)
 	c.issuer = issuer.NewIssuer(c.issuerStore, c, networkPublisher, openidHandlerFn, didResolver, c.keyStore, c.jsonldManager, c.trustConfig, status)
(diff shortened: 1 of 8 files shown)


vcr/vcr.go
```diff
@@ -225,7 +226,16 @@ func (c *vcr) Configure(config core.ServerConfig) error {
 		c.openidSessionStore = c.storageClient.GetSessionDatabase()
 	}
 
-	status := revocation.NewStatusList2021(c.storageClient.GetSQLDatabase(), client.NewWithCache(config.HTTPClient.Timeout))
+	status := revocation.NewStatusList2021(c.storageClient.GetSQLDatabase(), client.NewWithCache(config.HTTPClient.Timeout),
+		func(issuer did.DID, page int) (string, error) {
```
woutslakhorst, line 230: Any reason to make the URL resolution dynamic in this way? Are we going to magically configure any other implementation here?

reinkrul: Yes, but probably not needed. Simplified it with simply the base URL.

woutslakhorst (approved):

gerardsn:
I disagree with these changes. This makes it look like the statuslist domain name is configurable while it is not. Once a credential has been issued containing a statuslist entry, the status of this credential must forever be resolvable on the address listed in the credential.

By changing the configured URL 
- a new StatuslistCredential will be created for newly issued credentials, so issuance might work.
- resolving of existing credentials is broken. Even if you were to redirect from the old to the new domain in the proxy, the node will no longer be able to locate StatuslistCredential in the database since the change in this PR always looks for credential on the new domain.

PR #3221, opened 2024-07-03
Add URACredential to new 2024 JSONLD context

reinkrul:
To progress on authorization using logical identifiers (ABAC), it would be nice to have a credential with a well-understood organization number in the JSON-LD contexts that are packaged in the Nuts node. This can then be used for ABAC (e.g. "organization with URA 123456 can access this and that") in authz policies. This would also help Localisation/Authorization PoCs that require such an attribute (without having to fumble with cumbersome JSONLD configuration/mapping).

We have a current use case for this in Zorg bij jou/Shared Care Planning.

And as long as there is no official issuer for that number, I propose packaging a credential containing such number in the Nuts node. I know there's currently not 1 number to rule them all, but URA is probably the safest bet? We can always add more credentials (e.g. AGB).

Maybe we should introduce an "experimental" JSON-LD context for such "support" contexts that we don't intend to keep forever? E.g. https://nuts.nl/credentials/experimental


woutslakhorst:
I would propose to use https://nuts.nl/credentials/2024 since everything we come up with this year is probably experimental ;)

woutslakhorst:
and context will need to be published on that url as well.

reinkrul:
Also `URACredential2024` or just `URACredential`?

reinkrul:
It's `https://nuts.nl/credentials/2024#URACredential` now.

reinkrul:
> and context will need to be published on that url as well.

Will create a PR in nuts-webpresence after this has been merged.

rolandgroen:
My Feedback:

1. I'm not sure about the naming convention, but shouldn't it be NutsUraCredential instead of NutsURACredential?
2. It makes sense to me to add the name of the Organization to the credential field, just to make it somewhat human readable. The DEZI implementation of WVS does also release the name of the organization with the URA on UZI issuance.

reinkrul:
> My Feedback:
> 
> 1. I'm not sure about the naming convention, but shouldn't it be NutsUraCredential instead of NutsURACredential?
Don't know. I think in VC/DID it's generally what you suggest.

> 2. It makes sense to me to add the name of the Organization to the credential field, just to make it somewhat human readable. The DEZI implementation of WVS does also release the name of the organization with the URA on UZI issuance.

Just the name (NutsOrganizationCredential also has `city`)? Or do we want to always combine it with `NutsOrganizationCredential`? Having 1 credential might be more practical. 

So;
```json
{
  "ura": "123456",
  "name": "Dorpsziekenhuis"
}
```

?

rolandgroen:
Makes sense to me

rolandgroen (approved):

woutslakhorst (approved):

reinkrul:
Publishing it: https://github.com/nuts-foundation/nuts-webpresence/pull/264

stevenvegt:
The form of the credential is semantically not-optimal.
URA is a membership. The credentialSubject is an organisation which has a membership and this membership has a identifier. So I would like to propose the following structure:

https://tinyurl.com/29blhotl

```
{
  "id":"did:nuts:123#demo-uracredential",
  "type": [
    "VerifiableCredential",
    "NutsUraCredential"
  ],
  "credentialSubject": {
    "@id":"did:nuts:123",
    "type":"schema:Organisation",
    "memberOf": {
      "ura": "12345",
      "programName": "UZI Register Abonnee"
    }
  }
}
```

Issue #3208, opened 2024-06-24
Refactor VDR did management


woutslakhorst:
part of #3191

use sql DID storage as primary source of truth. Pass updates to that model to the different method managers.

PR #3226, opened 2024-07-09
Add didsubject DB management and interfaces

woutslakhorst:
part of #3208 

This is the first part of the VDR refactor to support management of DIDs via a DID subject. This PR represents about half of the added code.

- There's a lot of duplication from the didweb and mangement package. The DB will no longer be only for did:web, but will serve as primary storage for all future did methods. The added code is not yet used and can therefore be reviewed as standalone.
- The DIDChangeLog table is used for a 2-phase commit to keep did:nuts and did:web in sync. The mechanism that uses it will be added in another PR.
- The entries in the DID table are used to fill the `AlsoKnownAs` field. This is reflected by the `Aka` field added to the DID struct. This field is also preloaded in all queries.
- moved some jsonld constants around.
- The `DocumentManager` interface will be directly used by the V1 API while the V2 api will call the `SubjectManager` interface.
- The `MethodManager` interface needs to be implemented by each method. The nuts method implements both the `DocumentManager` and `MethodManager`.
- Future PRs will remove DocCreator, DocManipulator and DocUpdater


vdr/didsubject/did_document.go
```diff
@@ -0,0 +1,189 @@
...
+	ID        string `gorm:"primaryKey"`
+	DidID     string `gorm:"column:did"`
+	DID       DID    `gorm:"foreignKey:DidID;references:ID"`
+	CreatedAt int64  `gorm:"autoCreateTime:false"`
+	// UpdatedAt is the (unix) timestamp when the document was updated
+	// Also used to purge DID document changes that haven't been committed within a certain time frame
+	UpdatedAt           int64 `gorm:"autoUpdateTime:false"`
+	Version             int
+	VerificationMethods []VerificationMethod `gorm:"foreignKey:DIDDocumentID;references:ID"`
+	Services            []SqlService         `gorm:"foreignKey:DIDDocumentID;references:ID"`
```
gerardsn, line 44: I think the values for `foreignKey` and `references` are reversed

vdr/didsubject/did_document.go
```diff
@@ -0,0 +1,189 @@
...
+
+// DIDDocument is the gorm representation of the did_document_version table
+type DIDDocument struct {
+	ID        string `gorm:"primaryKey"`
+	DidID     string `gorm:"column:did"`
+	DID       DID    `gorm:"foreignKey:DidID;references:ID"`
+	CreatedAt int64  `gorm:"autoCreateTime:false"`
+	// UpdatedAt is the (unix) timestamp when the document was updated
+	// Also used to purge DID document changes that haven't been committed within a certain time frame
+	UpdatedAt           int64 `gorm:"autoUpdateTime:false"`
```
gerardsn, line 41: So `CreatedAt` is when the `did` was created and `UpdatedAt` is when this version of its `document` was created? (comment sounds like this entry is being updated)

vdr/didsubject/did_document.go
```diff
@@ -0,0 +1,189 @@
...
+type SqlDIDDocumentManager struct {
+	tx *gorm.DB
+}
+
+// NewDIDDocumentManager creates a new DIDDocumentManager for an open transaction
+func NewDIDDocumentManager(tx *gorm.DB) *SqlDIDDocumentManager {
+	return &SqlDIDDocumentManager{tx: tx}
+}
+
+func (s *SqlDIDDocumentManager) CreateOrUpdate(did DID, verificationMethods []VerificationMethod, services []SqlService) (*DIDDocument, error) {
```
gerardsn, line 78: I think there are race conditions possible on the update

vdr/didsubject/did_document.go
```diff
@@ -0,0 +1,189 @@
...
+// DIDDocument is the gorm representation of the did_document_version table
+type DIDDocument struct {
+	ID        string `gorm:"primaryKey"`
+	DidID     string `gorm:"column:did"`
+	DID       DID    `gorm:"foreignKey:DidID;references:ID"`
+	CreatedAt int64  `gorm:"autoCreateTime:false"`
+	// UpdatedAt is the (unix) timestamp when the document was updated
+	// Also used to purge DID document changes that haven't been committed within a certain time frame
+	UpdatedAt           int64 `gorm:"autoUpdateTime:false"`
+	Version             int
```
gerardsn, line 42: both are strictly increasing functions to determine the ordering of documents. Do we need both?

gerardsn (changes requested):
seeing only half of the changes for this refactor leaves a lot to the imagination, so most comments are questions

vdr/didsubject/interface.go
```diff
@@ -0,0 +1,43 @@
...
+
+// MethodManager keeps DID method specific state in sync with the DID sql database.
+type MethodManager interface {
+	// GenerateDocument generates a new DID document for the given subject.
+	// This is done by the method manager since the DID might depend on method specific rules.
+	GenerateDocument(ctx context.Context, keyFlags DIDKeyFlags) (*DIDDocument, error)
+	// GenerateVerificationMethod generates a new VerificationMethod for the given subject.
+	// This is done by the method manager since the VM ID might depend on method specific rules.
+	// If keyUsage includes management.KeyAgreement, an RSA key is generated, otherwise an EC key.
+	GenerateVerificationMethod(ctx context.Context, controller did.DID, keyUsage DIDKeyFlags) (*did.VerificationMethod, error)
```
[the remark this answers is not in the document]
woutslakhorst: `New` would be even better, since it signals only a struct is created, no side-effects in the db

storage/sql_migrations/003_did.sql
```diff
@@ -16,11 +16,25 @@ create table did_document_version
     -- id is v4 uuid
     id varchar(36) not null primary key,
     did varchar(370) not null,
+    created_at integer, -- make not nil in future PR
+    updated_at integer, -- make not nil in future PR
     version int not null,
+    raw $TEXT_TYPE, -- make not nil in future PR
```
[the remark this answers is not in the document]
woutslakhorst: this is the mechanism for injecting the DB specific data type (sql server vs mysql/sqlite/postgresql)

vdr/didsubject/changelog.go
```diff
@@ -0,0 +1,57 @@
...
+func (d DIDChangeLog) Method() string {
+	id, _ := did.ParseDID(d.DIDDocumentVersion.DID.ID)
+	return id.Method
+}
+
+// DID returns the did.DID of the DID Document
+func (d DIDChangeLog) DID() did.DID {
+	id, _ := did.ParseDID(d.DIDDocumentVersion.DID.ID)
+	return *id
+}
```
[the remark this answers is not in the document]
woutslakhorst: will change to something that causes an error down the line (eg: unknown method or empty did)

vdr/didsubject/management.go
```diff
@@ -0,0 +1,191 @@
...
+
+// CreationOptions defines options for creating DID Documents.
+type CreationOptions interface {
+	// With adds an option to the CreationOptions.
+	// It returns a new CreationOptions instance.
+	// If the same option is specified multiple times, the last instance will be used.
+	With(option CreationOption) CreationOptions
+	// All returns all the options.
+	All() []CreationOption
+}
```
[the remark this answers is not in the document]
woutslakhorst: didn't want to refactor things that are not directly related to what is needed.

storage/sql_migrations/003_did.sql
```diff
@@ -16,11 +16,25 @@ create table did_document_version
...
     unique (did, version),
     foreign key (did) references did (id) on delete cascade
 );
 
+-- this table is used for the poor-mans 2-phase commit
+create table did_change_log
+(
+    did_document_version_id varchar(36) not null primary key,
+    transaction_id varchar(36) not null,
+    type varchar(32) not null,
```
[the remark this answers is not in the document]
woutslakhorst: integration tests do not find it a problem

vdr/didsubject/did_document.go
woutslakhorst: correct

woutslakhorst: added extra commect

vdr/didsubject/did_document.go
woutslakhorst: and still it works:
```
time="2024-07-10T14:30:18+02:00" level=trace msg="Query (took 353.291µs): SELECT * FROM `did_document_version` WHERE did = \"did:web:example.com:iam:bob\" ORDER BY version desc,`did_document_version`.`id` LIMIT 1" module=Storage
time="2024-07-10T14:30:18+02:00" level=trace msg="Query (took 334.792µs): INSERT INTO `did` (`id`,`subject`) VALUES (\"did:web:example.com:iam:bob\",\"bob\") ON CONFLICT DO NOTHING" module=Storage
time="2024-07-10T14:30:18+02:00" level=trace msg="Query (took 48.25µs): INSERT INTO `did_verificationmethod` (`id`,`did_document_id`,`key_types`,`data`) VALUES (\"#1\",\"125fedb4-ee25-4933-a28b-55c01c2e435a\",31,\"{}\") ON CONFLICT (`id`) DO UPDATE SET `did_document_id`=`excluded`.`did_document_id`" module=Storage
time="2024-07-10T14:30:18+02:00" level=trace msg="Query (took 35.917µs): INSERT INTO `did_service` (`id`,`did_document_id`,`data`) VALUES (\"#2\",\"125fedb4-ee25-4933-a28b-55c01c2e435a\",\"{}\") ON CONFLICT (`id`) DO UPDATE SET `did_document_id`=`excluded`.`did_document_id`" module=Storage
time="2024-07-10T14:30:18+02:00" level=trace msg="Query (took 550.208µs): INSERT INTO `did_document_version` (`id`,`did`,`created_at`,`updated_at`,`version`,`raw`) VALUES (\"125fedb4-ee25-4933-a28b-55c01c2e435a\",\"did:web:example.com:iam:bob\",0,1720614618,1,\"{\"\"@context\"\":[\"\"https://w3c-ccg.github.io/lds-jws2020/contexts/lds-jws2020-v1.json\"\",\"\"https://www.w3.org/ns/did/v1\"\"],\"\"assertionMethod\"\":[{\"\"controller\"\":\"\"did:web:example.com:iam:bob\"\",\"\"id\"\":\"\"\"\"}],\"\"authentication\"\":[{\"\"controller\"\":\"\"did:web:example.com:iam:bob\"\",\"\"id\"\":\"\"\"\"}],\"\"capabilityDelegation\"\":[{\"\"controller\"\":\"\"did:web:example.com:iam:bob\"\",\"\"id\"\":\"\"\"\"}],\"\"capabilityInvocation\"\":[{\"\"controller\"\":\"\"did:web:example.com:iam:bob\"\",\"\"id\"\":\"\"\"\"}],\"\"id\"\":\"\"did:web:example.com:iam:bob\"\",\"\"keyAgreement\"\":[{\"\"controller\"\":\"\"did:web:example.com:iam:bob\"\",\"\"id\"\":\"\"\"\"}],\"\"service\"\":[{\"\"id\"\":\"\"\"\"}],\"\"verificationMethod\"\":[{\"\"controller\"\":\"\"did:web:example.com:iam:bob\"\",\"\"id\"\":\"\"\"\"}]}\")" module=Storage
```

vdr/didsubject/did_document.go
woutslakhorst: updatedAt could be the same for two versions. It's only seconds

vdr/didsubject/did_document.go
woutslakhorst: how exactly? The `SqlDIDDocumentManager` contains a reference to a transaction, so transaction isolation is in place.

vdr/didsubject/did_document.go
gerardsn: I compared it to some other foreign key relationships like the one on DIDDocument.DID, but those are "Belongs To" relationships while these are "Has Many" relationships. The direction of relationships are reversed explaining my confusion. This gorm magic naming gets me every time.

vdr/didsubject/did_document.go
gerardsn: you can start 2 simultaneous transactions on the same DID. Both will act on the same "latest" version. Unless this is solved in the code not part of this PR

gerardsn (commented):
just the race condition question remaining

vdr/didsubject/did_document.go
gerardsn: adding a uniqueness constraint on `version` would at least makes one transaction fail

gerardsn (approved):
Maybe add a comment in code explaining how the race condition is fixed. This will explain to out future self that an error is intentional

PR #3233, opened 2024-07-10
fix failing self-control migratoin on VDR

### LINKED ISSUE #3228: Migration for making did:nuts self-controlled fails to fast
It should consider, deactivated, missing keys, etc, log them and continue with the rest.

woutslakhorst:
fixes #3228 

added tests and continue on errors.
the test transaction is also required in the larger VDR refactor PR. Will come up when merging other PR.


vdr/vdr_test.go
```diff
@@ -565,3 +691,67 @@ type roundTripperFunc func(*http.Request) (*http.Response, error)
 func (fn roundTripperFunc) RoundTrip(r *http.Request) (*http.Response, error) {
 	return fn(r)
 }
+
+type testTransaction struct {
```
gerardsn, line 695: the only test using this struct also passes without specifying a return value from the mock. saves 60+ lines

gerardsn (approved):

vdr/vdr_test.go
woutslakhorst: future PRs will need it anyway

PR #3239, opened 2024-07-12
Refactor DocumentOwner

woutslakhorst:
Part 2 of #3208 

This PR changes the `DocumentOwner`. It's method independent with a fallback to the private key store.


vdr/did_owner.go
```diff
@@ -20,17 +20,81 @@ package vdr
...
+}
+
+func (m *MultiDocumentOwner) IsOwner(ctx context.Context, d did.DID) (bool, error) {
+	for _, do := range m.DocumentOwners {
+		owned, err := do.IsOwner(ctx, d)
+		if owned || err != nil {
+			return owned, err
+		}
+	}
+	return false, nil
```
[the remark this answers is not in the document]
gerardsn: On second thought. With the current implementations this is only possible in theory. I prefer splitting this in two separate conditions so it hurts my brain a bit less

vdr/didweb/manager.go
```diff
@@ -263,7 +238,7 @@ func (m Manager) createService(tx *gorm.DB, subjectDID did.DID, service did.Serv
 
 	_, err = didDocumentManager.CreateOrUpdate(current.DID, current.VerificationMethods, append(current.Services, sqlService))
 
-	return &service, nil
+	return &service, err
```
gerardsn, line 241: missing test?

gerardsn (approved):
codeclimate is complaining about `go fmt` issue.

vdr/didweb/manager.go
woutslakhorst: probably, but will be removed in next PR anyway

PR #3254, opened 2024-07-17
Add config for VDR

woutslakhorst:
to configure which did methods to use.

needed for #3245


gerardsn (approved):

vdr/cmd/cmd.go
```diff
@@ -41,6 +42,10 @@ import (
 // FlagSet contains flags relevant for the VDR instance
 func FlagSet() *pflag.FlagSet {
 	flagSet := pflag.NewFlagSet("vdr", pflag.ContinueOnError)
+
+	defs := vdr.DefaultConfig()
+
+	flagSet.StringSlice("didmethods", defs.DIDMethods, "Comma-separated list of DID methods (without did: prefix).")
```
reinkrul, line 48: must have `vdr` prefix. Does not describe what it does, not that order is important.

Issue #3251, opened 2024-07-17
Requesting unknown Discovery Service leads to HTTP error 500


reinkrul:
Requesting an unknown service leads to:

```json
{"detail":"discovery service not found","status":500,"title":"GetPresentations failed"}
```

(e.g.: `http://localhost:8080/discovery/non-existent`)

This should be status 400/404

PR #3257, opened 2024-07-20
Discovery: request unknown service should yield HTTP 404 instead of 500

reinkrul:
Fixes https://github.com/nuts-foundation/nuts-node/issues/3251


woutslakhorst (commented):
Could you explain how removing specific error code resolving to http.StatusNotFound fixes this problem? Also there's no test verifying this?

reinkrul:
Nothing's removed? Added a test.

woutslakhorst:
This is weird, it previously showed it as red (removed) in the github "Files changed" page....

PR #3263, opened 2024-07-23
move entity models to own package

### LINKED ISSUE #3261: Move entity models to ORM package inside storage engine
So more modules can make use of them without causing cyclic deps.

Required for some did to did subject refactors.

woutslakhorst:
closes #3261 

this PR only moves the models not the managers. Future changes might require a manager to be used across modules, then it could be moved as well.

Also moved the keyflags type, not sure if it's at the right location.


reinkrul (approved):

storage/orm/keyflag.go
gerardsn, line 1: (quack) Weird place for this. Would go-did be better for the type and consts, and crypto for the KeyUsage funcs?

Issue #3262, opened 2024-07-23
Auth: change API endpoint DID parameter to subject ID when requesting tokens/credentials


reinkrul:
The current v2 auth API endpoints have DID as parameter for multi-tenancy: access tokens issued by the Authz Server are scoped to this parameter. However, since we now have a subject instead of DID as tenant, having the DID in endpoints complicates for endpoint registration in DID documents:

- it 1:1 binds a specific DID (e.g. did:web) to an Authz Server endpoint scoped to the same DID (although the subject might have multiple DIDs)
- DID isn't the primary parameter used for communicating tenancy to the local Nuts node, so constructing the authz server URL from the DID might require more API calls and resolving

Since the DID in the API endpoints are really only used for tenancy, we should change it to subject ID instead.

Out of scope:
- https://github.com/nuts-foundation/nuts-node/issues/3278

woutslakhorst:
fixed bu #3333

PR #3264, opened 2024-07-23
Auth: replace DID with subject in OAuth2 client_id and URLs

reinkrul:
Fixes https://github.com/nuts-foundation/nuts-node/issues/3262

TODO:
- [ ] Move `selectDID()`, which's result is used to construct redirect URLs, to service/business layer (instead of API), e.g. by checking the Presentation Definition.
- [ ] Re-enable disabled user session checks (search for `#3264`)

### DIFF
--- a/auth/api/iam/access_token.go
+++ b/auth/api/iam/access_token.go
@@ -60,7 +60,7 @@ type AccessToken struct {
 }
 
 // createAccessToken is used in both the s2s and openid4vp flows
-func (r Wrapper) createAccessToken(issuer did.DID, walletDID did.DID, issueTime time.Time, scope string, pexState PEXConsumer, dpopToken *dpop.DPoP) (*oauth.TokenResponse, error) {
+func (r Wrapper) createAccessToken(subjectID string, walletDID did.DID, issueTime time.Time, scope string, pexState PEXConsumer, dpopToken *dpop.DPoP) (*oauth.TokenResponse, error) {
 	credentialMap, err := pexState.credentialMap()
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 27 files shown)


reinkrul:
I'm trying to replace `did` with `subject` in all `OAuth2` endpoints. This has the advantage of admin/ops not having to specify a DID in their OAuth2 issuer URLs, but a fixed `subject` (which may be assigned by the vendor or generated). Having a DID in one's OAuth2 issuer URL is a problem since the introduction of "subject", as the number of DIDs associated with a subject becomes more dynamic (`did:web` will be automatically created for `did:nuts` DIDs, `did:tdw` will be added later).

Thus, having a fixed subject in the OAuth2 issuer URL makes administration a lot easier, but then we don't have "OAuth2 client registration" anymore: parties can't deference the issuer URL to find the public keys to check signatures (relevant for JWT-Secured Authorization Requests, JARs).

Note that this only applies to the OpenID4VP flows.

It leads boils down to the following issue (`handleAuthorizeRequestFromHolder()`):
```
// TODO: issue #3216 (client_id is assumed to be did:web DID)
oauthIssuer, err := nutsOAuth2Issuer(*walletDID)
```

Options:
- Switch from DID to URL for `client_id`: put DIDs in the metadata, as a way for communicating DIDs that might be used for signing stuff. This should be safe enough (?) for checking JAR signatures, but other use cases could require a bi-directional relation (e.g. adding a service to the DID document that specifies the OAuth2 issuer URL). This would be non-standard.
- Keep DID as `client_id`: provide `metadata_url` in authorization request. This would be non-standard (although specified in OpenID4VP/VCI)
- ....

@gerardsn @woutslakhorst

reinkrul:
Fixed in other PR

PR #3268, opened 2024-07-23
Auth: move did:web public endpoints to VDR v2 API

reinkrul:
This PR moves the did:web endpoints (`/.well-known/did.json` and `/iam/<id>/did.json`) from Auth to VDR API, where they belong. This makes near-future subject-refactor PRs smaller.


woutslakhorst (approved):

woutslakhorst (commented):
It also generated client code, which it previously didn't.

PR #3270, opened 2024-07-24
change SignJWS and SignJWT signatures

woutslakhorst:
This PR is to prepare for the separation of `KID` (frontend) and `KeyName/Version` (backend). Two main things:

- Change signature of SignJWS and SignJWT so they are 'public' facing and always work on the kid.
- Reduce the usage of the TestKey for internal purposes:
  - the TestKey has a kid which is not always required
  - used real keys from the `memoryKeyStore` so key and keyStore are actually consistent in test.
- found some dead code.

### DIFF
--- a/auth/api/iam/api_test.go
+++ b/auth/api/iam/api_test.go
@@ -25,6 +25,7 @@ import (
 	"errors"
 	"fmt"
 	"github.com/nuts-foundation/nuts-node/core/to"
+	"github.com/nuts-foundation/nuts-node/crypto/storage/spi"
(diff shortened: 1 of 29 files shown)


reinkrul (approved):
Can we remove the ephemeral keys feature as well, or is it already gone? If it's still there, they now can't be used any more for signing.

woutslakhorst:
> Can we remove the ephemeral keys feature as well, or is it already gone? If it's still there, they now can't be used any more for signing.

in my ongoing work, it's already gone

Issue #3260, opened 2024-07-22
Persistent key ID and ordering.


woutslakhorst:
Some Key stores (Vault, Azure) require an ID before generating a key. Some DID methods (nuts) require a key to generate an ID for the DID document.

Alter interfaces in such a way that we can provide the subject, method and incrementing ID for a key to the sql DB and backend key storage. We can then use the sql records to generate a DID document entry according to the DID method.

For migration purposes we might need the actual keyID from vault/azure to be stored as well?

PR #3272, opened 2024-07-26
remove calls to deprecates documentManager for new key creation

woutslakhorst:
part of #3260 

Crypto will get a new DB table for connecting backend identifiers to `kid`. For all new keys, an entry must be created in the DB. This is a bit annoying for the legacy `NutsDocumentManager` because logic has to be moved around or duplicated. Fortunately it's also possible to route older key creation events via the newer logic.


test/node/server.go
```diff
@@ -75,6 +75,7 @@ func StartServer(t *testing.T, configFunc ...func(internalHttpServerURL, publicH
 	t.Setenv("NUTS_EVENTS_NATS_PORT", natsPort)
 	t.Setenv("NUTS_EVENTS_NATS_HOSTNAME", "localhost")
 	t.Setenv("NUTS_URL", publicHttpServerURL)
+	t.Setenv("NUTS_VDR_DIDMETHODS", "nuts")
```
reinkrul, line 78: this looks off; the RTD configuration reference indicates `didmethods` is a root property; https://nuts-node.readthedocs.io/en/latest/pages/deployment/configuration.html

vdr/vdr.go
```diff
@@ -272,6 +272,7 @@ func (r *Module) ConflictedDocuments() ([]did.Document, []resolver.DocumentMetad
 	return conflictedDocs, conflictedMeta, err
 }
 
+// Deprecated
```
reinkrul, line 275: follow `// <name> ...` convention

vdr/didsubject/management.go
```diff
@@ -55,28 +54,15 @@ type MethodManager interface {
 }
 
 // DocumentManager is the interface that groups several higher level methods to create and update DID documents.
+// Deprecated
 // Only used for V1 API calls.
 type DocumentManager interface {
```
reinkrul, line 59: Why are Update and RemoveVerificationMethod still relevant in this interface? I see a deprecation warning on Update, but not on the Remove function?

vdr/didnuts/manager.go
```diff
@@ -383,6 +261,8 @@ func (m Manager) NewDocument(ctx context.Context, keyFlags orm.DIDKeyFlags) (*or
 		return nil, err
 	}
 
+	keyFlags := DefaultKeyFlags()
```
reinkrul, line 264: do the default key flags indicate all key usages we need? I think we used to have a situation where the default key flags did not indicate one we actually needed for regular v5 Nuts use cases (keyAgreement? not sure which it was)

vdr/api/v1/api.go
```diff
@@ -120,12 +134,24 @@ func (a *Wrapper) CreateDID(ctx context.Context, request CreateDIDRequestObject)
...
 
-	doc, _, err := a.VDR.NutsDocumentManager().Create(ctx, options)
+	docs, _, err := a.SubjectManager.Create(ctx, options)
 	// if this operation leads to an error, it may return a 500
 	if err != nil {
 		return nil, err
 	}
+	var doc *did.Document
+	for _, m := range docs {
+		if strings.HasPrefix(m.ID.String(), "did:nuts:") {
```
reinkrul, line 146: `m.ID.Method == "nuts"`

vdr/api/v1/api.go
```diff
@@ -120,12 +134,24 @@ func (a *Wrapper) CreateDID(ctx context.Context, request CreateDIDRequestObject)
 	if keyFlags != defaultKeyFlags {
 		options = options.With(keyFlags)
 	}
+	options = options.With(didsubject.NutsLegacyNamingOption{})
 
-	doc, _, err := a.VDR.NutsDocumentManager().Create(ctx, options)
+	docs, _, err := a.SubjectManager.Create(ctx, options)
```
reinkrul, line 139: this also automatically creates a did:web DID?

Do we want that, or only if you use the v2 API?

vdr/api/v1/api.go
```diff
@@ -88,10 +91,21 @@ func (a *Wrapper) AddNewVerificationMethod(ctx context.Context, request AddNewVe
...
 	}
+	var vm *did.VerificationMethod
+	for _, m := range vms {
+		if m.ID.DID.String() == request.Did {
+			vm = &m
+			break
+		}
+	}
+	if vm == nil {
+		return nil, resolver.ErrNotFound
```
reinkrul, line 106: this would rather be a really weird internal error instead of not found? (you did something on a DID that didn't exist, but no error was returned).

reinkrul (changes requested):

gerardsn:
I also see no changes to statuscode mappings in the v1 API. Do these still match?

test/node/server.go
woutslakhorst: #3284

vdr/didnuts/manager.go
woutslakhorst: these are: all keyflags

vdr/api/v1/api.go
woutslakhorst: We'll add documentation on how to cope with the different APIs, the supported DID methods and v5/v6

woutslakhorst: #3285

vdr/api/v1/api.go
woutslakhorst: `"verification method added for subject: %s but not for DID: %s, do not use the V1 API for non-nuts DIDs"`

vdr/api/v1/api.go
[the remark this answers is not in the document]
woutslakhorst: #3285

woutslakhorst:
> I also see no changes to statuscode mappings in the v1 API. Do these still match?

Resolver is still used inside the SubjectManager

reinkrul (approved):

PR #3282, opened 2024-07-30
Auth: remove check wallet supports 'did' in supported_client_auth_methods

reinkrul:
When initiating OpenID4VP through Authorized Code Flow. Check can't be done at this point, since it requires dereferencing the client_id into OAuth2 Authorization Server metadata, which is only possible for `did:web` DIDs.

It is also unnecessary: the same check can be done in future when the client retrieves the OpenID4VP authorization request (which uses JAR), since it then provides its metadata (which can contain the supported client ID schemes property).

### DIFF
--- a/auth/api/iam/openid4vp.go
+++ b/auth/api/iam/openid4vp.go
@@ -106,28 +106,6 @@ func (r Wrapper) handleAuthorizeRequestFromHolder(ctx context.Context, verifier 
 		return nil, oauthError(oauth.InvalidRequest, "invalid value for code_challenge_method parameter, only S256 is supported")
 	}
 
-	// GET authorization server metadata for wallet
-	walletID := params.get(oauth.ClientIDParam)
-	// the walletDID must be a did:web
-	walletDID, err := did.ParseDID(walletID)
-	if err != nil || walletDID.Method != "web" {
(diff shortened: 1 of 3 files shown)


auth/api/iam/openid4vp.go
```diff
@@ -137,8 +115,7 @@ func (r Wrapper) handleAuthorizeRequestFromHolder(ctx context.Context, verifier
 	}
 
 	session := OAuthSession{
-		ClientID:          walletID,
-		IssuerURL:         oauthIssuer.String(),
+		ClientID:          params.get(oauth.ClientIDParam),
```
gerardsn, line 118: should check that client_id is not empty, it is a required field

gerardsn (changes requested):

auth/api/iam/openid4vp.go
gerardsn: 1 level higher checks that `client_id` starts with `did:web`. This means it can't be empty right now, but that check is likely to disappear soon

gerardsn: is checked in `jar.Parse`

gerardsn (approved):

PR #3286, opened 2024-07-30
Auth: have jar.Parse() take metadata instead of DID

reinkrul:
For forwards compatibility with other DID methods, and easier (future) enabling/disabling of Authorization Endpoint in metadata.


gerardsn (approved):
No issues with code changes, but I don't understand what (future) problem this solves? Can you elaborate

woutslakhorst (commented):
`ErrPullRequestTooSmall`

reinkrul:
> No issues with code changes, but I don't understand what (future) problem this solves? Can you elaborate

It currently assumes that the client_id, as given to `Parse()` is a DID, and is to be used as URL for OAuth2 endpoints in the metadata. This only applies to our current usage of `client_id` (DID) in one specific case (POST with metadata in OpenID4VP), so passing the function the client metadata instead is much cleaner.

PR #3288, opened 2024-07-30
VDR: Return better errors from SubjectManager, 

reinkrul:
And let List() return an error if subject doesn't exist (otherwise, we'll be checking `len(dids) == 0` everywhere).

### DIFF
--- a/vdr/api/v2/api.go
+++ b/vdr/api/v2/api.go
@@ -56,12 +56,13 @@ type Wrapper struct {
 // ResolveStatusCode maps errors returned by this API to specific HTTP status codes.
 func (w *Wrapper) ResolveStatusCode(err error) int {
 	return core.ResolveStatusCode(err, map[error]int{
+		didsubject.ErrSubjectNotFound:       http.StatusNotFound,
+		didsubject.ErrSubjectAlreadyExists:  http.StatusConflict,
 		resolver.ErrNotFound:                http.StatusNotFound,
 		resolver.ErrDIDNotManagedByThisNode: http.StatusForbidden,
 		did.ErrInvalidDID:                   http.StatusBadRequest,
 		didsubject.ErrInvalidService:        http.StatusBadRequest,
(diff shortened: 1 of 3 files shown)


gerardsn (changes requested):
The APIs to manipulate services and verification methods should also return a 404 when the subject does not exist. A single `len(dids)` check in `applyToDIDDocuments` should cover them all?

woutslakhorst:
> And let List() return an error if subject doesn't exist (otherwise, we'll be checking `len(dids) == 0` everywhere).

And now you'll be checking the error if you accept len==0..... where is `List()` used?

gerardsn:
in most cases these errors can be returned directly to the API layer (don't differentiate from other errors) where it is automatically converted to a 404

reinkrul:
> FindBySubject

We could use `List()` (optionally rename it to `getBySubject`) outside of the TX. I think it doesn't matter much for transaction integrity (or have a private `getBySubject` which takes the TX), and then we can re-use the error functionality.

In my opinion, functions that start with...:

- `Find...()` is a search operation, shouldn't return an error if nothing is found. Unless of course, you specify a partition/group/tenant you expect to exist (this is a case of that). But if the query doesn't match anything, it's not an error.
- Same applies to `List...()`
- `Get...()` indicates you expect the exact thing to exist, so always an error.

gerardsn:
> Same applies to `List...()`

Then the List here should also not return an error?

---
I was reasoning from the API perspective (subject is in the path so always returns a 404 if it does not exist), but I suppose that can be different from the methods in the business layer

reinkrul:
> > Same applies to `List...()`
> 
> Then the List here should also not return an error?
> 
> I was reasoning from the API perspective (subject is in the path so always returns a 404 if it does not exist), but I suppose that can be different from the methods in the business layer

Both are OK, potato-potato. Downside is by returning it from storage is that I consider public errors to be part of the business layer, upside is simplicity.

gerardsn (approved):

PR #3290, opened 2024-08-01
Docs: stable order for maps in configuration reference

reinkrul:
Prevents diffs without actual changes when regenerating the documentation.

Also removed the "Server Configuration Flags" section, since it basically repeats the Configuration page, and I couldn't easily fix the ordering there.


docs/pages/deployment/cli-reference.rst
gerardsn, line 1: I think this is the only place showing how the config parameters translate to CLI parameters. It's just a `--` prefix, but we can't expect users to just guess that. At some point we said we are going to drop support for the CLI, so maybe removing this page is also an option?

gerardsn: https://github.com/nuts-foundation/nuts-node/pull/3291#issuecomment-2262883379

gerardsn (approved):

docs/pages/deployment/cli-reference.rst
reinkrul: The ordering is controlled by a config flag, and it impacts the order of results presented to the user. Should this be in config description or do we not want to make such guarantees

CLI passing is (also) documented in https://nuts-node.readthedocs.io/en/latest/pages/deployment/configuration.html

PR #3291, opened 2024-08-01
VDR: SubjectManager returns DIDs in preferred order

reinkrul:


vdr/didsubject/manager.go
```diff
@@ -522,3 +527,44 @@ func (r *Manager) Rollback(ctx context.Context) {
...
+	orderedList := make([]did.Document, len(list))
+	for i, id := range listOfDIDs {
+	inner:
+		for _, doc := range list {
+			if doc.ID == id {
+				orderedList[i] = doc
+				break inner
+			}
+		}
+	}
```
gerardsn, line 568: This fails if `listOfDIDs` contains duplicates. Add a godoc describing this incase we re-use the method later in a scenario where this is possible

gerardsn (changes requested):
The ordering is controlled by a config flag, and it impacts the order of results presented to the user. Should this be in config description or do we not want to make such guarantees

vdr/didsubject/manager.go
```diff
@@ -522,3 +527,44 @@ func (r *Manager) Rollback(ctx context.Context) {
 		log.Logger().WithError(err).Error("failed to rollback DID documents")
 	}
 }
+
+func sortDIDs(list []did.DID, order []string) {
+	// Thank you, Copilot
```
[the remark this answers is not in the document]
reinkrul: No credit? ;-p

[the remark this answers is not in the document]
gerardsn: only if it prevents copilot from learning from its own output

vdr/didsubject/manager.go
reinkrul: would be weird, but fixed it just in kees.

vdr/didsubject/manager_test.go
```diff
@@ -420,3 +420,20 @@ func (t testMethod) Commit(_ context.Context, _ orm.DIDChangeLog) error {
...
+			{ID: did.MustParseDID("did:example:1")},
+			{ID: did.MustParseDID("did:test:1")},
+		}
+
+		sortDIDDocuments(documents, []string{"test", "example"})
+
+		require.Len(t, documents, 3)
+		assert.Equal(t, "did:test:1", documents[0].ID.String())
+		assert.Equal(t, "did:example:1", documents[1].ID.String())
+		assert.Equal(t, "did:example:1", documents[2].ID.String())
```
gerardsn, line 437: bad test
the sorted results contain compared to the original list documents [2,0,0], document on original index [1] has disappeared. (possible scenario where this may fail in the future (so godoc is fine, but this test is not) is when the list contains different versions of a DID document

gerardsn (changes requested):

reinkrul:
> The ordering is controlled by a config flag, and it impacts the order of results presented to the user. Should this be in config description or do we not want to make such guarantees

CLI passing is (also) documented in https://nuts-node.readthedocs.io/en/latest/pages/deployment/configuration.html

gerardsn:
last comment is in the wrong PR?

reinkrul:
Oops, comment moved to https://github.com/nuts-foundation/nuts-node/pull/3290

woutslakhorst (changes requested):
This will probably also impact which DID is used to sign a VP? If so change docs for config accordingly.

vdr/didsubject/manager_test.go
gerardsn: now failing after I made some changes

woutslakhorst:
around line 660 of `auth/api/iam.go` the signingKeyis set for the openid configuration (`signingKey = kid`). It uses the last key from the list, should become first one.

reinkrul:
I don't think this should matter any more, since any use of DIDs is bound by what the other side accepts. So there's not really a place for "preferred order". Although weird, if a verifier accepts 2 DID methods and the holder has both, it doesn't matter if it presents the first DID initially, and the second DID at some later point?

vdr/didsubject/manager_test.go
reinkrul: This is not a valid use case for the function, I made its description more precise to avoid future mistakes about its usage.

auth/api/iam/api.go
```diff
@@ -628,37 +628,37 @@ func (r Wrapper) OpenIDConfiguration(ctx context.Context, request OpenIDConfigur
...
+		return nil, oauth.OAuth2Error{
+			Code:          oauth.ServerError,
+			InternalError: err,
 		}
-		_ = jwkKey.Set(jwk.KeyIDKey, kid)
-		_ = set.AddKey(jwkKey)
-		signingKey = kid
 	}
+	_ = jwkKey.Set(jwk.KeyIDKey, kid)
+	_ = set.AddKey(jwkKey)
```
gerardsn, line 659: `set` needs to contain all keys in `dids`

gerardsn (changes requested):

auth/api/iam/api.go
reinkrul: aah, now I udnerstand Wout's comment

gerardsn (approved):

reinkrul:
@woutslakhorst would you like to re-review?

woutslakhorst (approved):

PR #3305, opened 2024-08-09
v5.4: fix error message in private key existence check

gerardsn:
aligned the message with v6. 
In V6 `Exists` also returns an error clarifying this even further.


stevenvegt:
I would suggest changing the interface to return the error instead. Helps us to debug in the future. Perhaps the vault service is down or some other thing has happened.

gerardsn:
That would have to be a new issue that will eventually end up on the `wontfix` pile as this is already fixed in v6. I was just going for a message that is less deterministic about the cause of the failure.

stevenvegt (approved):
Since this is rewritten in v6, this should do the job for the remainder of the lifetime of the v5 branch.

PR #3312, opened 2024-08-19
Fix dev image persistent tunnel

gerardsn:
This PR contains 3 changes to the `entrypoint.sh` script. I'm happy to only merge the ones that make sense
- commit 1: fix script to first [create a persistent tunnel before hosting](https://learn.microsoft.com/en-us/azure/developer/dev-tunnels/faq#how-can-i-create-a-persistent-tunnel) it. Calling `devtunnel host` without a `TUNNEL_ID` should create a temporary tunnel, but the behavior seems inconsistent as the tunnel is not always deleted. This commit should fix the inconsistency and always persist the created tunnel.
- commit 2: move tunnel variables from `/devtunnel/...` to `~/config/nuts-devtunnel`. It is likely that the `./config` folder is mounted anyway, so this saves a volume mount. However, the `config` directory inside the `nuts-node` is defined relative to the configured `datadir`, which could be set to a different path as the current working directory at the start of the container. This could be confusing so I'm not sure if it's worth the trouble.
- commit 3: devtunnel stores its user session info in `/nuts/DevTunnel`. When mounting `/nuts` (the default nuts datadir), this session info is also persisted over restarts, so we can check if there is a session and only prompt for a login if there is no session.

devtunnel has several [service limits](https://learn.microsoft.com/en-us/azure/azure-resource-manager/management/azure-subscription-service-limits#dev-tunnels-limits). The one I think we are most likely to hit first during a hackathon is 
> Active connections	20 per port

TODO in documentation
- [x] commit 2 and 3 require updates if approved
- [x] add link to service limits
- [x] link to devtunnel docs in case the devtunnel fails (not mounting `/devtunnel` may exceed max tunnels if the container is restarted often enough)

### DIFF
--- a/development/dev-image/Dockerfile
+++ b/development/dev-image/Dockerfile
@@ -1,4 +1,4 @@
-FROM nutsfoundation/nuts-node:master as nutsnode
+FROM nutsfoundation/nuts-node:master AS nutsnode
 FROM debian:buster-slim
 
 RUN apt-get update && apt-get install -y curl sudo musl
--- a/development/dev-image/entrypoint.sh
+++ b/development/dev-image/entrypoint.sh
@@ -7,15 +7,24 @@ mkdir -p /devtunnel
 # login with github user
 devtunnel user login -d -g
 
+# clear log without error if does not exist
+rm -f /devtunnel/tunnel.log
+
 # read from /devtunnel/tunnelid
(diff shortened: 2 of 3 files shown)


woutslakhorst (approved):
I think some aspects of devtunnel changed in the past x months.

PR #3311, opened 2024-08-19
Update to go 1.23

gerardsn:
- [x] remove `golang.org/x/exp` added in #3374

PR #3313, opened 2024-08-19
don't ignore errors

gerardsn:
ignoring errors results in a goose chase in #3311


reinkrul (approved):
can `Setenv` fail?

gerardsn:
> can `Setenv` fail?

The function seems to have some error conditions. I don't know if they apply to us, and I don't care. Go is optimized to assume the condition evaluates to `err==nil` (this only runs once anyway), and the IDE hides the `if err != nil` block, so when a signature contains an error I just return it.

PR #3319, opened 2024-08-22
IAM: Introduce new 'supported_client_id_did_methods' OAuth2 AS metadata property

reinkrul:


woutslakhorst (approved):
ok since it's small and doesn't hurt

PR #3328, opened 2024-08-29
Storage: enable Azure SQL Support

reinkrul:
The driver already supports it, we just needed to recognize the driver scheme (`azuresql://`)

### DIFF
--- a/storage/engine.go
+++ b/storage/engine.go
@@ -22,6 +22,8 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"github.com/microsoft/go-mssqldb/azuread"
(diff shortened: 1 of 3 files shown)


storage/engine.go
```diff
@@ -245,6 +247,8 @@ func (e *engine) initSQLDatabase() error {
 			return err
 		}
 		dialect = goose.DialectPostgres
+	case azuread.DriverName:
```
gerardsn, line 250: I _prefer_ `"azuresql"` for consistency and so I don't have to check the value of something that never changes anyway

reinkrul:
> e2e test would be nice, but not feasible I guess?

Would've been nice, but not feasible, no :-(

PR #3333, opened 2024-09-02
Change remaining DIDs in API path to subject and use client_id (from url)

### LINKED ISSUE #3216: IAM: client_id is assumed to be did:web DID in OpenID4VP flow
When an auth server received an OAuth2 Authz Code Flow request, it switches to OpenID4VP to request a presentation from the client requesting the access token. For this, it needs the AS metadata of the client, to sent the OpenDI4VP Authz Request to.

The metadata URL is determined by taking the `client_id`, which currently always is a did:web DID, converting it to a URL and then fetching the metadata from the well-known endpoint. Since we're moving towards explicitly specifying Auth Server URLs, away from assuming convertability from the DID to a URL, this should be changed.

Options:
- Pass client metadata URL in authz request (not sure if it's in core OAuth2 spec?)
- Change client_id to a URL
- ...?

### LINKED ISSUE #3258: Subject should be provided to Resource Server by PEP
There should be a way to derive the subject ID from a DID: when a remote party accesses data using a service-to-service access token, the PEP extracts the DID of the local party, which is then passed to the upstream resource server.

But the vendor applications now interact with the Nuts node using the subject ID instead of DID, so applications administer subject ID of DID (e.g. in mapping to a tenant/customer-specific FHIR store). Thus; it needs to know which subject the token is about, instead of just the DID. Options:

- Have an API that returns the subject ID for a particular DID (and all related DIDs and other useful info)
- Have Token Introspection provide the subject as issuer

2nd is probably the most correct.

woutslakhorst:
Closes #3216
Closes #3258

- the `client_id` in JAR uses the tenant URL (ex: `https://example.com/oauth2/tenant`) and not a `client_id` from the app. This is done so the JAR validate can use the OpenID federation configuration endpoint to resolve the `JWKs`. Future changes could generate a deployment specific client_id and keys.
- there are 3 or 4 instances of `subject.ListDIDs()[0]`. It does not take into account which methods the other side supports. This is a general todo for us.
- I removed `did` client_id scheme which interferes with PoCs

### DIFF
--- a/auth/api/iam/api.go
+++ b/auth/api/iam/api.go
@@ -734,25 +734,6 @@ func (r Wrapper) PresentationDefinition(ctx context.Context, request Presentatio
 	return PresentationDefinition200JSONResponse(result), nil
 }
 
-// toOwnedDIDForOAuth2 is like toOwnedDID but wraps the errors in oauth.OAuth2Error to make sure they're returned as specified by the OAuth2 RFC.
-func (r Wrapper) toOwnedDIDForOAuth2(ctx context.Context, didAsString string) (*did.DID, error) {
-	result, err := r.toOwnedDID(ctx, didAsString)
-	if err != nil {
-		if strings.HasPrefix(err.Error(), "DID resolution failed") {
-			return nil, oauth.OAuth2Error{
-				Code:        oauth.ServerError,
-				Description: err.Error(),
-			}
-		} else {
-			return nil, oauth.OAuth2Error{
-				Code:        oauth.InvalidRequest,
-				Description: err.Error(),
-			}
-		}
-	}
-	return result, nil
-}
-
 func (r Wrapper) toOwnedDID(ctx context.Context, didAsString string) (*did.DID, error) {
 	ownDID, err := did.ParseDID(didAsString)
 	if err != nil {
--- a/auth/api/iam/api.go
+++ b/auth/api/iam/api.go
@@ -37,6 +37,7 @@ import (
 	"net/http"
 	"net/url"
 	"strings"
+	"sync/atomic"
 	"time"
 
 	"github.com/lestrrat-go/jwx/v2/jwt"
@@ -101,8 +102,8 @@ type Wrapper struct {
 	vdr            vdr.VDR
 	jwtSigner      nutsCrypto.JWTSigner
 	keyResolver    resolver.KeyResolver
-	jar            JAR
 	subjectManager didsubject.SubjectManager
+	_jar           atomic.Value
 }
 
 func New(
@@ -114,6 +115,7 @@ func New(
 	if err != nil {
 		panic(err)
 	}
+	keyResolver := resolver.DIDKeyResolver{Resolver: vdrInstance.Resolver()}
 	return &Wrapper{
 		auth:           authInstance,
 		policyBackend:  policyBackend,
@@ -123,14 +125,19 @@ func New(
 		subjectManager: subjectManager,
 		jsonldManager:  jsonldManager,
 		jwtSigner:      jwtSigner,
-		keyResolver:    resolver.DIDKeyResolver{Resolver: vdrInstance.Resolver()},
-		jar: &jar{
-			auth:        authInstance,
-			jwtSigner:   jwtSigner,
-			keyResolver: resolver.DIDKeyResolver{Resolver: vdrInstance.Resolver()},
-			resolver:    vdrInstance.Resolver(),
-		},
+		keyResolver:    keyResolver,
 	}
+}
+
+func (r Wrapper) jar() JAR {
+	// so we can mock it in tests
+	var current JAR
+	var ok bool
+	if current, ok = r._jar.Load().(JAR); !ok {
+		current = NewJAR(r.auth, r.jwtSigner, r.keyResolver, r.auth.IAMClient())
+		r._jar.Store(current)
+	}
+	return current
 }
 
 func (r Wrapper) Routes(router core.EchoRouter) {
@@ -235,7 +242,6 @@ func (r Wrapper) HandleTokenRequest(ctx context.Context, request HandleTokenRequ
 				Description: "missing required parameters",
 			}
 		}
-		// todo pass clientID? (2b added to spec)
 		return r.handleS2SAccessTokenRequest(ctx, request.Subject, *request.Body.Scope, *request.Body.PresentationSubmission, *request.Body.Assertion)
 	default:
 		return nil, oauth.OAuth2Error{
@@ -457,7 +463,7 @@ func (r Wrapper) HandleAuthorizeRequest(ctx context.Context, request HandleAutho
 // The caller must ensure ownDID is actually owned by this node.
 func (r Wrapper) handleAuthorizeRequest(ctx context.Context, subject string, ownMetadata oauth.AuthorizationServerMetadata, request url.URL) (HandleAuthorizeRequestResponseObject, error) {
 	// parse and validate as JAR (RFC9101, JWT Authorization Request)
-	requestObject, err := r.jar.Parse(ctx, ownMetadata, request.Query())
+	requestObject, err := r.jar().Parse(ctx, ownMetadata, request.Query())
 	if err != nil {
 		// already an oauth.OAuth2Error
 		return nil, err
@@ -523,7 +529,7 @@ func (r Wrapper) RequestJWTByGet(ctx context.Context, request RequestJWTByGetReq
 	}
 
 	// TODO: supported signature types should be checked
-	token, err := r.jar.Sign(ctx, ro.Claims)
+	token, err := r.jar().Sign(ctx, ro.Claims)
 	if err != nil {
 		return nil, oauth.OAuth2Error{
 			Code:          oauth.ServerError,
@@ -577,7 +583,7 @@ func (r Wrapper) RequestJWTByPost(ctx context.Context, request RequestJWTByPostR
 	}
 
 	// TODO: supported signature types should be checked
-	token, err := r.jar.Sign(ctx, ro.Claims)
+	token, err := r.jar().Sign(ctx, ro.Claims)
 	if err != nil {
 		return nil, oauth.OAuth2Error{
 			Code:          oauth.ServerError,
@@ -601,7 +607,7 @@ func (r Wrapper) OAuthAuthorizationServerMetadata(_ context.Context, request OAu
 	return OAuthAuthorizationServerMetadata200JSONResponse(*md), nil
 }
 
-func (r Wrapper) oauthAuthorizationServerMetadata(clientID *url.URL) (*oauth.AuthorizationServerMetadata, error) {
+func (r Wrapper) oauthAuthorizationServerMetadata(clientID url.URL) (*oauth.AuthorizationServerMetadata, error) {
 	md := authorizationServerMetadata(clientID, r.vdr.SupportedMethods())
 	if !r.auth.AuthorizationEndpointEnabled() {
 		md.AuthorizationEndpoint = ""
@@ -618,7 +624,7 @@ func (r Wrapper) OAuthClientMetadata(ctx context.Context, request OAuthClientMet
 
 	identityURL := r.subjectToBaseURL(request.Subject)
 
-	return OAuthClientMetadata200JSONResponse(clientMetadata(*identityURL)), nil
+	return OAuthClientMetadata200JSONResponse(clientMetadata(identityURL)), nil
 }
 
 func (r Wrapper) OpenIDConfiguration(ctx context.Context, request OpenIDConfigurationRequestObject) (OpenIDConfigurationResponseObject, error) {
@@ -896,7 +902,7 @@ func (r Wrapper) createAuthorizationRequest(ctx context.Context, subject string,
 
 	// request_uri
 	requestURIID := nutsCrypto.GenerateNonce()
-	requestObj := r.jar.Create(*clientDID, clientID.String(), audience, modifier)
+	requestObj := r.jar().Create(*clientDID, clientID.String(), audience, modifier)
 	if err := r.authzRequestObjectStore().Put(requestURIID, requestObj); err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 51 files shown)


auth/api/iam/api.go
```diff
@@ -155,8 +165,8 @@ func (r Wrapper) Routes(router core.EchoRouter) {
 			// The following URLs require a user session:
 			paths := []string{
 				"/oauth2/:subjectID/user",
-				"/oauth2/:did/authorize",
-				"/oauth2/:did/callback",
+				"/oauth2/:subjectID/authorize",
```
reinkrul, line 168: in the other URLs they're called `:subject`, might be good to keep them constant

auth/api/iam/api.go
```diff
@@ -226,13 +236,13 @@ func (r Wrapper) HandleTokenRequest(ctx context.Context, request HandleTokenRequ
 		}
 	case oauth.VpTokenGrantType:
 		// Nuts RFC021 vp_token bearer flow
-		if request.Body.PresentationSubmission == nil || request.Body.Scope == nil || request.Body.Assertion == nil {
+		if request.Body.PresentationSubmission == nil || request.Body.Scope == nil || request.Body.Assertion == nil || request.Body.ClientId == nil {
```
reinkrul, line 239: derive client_id from subject in URL

auth/api/iam/api.go
```diff
@@ -433,26 +443,27 @@ func (r Wrapper) HandleAuthorizeRequest(ctx context.Context, request HandleAutho
 			Description: "authorization endpoint is disabled",
 		}
 	}
-	ownDID, err := r.toOwnedDIDForOAuth2(ctx, request.Did)
+	err := r.subjectExists(ctx, request.SubjectID)
 	if err != nil {
 		return nil, err
 	}
-	metadata, err := r.oauthAuthorizationServerMetadata(ownDID)
+	clientID := r.subjectToBaseURL(request.SubjectID)
```
reinkrul, line 450: `getClientID()`?

auth/api/iam/api.go
```diff
@@ -467,20 +478,8 @@ func (r Wrapper) handleAuthorizeRequest(ctx context.Context, ownDID did.DID, own
...
-		if strings.HasPrefix(clientId, "did:web:") {
-			// client is a cloud wallet with user
-			return r.handleAuthorizeRequestFromHolder(ctx, ownDID, requestObject)
-		} else {
-			return nil, oauth.OAuth2Error{
-				Code:        oauth.InvalidRequest,
-				Description: "client_id must be a did:web",
-			}
-		}
+		// if client_id is a url, we can use OpenID federation for automatic client registration
```
reinkrul, line 481: ?

auth/api/iam/api.go
```diff
@@ -623,18 +617,90 @@ func (r Wrapper) oauthAuthorizationServerMetadata(ownDID *did.DID) (*oauth.Autho
...
+}
+
+func (r Wrapper) OpenIDConfiguration(ctx context.Context, request OpenIDConfigurationRequestObject) (OpenIDConfigurationResponseObject, error) {
+	// find subject
+	exists, err := r.subjectManager.Exists(ctx, request.Subject)
 	if err != nil {
-		return nil, err
+		return nil, oauth.OAuth2Error{
+			Code:          oauth.ServerError,
+			Description:   "internal server error", // security reasons
```
reinkrul, line 636: coudl leave out description then, not mandatory

auth/api/iam/api.go
```diff
@@ -623,18 +617,90 @@ func (r Wrapper) oauthAuthorizationServerMetadata(ownDID *did.DID) (*oauth.Autho
...
+		}
+	}
+	if !exists {
+		return nil, oauth.OAuth2Error{
+			Code:        oauth.InvalidRequest,
+			Description: "subject not found",
+		}
+	}
+	// find DIDs for subject
+	dids, err := r.subjectManager.List(ctx, request.Subject)
```
reinkrul, line 647: 2nd database call, could leave out `Exists` all and check for `ErrUnknownSubject` (or something)

auth/api/iam/api.go
```diff
@@ -623,18 +617,90 @@ func (r Wrapper) oauthAuthorizationServerMetadata(ownDID *did.DID) (*oauth.Autho
...
+	if err != nil {
+		return nil, oauth.OAuth2Error{
+			Code:          oauth.ServerError,
+			Description:   "internal server error", // security reasons
```
reinkrul, line 651: leave out?

auth/api/iam/api.go
```diff
@@ -623,18 +617,90 @@ func (r Wrapper) oauthAuthorizationServerMetadata(ownDID *did.DID) (*oauth.Autho
...
+	}
+	// resolve DID keys
+	set := jwk.NewSet()
+	var signingKey string
+	for _, did := range dids {
+		kid, key, err := r.keyResolver.ResolveKey(did, nil, resolver.AssertionMethod)
+		if err != nil {
+			return nil, oauth.OAuth2Error{
+				Code:          oauth.ServerError,
+				Description:   "internal server error", // security reasons
```
reinkrul, line 663: again

auth/api/iam/api.go
```diff
@@ -623,18 +617,90 @@ func (r Wrapper) oauthAuthorizationServerMetadata(ownDID *did.DID) (*oauth.Autho
...
+				InternalError: err,
+			}
+		}
+		// create JWK and add to set
+		jwkKey, err := jwk.FromRaw(key)
+		if err != nil {
+			return nil, oauth.OAuth2Error{
+				Code:          oauth.ServerError,
+				Description:   "internal server error", // security reasons
```
reinkrul, line 672: and again

auth/api/iam/api.go
```diff
@@ -888,25 +929,13 @@ func (r Wrapper) authzRequestObjectStore() storage.SessionStore {
...
-	if tenantIdx != -1 {
-		basePath = didURL.Path[:tenantIdx]
-	}
-	result, err := didURL.Parse("/")
-	if err != nil {
-		return nil, err
+func (r Wrapper) subjectToBaseURL(subject string) url.URL {
+	u := &url.URL{}
+	publicURL := r.auth.PublicURL()
+	if publicURL != nil {
```
reinkrul, line 936: It's mandatory, so can't be nil

reinkrul (changes requested):

auth/api/iam/jar.go
```diff
@@ -92,14 +105,14 @@ func createJarRequest(client did.DID, authServerURL string, modifier requestObje
...
-		Client:           client,
+		Client:           clientID,
 		RequestURIMethod: requestURIMethod,
 	}
 }
 
 func (j jar) Sign(ctx context.Context, claims oauthParameters) (string, error) {
-	clientID := claims.get(oauth.ClientIDParam)
-	clientDID, err := did.ParseDID(clientID)
+	issuerID := claims.get(jwt.IssuerKey)
```
reinkrul, line 114: future refactor we'll introduce a `Client` that has an `ID` property and `SigningKey()` function..

auth/api/iam/jar.go
```diff
@@ -163,14 +179,36 @@ func (j jar) validate(ctx context.Context, rawToken string, clientId string) (oa
 	if clientId != params.get(oauth.ClientIDParam) {
 		return nil, oauth.OAuth2Error{Code: oauth.InvalidRequestObject, Description: "invalid client_id claim in signed authorization request"}
 	}
-	// check if the signer of the JWT is the client
-	signer, err := did.ParseDIDURL(signerKid)
+	configuration, err := j.client.OpenIDConfiguration(ctx, clientId)
```
reinkrul, line 182: can't remember, is this the caching HTTP client? Doesn't matter for now, might be a future TODO (nobody is going to use this in production anyways)

auth/api/iam/openid4vp.go
```diff
@@ -193,15 +187,22 @@ func (r Wrapper) nextOpenID4VPFlow(ctx context.Context, state string, session OA
...
-		// See https://github.com/nuts-foundation/nuts-node/issues/3216
-		authServerURL, _ := nutsOAuth2Issuer(*walletDID)
-		redirectURL, err = r.createAuthorizationRequest(ctx, *session.OwnDID, authServerURL.String(), modifier)
+		// fetch openId configuration
+		configuration, innerErr := r.auth.IAMClient().OpenIDConfiguration(ctx, session.ClientID)
+		if innerErr != nil {
+			return nil, oauth.OAuth2Error{
+				Code:          oauth.ServerError,
+				Description:   "failed to authorize client",
+				InternalError: fmt.Errorf("failed to fetch OpenIDConfiguration: %w", err),
```
reinkrul, line 201: shouldn't it wrap `innerErr`?

auth/api/iam/user.go
```diff
@@ -208,14 +200,3 @@ func (r Wrapper) issueEmployeeCredential(ctx context.Context, session user.Sessi
 	}
 	return employeeCredential, nil
 }
-
-// nutsOAuth2Issuer returns the URL of the OAuth2 issuer for the given DID.
-// It's temporary: soon to be released by the application explicitly specifying the OAuth2 issuer URL.
-func nutsOAuth2Issuer(subject did.DID) (*url.URL, error) {
```
reinkrul, line 214: it's gooooonee!

auth/oauth/types.go
```diff
@@ -397,3 +400,46 @@ type OpenIDCredentialIssuerMetadata struct {
...
+
+// OpenIDConfiguration represents the OpenID configuration
+// It contains the minimal information required for OpenID4VP, the required `jwks` is also omitted
+// see https://openid.net/specs/openid-connect-federation-1_0-29.html#entity-statement
+type OpenIDConfiguration struct {
+	// - Issuer: an url representing the issuer of the entity statement
+	// for now we keep it teh same as the subject, eg the subject/tenant
+	Issuer string `json:"iss"`
+	// - Subject: an url representing the subject of the entity statement
+	Subject string `json:"sub"`
```
reinkrul, line 412: weird godoc format

auth/oauth/types.go
```diff
@@ -397,3 +400,46 @@ type OpenIDCredentialIssuerMetadata struct {
...
+	// IssuedAt: the time the entity statement was issued
+	IssuedAt int64 `json:"iat"`
+	// JWKs is the JSON Web Key Set of the entity statement. Contains keys of all DIDs for the subject
+	JWKs jwk.Set `json:"jwks"`
+	// OpenIDProvider: the metadata of the OpenID provider. We need it as Authorization server metadata
+	OpenIDProvider AuthorizationServerMetadata `json:"openid_provider"`
```
reinkrul, line 418: shouldn't this be wrapped in a `metadata` field? (according to the example in https://openid.net/specs/openid-connect-federation-1_0-29.html#entity-statement)

auth/oauth/types.go
```diff
@@ -397,3 +400,46 @@ type OpenIDCredentialIssuerMetadata struct {
...
+}
+
+// UnmarshalJSON parses the OpenIDConfiguration from JSON
+func (j *OpenIDConfiguration) UnmarshalJSON(bytes []byte) error {
+	claims := make(map[string]interface{})
```
reinkrul, line 423: regular marshal won't work?

auth/oauth/types.go
```diff
@@ -397,3 +400,46 @@ type OpenIDCredentialIssuerMetadata struct {
 	// - Display: a slice of maps where each map represents the display information (optional)
 	Display []map[string]string `json:"display,omitempty"`
 }
+
+// OpenIDConfiguration represents the OpenID configuration
+// It contains the minimal information required for OpenID4VP, the required `jwks` is also omitted
+// see https://openid.net/specs/openid-connect-federation-1_0-29.html#entity-statement
+type OpenIDConfiguration struct {
```
reinkrul, line 407: Why do we need this again, if we're changing client_id from DID to URL?

auth/api/iam/api.go
woutslakhorst: Will switch to `subjectID` since `subject` might be confusing for people not linking it to *DID subject*

auth/api/iam/api.go
woutslakhorst: the `client_id` here is from the requester, the subject in the path is the issuer of the token

auth/api/iam/api.go
woutslakhorst: additional comment on how to resolve client based on client_id. This snippet refers to openID federation

auth/api/iam/api.go
```diff
@@ -818,45 +865,39 @@ func (r Wrapper) openid4vciMetadata(ctx context.Context, issuer string) (*oauth.
...
+		return nil, fmt.Errorf("no authorization endpoint found in metadata for %s", metadata.Issuer)
 	}
+
 	endpoint, err := url.Parse(metadata.AuthorizationEndpoint)
 	if err != nil {
 		return nil, fmt.Errorf("failed to parse authorization endpoint URL: %w", err)
 	}
 
+	clientID := r.subjectToBaseURL(subject)
+	clientDID, err := r.determineClientDID(ctx, metadata, subject)
```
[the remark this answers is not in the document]
woutslakhorst: renamed

auth/api/iam/api.go
woutslakhorst: yes, but the road from core serverConfig to place of use is long and there are many tests that ignore it (but do call it). Changed it to a panic for the API.

auth/api/iam/api.go
woutslakhorst: ?

auth/oauth/types.go
woutslakhorst: `jwk.Set` is an interface.....

auth/oauth/types.go
woutslakhorst: damn copilot

auth/oauth/types.go
woutslakhorst: for the JAR key lookup

reinkrul (approved):

auth/api/iam/api.go
```diff
@@ -579,7 +572,9 @@ func (r Wrapper) RequestJWTByPost(ctx context.Context, request RequestJWTByPostR
 			ro.Claims[oauth.WalletNonceParam] = *request.Body.WalletNonce
 		}
 	}
-	ro.Claims[jwt.AudienceKey] = walletMetadata.Issuer
+	if walletMetadata.Issuer != "https://self-issued.me/v2" {
+		ro.Claims[jwt.AudienceKey] = walletMetadata.Issuer
+	}
```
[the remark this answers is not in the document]
woutslakhorst: logic hasn't changed

auth/api/iam/api.go
```diff
@@ -623,18 +611,77 @@ func (r Wrapper) oauthAuthorizationServerMetadata(ownDID *did.DID) (*oauth.Autho
...
+		jwkKey, err := jwk.FromRaw(key)
+		if err != nil {
+			return nil, oauth.OAuth2Error{
+				Code:          oauth.ServerError,
+				InternalError: err,
+			}
+		}
+		_ = jwkKey.Set(jwk.KeyIDKey, kid)
+		_ = set.AddKey(jwkKey)
+		signingKey = kid
```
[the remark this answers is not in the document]
woutslakhorst: I'll add comment to that PR

PR #3338, opened 2024-09-03
VDR: Validate subject to a sensible pattern

reinkrul:
I encountered this error while developing Nuts Admin, which send an empty string when it wasn't filled in, in the UI.

To prevent hard-to-read subjects which then need to be URL encoded, I think it makes sense to limit allowed characters to those that aren't URL encoded.

This also disallows empty "subject", which is currently allowed (which will break at some point).


woutslakhorst:
`');drop table did;`

reinkrul:
> `');drop table did;`

this would be another invalid subject indeed

woutslakhorst (approved):

PR #3342, opened 2024-09-04
Replace did param in DPoP proof API to kid

### LINKED ISSUE #3314: DPoP header API needs to know what key to use for signing
DPoP consists of 2 steps
1. During the access_token request, the client communicates a key that the token will be bound to.
2. During a resource request, the client signs a DPoP header with the key that the access_token is bound to. The authorization server should only grant access if this is true.

We currently provide API `/interal/auth/v2/{did}/dpop` to create this header. This selects a key based on the DID in the path. There are several issues with this:
- the XIS that makes this call has requested an access_token for a subject, it is not aware of the DID that was actually used
- even if it does know the DID, the DID document may contain multiple keys that could be used

To solve this we need to make some changes to the `TokenResponse` schema returned from
`/internal/auth/v2/accesstoken/{sessionID}` and `/internal/auth/v2/{subjectID}/request-service-access-token`. 
The `TokenResponse` could include the `verificationMethod.id` of the key used in DPoP step 1.
Alternatively we could store some state in the Nuts node to allow returning both `DPoP` and `Authorization` headers using a session ID.

woutslakhorst:
closes #3314 

depends on #3333 

The access token request now also returns a kid that is used as path param.

### DIFF
--- a/auth/api/iam/jar.go
+++ b/auth/api/iam/jar.go
@@ -27,6 +27,7 @@ import (
 	"github.com/lestrrat-go/jwx/v2/jwt"
 	"github.com/nuts-foundation/go-did/did"
 	"github.com/nuts-foundation/nuts-node/auth"
+	"github.com/nuts-foundation/nuts-node/auth/client/iam"
 	"github.com/nuts-foundation/nuts-node/auth/oauth"
 	cryptoNuts "github.com/nuts-foundation/nuts-node/crypto"
 	"github.com/nuts-foundation/nuts-node/vdr/resolver"
@@ -48,6 +49,7 @@ type jar struct {
 	auth        auth.AuthenticationServices
 	jwtSigner   cryptoNuts.JWTSigner
 	keyResolver resolver.KeyResolver
+	client      iam.Client
 }
 
 type JAR interface {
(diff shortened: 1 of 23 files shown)


crypto/dpop/dpop.go
```diff
@@ -59,6 +59,7 @@ const maxJtiLength = 256
 type DPoP struct {
 	raw     string
 	Headers jws.Headers `json:"-"`
+	Kid     string      `json:"-"`
 	Token   jwt.Token   `json:"-"`
```
gerardsn, line 63: why are there tags if all fields are ignored

auth/client/iam/openid4vp.go
```diff
@@ -339,15 +343,19 @@ func (c *OpenID4VPClient) walletWithExtraCredentials(ctx context.Context, subjec
 	}), nil
 }
 
-func (c *OpenID4VPClient) dpop(ctx context.Context, requester did.DID, request http.Request) (string, error) {
+func (c *OpenID4VPClient) dpop(ctx context.Context, requester did.DID, request http.Request) (string, *string, error) {
```
gerardsn, line 346: weird signature. Unclear what the outputs are and why one is a pointer if it is always used. change to struct?

auth/api/iam/jar.go
gerardsn, line 1: remove changes

gerardsn (approved):

PR #3354, opened 2024-09-10
remove sub from access token introspection

### LINKED ISSUE #3351: make "sub" empty in token introspection
`sub` is used as the owner of the resource. In OAuth usage this is the user that owns it regardless of which system it logs into. Currently it's the same as the `iss` field. For future use and to prevent mistakes, we empty it.

woutslakhorst:
closes #3351

### DIFF
--- a/docs/_static/auth/v2.yaml
+++ b/docs/_static/auth/v2.yaml
@@ -594,10 +594,6 @@ components:
           type: string
           description: Contains the DID of the authorizer. Should be equal to 'sub'
           example: did:web:example.com:resource-owner
-        sub:
-          type: string
-          description: Contains the DID of the resource owner
-          example: did:web:example.com:resource-owner
         aud:
           type: string
           description: RFC7662 - Service-specific string identifier or list of string identifiers representing the intended audience for this token, as defined in JWT [RFC7519].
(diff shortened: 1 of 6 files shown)


docs/_static/auth/v2.yaml
```diff
@@ -594,10 +594,6 @@ components:
           type: string
           description: Contains the DID of the authorizer. Should be equal to 'sub'
           example: did:web:example.com:resource-owner
-        sub:
-          type: string
-          description: Contains the DID of the resource owner
-          example: did:web:example.com:resource-owner
         aud:
           type: string
           description: RFC7662 - Service-specific string identifier or list of string identifiers representing the intended audience for this token, as defined in JWT [RFC7519].
```
gerardsn, line 599: `aud` should also be removed?

gerardsn (changes requested):
failing test

docs/_static/auth/v2.yaml
woutslakhorst: is not used in s2s flow afaik

reinkrul (approved):

gerardsn (approved):

PR #3357, opened 2024-09-11
allow for additional parameters when activating discovery service

### LINKED ISSUE #3352: add issuer_id/client_id to discovery service
Add the fully qualified entity_id/client_id (`https://example.com/oauth2/1234-1234-12345678-1234`) to the VP of the service discovery.

This must be done in a "holder credential". The type of this credential would be `NutsOAuthCredential` with a single `authServerURL` field under credentialSubject.

### LINKED ISSUE #3353: Allow addition of holder credentials to use case activation
These credentials can then also be required from the Presentation Definition. Some DB changes are probably required.

- This makes sure that the endpoints are signed.
- One less registration step for users.
- One less API needed (endpoint resolve)
- Check on available endpoints via PD

As examples we'll use a NutsEndpointCredential with `{"type":"XXX", "endpoint":"YYYYY"}` as credentialSubject.

woutslakhorst:
closes #3353 
closes #3352 

also returns the params with a search.

### DIFF
--- a/discovery/module.go
+++ b/discovery/module.go
@@ -35,6 +35,7 @@ import (
 	"net/url"
 	"os"
 	"path"
+	"slices"
 	"strings"
 	"sync"
 	"time"
@@ -92,6 +93,7 @@ type Module struct {
 	ctx                 context.Context
 	cancel              context.CancelFunc
 	routines            *sync.WaitGroup
+	publicURL           *url.URL
 }
 
 func (m *Module) Configure(serverConfig core.ServerConfig) error {
@@ -106,6 +108,11 @@ func (m *Module) Configure(serverConfig core.ServerConfig) error {
 			return nil
 		}
 		return fmt.Errorf("failed to load discovery defintions: %w", err)
+	}
+
+	m.publicURL, err = serverConfig.ServerURL()
+	if err != nil {
+		return err
 	}
 
 	m.allDefinitions, err = loadDefinitions(m.config.Definitions.Directory)
@@ -333,10 +340,16 @@ func forwardedHost(ctx context.Context) string {
 
 // ActivateServiceForSubject is a Discovery Client function that activates a service for a subject.
 // See interface.go for more information.
-func (m *Module) ActivateServiceForSubject(ctx context.Context, serviceID, subjectID string) error {
+func (m *Module) ActivateServiceForSubject(ctx context.Context, serviceID, subjectID string, parameters map[string]interface{}) error {
 	log.Logger().Debugf("Activating service for subject (subject=%s, service=%s)", subjectID, serviceID)
 
-	err := m.registrationManager.activate(ctx, serviceID, subjectID)
+	// create authServerURL and add to parameters
+	if parameters == nil {
+		parameters = make(map[string]interface{})
+	}
+	parameters[authServerURLField] = m.publicURL.JoinPath("/oauth2/subjectID").String()
+
+	err := m.registrationManager.activate(ctx, serviceID, subjectID, parameters)
 	if err != nil {
 		if errors.Is(err, ErrPresentationRegistrationFailed) {
 			log.Logger().WithError(err).Warnf("Presentation registration failed, will be retried later (subject=%s,service=%s)", subjectID, serviceID)
@@ -449,12 +462,38 @@ func (m *Module) Search(serviceID string, query map[string]string) ([]SearchResu
 			}
 		}
 
+		// extract registrationParameters from VP
+		registrationParameters := extractParameters(matchingVP)
+
 		result = append(result, SearchResult{
 			Presentation: matchingVP,
 			Fields:       fields,
+			Parameters:   registrationParameters,
 		})
 	}
 	return result, nil
+}
+
+func extractParameters(vp vc.VerifiablePresentation) map[string]interface{} {
+	result := make(map[string]interface{})
+	credentials := vp.VerifiableCredential
+	for _, cred := range credentials {
+		if slices.Contains(cred.Type, credential.DiscoveryRegistrationCredentialTypeV1URI()) {
+			credentialSubject := make([]credential.DiscoveryRegistrationCredentialSubject, 0)
+			err := cred.UnmarshalCredentialSubject(&credentialSubject)
+			if err != nil {
+				log.Logger().WithError(err).Error("Failed to unmarshal Discovery Registration Credential Subject")
+				continue
+			}
+			if len(credentialSubject) > 0 {
+				result = credentialSubject[0]
+				// remove id
+				delete(result, "id")
+			}
+		}
+	}
+
+	return result
 }
 
 func (m *Module) update() {
--- a/discovery/module.go
+++ b/discovery/module.go
@@ -343,11 +343,13 @@ func forwardedHost(ctx context.Context) string {
 func (m *Module) ActivateServiceForSubject(ctx context.Context, serviceID, subjectID string, parameters map[string]interface{}) error {
 	log.Logger().Debugf("Activating service for subject (subject=%s, service=%s)", subjectID, serviceID)
 
-	// create authServerURL and add to parameters
 	if parameters == nil {
 		parameters = make(map[string]interface{})
 	}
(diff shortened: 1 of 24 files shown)


gerardsn (changes requested):

discovery/module.go
```diff
@@ -333,10 +340,16 @@ func forwardedHost(ctx context.Context) string {
...
-func (m *Module) ActivateServiceForSubject(ctx context.Context, serviceID, subjectID string) error {
+func (m *Module) ActivateServiceForSubject(ctx context.Context, serviceID, subjectID string, parameters map[string]interface{}) error {
 	log.Logger().Debugf("Activating service for subject (subject=%s, service=%s)", subjectID, serviceID)
 
-	err := m.registrationManager.activate(ctx, serviceID, subjectID)
+	// create authServerURL and add to parameters
+	if parameters == nil {
+		parameters = make(map[string]interface{})
+	}
+	parameters[authServerURLField] = m.publicURL.JoinPath("/oauth2/subjectID").String()
```
reinkrul: Only do this if not already set?

discovery/module.go
```diff
@@ -449,14 +462,40 @@ func (m *Module) Search(serviceID string, query map[string]string) ([]SearchResu
...
 
+func extractParameters(vp vc.VerifiablePresentation) map[string]interface{} {
+	result := make(map[string]interface{})
+	credentials := vp.VerifiableCredential
+	for _, cred := range credentials {
+		if slices.Contains(cred.Type, credential.DiscoveryRegistrationCredentialTypeV1URI()) {
+			credentialSubject := make([]credential.DiscoveryRegistrationCredentialSubject, 0)
+			err := cred.UnmarshalCredentialSubject(&credentialSubject)
+			if err != nil {
+				log.Logger().WithError(err).Error("Failed to unmarshal Discovery Registration Credential Subject")
```
reinkrul, line 485: maybe add VP ID, so you can track which one it was. Is also an INFO level message at most, I'd say. Since t's not something you can fix in your own infra (probably issue at another participant).

docs/_static/discovery/v1.yaml
```diff
@@ -215,18 +221,41 @@ components:
...
         - fields
+        - registrationParameters
       properties:
         id:
           type: string
           description: The ID of the Verifiable Presentation.
         credential_subject_id:
           type: string
           description: The ID of the Verifiable Credential subject (holder), typically a DID.
+        registrationParameters:
```
reinkrul, line 232: Why not have it extracted through the Service Discovery definition, just like the other claims? Then you can address it as "auth_server_url", for instance.

discovery/module.go
```diff
@@ -449,14 +462,40 @@ func (m *Module) Search(serviceID string, query map[string]string) ([]SearchResu
...
 		result = append(result, SearchResult{
 			Presentation: matchingVP,
 			Fields:       fields,
+			Parameters:   registrationParameters,
 		})
 	}
 	return result, nil
 }
 
+func extractParameters(vp vc.VerifiablePresentation) map[string]interface{} {
```
reinkrul, line 477: I'd not do this, and use the same principle with other extracted fields, using input descriptor field constraint IDs (like organization_name or ura)

discovery/store.go
```diff
@@ -296,14 +299,18 @@ func (s *sqlStore) removeExpired() (int, error) {
...
 	return s.db.Transaction(func(tx *gorm.DB) error {
 		if nextRefresh == nil {
 			// Delete registration
 			return tx.Delete(&presentationRefreshRecord{}, "service_id = ? AND subject_id = ?", serviceID, subjectID).Error
 		}
 		// Create or update it
-		return tx.Save(presentationRefreshRecord{SubjectID: subjectID, ServiceID: serviceID, NextRefresh: nextRefresh.Unix()}).Error
+		var bytes []byte
+		if parameters != nil {
+			bytes, _ = json.Marshal(parameters)
```
reinkrul, line 311: returning err won't hurt?

vcr/pe/presentation_definition.go
```diff
@@ -294,6 +294,11 @@ func matchFormat(format *PresentationDefinitionClaimFormatDesignations, credenti
 		return true
 	}
 
+	if credential.Format() == "" {
+		// holder credential
```
reinkrul, line 298: what has holder credential (actually "issuer is holder"-credential) to do with format? Weird comment?

reinkrul (changes requested):

docs/_static/discovery/v1.yaml
woutslakhorst: because you always have to add all registered params in the PD. You can still do this, this just helps in getting it all. Very handy for the `authServerURL` since it's always present. Adding that to the PD every time is annoying.

vcr/pe/presentation_definition.go
woutslakhorst: Format is extracted from the proof. Holder credential has no Proof and thus no format.

discovery/module.go
```diff
@@ -449,14 +462,40 @@ func (m *Module) Search(serviceID string, query map[string]string) ([]SearchResu
...
+			err := cred.UnmarshalCredentialSubject(&credentialSubject)
+			if err != nil {
+				log.Logger().WithError(err).Error("Failed to unmarshal Discovery Registration Credential Subject")
+				continue
+			}
+			if len(credentialSubject) > 0 {
+				result = credentialSubject[0]
+				// remove id
+				delete(result, "id")
+			}
```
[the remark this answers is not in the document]
woutslakhorst: added more comments

discovery/module.go
```diff
@@ -449,14 +462,40 @@ func (m *Module) Search(serviceID string, query map[string]string) ([]SearchResu
...
+	result := make(map[string]interface{})
+	credentials := vp.VerifiableCredential
+	for _, cred := range credentials {
+		if slices.Contains(cred.Type, credential.DiscoveryRegistrationCredentialTypeV1URI()) {
+			credentialSubject := make([]credential.DiscoveryRegistrationCredentialSubject, 0)
+			err := cred.UnmarshalCredentialSubject(&credentialSubject)
+			if err != nil {
+				log.Logger().WithError(err).Error("Failed to unmarshal Discovery Registration Credential Subject")
+				continue
+			}
```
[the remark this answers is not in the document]
woutslakhorst: added comment

docs/_static/discovery/v1.yaml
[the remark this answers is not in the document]
woutslakhorst: added comment on the const

storage/sql_migrations/004_discoveryservice_client_registration.sql
```diff
@@ -7,6 +8,8 @@ create table discovery_presentation_refresh
     service_id   varchar(200) not null,
     -- subject_id is the subject that should be registered on the Discovery Service.
     subject_id   varchar(370) not null,
+    -- parameters contains the registration parameters passed at activation.
```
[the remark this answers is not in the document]
woutslakhorst: added comment (JSON)

vcr/assets/assets/contexts/nuts.ldjson
```diff
@@ -83,6 +83,15 @@
         "@vocab": "schema"
       }
     },
+    "DiscoveryRegistrationCredential": {
```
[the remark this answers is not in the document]
woutslakhorst: added issue for it #3360

discovery/module.go
woutslakhorst: see other comment, using input descriptor field constraint IDs makes the PD too bloated.

discovery/store.go
woutslakhorst: doesn't hurt but in general only possible when you put a function ptr in the map.

vcr/pe/presentation_definition.go
reinkrul: The credential can still be in either JWT (unsigned) or JSON-LD format?

woutslakhorst: We have to rewrite some stuff. Let's hold that until VC data model 2 and it's new VC types. The two formats we're using now are gone in 202x?

PR #3363, opened 2024-09-13
backport - write DAG XOR tree consistency fixes to disk

gerardsn:
does this need a patch release, or should we just leave it in this branch to be released together with some other fix?


woutslakhorst (approved):

woutslakhorst:
probably release it too?

PR #3365, opened 2024-09-13
Auto set CreatedAt for new DID documents

### LINKED ISSUE #3364: VDR v2 DID creation does not set CreatedAt for new documents
since all DID updates copy the value from the previous document version, the CreatedAt value is 0 for all versions

gerardsn:
closes #3364

### DIFF
--- a/vdr/didsubject/did_document_test.go
+++ b/vdr/didsubject/did_document_test.go
@@ -83,9 +83,13 @@ func TestSqlDIDDocumentManager_CreateOrUpdate(t *testing.T) {
 		require.NoError(t, err)
 		require.NoError(t, tx.Commit().Error)
 
+		// rewrite timestamps to make sure docRoot.CreatedAt < doc.UpdatedAt
+		createdAt := time.Now().Add(-2 * time.Second).Unix()
+		docRoot.CreatedAt, docRoot.UpdatedAt = createdAt, createdAt
+		db.Save(&docRoot)
+
 		docManager = NewDIDDocumentManager(transaction(t, db))
 		require.NoError(t, err)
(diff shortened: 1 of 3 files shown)


vdr/didsubject/did_document_test.go
```diff
@@ -76,18 +79,22 @@ func TestSqlDIDDocumentManager_CreateOrUpdate(t *testing.T) {
...
 		tx := db.Begin()
 		docManager := NewDIDDocumentManager(tx)
-		_, err := docManager.CreateOrUpdate(sqlDidBob, []orm.VerificationMethod{vm}, []orm.Service{service})
+		docRoot, err := docManager.CreateOrUpdate(sqlDidBob, []orm.VerificationMethod{vm}, []orm.Service{service})
 		require.NoError(t, err)
 		require.NoError(t, tx.Commit().Error)
 
 		docManager = NewDIDDocumentManager(transaction(t, db))
 		require.NoError(t, err)
+		time.Sleep(time.Second) // make sure doc.CreatedAt < doc.UpdatedAt
```
woutslakhorst, line 88: try to prevent this. Could be done by setting a timeNow func in the package.

woutslakhorst (changes requested):

vdr/didsubject/did_document_test.go
woutslakhorst: but using gorm autoCreate makes this totally impossible...

woutslakhorst: you could manually adjust the field value in the DB (low level) and then call update

gerardsn:
Now also start counting document versions at 0 instead of 1

woutslakhorst (approved):

PR #3372, opened 2024-09-13
have wallet.List() only return non-revoked/non-expired credentials

### LINKED ISSUE #3362: Discovery registration selects expired credential
**Description**:

I issued a credential with an `expirationDate` in the past. Then issued a new credential with the date in the future. When registering the subject for the discovery service, it replies with the following:

```
Unexpected HTTP response (len=169): {\"detail\":\"presentation is invalid for registration\\npresentation is valid longer than the credential(s) it contains\",\"status\":400,\"title\":\"RegisterPresentation failed\"}" http_request_path=/discovery/hackathon_v2024 module=Discovery
```

**Current behavior:**
It seems the wallet selects the first matching credential without looking if it is still valid.

**Expected behavior:**
The wallet selects the first valid credential.

woutslakhorst:
fixes #3362 

`wallet.List()` now only returns non-revoked, non-expired credentials. Trust and signatures are not checked. Trust checking is done via PEX and signatures are checked at upload time.

### DIFF
--- a/vcr/holder/sql_wallet.go
+++ b/vcr/holder/sql_wallet.go
@@ -113,7 +113,23 @@ func (h sqlWallet) Put(_ context.Context, credentials ...vc.VerifiableCredential
 }
 
 func (h sqlWallet) List(_ context.Context, holderDID did.DID) ([]vc.VerifiableCredential, error) {
-	return h.walletStore.list(holderDID)
+	credentials, err := h.walletStore.list(holderDID)
+	if err != nil {
+		return nil, err
+	}
+	// now validate credentials and remove invalid ones
+	validCredentials := make([]vc.VerifiableCredential, 0, len(credentials))
+	for _, credential := range credentials {
+		// we only want to check expiration and revocation status
+		if err = h.verifier.Verify(credential, true, false, nil); err == nil {
+			validCredentials = append(validCredentials, credential)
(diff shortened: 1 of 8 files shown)


vcr/holder/sql_wallet.go
```diff
@@ -113,7 +113,23 @@ func (h sqlWallet) Put(_ context.Context, credentials ...vc.VerifiableCredential
 }
 
 func (h sqlWallet) List(_ context.Context, holderDID did.DID) ([]vc.VerifiableCredential, error) {
-	return h.walletStore.list(holderDID)
+	credentials, err := h.walletStore.list(holderDID)
+	if err != nil {
+		return nil, err
+	}
+	// now validate credentials and remove invalid ones
```
gerardsn, line 120: The `NutsAuthorizationCredential` is still heavily in use meaning that `credentials` is going to contain a lot of invalid credentials. Can we filter this during the SQL query?

reinkrul: I think they are not in the SQL wallet?

woutslakhorst: `vcr.go#canWalletHoldCredential` filters NutsAuthCreds so they won't go to the wallet. @gerardsn

gerardsn (approved):

PR #3375, opened 2024-09-16
fix ordering of DID document context

### LINKED ISSUE #3368: DID @context are added in the wrong order
what we produce:
did:nuts: `"@context":["https://nuts.nl/did/v1","https://w3c-ccg.github.io/lds-jws2020/contexts/lds-jws2020-v1.json","https://www.w3.org/ns/did/v1"]`
did:web: `"@context":["https://w3c-ccg.github.io/lds-jws2020/contexts/lds-jws2020-v1.json","https://www.w3.org/ns/did/v1"`

according to the DID spec https://www.w3.org/TR/did-core/#production-0:
> The serialized value of @ context MUST be the JSON String https://www.w3.org/ns/did/v1, or a JSON Array where **the first item** is the JSON String https://www.w3.org/ns/did/v1 and the subsequent items are ...

I believe this has to do with security context to prevent overloading values or something like that?

It's too late to fix this for `did:nuts`, but we should consider fixing it now that we are introducing new did methods.

(We've already fixed this for VC's)

woutslakhorst:
fixes #3368 

note: nuts did context is no longer used

### DIFF
--- a/e2e-tests/ops/create-subject/docker-compose.yml
+++ b/e2e-tests/ops/create-subject/docker-compose.yml
@@ -1,0 +1,11 @@
+services:
+  nodeA:
+    image: "${IMAGE_NODE_A:-nutsfoundation/nuts-node:master}"
+    ports:
+      - "18081:8081"
+    environment:
+      NUTS_CONFIGFILE: /opt/nuts/nuts.yaml
+    volumes:
+      - "./nuts.yaml:/opt/nuts/nuts.yaml:ro"
+    healthcheck:
+      interval: 1s # Make test run quicker by checking health status more often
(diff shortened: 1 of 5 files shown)


gerardsn (changes requested):
this should have a regression test on the first context value

gerardsn (approved):
**!!! I added 1 commit !!!**

An e2e-test seemed like the most reliable place to check all current and future DID docs. (i.e., no relevant parts are mocked and it does not require a list that needs to be updated to include new methods; assuming the default is to create a DID method for all available methods.) I'm fine leaving this out though

`did:jwk` does not contain contexts at all while the spec does list them;
`did:key` also lists contexts in the wrong order;
since both are only used internally this probably doesn't break anything

PR #3378, opened 2024-09-16
pass matcher details from PEX in error for non-matches

### LINKED ISSUE #3007: Error "invalid_request - no credentials available" needs more pointers for fixing
When the Nuts node wallet can't match a Presentation Definition during an OpenID4VP flow, the following error occurs:

```
invalid_request - no credentials available
```

This is returned to the verifier and ultimately shown to the user. I forecast this is one of the most often seen errors, but probably does not give enough info to developers to fix this.

Users might also see this error, so we should either use error URIs (so developers can map the error for their users), or make it more user-friendly.

### LINKED ISSUE #3025: PEX: add verbose PD matching option for analysis
Presentation Definition matching failures (when a wallet does not yield the required/expected credentials) is sometimes really hard (and an arcane craft), since:
- You don't know exactly what credentials are in the wallet, and how they're build
- You don't know which constraint failed why

Would be nice to have a report returned by `PresentationDefinition.Match()` that contains which credentials have been considered, and which have been rejected why.

woutslakhorst:
closes #3007 
closes #3025 

I changed the PEX implementation so it always returns a `errors.Join(pe.ErrNoCredentials, reason)` when there's no match. For RFC021 this is returned as is. For the discovery service this is placed in trace logging to get details when needed.

### DIFF
--- a/auth/api/iam/api.go
+++ b/auth/api/iam/api.go
@@ -31,7 +31,6 @@ import (
 	"github.com/lestrrat-go/jwx/v2/jwk"
 	"github.com/nuts-foundation/nuts-node/http/cache"
 	"github.com/nuts-foundation/nuts-node/http/user"
-	"github.com/nuts-foundation/nuts-node/vcr/holder"
 	"github.com/nuts-foundation/nuts-node/vdr/didsubject"
 	"html/template"
 	"net/http"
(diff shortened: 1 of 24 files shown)


gerardsn (approved):
Nice cleanup of the ErrNoCredentials check

reinkrul (approved):
nice

reinkrul:
This might be enough to close https://github.com/nuts-foundation/nuts-node/issues/3025 as well

PR #3384, opened 2024-09-17
add DIDMethods config param to discovery service

### LINKED ISSUE #3383: Add supported DID methods to discovery service
This will enable use cases to dictate the DID methods that may be used.

This holds for both the issuer of the VP and the credentialSubject.Id of the VC

woutslakhorst:
closes #3383

known under JSON as `did_methods`. Direct activation and deactivation fails with `ErrDIDMethodsNotSupported`. If this error occurs in a refresh, the registration is removed.


gerardsn (approved):
just to check, DID methods that are no longer accepted after a service definition update will not be removed but rely on the max registration window to disappear? (which means we rely on use-cases set a reasonable max registration window)

Issue #3366, opened 2024-09-13
VDR: v2 service APIs issues


gerardsn:
note: `service.id` == `serviceId`

### GET / search `/internal/vdr/v2/subject/{id}/service`
search based on `service.type`
`type` parameter is required in business logic, but optional in API

### POST / create `/internal/vdr/v2/subject/{id}/service`
It contains the following example for the body:
```
{
  "id": "string",
  "type": "string",
  "serviceEndpoint": "string"
}
```
`id` value is overwritten by the API.
The documentations says
> The ID is always generated. Duplicate services are ignored.

So `id` field should be removed from the example.
The definition of duplicates is unclear, is this based on `type` or in combination with `serviceEndpoint`?

~The generated `id` is a uuid that is the same for all DID docs in the subject. However, our did:nuts spec requires the `id` to be a hash of service.~ it is the hash of the document

### PUT / update `/internal/vdr/v2/subject/{id}/service/{serviceId}`
Same issues as POST.

Lookup is based on `service.id`, but we used to manage services using `service.type` and still use this in GET/POST. Doing everything by `service.id` would be more in line with the DID spec, but this requires us to change the way we work with services.

PUT first does a search based on the `serviceId` and then generates a new `serviceId` for the updated service. So the `serviceId` PUT is being called on is removed and replaced with a new one. Is this in line with REST?

### DELETE `/internal/vdr/v2/subject/{id}/service/{serviceId}`
Also managed by `serviceId`, which does not match the way we work with services. I also don't know how users are going to find the `serviceId` if it is not stored directly after creating/updating a service

woutslakhorst:
> the way we work with services?

what does that mean?

woutslakhorst:
I believe the awkwardness comes from the fact that we only look at the fragment of the given serviceID.

gerardsn:
> > the way we work with services
> 
> what does that mean?

In did:nuts we reference services by `type` which MUST be unique in a DID document. We currently still create service based on type. If the nuts-node manages DID Documents we can also guarantee there is only 1 service per type. 

It is weird to create/search a service by type and then update/delete it by ID. 

Usage would look like
- A user creates a service on a subject with a certain type
- It then receives the created services, one per did in the subject, which is the input that now contains a service.id. The service.id is equal to full-did#serviceID, making users handle DIDs again.
- The user now has to store a mapping from `type` to `serviceID` to be able to update/delete it in the future
[...]

PR #3385, opened 2024-09-17
Change UpdateService API to POST

gerardsn:
PUT should idempotent, but calling this on an existing service is effectively a DELETE + POST to create a new service with a new ID. Hence first call would be 200, second call using same ID would be a 404 since the service no longer exists.

Also includes some documentation, cleanup, and a bugfix.


woutslakhorst:
This closes #3366?

woutslakhorst (approved):

PR #3386, opened 2024-09-18
removed incorrect stuff from docs for v6 release. added discovery examples

woutslakhorst:
This is some minor cleanup. It doesn't yet cover what is potentially missing.

PR #3389, opened 2024-09-19
Mark user flow and openid4vci flow experimental

gerardsn:
builds on #3386

removed experimental label of all APIs except the user and openid4vci flow

moved the openid4vci flow up in the document so it is grouped with the other `/internal/auth/v2/{subjectID}/...` paths making it easier to scan the document.

**should we also change:**
`/internal/auth/v2/{kid}/dpop` -> `/internal/auth/v2/dpop/{kid}` 
`/internal/auth/v2/dpop_validate` -> `/internal/auth/v2/dpop/validate`


woutslakhorst (approved):

woutslakhorst:
> **should we also change:**
> `/internal/auth/v2/{kid}/dpop` -> `/internal/auth/v2/dpop/{kid}`
> `/internal/auth/v2/dpop_validate` -> `/internal/auth/v2/dpop/validate`

is a bit better on the eyes. I don't think anyone is using it yet so we can still change it for the RCs. and give them a `dpop` label?

PR #3392, opened 2024-09-19
Migrate owned did nuts to SQL

### LINKED ISSUE #3329: migrate storage of did:nuts to sql, including history
But how to differ between owned and other DID docs?

gerardsn:
closes #3329

In did:nuts a deactivated DID can never be reused. It is still possible to publish updates on the DID document after this on the network, but the node will forever consider the document deactivated. The deactivation status is stored in the document metadata that no longer exists in SQL, which means that after migrating the DID document history to SQL it will appear active again.

Possible solutions
- add metadata to SQL
- don't migrate any document versions after the deactivation

TODO:
- [x] tests
- [x] DB checks in e2e-test + describe test plan
- [x] handle deactivated documents
- [x] documentation


woutslakhorst:
> In did:nuts a deactivated DID can never be reused. It is still possible to publish updates on the DID document after this on the network, but the node will forever consider the document deactivated. The deactivation status is stored in the document metadata that no longer exists in SQL, which means that after migrating the DID document history to SQL it will appear active again.
> 
> Possible solutions
> 
> * add metadata to SQL
> * don't migrate any document versions after the deactivation

This will only affect the owner. Others will always regard the DID as deactivated. So I would say don't migrate past that point. Add the behaviour to the release notes/documentation (eg: no conflicts may exist before updating to v6)

storage/test.go
```diff
@@ -70,6 +71,7 @@ func NewTestStorageEngineInDir(t testing.TB, dir string) Engine {
 	t.Cleanup(func() {
 		_ = result.Shutdown()
 	})
+	fmt.Printf("Created test storage engine in %s\n", dir)
```
woutslakhorst, line 74: You also get this information if you enable debug/trace logging on gorm.

e2e-tests/migration/run-test.sh
```diff
@@ -0,0 +1,161 @@
...
+echo "------------------------------------"
+echo "Stopping Docker containers..."
+echo "------------------------------------"
+docker compose -f docker-compose-post-migration.yml stop
+
+echo "------------------------------------"
+echo "Verifying migration results..."
+echo "------------------------------------"
+# all 'waitForTXCount' calls have confirmed the didstore is (likely to be) in the correct state. Now check SQL store.
+# TODO
```
woutslakhorst, line 155: or done?

woutslakhorst (changes requested):

storage/test.go
gerardsn: I knew I had seen it before, just didn't know how to activate it.
Since this is a test function, I would always like to know the path

woutslakhorst (approved):

Issue #3393, opened 2024-09-19
Storage: MS SQL Server does not properly support SELECT FOR UPDATE


reinkrul:
Yields: `mssql: Line 1: FOR UPDATE clause allowed only for DECLARE CURSOR.`

Also, SELECT FOR UPDATE will be removed from MS SQL Server: https://learn.microsoft.com/en-us/answers/questions/768729/sql-select-for-update

https://learn.microsoft.com/en-us/sql/relational-databases/sql-server-transaction-locking-and-row-versioning-guide?view=azuresqldb-current

Got solution from irmago; use `WITH(UDPLOCK, ROWLOCK)`

PR #3391, opened 2024-09-19
VCR: Issuing VC on v6.0.0-rc.1 on Postgres fails

reinkrul:
Issuing a VC with v6.0.0-rc.1 on Postgres fails terribly with the following error:

```
[...]

PR #3394, opened 2024-09-19
SQL: Fix locking with MS SQL Server

reinkrul:
Fixes https://github.com/nuts-foundation/nuts-node/issues/3393

### DIFF
--- a/discovery/store.go
+++ b/discovery/store.go
@@ -233,12 +233,8 @@ func (s *sqlStore) search(serviceID string, query map[string]string) ([]vc.Verif
 
 // incrementTimestamp increments the last_timestamp of the given service.
 func (s *sqlStore) incrementTimestamp(tx *gorm.DB, serviceID string) (*int, error) {
-	var service serviceRecord
-	// Lock (SELECT FOR UPDATE) discovery_service row to prevent concurrent updates to the same list, which could mess up the last Timestamp.
-	if err := tx.Clauses(clause.Locking{Strength: "UPDATE"}).
-		Where(serviceRecord{ID: serviceID}).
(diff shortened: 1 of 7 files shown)


discovery/store.go
```diff
@@ -252,19 +248,35 @@ func (s *sqlStore) incrementTimestamp(tx *gorm.DB, serviceID string) (*int, erro
...
+	service, err := s.findServiceById(tx, serviceID)
+	if err != nil {
 		return err
 	}
 	service.ID = serviceID
 	service.LastLamportTimestamp = timestamp
 	return tx.Save(service).Error
 }
 
+func (s *sqlStore) findServiceById(tx *gorm.DB, serviceID string) (serviceRecord, error) {
```
woutslakhorst, line 260: let the name reflect that it uses locking.

woutslakhorst (changes requested):
There's no coverage on the status list MSSql logic. Could you extend the oauth/statuslist e2e tests with mssql?

reinkrul:
> There's no coverage on the status list MSSql logic. Could you extend the oauth/statuslist e2e tests with mssql?

I did that in the RFC021 e2e-test as part of https://github.com/nuts-foundation/nuts-node/pull/3391 by enabling revocation status for the issued credentials. Although I don't necessarily think that's the right test (because it was originally intended to be RFC021-specific), having a varying set of databases tested in different tests here and there, is not very maintainable. As RFC021 is actually the "only" use case/flow we truly support and use for v6, maybe we can keep all variants we need (MySql/Postgres/MSSQL Server) there?

woutslakhorst (approved):

PR #3399, opened 2024-09-20
added security considerations page with single section on max upload sizes

woutslakhorst:

PR #3400, opened 2024-09-20
Add VDR API version usage with regards to DID methods to docs

### LINKED ISSUE #3285: Add documentation on enabled DID methods and VDR API usage for v6
There could be harmful combinations when enabling did:web and using the VDR V1 API.

woutslakhorst:
closes #3285 

builds on: #3399


docs/pages/deployment/security-considerations.rst
```diff
@@ -0,0 +1,37 @@
...
+
+Using did:web
+*************
+
+The ``did:web`` method allows for easier integration with existing web infrastructure. However, it's also less secure and vulnerable to domain takeover.
+When using ``did:web``, you should consider the following:
+
+- Protect your domain from takeover. Make sure it's locked for a year after cancelling the domain.
+- Monitor calls to ``**/did.json`` on the domain and make sure they are handled by the Nuts Node.
+- Use Hashicorp Vault or Microsoft Azure Key Vault to store the private keys.
```
gerardsn, line 37: type of key storage is not did method specific

docs/pages/deployment/security-considerations.rst
```diff
@@ -0,0 +1,37 @@
+.. _security-considerations::
+
+Security Considerations
+#######################
```
gerardsn, line 4: [did:web](https://w3c-ccg.github.io/did-method-web/#dns-considerations) recommends? [DNS over HTTPS](https://datatracker.ietf.org/doc/html/rfc8484) and DNSSEC [RFC4033](https://tools.ietf.org/html/rfc4033), [RFC4034](https://tools.ietf.org/html/rfc4034), and [RFC4035](https://tools.ietf.org/html/rfc4035).

[openid4vp](https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#name-normative-references) requires [BCP195](https://www.rfc-editor.org/info/bcp195)

I think we should therefore include these in our recommendations/requirements as well

gerardsn (changes requested):
maybe rebase first, because I have now reviewed the outdated file that is part of a merged PR. (Comments still stand though)

docs/pages/deployment/security-considerations.rst
woutslakhorst: true but more important for did:web. Changed point slightly

PR #3403, opened 2024-09-20
fix spec example

### LINKED ISSUE #3401: API params wrong disovery service registration
For `POST ​/internal​/discovery​/v1​/{serviceID}​/{subjectID}`

Docs say:
```
{
  "registrationParameters": [
    {
      "some-endpoint-type": "https://example.com/some-endpoint"
    }
  ]
}
```

Nuts node wants:

```
{
  "registrationParameters": {
      "some-endpoint-type": "https://example.com/some-endpoint"
    }
}
```

woutslakhorst:
fixes #3401


stevenvegt:
Since it occurs twice in the spec, shouldn't this be a type?

PR #3418, opened 2024-09-25
Alter discovery activation

### LINKED ISSUE #3408: Discovery service retry for failed registration is unclear
The discovery service client will try to register VPs even if the holder doesn't have the right credentials. This is strange behaviour. Proposal is to return a 409 if the PD is not satisfied and only refresh when still valid. In all other cases don't retry.

If for whatever reason the PD is updated which causes current registrations to become invalid, this will be a manual check.`GetServiceActivation` can be used for this.

woutslakhorst:
closes #3408 

- activation through the API is now a fail/success API, no magic is done at a later point if it fails.
- the refresh adds `presentationRefreshError` records if it fails for a candidate.
- refresh success or activation through the API clears the `presentationRefreshError` records.
- The `getServiceActivation` API returns an additional `status` and `error` field.

### DIFF
--- a/discovery/api/v1/api.go
+++ b/discovery/api/v1/api.go
@@ -42,10 +42,9 @@ type Wrapper struct {
 
 func (w *Wrapper) ResolveStatusCode(err error) int {
 	switch {
-	case errors.Is(err, discovery.ErrServiceNotFound):
-		return http.StatusNotFound
-	case errors.Is(err, didsubject.ErrSubjectNotFound):
-		return http.StatusNotFound
+	case errors.Is(err, discovery.ErrServiceNotFound):return http.StatusNotFound
+	case errors.Is(err, didsubject.ErrSubjectNotFound):return http.StatusNotFound
+	case errors.Is(err, discovery.ErrPresentationRegistrationFailed):return http.StatusPreconditionFailed
 	default:
 		return http.StatusInternalServerError
 	}
@@ -108,12 +107,6 @@ func (w *Wrapper) ActivateServiceForSubject(ctx context.Context, request Activat
 	}
 
 	err := w.Client.ActivateServiceForSubject(ctx, request.ServiceID, request.SubjectID, parameters)
-	if errors.Is(err, discovery.ErrPresentationRegistrationFailed) {
-		// registration failed, but will be retried
-		return ActivateServiceForSubject202JSONResponse{
-			Reason: err.Error(),
-		}, nil
-	}
 	if err != nil {
 		// other error
 		return nil, err
@@ -123,12 +116,6 @@ func (w *Wrapper) ActivateServiceForSubject(ctx context.Context, request Activat
 
 func (w *Wrapper) DeactivateServiceForSubject(ctx context.Context, request DeactivateServiceForSubjectRequestObject) (DeactivateServiceForSubjectResponseObject, error) {
 	err := w.Client.DeactivateServiceForSubject(ctx, request.ServiceID, request.SubjectID)
-	if errors.Is(err, discovery.ErrPresentationRegistrationFailed) {
-		// deactivation succeeded, but not all Verifiable Presentation couldn't be removed from remote Discovery Server.
-		return DeactivateServiceForSubject202JSONResponse{
-			Reason: err.Error(),
-		}, nil
-	}
 	if err != nil {
 		return nil, err
 	}
--- a/discovery/api/v1/api.go
+++ b/discovery/api/v1/api.go
@@ -24,6 +24,7 @@ import (
 	"github.com/labstack/echo/v4"
 	"github.com/nuts-foundation/nuts-node/audit"
 	"github.com/nuts-foundation/nuts-node/core"
+	"github.com/nuts-foundation/nuts-node/core/to"
 	"github.com/nuts-foundation/nuts-node/discovery"
 	"github.com/nuts-foundation/nuts-node/vcr/credential"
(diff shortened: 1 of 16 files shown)


docs/_static/discovery/v1.yaml
```diff
@@ -148,6 +159,7 @@ paths:
         
         error returns:
         * 400 - incorrect input: invalid/unknown service or subject.
+        * 412 - precondition failed: subject doesn't have the required certificates.
```
reinkrul, line 162:
```suggestion
        * 412 - precondition failed: subject doesn't have the required credentials.
```

e2e-tests/discovery/run-test.sh
```diff
@@ -15,6 +15,7 @@ docker compose down
 echo "------------------------------------"
 echo "Starting Docker containers..."
 echo "------------------------------------"
+docker compose up --wait nodeA-backend nodeA
 docker compose up --wait nodeB-backend nodeB
```
reinkrul, line 19: could collapse this 2 into 1, which speeds it up (since they're started in parallel)

discovery/api/v1/api.go
```diff
@@ -145,8 +136,17 @@ func (w *Wrapper) GetServiceActivation(ctx context.Context, request GetServiceAc
...
 		return nil, err
 	}
-	return GetServiceActivation200JSONResponse{
+	response := GetServiceActivation200JSONResponse{
 		Activated: activated,
+		Status:    ServiceStatusActive,
 		Vp:        &presentations,
-	}, nil
+	}
+	if refreshError := w.Client.GetServiceRefreshError(ctx, request.ServiceID, request.SubjectID); refreshError != nil {
```
reinkrul, line 144: You should only do this when not properly activated? Or can it be both activated (valid VPs), but have an error? (1 DID failed, 1 succeeded)

discovery/store.go
```diff
@@ -95,6 +95,23 @@ func (l presentationRefreshRecord) TableName() string {
 	return "discovery_presentation_refresh"
 }
 
+// presentationRefreshError is a record of a failed refresh attempt.
+type presentationRefreshError struct {
```
reinkrul, line 99: having this as child resource of the Presentation speeds up fetching , if preloading (now getting activation status is always 1+1 SQL query)

discovery/client.go
```diff
@@ -132,6 +123,10 @@ func (r *defaultClientRegistrationManager) activate(ctx context.Context, service
 	if err := r.store.updatePresentationRefreshTime(serviceID, subjectID, parameters, &refreshVPAfter); err != nil {
 		return fmt.Errorf("unable to update Verifiable Presentation refresh time: %w", err)
 	}
+	// clear any previous presentationRefreshErrors here so it's triggered by both the refresh and API call
```
reinkrul, line 126: The API calls `activate()`, the timer calls `refresh()` (which in turn calls `activate()`). `setPresentationRefreshError` is only called by `refresh()` if an error occurs (and never by `activate()`), this means;
- the API call will never cause the error to be set, even one of the DIDs fails and one of them succeeds
- the same applies for `refresh()`

reinkrul (changes requested):

discovery/client.go
woutslakhorst: That's the idea. `activate()` via the API does not create a refresh record when it fails, so also not an error record. The error is passed back through the API.

So you only want to set an error (and thus being able to read it) when the refresh of a active registration fails.

discovery/api/v1/api.go
woutslakhorst: You only have an error when all DIDs failed. The idea is that you always call this API from your administration to check for green √. Combinations that could occur:

- not activated (yellow)
- activated but error (red)
- activated, no error (green)

discovery/store.go
woutslakhorst: and cascading delete would cleanup any lingering errors

reinkrul: Although foreign keys and cascading deletes are not required for the relationship on GORM level to work

woutslakhorst:
@reinkrul I exposed the error through `GetServiceActivation()`. Less methods in the interface and use of a `Preload`

storage/sql_migrations/008_discoveryservice_client_error.sql
```diff
@@ -13,7 +13,7 @@ create table discovery_presentation_error
     -- last_occurrence is the timestamp (seconds since Unix epoch) when the registration last failed.
     last_occurrence integer      not null,
     primary key (service_id, subject_id),
-    constraint fk_discovery_presentation_error_service foreign key (service_id) references discovery_service (id) on delete cascade
+    constraint fk_discovery_presentation_error_service foreign key (service_id, subject_id) references discovery_presentation_refresh (service_id, subject_id) on delete cascade
```
reinkrul, line 16: can we do a migration 9 altering the foreign key? Recreating schema starts to hurt...

woutslakhorst: migration 8 is already new. So this shouldn't give any conflict with existing DBs

reinkrul: Aaah yes, nvm

woutslakhorst: > Aaah yes, nvm

click *approve*

reinkrul (approved):
_approve_

PR #3429, opened 2024-10-01
fix status codes on did service APIs

### LINKED ISSUE #3428: Adding invalid service to DID subject results in 500
Adding invalid JSON results in:

```
{
    "detail": "could not add service to DID Documents: DID Document validation failed: invalid service",
    "status": 500,
    "title": "CreateService failed"
}
```

the `invalid service` comes from the go-did library. It's a `ErrInvalidService`.
~~Also the API spec doesn't generate a correct example body in my rest client.~~

and for adding the same type twice:
```
{
    "detail": "could not update service for DID Documents: invalid service: service type is duplicate",
    "status": 500,
    "title": "UpdateService failed"
}
```

Also encoding the path param for update is tricky. The `#` has to be replaced with `%23` but the `:`s can't be encoded.

woutslakhorst:
fixes #3428 

added some error to status codes mappings. Added some doc to the update/delete api spec on escaping the service ID.

### DIFF
--- a/docs/_static/vdr/v2.yaml
+++ b/docs/_static/vdr/v2.yaml
@@ -298,7 +298,8 @@ paths:
       summary: Delete a specific service from the subject
       description: |
         Removes the service from all DID Documents in the subject. Matching is done on the fragment of the id.
-        No cascading will happen for references to the service. 
+        No cascading will happen for references to the service.
+        Make sure to only URL encode the pound (#) as %23 in the serviceId. Do not encode the colons (:). 
 
         error returns:
         * 400 - Returned in case of malformed subject or service ID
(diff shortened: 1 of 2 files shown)


docs/_static/vdr/v2.yaml
```diff
@@ -298,7 +298,8 @@ paths:
       summary: Delete a specific service from the subject
       description: |
         Removes the service from all DID Documents in the subject. Matching is done on the fragment of the id.
-        No cascading will happen for references to the service. 
+        No cascading will happen for references to the service.
+        Make sure to only URL encode the pound (#) as %23 in the serviceId. Do not encode the colons (:).
```
reinkrul, line 302: isn't this default behavior? Standard API clients should do this?

woutslakhorst: If I do an encode URI component in Postman, it escapes both `:` and `#` so not all tools are equal

gerardsn (approved):

PR #3432, opened 2024-10-01
add migration notes to docs

gerardsn:
This probably doesn't fit under `Version Incompatibilities`. We need to move it elsewhere, or rename the page and make Version Incompatibilities a subsection...

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -23,6 +23,7 @@ Nuts documentation
 
     pages/deployment/domain.rst
     pages/deployment/configuration.rst
+    pages/deployment/migration.rst
     pages/deployment/recommended-deployment.rst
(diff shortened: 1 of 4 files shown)


woutslakhorst (commented):
The incompatibilities should list the fact that new [did:web, did:nuts] + private credentials over the network won't work.

The rest can go under a _migrating from v5_ page under the _deployment&config_ section. As 2nd or 3rd item. Then link release notes to that page.

gerardsn:
> The incompatibilities should list the fact that new [did:web, did:nuts] + private credentials over the network won't work.

this works as before since new did:nuts DIDs still get an (elliptic curve) encryption key

woutslakhorst (approved):

PR #3433, opened 2024-10-01
fixes for rfc021 client status codes

### LINKED ISSUE #3430: Incorrect status return codes for request-service-access-token
With wrong authorization_server:
```
{
  "authorization_server": "http://nuts.local:8080/oauth2/test2",
  "scope": "test",
  "tokenType": "Bearer"
}
```

If the metadata returns a `404` or on any http client error (on the metadata), this should yield a `400`. If the metadata returns 50x, then it's a 502.

And:
```
{
    "detail": "failed to retrieve presentation definition: failed to call endpoint: Get \"https://nuts.local:8080/oauth2/test2/presentation_definition?scope=test\": http: server gave HTTP response to HTTPS client",
    "status": 500,
    "title": "RequestServiceAccessToken failed"
}
```
A `502` is justified for all non-http errors from the http client.

A mismatch on DID methods also results in a 500. This should be a 412 Precondition failed.

woutslakhorst:
fixes #3430 

some 500 => correct status codes based on early findings. I don't think this is the complete set. Will have to do some more tests using a fully setup https environment.

### DIFF
--- a/auth/client/iam/client.go
+++ b/auth/client/iam/client.go
@@ -22,6 +22,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"github.com/lestrrat-go/jwx/v2/jws"
 	"github.com/lestrrat-go/jwx/v2/jwt"
@@ -39,6 +40,12 @@ import (
 	"github.com/nuts-foundation/nuts-node/core"
 	"github.com/nuts-foundation/nuts-node/vcr/pe"
 )
+
+// ErrInvalidClientCall is returned when the node makes a http call as client based on wrong information passed by the client.
+var ErrInvalidClientCall = errors.New("invalid client call")
+
+// ErrBadGateway is returned when the node makes a http call as client and the upstream returns an unexpected result.
+var ErrBadGateway = errors.New("upstream returned unexpected result")
 
 // HTTPClient holds the server address and other basic settings for the http client
 type HTTPClient struct {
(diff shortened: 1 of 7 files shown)


docs/_static/auth/v2.yaml
```diff
@@ -14,9 +14,10 @@ paths:
         It'll initiate a s2s (RFC021) flow.
 
         error returns:
-        * 400 - one of the parameters has the wrong format or an OAuth error occurred
-        * 412 - the organization wallet does not contain the correct credentials
-        * 503 - the authorizer could not be reached or returned an error
+        * 400 - one of the parameters has the wrong format, an OAuth error occurred or the http client calling the authorizer returned an error due to incorrect input
```
gerardsn, line 17:
```suggestion
        * 400 - one of the parameters has the wrong format, an OAuth error occurred, or the http client calling the authorizer returned an error due to incorrect input
```

auth/client/iam/client.go
```diff
@@ -63,7 +70,14 @@ func (hb HTTPClient) OAuthAuthorizationServerMetadata(ctx context.Context, oauth
 	}
 	var metadata oauth.AuthorizationServerMetadata
 	if err = hb.doGet(ctx, metadataURL.String(), &metadata); err != nil {
-		return nil, err
+		// if this is a core.HttpError and the status code >= 500 then we want the caller to receive a 502 Bad Gateway
+		// we do this by changing the status code of the error
+		// any other error should result in a 400 Bad Request
+		if httpErr, ok := err.(core.HttpError); ok && httpErr.StatusCode >= 500 {
+			httpErr.StatusCode = http.StatusBadGateway
```
gerardsn, line 77: untested; don't know how hard it is to test

gerardsn (approved):

auth/client/iam/client.go
woutslakhorst: I selected code and asked CoPilot to generate the test for me....

PR #3434, opened 2024-10-01
Tools: golangci lint

gerardsn:
remaining issues:
```
http/user/session.go:97:99: SA1029: should not use empty anonymous struct as key for value; define your own type to avoid collisions (staticcheck)
                echoCtx.SetRequest(echoCtx.Request().WithContext(context.WithValue(echoCtx.Request().Context(), userSessionContextKey{}, sessionData)))
                                                                                                                ^
```
I don't know what is causing this

```
discovery/store.go:387:20: func `(*sqlStore).getPresentationRefreshError` is unused (unused)
func (s *sqlStore) getPresentationRefreshError(serviceID string, subjectID string) (*presentationRefreshError, error) {
```
This is used in tests, but not in prodcution code. @ wout, was there a plan to use this in an API (possibly new) where you can request the status + error of a registration, or should I move this to a test file.


woutslakhorst:
> ```
> discovery/store.go:387:20: func `(*sqlStore).getPresentationRefreshError` is unused (unused)
> func (s *sqlStore) getPresentationRefreshError(serviceID string, subjectID string) (*presentationRefreshError, error) {
> ```
>   
> 
> This is used in tests, but not in prodcution code. @ wout, was there a plan to use this in an API (possibly new) where you can request the status + error of a registration, or should I move this to a test file.

Can be removed, was part of older proposal

woutslakhorst (approved):
changes OK, another PR may remove that refresh error func. (or add to this one)

PR #3439, opened 2024-10-02
Move didmethods to config root

gerardsn:


woutslakhorst (approved):

reinkrul:
Why?

gerardsn:
`didmethods` is used in multiple modules. Should have been on root from the start to prevent other modules having a dependency on VDR just for this.

reinkrul (approved):

PR #3441, opened 2024-10-02
update koanf major

woutslakhorst:
also fixes loading of slice values


e2e-tests/browser/rfc019_selfsigned/config/node/nuts.yaml
```diff
@@ -6,4 +6,4 @@ http:
     address: :8081
 auth:
   contractvalidators:
-    - selfsigned
+    - employeeid
```
reinkrul, line 9: why did this change?

gerardsn: test just happend to work because of koanf bug, default is `[irma,dummy,employeeid]` this was overwitten to `[selfsigned,dummy,employeeid]`. Fixed koanf bug means its nog `[selfsigned]` which does not exist

PR #3446, opened 2024-10-03
Introduce x509 DID method with PKI validation.

rolandgroen:
Added support for x509 DID method including a new resolver and PKI validation. Adjusted metadata handling in key resolution across the codebase to include x509 specific fields and validation logic.

- Added support for did:x509 as a new DID resolver.
- Enabled validation of did:x509 VCs and VPs.
- Added corresponding tests.


vcr/verifier/signature_verifier.go
```diff
@@ -116,7 +120,19 @@ func (sv *signatureVerifier) jwtSignature(jwtDocumentToVerify string, issuer str
...
+		}
+		if strings.HasPrefix(keyID, "did:x509:") {
+			message, _ := jws.ParseString(jwtDocumentToVerify)
+			if message != nil && len(message.Signatures()) > 0 {
+				headers := message.Signatures()[0].ProtectedHeaders()
+				metadata.X509CertThumbprint = headers.X509CertThumbprint()
+				metadata.X509CertChain = headers.X509CertChain()
+				metadata.X509CertThumbprintS256 = headers.X509CertThumbprintS256()
+			}
+		}
```
reinkrul, line 134: why not always do this, regardless of the DID method? Probably less complex.

Although: you could decide to only do this when it's did:x509, then you can (more) safely return an error when `ParseString()` fails. Since the x5c chain is required for the process, I think it should fail if it can't parse the JWS? (or maybe even when x5c isn't there?)

vcr/verifier/verifier.go
```diff
@@ -161,7 +162,21 @@ func (v verifier) Verify(credentialToVerify vc.VerifiableCredential, allowUntrus
...
+			if rawJwt != "" {
+				message, _ := jws.ParseString(rawJwt)
+				if message != nil && len(message.Signatures()) > 0 {
+					headers := message.Signatures()[0].ProtectedHeaders()
+					metadata.X509CertThumbprint = headers.X509CertThumbprint()
+					metadata.X509CertChain = headers.X509CertChain()
+					metadata.X509CertThumbprintS256 = headers.X509CertThumbprintS256()
+				}
+
+			}
```
reinkrul, line 177: duplicate, could use a util func, maybe in the x509 package?

vdr/didx509/resolver.go
```diff
@@ -0,0 +1,232 @@
...
+	if err != nil {
+		return nil, nil, err
+	}
+	_, err = findCertificateByHash(chain, ref.RootCertRef, ref.Method)
+	if err != nil {
+		return nil, nil, err
+	}
+	validationCert, err := findCertificateByHash(chain, metadata.X509CertThumbprint, "sha1")
+	if err != nil {
+		validationCert, err = findCertificateByHash(chain, metadata.X509CertThumbprintS256, "sha256")
```
reinkrul, line 76: are these hashing algorithms fixed?

vdr/didx509/resolver.go
```diff
@@ -0,0 +1,232 @@
...
+}
+
+// validatePolicy validates a certificate against a given X509DidReference and its policy.
+func validatePolicy(ref *X509DidReference, cert *x509.Certificate) error {
+
+	switch ref.PolicyName {
+	case "subject":
+		keyValue := strings.Split(ref.PolicyValue, ":")
+		if len(keyValue)%2 != 0 {
+			return ErrDidSanMalformed
```
reinkrul, line 105: I think this should've been subject malformed?

vdr/didx509/resolver.go
```diff
@@ -0,0 +1,232 @@
...
+					return ErrWrongLocality
+				}
+			case "O":
+				if !slices.Contains(subject.Organization, value) {
+					return ErrWrongOrganization
+				}
+			case "OU":
+				if !slices.Contains(subject.OrganizationalUnit, value) {
+					return ErrWrongOrganizationalUnit
+				}
```
reinkrul, line 131: this allows unsupported/unknown RDNs in the subject DN. There should be a default case that rejects the did:x509 (since it's not something checked against the certificate).

reinkrul: Also, it's missing `ST` and `STREET` (https://trustoverip.github.io/tswg-did-x509-method-specification/#json-data-model-for-x509-certificate-chains)

vdr/didx509/resolver.go
```diff
@@ -0,0 +1,232 @@
...
+
+		}
+	}
+
+	return nil
+}
+
+// createDidDocument generates a new DID Document based on the provided DID identifier and validation certificate.
+func createDidDocument(id did.DID, validationCert *x509.Certificate) (*did.Document, error) {
+	verificationMethod, err := did.NewVerificationMethod(did.DIDURL{DID: id}, ssi.JsonWebKey2020, id, validationCert.PublicKey)
```
reinkrul, line 183: The ID of the verification method needs to be the DID (you ahve this), appended with the `#0` fragment. I think you can set `Fragment` on the first argument to `0` to achieve this.

vdr/didx509/resolver.go
```diff
@@ -0,0 +1,232 @@
...
+			ssi.MustParseURI("https://www.w3.org/ns/did/v1"),
+		},
+		ID:                 id,
+		Controller:         []did.DID{id},
+		VerificationMethod: did.VerificationMethods{verificationMethod},
+	}
+	document.AddAssertionMethod(verificationMethod)
+	document.AddCapabilityDelegation(verificationMethod)
+	document.AddCapabilityInvocation(verificationMethod)
+	document.AddKeyAgreement(verificationMethod)
```
reinkrul, line 198: missing `AddAuthenticationMethod`

vdr/didx509/resolver.go
```diff
@@ -0,0 +1,232 @@
...
+// parseX509Did parses a DID (Decentralized Identifier) in the x509 format and returns a corresponding X509DidReference.
+func parseX509Did(id did.DID) (*X509DidReference, error) {
+	ref := X509DidReference{}
+	fullDidString := id.ID
+	policyString := ""
+	didString := ""
+	fullDidParts := strings.Split(fullDidString, "::")
+	if len(fullDidParts) == 1 {
+		didString = fullDidParts[0]
+	} else if len(fullDidParts) == 2 {
```
reinkrul, line 211: missing test coverage

vdr/didx509/resolver_test.go
```diff
@@ -0,0 +1,121 @@
...
+		assert.Equal(t, err.Error(), errors.New("unsupported hash algorithm: kaas").Error())
+	})
+	t.Run("wrong hash value", func(t *testing.T) {
+		rootDID := did.MustParseDID(fmt.Sprintf("did:x509:0:%s:%s::san:otherName:%s", "sha256", "kaas", "ANOTHER_BIG_STRING"))
+		_, _, err := resolver.Resolve(rootDID, &metadata)
+		require.Error(t, err)
+		assert.Equal(t, err.Error(), errors.New("cannot find a certificate with alg: sha256 hash: kaas").Error())
+	})
+	t.Run("wrong DID type", func(t *testing.T) {
+		rootDID := did.MustParseDID("did:kaas:gouda.nl:jonge")
```
reinkrul, line 93: however funny, better use non-existing/example domain (`example.com`)
```suggestion
		rootDID := did.MustParseDID("did:kaas:example.com:jonge")
```

vdr/didx509/resolver_test.go
```diff
@@ -0,0 +1,121 @@
...
+	t.Run("happy flow 2", func(t *testing.T) {
+		metadata.X509CertThumbprintS256 = ""
+		validator.EXPECT().Validate(gomock.Any()).Return(nil)
+		resolve, documentMetadata, err := resolver.Resolve(rootDID, &metadata)
+		require.NoError(t, err)
+		assert.NotNil(t, resolve)
+		assert.NotNil(t, documentMetadata)
+		metadata.X509CertThumbprintS256 = sha256Sum(signingCert.Raw)
+	})
+	t.Run("happy flow 2", func(t *testing.T) {
```
reinkrul, line 58: duplicate test name

vdr/didx509/resolver_test.go
```diff
@@ -0,0 +1,121 @@
...
+
+	})
+	t.Run("happy flow", func(t *testing.T) {
+		validator.EXPECT().Validate(gomock.Any()).Return(nil)
+		resolve, documentMetadata, err := resolver.Resolve(rootDID, &metadata)
+		require.NoError(t, err)
+		assert.NotNil(t, resolve)
+		assert.NotNil(t, documentMetadata)
+	})
+	t.Run("happy flow 2", func(t *testing.T) {
```
reinkrul, line 49: "2" isn't really informative

vdr/didx509/x509_utils.go
```diff
@@ -0,0 +1,173 @@
...
+				return "", err
+			}
+			return value, err
+		}
+	}
+	return "", nil
+}
+
+// forEachSAN processes each SAN extension in the certificate
+func forEachSAN(extension []byte, callback func(tag int, data []byte) error) error {
```
reinkrul, line 69: SAN -> San (Golang convention)

vdr/didx509/x509_utils.go
```diff
@@ -0,0 +1,173 @@
...
+		if bytes.Equal(targetHash, certHash) {
+			return cert, nil
+		}
+	}
+
+	return nil, fmt.Errorf("cannot find a certificate with alg: %s hash: %s", alg, targetHashString)
+}
+
+// hash computes and returns the hash of the given data using the specified algorithm.
+// Supported algorithms are: "sha1", "sha256Sum", "sha384", "sha512".
```
reinkrul, line 154:
```suggestion
// Supported algorithms are: "sha1", "sha256", "sha384", "sha512".
```

vcr/verifier/signature_verifier.go
rolandgroen: It does not make a lot of sense to do this whenever the method is not x509, but somehow understandable if you would fill the headers anyway. There MIGHT be a situation where the fields are used outside the x509 spec..

About failing: 
I think it will fail later on when X509CertThumbprint et. al. is not filled and the method is x509.

rolandgroen: Maybe the cleanest way would be to just stuff in the headers into the metadata. That would be way more generic.

vdr/didx509/resolver.go
rolandgroen: Yes.

vdr/didx509/resolver.go
rolandgroen: Correct. Fixed.

vdr/didx509/resolver.go
rolandgroen: First thing turned up at testing.

rolandgroen: The STREET and ST are added.

vdr/didx509/resolver_test.go
rolandgroen: Updated most domains and emails to example.com

vdr/didx509/resolver_test.go
rolandgroen: Resolved

vdr/didx509/x509_utils.go
rolandgroen: Changed.

vdr/didx509/resolver_test.go
rolandgroen: Updated all test names to more semantic names

vdr/didx509/x509_test_utils.go
gerardsn, line 1: filename should end in `_test.go` to prevent it from being used outside of tests

vdr/didx509/resolver.go
rolandgroen: Check done

vdr/didx509/resolver.go
rolandgroen: Set the fragment, added a check to the tests.

vdr/didx509/resolver.go
rolandgroen: Added tests, and PolicyNone. The latter will get removes as soon as we support multiple policies.

vdr/didx509/x509_test_utils.go
rolandgroen: Renamed the file.

vdr/didx509/jwt_utils.go
```diff
@@ -0,0 +1,27 @@
...
+	"github.com/lestrrat-go/jwx/v2/jws"
+	"strings"
+)
+
+// ExtractProtectedHeaders extracts the protected headers from a JWT string.
+// The function takes a JWT string as input and returns a map of the protected headers.
+// Note that:
+//   - This method ignores strings that don't look like JWTs and returns an empty map.
+//   - This method ignores any parsing errors and returns an empty map instead of an error.
+func ExtractProtectedHeaders(jwt string) (map[string]interface{}, error) {
```
gerardsn, line 14: inverting the conditions / removing all nested statements makes this a lot easier to read / maintain

vdr/didx509/resolver.go
```diff
@@ -0,0 +1,316 @@
...
+	if err != nil {
+		return nil, nil, err
+	}
+
+	err = validatePolicy(ref, validationCert)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	err = r.pkiValidator.Validate(chain)
```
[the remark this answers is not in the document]
gerardsn: Discussed with @stevenvegt: No. this is not acceptable. The `pkiValidator` must use hardfail: return an error when it has no CRL cache for the chain, or when the cache is stale, i.e. older than `pki.maxupdatefailhours`. For other/existing use-cases the failure mode should still be configurable.

vdr/didx509/resolver_test.go
```diff
@@ -0,0 +1,514 @@
...
+	metadata.JwtProtectedHeaders[X509CertThumbprintS256Header] = sha256Sum(signingCert.Raw)
+
+	rootDID := did.MustParseDID(fmt.Sprintf("did:x509:0:%s:%s::san:otherName:%s", "sha256", sha256Sum(rootCertificate.Raw), otherNameValue))
+
+	t.Run("test nulls", func(t *testing.T) {
+		chain, _ := metadata.GetProtectedHeaderChain(X509CertChainHeader)
+		delete(metadata.JwtProtectedHeaders, X509CertChainHeader)
+		_, _, err := resolver.Resolve(rootDID, &metadata)
+		require.Error(t, err)
+		assert.Equal(t, ErrX509ChainMissing.Error(), err.Error())
```
gerardsn, line 41: FYI: Normally you would use `assert.ErrorIs(...)` to check for errors. `ErrorsIs` also works if the error is wrapped by another error

vdr/didx509/resolver_test.go
```diff
@@ -0,0 +1,514 @@
...
+		require.NoError(t, err)
+		assert.NotNil(t, resolve)
+		require.NoError(t, err)
+		assert.NotNil(t, documentMetadata)
+		// Check that the DID url is did#0
+		didUrl, err := did.ParseDIDURL(rootDID.String() + "#0")
+		assert.NotNil(t, resolve.VerificationMethod.FindByID(*didUrl))
+	})
+	t.Run("happy flow, policy depth of 1", func(t *testing.T) {
+		validator.EXPECT().Validate(gomock.Any()).Return(nil)
```
gerardsn, line 68: FYI: mocks always return the default value if none is specified. So you don't need to add `.Return(nil)`, only if you actually want to return an error.

vdr/didx509/x509_utils.go
gerardsn, line 1: seems to be missing a bunch of non-happy flow test cases

vdr/resolver/did.go
gerardsn, line 1: untested

gerardsn (changes requested):
Very limited review since there is no reference to the spec + version being implemented.

Some general comments:
- files in `vdr/didx509` need more comments. I currently have no information on choices that were made to maintain this in the future.
- seems to be missing a bunch of non-happy flow tests, but the code coverage tool is also marking things untested that are tested... The truth is probably somewhere in between.

vcr/verifier/signature_verifier.go
```diff
@@ -116,7 +120,15 @@ func (sv *signatureVerifier) jwtSignature(jwtDocumentToVerify string, issuer str
...
 	_, err := crypto.ParseJWT(jwtDocumentToVerify, func(kid string) (crypt.PublicKey, error) {
 		keyID = kid
-		return sv.resolveSigningKey(kid, issuer, at)
+		metadata := &resolver.ResolveMetadata{
+			ResolveTime: at,
+		}
+		headers, err := didx509.ExtractProtectedHeaders(jwtDocumentToVerify)
+		if err != nil {
+			return nil, err
+		}
```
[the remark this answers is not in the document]
rolandgroen: Not completely sure I mocked the package function the right way, but a test is added.

vcr/verifier/verifier.go
```diff
@@ -161,7 +162,17 @@ func (v verifier) Verify(credentialToVerify vc.VerifiableCredential, allowUntrus
...
 		issuerDID, _ := did.ParseDID(credentialToVerify.Issuer.String())
-		_, _, err := v.didResolver.Resolve(*issuerDID, &resolver.ResolveMetadata{ResolveTime: validAt, AllowDeactivated: false})
+		metadata := resolver.ResolveMetadata{ResolveTime: validAt, AllowDeactivated: false}
+		rawJwt := credentialToVerify.Raw()
+		if rawJwt != "" {
+			headers, err := didx509.ExtractProtectedHeaders(rawJwt)
+			if err != nil {
+				return err
+			}
+			metadata.JwtProtectedHeaders = headers
```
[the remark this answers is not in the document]
rolandgroen: Added tests

vdr/didx509/jwt_utils.go
```diff
@@ -0,0 +1,27 @@
...
+	headers := make(map[string]interface{})
+	if jwt != "" && strings.Count(jwt, ".") > 1 && strings.HasPrefix(jwt, "ey") {
+		message, _ := jws.ParseString(jwt)
+		if message != nil && len(message.Signatures()) > 0 {
+			var err error
+			headers, err = message.Signatures()[0].ProtectedHeaders().AsMap(context.Background())
```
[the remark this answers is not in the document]
rolandgroen: Introduced a check for message.Signatures() == 1 and included tests for the case.

vdr/didx509/resolver.go
```diff
@@ -0,0 +1,316 @@
...
+		}
+		if validationCert == nil {
+			validationCert = otherValidationCert
+		} else {
+			if !otherValidationCert.Equal(validationCert) {
+				return nil, errors.New("x5t#S256 header does not match the certificate from the x5t headers")
+			}
+		}
+	}
+	return validationCert, nil
```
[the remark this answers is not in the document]
rolandgroen: Added a check, the nil is possible.

vdr/didx509/jwt_utils.go
rolandgroen: The conditions are removed, let the parse method figure it out.

vdr/didx509/resolver_test.go
rolandgroen: I did refactor the assertions, most to assert.ErrorIs, some didn't work that way. These are refactored to `assert.EqualError(t, err, str)`

vdr/didx509/resolver_test.go
rolandgroen: Removed the .Return(nil)

vdr/didx509/resolver.go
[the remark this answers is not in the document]
reinkrul: @gerardsn @stevenvegt why is this different from the other cases where this is used (why hard-fail here?).

I find it dangerous to choose to hard-fail here, since it causes data exchanges to be blocked when the issuer of the UZI cert backing the credential's CRL is unreachable or expired. It's already quite a long chain, with more and more failure modes (we also have the revocation status of the credential itself to check), so hard-fail makes it less robust.
That's why we implemented soft-fail in the first place, so how is it different here?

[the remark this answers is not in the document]
gerardsn: - before it was 'just' one of the layers of security. (Still requires did:nuts signatures to do anything)
- here it is the root of trust for your credentials. A CA should always be online, and even if it is offline for a few minutes, the cache is still available. Using softfail here means you accept credentials without checking the trust

I can make a PR that will make the required methods available

gerardsn:
This is missing an e2e-test.

Certificate revocation is currently not checked. The `pki.Validator` can only check CRLs for CAs certs in the configured truststore. Validation fails if the CA is not in the truststore, but this specific error is ignored when configured for `softfail` (default).

An e2e-test should show 
- warning logs containing `certificate validation failed because CRL cannot be updated`
- error logs containing `Certificate CRL check softfail bypass. Might be unsafe, find cause of failure!`

vdr/didx509/resolver.go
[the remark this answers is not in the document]
rolandgroen: @gerardsn I get what the general idea is, I just could use some help on how to implement this correctly. Should I create a new first class citizen for the pkiValidator instance with configuation elements, new defaults etc.? Or should I create my own instance with a specific configuration? Or should I clone / copy the TLS instance and just set the softfail = false?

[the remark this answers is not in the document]
gerardsn: I will create a PR later today that you can use here.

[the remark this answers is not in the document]
gerardsn: #3531 should give you `ValidateStrict` to use here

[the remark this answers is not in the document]
rolandgroen: What is procedure from here? Move on the this PR and add it later or wait for the https://github.com/nuts-foundation/nuts-node/pull/3531 to merge into main/master?

In the mean time I'm working on the latest issues.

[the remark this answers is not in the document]
gerardsn: PR is merged so yo can use `ValidateStrict` now. I believe @reinkrul was still going to compare this with the spec, so maybe resolve all conflicts in the meantime so the tests can run again.

vdr/didx509/x509_utils.go
rolandgroen: Added tests 😅

vdr/resolver/did.go
rolandgroen: Added tests.

vdr/didx509/resolver.go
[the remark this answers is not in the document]
rolandgroen: The master + PR is merged into this PR, code is changed to `ValidateStrict`.

reinkrul (approved):

vdr/didx509/jwt_utils.go
```diff
@@ -0,0 +1,33 @@
...
+
+var (
+	ErrorInvalidNumberOfSignatures = errors.New("invalid number of signatures")
+)
+
+// ExtractProtectedHeaders extracts the protected headers from a JWT string.
+// The function takes a JWT string as input and returns a map of the protected headers.
+// Note that:
+//   - This method ignores any parsing errors and returns an empty map instead of an error.
+func ExtractProtectedHeaders(jwt string) (map[string]interface{}, error) {
```
woutslakhorst, line 17: should be private otherwise it's in the wrong place

vdr/didx509/jwt_utils.go
```diff
@@ -0,0 +1,33 @@
+package didx509
```
woutslakhorst, line 1: missing copyrights

vdr/didx509/resolver.go
```diff
@@ -0,0 +1,182 @@
...
+
+const (
+	MethodName                   = "x509"
+	X509CertChainHeader          = "x5c"
+	X509CertThumbprintHeader     = "x5t"
+	X509CertThumbprintS256Header = "x5t#S256"
+)
+
+var (
+	ErrX509ChainMissing            = errors.New("x509 rootCert chain is missing")
```
woutslakhorst, line 22: Missing godoc

vdr/didx509/resolver.go
```diff
@@ -0,0 +1,182 @@
...
+	"fmt"
+	ssi "github.com/nuts-foundation/go-did"
+	"github.com/nuts-foundation/go-did/did"
+	"github.com/nuts-foundation/nuts-node/pki"
+	"github.com/nuts-foundation/nuts-node/vdr/resolver"
+	"strings"
+)
+
+const (
+	MethodName                   = "x509"
```
woutslakhorst, line 15: missing godoc

vdr/didx509/resolver.go
```diff
@@ -0,0 +1,182 @@
...
+	Name  PolicyName
+	Value string
+}
+
+type X509DidReference struct {
+	Method      HashAlgorithm
+	RootCertRef string
+	Policies    []X509DidPolicy
+}
+
```
woutslakhorst, line 50: godoc on public structs

woutslakhorst: and missing reference to w3c spec

vdr/didx509/validation.go
```diff
@@ -0,0 +1,226 @@
...
+		subject := cert.Subject
+		if !slices.Contains(subject.OrganizationalUnit, value) {
+			return fmt.Errorf("query does not match the subject : %s", key)
+		}
+		return nil
+	},
+}
+
+// ValidatePolicy validates a certificate against a given X509DidReference and its policies.
+func ValidatePolicy(ref *X509DidReference, cert *x509.Certificate) error {
```
woutslakhorst, line 182: public or private?

vdr/resolver/did.go
```diff
@@ -189,3 +192,27 @@ func GetDIDFromURL(didURL string) (did.DID, error) {
 func IsDeactivated(document did.Document) bool {
 	return len(document.Controller) == 0 && len(document.CapabilityInvocation) == 0
 }
+
+func (m *ResolveMetadata) GetProtectedHeaderString(key string) (string, bool) {
```
woutslakhorst, line 196: godoc

vdr/resolver/did.go
```diff
@@ -189,3 +192,27 @@ func GetDIDFromURL(didURL string) (did.DID, error) {
...
+	}
+	value, ok := m.JwtProtectedHeaders[key]
+	if !ok {
+		return "", false
+	}
+	str, ok := value.(string)
+	return str, ok
+}
+
+func (m *ResolveMetadata) GetProtectedHeaderChain(key string) (*cert.Chain, bool) {
```
woutslakhorst, line 208: godoc

woutslakhorst (changes requested):
- refer to which spec is implemented in godoc comment
- copyrights
- godoc on public vars/functions/structs/methods

vdr/didx509/jwt_utils.go
rolandgroen: It's not private, so where should it go? `crypto/jwx.go`?

vdr/didx509/jwt_utils.go
gerardsn: run `make fix-copyright`

vdr/didx509/jwt_utils.go
rolandgroen: This method is made private.

vdr/didx509/validation.go
rolandgroen: Made private.

vdr/resolver/did.go
rolandgroen: Added doc.

vdr/resolver/did.go
rolandgroen: Added doc.

vdr/didx509/resolver.go
rolandgroen: Added docs

vdr/didx509/jwt_utils.go
rolandgroen: Added manually.

vdr/didx509/jwt_utils.go
rolandgroen: Done.

woutslakhorst (approved):

gerardsn (approved):
Last comment: add did:x509 to the supported DID methods in `docs/pages/integrating/supported-protocols-formats.rst`

vdr/didx509/resolver.go
```diff
@@ -0,0 +1,222 @@
...
+	if didParts[0] != "0" {
+		return nil, ErrDidVersion
+	}
+	ref.Method = HashAlgorithm(didParts[1])
+	ref.RootCertRef = didParts[2]
+
+	for _, policyString := range policyStrings {
+		policyFragments := strings.Split(policyString, ":")
+		if len(policyFragments) > 1 {
+			policy := X509DidPolicy{Name: PolicyName(policyFragments[0]), Value: strings.Join(policyFragments[1:], ":")}
```
reinkrul, line 215: Does this work with percent-encoded values in the `did:x509`? The spec (https://trustoverip.github.io/tswg-did-x509-method-specification/#san-policy) states:

> ```
policy-name     = "san"
policy-value    = san-type ":" san-value
san-type        = "email" / "dns" / "uri"
san-value       = 1*idchar
```
> san-type is the SAN type and must be one of email, dns, or uri. Note that dn is not supported.
> san-value is the SAN value, percent-encoded.

**Review by stevenvegt — CHANGES_REQUESTED** [review]
Wait for it....
Let me try it out before merging.

File: vdr/didx509/x509_utils.go [hunk 31]
```diff
@@ -0,0 +1,236 @@
...
+			return findSanValue(extension)
+		}
+	}
+	return "", nil
+}
+
+// findSanValue extracts the SAN value from a given pkix.Extension, returning the resulting value or an error.
+func findSanValue(extension pkix.Extension) (string, error) {
+	value := ""
+	err := forEachSan(extension.Value, func(data []byte) error {
```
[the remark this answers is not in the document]
rolandgroen: Well, the last one, actually. I'll take a look at this.

rolandgroen:
The file `docs/pages/integrating/supported-protocols-formats.rst` file has been modified.

rolandgroen:
The command `go fmt vdr/vdr.go` has been applied.

stevenvegt (approved):
I don't want to hold this one up longer than needed. We can finetune the logic in a later phase, since this feature is still very experimental.

rolandgroen:
I did some end-to-end testing and some issues where found & fixed:
 *  The #0 fragment was missing in the generated vc, fixed that in the issuer code.
 *  The issuer added the permanentIdentifier by default, that does not work with the current NUTS code, so that feature has been hidden behind a flag in the issuer.
 * The x509 otheName extraction code did not check for other types of otheName values and crashed on the alternative host name. That has been fixed + tests have been added.

PR #3449, opened 2024-10-04
Disable network layer if 'didmethods' does not contain 'nuts'

gerardsn:


woutslakhorst (approved):
Feels like a trial and error exercise. But I have faith in our test suite.

PR #3465, opened 2024-10-08
introspection request returns error if missing a token

### LINKED ISSUE #3464: introspection returns no error when wrong media type is used
request is specced as `Content-Type: application/x-www-form-urlencoded`.
Using `application/json` is a common mistake that result in no token being received by the node, and a response `{active: false}`.

gerardsn:
closes #3464 

now returns a Warn message if there is no token (should only happen during development) and 415 if using the wrong content type

### DIFF
--- a/auth/api/iam/api.go
+++ b/auth/api/iam/api.go
@@ -30,6 +30,7 @@ import (
 	"html/template"
 	"net/http"
 	"net/url"
+	"slices"
 	"strings"
 	"time"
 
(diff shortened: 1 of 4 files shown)


woutslakhorst (changes requested):
This is not according to RFC7662. any wrong content (with the correct content-type) must result in a 200 with active set to false.

415 is allowed for the wrong content-type.

If more feedback is needed, check the logging.

woutslakhorst (approved):

PR #3474, opened 2024-10-09
Add remove credential based on subject and deprecate based on DID

### LINKED ISSUE #3462: API proposal: DELETE /internal/vcr/v2/holder/{subjectID}/vc/{id}
we currently have 
```
DELETE /internal/vcr/v2/holder/{did}/vc/{id}
```
which is the only API that still uses `{did}`.

probably a 'nice to have' to make the API's consistent

gerardsn:
closes #3462

deprecates: `DELETE /internal/vcr/v2/holder/{did}/vc/{id}`
adds: `DELETE /internal/vcr/v2/holder/{subjectID}/vc/{id}`

### DIFF
--- a/docs/_static/vcr/vcr_v2.yaml
+++ b/docs/_static/vcr/vcr_v2.yaml
@@ -530,6 +530,42 @@ paths:
           description: The credential will not be altered in any way, so no need to return it.
         default:
           $ref: '../common/error_response.yaml'
+  /internal/vcr/v2/holder/{subjectID}/vc/{id}:
+    parameters:
+      - name: subjectID
+        in: path
+        description: Subject ID of the wallet owner at this node.
+        required: true
+        content:
+          plain/text:
+            schema:
+              type: string
+              example: 90BC1AE9-752B-432F-ADC3-DD9F9C61843C
+      - name: id
+        in: path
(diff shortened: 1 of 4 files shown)


woutslakhorst (changes requested):
This can't work. Interesting echo doesn't give an error on this:

`/internal/vcr/v2/holder/{subjectID}/vc/{id}` is the exact same mapping as`/internal/vcr/v2/holder/{did}/vc/{id}`. Both have a variable, so if both mappings are active, which one processes the request? I think only the latest one survives.

Where does `/internal/vcr/v2/holder/1234-12345678-1234-1234/vc/1` map to?

gerardsn:
>Where does /internal/vcr/v2/holder/1234-12345678-1234-1234/vc/1 map to?

Both routes are registered in a map which does not have a deterministic order. I threw a performance test at it for 30 seconds and 400 simultaneous connections. All ±0.5M request ended up at the subjectID endpoint. When switching route registration order the other endpoint wins. So it appears that the first registered route will always win.

Anyway, Rein pointed out that the endpoint has not been in a previous release, so we can just change it.

woutslakhorst (approved):

PR #3476, opened 2024-10-09
pass problem details when discovery service returns them

### LINKED ISSUE #3471: Pass error details from discovery server to client and to user.
The spec doesn't contain error responses yet. If we add them then we can pass on any errors to the client.

woutslakhorst:
fixes #3471 

when the discovery server returns problem details, the client will try to parse. If successful, it'll pass on the error.

### DIFF
--- a/discovery/api/server/client/http.go
+++ b/discovery/api/server/client/http.go
@@ -109,5 +109,5 @@ func problemResponseToError(httpErr core.HttpError) string {
 	if err := json.Unmarshal(httpErr.ResponseBody, &problemDetails); err != nil {
 		return fmt.Sprintf("%s: %s", httpErr.Error(), httpErr.ResponseBody)
 	}
-	return fmt.Sprintf("server returned HTTP %d: %s: %s", problemDetails.Status, problemDetails.Title, problemDetails.Description)
+	return fmt.Sprintf("server returned HTTP status code %d: %s: %s", problemDetails.Status, problemDetails.Title, problemDetails.Description)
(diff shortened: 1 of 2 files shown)


discovery/api/server/client/http.go
```diff
@@ -95,3 +97,17 @@ func (h DefaultHTTPClient) Get(ctx context.Context, serviceEndpointURL string, t
...
+func problemResponseToError(httpErr core.HttpError) string {
+	var problemDetails struct {
+		Title       string `json:"title"`
+		Description string `json:"detail"`
+		Status      int    `json:"status"`
+	}
+	if err := json.Unmarshal(httpErr.ResponseBody, &problemDetails); err != nil {
+		return fmt.Sprintf("%s: %s", httpErr.Error(), httpErr.ResponseBody)
+	}
+	return fmt.Sprintf("server returned HTTP %d: %s: %s", problemDetails.Status, problemDetails.Title, problemDetails.Description)
```
gerardsn, line 112:
```suggestion
	return fmt.Sprintf("server returned HTTP status code %d: %s: %s", problemDetails.Status, problemDetails.Title, problemDetails.Description)
```

gerardsn (approved):
I'm approving the change, but it doesn't fix the issue

woutslakhorst:
> I'm approving the change, but it doesn't fix the issue

The title of the issue was wrong. The goal was to return the error description at the server to the client and beyond. Not create specific error titles.

PR #3478, opened 2024-10-10
Check RW permissions in datadir

### LINKED ISSUE #3470: Check write rights for user on startup
to prevent "out-of-memory" error

gerardsn:
closes #3470 

I'm not convinced this is actually an improvement. The 2 most likely errors are
```
level=fatal msg="Could not start the server" error="unable to create datadir (dir=./data): mkdir ./data: read-only file system"
```
and
```
level=fatal msg="Could not start the server" error="unable to configure Storage: open data/rw-access-test-file: read-only file system"
```
Both are pointing to read-only file system issue, but the real issue is an unprivileged user.
The unprivileged user is what also caused `"failed to initialize database, got error unable to open database file: out of memory (14)"` that resulted in this issue. This check does prevent the OOM error from ever occurring again though.

I don't see how to clarify the error without adding a very long error description since we can't differentiate between unprivileged users and an actual read-only file systems.

### DIFF
--- a/storage/engine.go
+++ b/storage/engine.go
@@ -24,6 +24,7 @@ import (
 	"fmt"
 	"os"
 	"path"
+	"path/filepath"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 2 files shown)


reinkrul (commented):
Maybe add it to the frequently encountered error page? (If we still have it)

woutslakhorst:
an error for read-only file system is much closer to anything rights related then out-of-memory. So already an improvement.

woutslakhorst:
There are some examples on how to use `os.Stat` and check for permission. But this only works for the permission bits. ACLs would still fail.

gerardsn:
ok, so hit that green button ;)

---

> Maybe add it to the frequently encountered error page? (If we still have it)

Page has been removed

reinkrul (approved):

PR #3483, opened 2024-10-14
move discovery VP validation to background process to prevent errors

### LINKED ISSUE #3468: If revocation status list is down, discovery registration fails
Should be a default-open state, if it can't fetch the status list? (just like the CRLs)

woutslakhorst:
fixes #3468 

any did:web or revocation downtime will not block discovery entries. Entries are invalid until VerifyVP succeeds at least once. After that a revocation checker will remove entries when a VC is revoked.

Only downside is a full table scan on presentation records every 10 minutes. If this becomes a problem (eg > 100k results) then `Search` has to verify the VPs and delete entries as a side effect.... (or some elaborate eventing thing from revocationList to discovery service.)

### DIFF
--- a/discovery/store.go
+++ b/discovery/store.go
@@ -479,7 +479,7 @@ func (s *sqlStore) setPresentationRefreshError(serviceID string, subjectID strin
 		}
 
 		if refreshErr == nil {
-			// a deletePresentationRecord
+			// a delete
 			return nil
 		}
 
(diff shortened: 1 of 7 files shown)


discovery/store.go
```diff
@@ -410,7 +479,7 @@ func (s *sqlStore) setPresentationRefreshError(serviceID string, subjectID strin
 		}
 
 		if refreshErr == nil {
-			// a delete
+			// a deletePresentationRecord
```
gerardsn, line 482: replace all error?

gerardsn (approved):

PR #3499, opened 2024-10-17
SignJWS: always add kid to header

### LINKED ISSUE #3488: Crypto: SignJWS and SignJWT don't log all key operations
SignJWT logging:
```
audit.Log(ctx, log.Logger(), audit.CryptoSignJWTEvent).Infof("Signing a JWT with key: %s (issuer: %s, subject: %s)", headers["kid"], claims["iss"], claims["sub"])
```

SignJWS logging:
```
audit.Log(ctx, log.Logger(), audit.CryptoSignJWSEvent).Infof("Signing a JWS with key: %s", protectedHeaders["kid"])
```

both use `headers["kid"]` which does not always exists. This results in things like `level=audit msg="Signing a JWS with key: %!s(<nil>)" actor=4.3.2.1 event=SignJWS module=Crypto operation=VDR.CreateDID`

gerardsn:
closes #3488

### DIFF
--- a/crypto/jwx.go
+++ b/crypto/jwx.go
@@ -82,9 +82,7 @@ func (client *Crypto) SignJWS(ctx context.Context, payload []byte, headers map[s
 		return "", err
 	}
 
-	if _, ok := headers["jwk"]; !ok {
-		headers["kid"] = kid
-	}
+	headers["kid"] = kid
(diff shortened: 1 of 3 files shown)


woutslakhorst (approved):
We might want to mark those sign APIs as deprecated in the futuer (not used)

PR #3508, opened 2024-10-21
secure outgoing http client with max connections

woutslakhorst:
fixes outcome from Elevation Of Privilege.

checked all creations of `http.Client` and replaced with `client.New`, `client.NewWithCache` or `client.NewWithTLSConfig`.

Also connected the limitreader to the strictClient.

### DIFF
--- a/core/http_client.go
+++ b/core/http_client.go
@@ -31,21 +31,6 @@ import (
 // If the response body is longer than this, it will be truncated.
 const HttpResponseBodyLogClipAt = 200
 
-// DefaultMaxHttpResponseSize is a default maximum size of an HTTP response body that will be read.
-// Very large or unbounded HTTP responses can cause denial-of-service, so it's good to limit how much data is read.
-// This of course heavily depends on the use case, but 1MB is a reasonable default.
-const DefaultMaxHttpResponseSize = 1024 * 1024
-
-// LimitedReadAll reads the given reader until the DefaultMaxHttpResponseSize is reached.
-// It returns an error if more data is available than DefaultMaxHttpResponseSize.
-func LimitedReadAll(reader io.Reader) ([]byte, error) {
-	result, err := io.ReadAll(io.LimitReader(reader, DefaultMaxHttpResponseSize+1))
-	if len(result) > DefaultMaxHttpResponseSize {
-		return nil, fmt.Errorf("data to read exceeds max. safety limit of %d bytes", DefaultMaxHttpResponseSize)
-	}
-	return result, err
-}
-
 // HttpError describes an error returned when invoking a remote server.
 type HttpError struct {
 	error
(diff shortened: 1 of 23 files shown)


http/client/client.go
```diff
@@ -19,23 +19,59 @@
...
+	SafeHttpTransport = http.DefaultTransport.(*http.Transport).Clone()
+	if SafeHttpTransport.TLSClientConfig == nil {
+		SafeHttpTransport.TLSClientConfig = &tls.Config{}
 	}
-	httpTransport.TLSClientConfig.MinVersion = tls.VersionTLS12
+	SafeHttpTransport.TLSClientConfig.MinVersion = tls.VersionTLS12
+	// to prevent slow responses from public clients to have significant impact (default was unlimited)
+	SafeHttpTransport.MaxConnsPerHost = 5
+
+	DefaultCachingTransport = SafeHttpTransport
```
gerardsn, line 44: `Engine.configureClient` also sets `DefaultCachingTransport` and comes after this. Does this do anything? Maybe add a comment that the real value is set somewhere else

http/client/caching.go
```diff
@@ -32,8 +32,8 @@ import (
 
 // DefaultCachingTransport is a http.RoundTripper that can be used as a default transport for HTTP clients.
 // If caching is enabled, it will cache responses according to RFC 7234.
-// If caching is disabled, it will behave like http.DefaultTransport.
-var DefaultCachingTransport = http.DefaultTransport
+// If caching is disabled, it will behave like our safe http.DefaultTransport.
```
gerardsn, line 35:
```suggestion
// If caching is disabled, it will behave like our safe http.SafeHttpTransport.
```

docs/pages/deployment/security-considerations.rst
```diff
@@ -19,6 +19,13 @@ D(D)oS Protection
 *****************
 
 Consider implementing (D)DoS protection on the application layer for all public endpoints.
+Monitor and og the following metrics:
```
gerardsn, line 22:
```suggestion
Monitor and log the following metrics:
```

pki/denylist.go
```diff
@@ -219,6 +219,7 @@ func (b *denylistImpl) Subscribe(f func()) {
 // download retrieves and parses the denylist
 func (b *denylistImpl) download() ([]byte, error) {
 	// Make an HTTP GET request for the denylist URL
+	// We do not use our safe http client here since we're downloading from our own resource
```
gerardsn, line 222: download is from a configurable endpoint. (hidden config flag though)
Probably doesn't matter since it is a trusted endpoint.

core/http_client.go
```diff
@@ -104,16 +89,18 @@ func (w httpRequestDoerAdapter) Do(req *http.Request) (*http.Response, error) {
...
-// CreateHTTPClient creates a new HTTP client with the given client configuration.
+// CreateHTTPInternalClient creates a new HTTP client with the given client configuration.
+// This client is to be used for internal API calls (CMDs and such)
 // The result HTTPRequestDoer can be supplied to OpenAPI generated clients for executing requests.
 // This does not use the generated client options for e.g. authentication,
 // because each generated OpenAPI client reimplements the client options using structs,
 // which makes them incompatible with each other, making it impossible to use write generic client code for common traits like authorization.
 // If the given authorization token builder is non-nil, it calls it and passes the resulting token as bearer token with requests.
-func CreateHTTPClient(cfg ClientConfig, generator AuthorizationTokenGenerator) (HTTPRequestDoer, error) {
+func CreateHTTPInternalClient(cfg ClientConfig, generator AuthorizationTokenGenerator) (HTTPRequestDoer, error) {
```
gerardsn, line 99: a bit weird that this is under `core` and all the other under `http` like the rest. Import cycle on config?

gerardsn (approved):

core/http_client.go
woutslakhorst: these are for our commands, the other are outgoing.

PR #3510, opened 2024-10-22
remove network migration and optimize network event retry

woutslakhorst:
Removed all logic referred to by *remove for v5/v6 release*

Optimised network event retry at startup causing deadlock between long running read and write lock


network/dag/dag.go
```diff
@@ -96,69 +96,6 @@ func newDAG(db stoabs.KVStore) *dag {
...
-			log.Logger().Info("Number of transactions not stored, migrating...")
-			numberOfTXs := d.getNumberOfTransactionsLegacy(tx)
-			err = d.setNumberOfTransactions(tx, numberOfTXs)
-		}
-		if err != nil {
-			return err
-		}
-
-		// Migrate headsLegacy to single head
-		// Todo: remove after V6 release => then remove headsShelf
```
gerardsn, line 127: I'm assuming the second half of this todo points to a now unused constant that should be removed?

gerardsn (approved):

network/dag/dag.go
woutslakhorst: unused shelf, but since the entire DB will be removed in the future...

PR #3513, opened 2024-10-23
status codes for discovery client

### LINKED ISSUE #3502: Discovery Service: check API responses for correct status codes

gerardsn:
closes #3502 

PR contains
- add clientRegistrationManager.getServiceAndSubject to check for service and subject
- Add 404 for unknown subjectID and serviceID
- Add 202 on DELETE as documented in OAPI spec
- Log ignored errors
- cleanup `api/server#client.New`


discovery/client.go
```diff
@@ -52,6 +52,8 @@ type clientRegistrationManager interface {
 	validate() error
 	// removeRevoked removes all revoked presentations from the store
 	removeRevoked() error
+	// getServiceAndSubject returns the service and subject, or ErrServiceNotFound / didsubject.ErrSubjectNotFound if either does not exist
+	getServiceAndSubject(ctx context.Context, serviceID, subjectID string) (ServiceDefinition, []did.DID, error)
```
woutslakhorst, line 56: why add it to the interface? There's no mock and only use is at the same struct.

discovery/client.go
```diff
@@ -116,7 +110,7 @@ func (r *defaultClientRegistrationManager) activate(ctx context.Context, service
 	subjectDIDs = subjectDIDs[:j]
 
 	if len(subjectDIDs) == 0 {
-		return fmt.Errorf("%w: %w for %s", ErrPresentationRegistrationFailed, didsubject.ErrSubjectNotFound, subjectID)
+		return fmt.Errorf("%w: %w for %s", ErrPresentationRegistrationFailed, ErrDIDMethodsNotSupported, subjectID)
```
woutslakhorst, line 113: Better to change the error to something like: "no active DIDs found for subject"

woutslakhorst (changes requested):

discovery/client.go
gerardsn: removed interface since it only had 1 implementation

woutslakhorst (approved):

discovery/client.go
gerardsn: added a new error for this with message `subject has no (active) DIDs matching the service`

PR #3529, opened 2024-11-01
PKI Valdiator always fails on unknown CAs

### LINKED ISSUE #3528: PKI: accepts certs not in truststore when softfail=true (default)
PKI validator should only accept certificates from CAs that are in the truststore, also when set to softfail.

this is currently working as designed, so I removed the bug label
From the `Validator.Validate` interface:
```
Ignoring all errors except ErrCertRevoked changes the behavior from hard-fail to soft-fail. Without a truststore, the Validator is a noop if set to soft-fail
```
Is there a reason we needed it this way?

backport to V5.4 and V6.0 if accepted

gerardsn:
closes #3528 

changed the accepted errors list from fail open to fail closed

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -31,6 +31,7 @@ Nuts documentation
     pages/deployment/configuration.rst
     pages/deployment/migration.rst
     pages/deployment/recommended-deployment.rst
+    pages/deployment/certificates.rst
     pages/deployment/docker.rst
     pages/deployment/storage.rst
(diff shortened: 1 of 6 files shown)


woutslakhorst (changes requested):
requires some clear documentation, maybe a new page _certificates_ and some addition to recommended deployment checlist.

gerardsn:
what should be added to the recommended deployment?

woutslakhorst (approved):

PR #3565, opened 2024-11-18
Cache access tokens client side

### LINKED ISSUE #3552: Cache access tokens
We can cache access tokens for their lifetime based on the service-access-token-request. This would result in automatic reuse of access tokens during their 15 minute lifespan while being transparent to the requestor.

support Redis(with sentinal) and memcached

woutslakhorst:
Cache access tokens per request. Subsequent requests are fulfilled from cache.
Added SessionOption to allow for custom `ttl` on session store.
Only implemented for `auth.iam`, not for `auth/v1`

closes #3552

### DIFF
--- a/auth/api/iam/api.go
+++ b/auth/api/iam/api.go
@@ -74,6 +74,10 @@ type httpRequestContextKey struct{}
 // TODO: Might want to make this configurable at some point
 const accessTokenValidity = 15 * time.Minute
 
+// accessTokenCacheOffset is used to reduce the ttl of the access token to ensure it is still valid when the client receives it.
+// this to offset clock skew and roundtrip times
+const accessTokenCacheOffset = 30 * time.Second
+
 // cacheControlMaxAgeURLs holds API endpoints that should have a max-age cache control header set.
 var cacheControlMaxAgeURLs = []string{
 	"/oauth2/:subjectID/presentation_definition",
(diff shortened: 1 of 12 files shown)


auth/api/iam/api.go
```diff
@@ -732,11 +751,22 @@ func (r Wrapper) RequestServiceAccessToken(ctx context.Context, request RequestS
...
 	clientID := r.subjectToBaseURL(request.SubjectID)
-	tokenResult, err := r.auth.IAMClient().RequestRFC021AccessToken(ctx, clientID.String(), request.SubjectID, request.Body.AuthorizationServer, request.Body.Scope, useDPoP, credentials)
+	tokenResult, err = r.auth.IAMClient().RequestRFC021AccessToken(ctx, clientID.String(), request.SubjectID, request.Body.AuthorizationServer, request.Body.Scope, useDPoP, credentials)
 	if err != nil {
 		// this can be an internal server error, a 400 oauth error or a 412 precondition failed if the wallet does not contain the required credentials
 		return nil, err
 	}
+	ttl := accessTokenValidity
+	if tokenResult.ExpiresIn != nil {
+		ttl = time.Second * time.Duration(*tokenResult.ExpiresIn)
```
reinkrul, line 761: when could it not be set 🤔 ?

storage/session.go
```diff
@@ -85,16 +85,31 @@ func (s SessionStoreImpl[T]) Get(key string, target interface{}) error {
...
 
-func (s SessionStoreImpl[T]) Put(key string, value interface{}) error {
+func (s SessionStoreImpl[T]) Put(key string, value interface{}, options ...SessionOption) error {
+	opts := s.defaultOptions()
+	for _, opt := range options {
+		opt(&opts)
+	}
+	// TTL can't go below 0 because that is translated to "no expiration" by the library
+	// in that case it should be 1 nanosecond
+	if opts.ttl < 0 {
```
reinkrul, line 95: or just not put it in cache?

reinkrul (commented):
small remarks/questions

reinkrul (approved):

storage/session.go
woutslakhorst: 🤦

auth/api/iam/api.go
woutslakhorst: It's a pointer, probably because the spec says it's optional. In our implementation it's not but you never know.

PR #3571, opened 2024-11-19
reenable CORS for IRMA

### LINKED ISSUE #3570: Proposal of replacing CORS with config in reverse proxy is not a 1-on-1 replacement.
the echo middleware used regexes to check the origin header and add a `Access-Control-Allow-Origin` equal to the `Origin` header if it matches the regex.

`*.example.com` was allowed in the Echo CORS config but not as config value for Nginx for example. Only a concrete list or `*` is allowed. 

Maybe backport

woutslakhorst:
fixes #3570 

Can be released with v6.1

Tested by hand with 
```
curl localhost:8080/public/auth/irmaclient/session -X OPTIONS -H "Origin: https://sub.example.com" -v
```
and config
```
auth.irma.cors.origin: "https://*.example.com"
```


auth/cmd/cmd.go
```diff
@@ -51,6 +54,7 @@ func FlagSet() *pflag.FlagSet {
 	defs := auth.DefaultConfig()
 	flags.String(ConfIrmaSchemeManager, defs.Irma.SchemeManager, "IRMA schemeManager to use for attributes. Can be either 'pbdf' or 'irma-demo'.")
 	flags.Bool(ConfAutoUpdateIrmaSchemas, defs.Irma.AutoUpdateSchemas, "set if you want automatically update the IRMA schemas every 60 minutes.")
+	flags.StringSlice(ConfIrmaCorsOrigin, defs.Irma.CORS.Origin, "sets the allowed CORS origins for the IRMA server")
```
reinkrul, line 57: A note for what the behavior is when not set?

reinkrul (approved):

PR #3573, opened 2024-11-19
Support did:x509 in Authorization Server metadata

reinkrul:
Allowing credentials to be issued from it.

This is probably not the right solution, since the actual issue is that `didmethods` indicates 2 things:
- DIDs the Nuts node creates for its subjects
- DID methods it accepts when other parties present credentials

These should probably be 2 lists. The 2nd list could actually be all supported DID methods (nuts, web, x509, jwk, key) minus those we can't resolve if not configured properly (nuts).
@woutslakhorst ?

### DIFF
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -179,7 +179,7 @@ func (auth *Auth) Configure(config core.ServerConfig) error {
 }
 
 func (auth *Auth) SupportedDIDMethods() []string {
-	return auth.supportedDIDMethods
+	return append(auth.supportedDIDMethods, "x509")
 }
 
 // Start starts the Auth engine (Noop)
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -189,9 +189,12 @@ func (auth *Auth) Configure(config core.ServerConfig) error {
 func (auth *Auth) SupportedDIDMethods() []string {
 	// DID methods that don't require additional resources/configuration in the Nuts node are always supported.
 	// Other DID methods (did:nuts), are only supported if explicitly enabled.
-	result := []string{didweb.MethodName, didjwk.MethodName, didkey.MethodName, didx509.MethodName}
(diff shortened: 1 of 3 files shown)


woutslakhorst:
The static set (resolvable by code) of jwk, key and x509 should always be added. This is also true for the discovery service definition `did_methods`

reinkrul:
Discovery has slightly different rules, made an issue https://github.com/nuts-foundation/nuts-node/issues/3575

Although Discovery might not actually need `did_methods`, and could use a prefix with wildcard in PEX..

auth/auth.go
```diff
@@ -179,7 +187,13 @@ func (auth *Auth) Configure(config core.ServerConfig) error {
 }
 
 func (auth *Auth) SupportedDIDMethods() []string {
-	return auth.supportedDIDMethods
+	// DID methods that don't require additional resources/configuration in the Nuts node are always supported.
+	// Other DID methods (did:nuts), are only supported if explicitly enabled.
+	result := []string{didweb.MethodName, didjwk.MethodName, didkey.MethodName, didx509.MethodName}
```
woutslakhorst, line 192: did:web can also be disabled

woutslakhorst (approved):

auth/auth.go
reinkrul: Thinking out loud: what would be the use case for NOT accepting did:web credentials?

- did:web is arguably unsafer compared to jwk, key, x509 (since these aren't suspectible to network-bound attacks, e.g. DNS or domain hijacking). So verifiers might not want to accept it.
- On the other hand, parties might want to accept it but might not want their Nuts node to create did:web DIDs.

Both are edge cases i.m.o., but I think the first argument is more credible. What do you think?

reinkrul:
Updated

Issue #3583, opened 2024-12-02
pki.AddTruststore is never called if tls.{certfile,certkeyfile} is not configured


gerardsn:
this causes `did:x509` certificate validation to fail because it needs the truststore

woutslakhorst:
meaning **"did:web only"** people can never use it.

reinkrul:
In my experience, you only need to configure `tls.truststorefile`?

tomelfring:
`pki` loads the TLSConfig in `CreateTLSConfig`. Which checks for `t.Enabled()`. That one does the following check:
```
// Enabled returns whether TLS should be enabled, according to the global config.
func (t TLSConfig) Enabled() bool {
	return len(t.CertFile) > 0 || len(t.CertKeyFile) > 0
}
```

reinkrul:
> `pki` loads the TLSConfig in `CreateTLSConfig`. Which checks for `t.Enabled()`. That one does the following check:
> 
> ```
> // Enabled returns whether TLS should be enabled, according to the global config.
> func (t TLSConfig) Enabled() bool {
> 	return len(t.CertFile) > 0 || len(t.CertKeyFile) > 0
> }
> ```
> 
> So setting only the `tls.truststorefile` doesn't make it load the truststore. Currently you do also need to set `certfile` and `certkeyfile`.

You're right. We looking into this to fix it.

gerardsn:
Proposed solution is to call `AddTruststore` at runtime from did:x509 logic **after** it has been verified that the chain is valid (minus CRL check) and allowed for a use-case. This also eliminates the requirement to configure a truststore for did:x509

rolandgroen:
Depends on https://github.com/nuts-foundation/nuts-node/issues/3588

rolandgroen:
The "truststore" from TLS is something else than the "cache/list of CRLs and CA-certs" for the x509 validation. It makes sense to separate both lists of CAs.

PR #3586, opened 2024-12-03
Rename pki.Validate to pki.CheckCRL for clarity on its purpose

gerardsn:


reinkrul:
Go style would be `CheckCrl`

gerardsn:
https://go.dev/wiki/CodeReviewComments disagrees
> Words in names that are initialisms or acronyms (e.g. “URL” or “NATO”) have a consistent case. For example, “URL” should appear as “URL” or “url” (as in “urlPony”, or “URLPony”), never as “Url”. As an example: ServeHTTP not ServeHttp. For identifiers with multiple initialized “words”, use for example “xmlHTTPRequest” or “XMLHTTPRequest”.
>
>This rule also applies to “ID” when it is short for “identifier” (which is pretty much all cases when it’s not the “id” as in “ego”, “superego”), so write “appID” instead of “appId”.
>
>Code generated by the protocol buffer compiler is exempt from this rule. Human-written code is held to a higher standard than machine-written code.

tomelfring:
Just curious as I digged through this method to debug #3583, is CheckCRL really the only intended feature of this method? It does also check if the certificate is on a denylist (arguably some kind of CRL):
https://github.com/nuts-foundation/nuts-node/blob/51cb9495c607f2f358b68299912f10e5027fb7a5/pki/validator.go#L176

And if the certificate is known in the truststore:
https://github.com/nuts-foundation/nuts-node/blob/51cb9495c607f2f358b68299912f10e5027fb7a5/pki/validator.go#L191

To me it looks like it does more than just a certificate revocation check.

reinkrul:
> https://go.dev/wiki/CodeReviewComments disagrees
> 
> > Words in names that are initialisms or acronyms (e.g. “URL” or “NATO”) have a consistent case. For example, “URL” should appear as “URL” or “url” (as in “urlPony”, or “URLPony”), never as “Url”. As an example: ServeHTTP not ServeHttp. For identifiers with multiple initialized “words”, use for example “xmlHTTPRequest” or “XMLHTTPRequest”.
> > This rule also applies to “ID” when it is short for “identifier” (which is pretty much all cases when it’s not the “id” as in “ego”, “superego”), so write “appID” instead of “appId”.
> > Code generated by the protocol buffer compiler is exempt from this rule. Human-written code is held to a higher standard than machine-written code.
> 
> the linter doesn't know it's an acronym though

Oh... I'm almost sure someone pointed that out to me, but I can't remember now 🤦

gerardsn:
> Just curious as I digged through this method to debug https://github.com/nuts-foundation/nuts-node/issues/3583, is CheckCRL really the only intended feature of this method? It does also check if the certificate is on a denylist (arguably some kind of CRL):

The denylist is a CRL under control of the user instead of the issuer. It should not be used but exists in case we need it in the future.

> And if the certificate is known in the truststore:

We need this certificate to validate the signature on the CRL. So it is more a lookup of a trusted key than confirming the issuer is in the truststore. The trustchain is verified outside of this module and that is one of the reasons I wanted to change the name.

PR #3589, opened 2024-12-03
#3587: validate certificate chain when resolving did:x509 DIDs

### LINKED ISSUE #3587: did:x509 issuer and resolver assume the root CA as ca-fingerprint
Which is wrong: in practice, this will be PKIOverheid Private Service Root CA, meaning _any_ intermediate CA certificate is accepted. This would allow more intermediate CAs than CIBG UZI (through KPN) to issue certificates that can used to create `UziServerCertificateCredential`.

The did:x509 spec states that the ca-fingerprint property can be of any root or intermediate CA. It's then up to the verifier (in our case through PEX) to restrict this to the accepted trust anchors

Meaning:
- The uzi-x509-issuer tool should take the direct issuer of the UZI server certificate (deepest intermediate CA), since this (probably) what use cases/Bolts want to configure in their PEX PDs
- The DID resolver in the Nuts node should take any ca-fingerprint to validate the cert chain, instead of hardcoded assume it's the root CA's ca-fingerprint.

### LINKED ISSUE #3588: Validate x5c chain of JWT credentials
Credentials issued by a did:x509 issuer contain a certificate chain in the jwt x5c header. This chain must be validated.

reinkrul:
Fixes #3588


core/tls.go
```diff
@@ -87,13 +87,23 @@ func LoadTrustStore(trustStoreFile string) (*TrustStore, error) {
...
-	trustStore.certificates, err = ParseCertificates(data)
+	certificates, err := ParseCertificates(data)
 	if err != nil {
 		return nil, err
 	}
+	trustStore := BuildTrustStore(certificates)
+
+	if err = validate(trustStore); err != nil {
+		return nil, err
+	}
```
gerardsn, line 98: unchanged behavior, but the error is marked as not having code coverage.

vcr/verifier/signature_verifier_test.go
```diff
@@ -361,71 +359,65 @@ func buildCertChain(ura string) (*cert.Chain, *x509.Certificate, *rsa.PrivateKey
...
 	if err != nil {
 		return nil, nil, nil, nil, err
 	}
-	chain[0] = signingCert
-	err = chainPems.Add(signingPEM)
-	if err != nil {
-		return nil, nil, nil, nil, err
+
+	certChain := &cert.Chain{}
+	for _, str := range []string{signingPEM, intermediateL2Pem, intermediateL1Pem, rootPem} {
```
gerardsn, line 412: the `x5c` header is specified to be DER encoded, not PEM
https://www.rfc-editor.org/rfc/rfc7515.html#section-4.1.6

vdr/didx509/resolver.go
```diff
@@ -115,12 +116,29 @@ func (r Resolver) Resolve(id did.DID, metadata *resolver.ResolveMetadata) (*did.
...
+		return nil, nil, fmt.Errorf("did:509 certificate chain validation failed: %w", err)
+	}
+
 	err = validatePolicy(ref, validationCert)
 	if err != nil {
 		return nil, nil, err
 	}
 
-	err = r.pkiValidator.ValidateStrict(chain)
+	err = r.pkiValidator.ValidateStrict(verifiedChains[0])
```
gerardsn, line 141: this only works if the leaf cert itself is revoked. I think the entire chain should be checked to make sure intermediates are not revoked?

gerardsn (changes requested):

vdr/didx509/resolver_test.go
```diff
@@ -227,6 +229,48 @@ func TestManager_Resolve_OtherName(t *testing.T) {
 		assert.ErrorIs(t, err, ErrNoMatchingHeaderCredentials)
 		metadata.JwtProtectedHeaders[X509CertThumbprintS256Header] = sha256Sum(signingCert.Raw)
 	})
+	t.Run("invalid signature of root certificate", func(t *testing.T) {
+		t.Skip("Can't test this right now, enable after https://github.com/nuts-foundation/nuts-node/issues/3587 has been fixed")
```
gerardsn, line 233: this PR is claiming to close referenced issue, so test should work?

vcr/verifier/signature_verifier_test.go
reinkrul: I wanted to keep that out of scope of this issue, just simplified the test a bit.

woutslakhorst: it MUST be base64 DER encoding. Also removes that newline replaces.

reinkrul: Yes, but I don't want to fix all issues in 1 big PR, to keep things reviewable and so we can collaborate on it?

woutslakhorst: Then an issue needs to be created stating `parseChain` in x509_utils is parsing PEM instead of DER.

reinkrul: Ah, we only had one for the issuer-tool. I created one for Nuts node as well: https://github.com/nuts-foundation/nuts-node/issues/3596

vdr/didx509/resolver.go
reinkrul: nah, validateCert verifies the whole chain. But `verifiedChains` is defined as `verifiedChains [][]x509.Certificate` (it's a slice of chains)

vdr/didx509/resolver_test.go
reinkrul: Good that you saw this! I referenced the wrong issue, fixed now (it's actually #3588).

gerardsn (approved):

PR #3607, opened 2024-12-11
did:x509: improved ca-fingerprint validation

### LINKED ISSUE #3587: did:x509 issuer and resolver assume the root CA as ca-fingerprint

reinkrul:
Fixes #3587 for the parsing in the Nuts node. Issuer tool will still have to be fixed.

It was actually already supported, just the field name was incorrect. But, it was missing a validation check; ca-fingerprint is not allowed to reference the leaf certificate (intermediate CA or root CA only).

### DIFF
--- a/vdr/didx509/resolver.go
+++ b/vdr/didx509/resolver.go
@@ -77,11 +77,14 @@ type X509DidPolicy struct {
 	Value string
 }
 
-// X509DidReference represents a reference for an X.509 Decentralized Identifier (DID) including method, root certificate, and policies.
+// X509DidReference represents a reference for an X.509 Decentralized Identifier (DID).
 type X509DidReference struct {
-	Method        HashAlgorithm
+	// Method specifies the hash algorithm that was used to generate CAFingerprint from the raw DER bytes of the CA certificate.
+	Method HashAlgorithm
+	// CAFingerprint is the fingerprint of the CA certificate.
 	CAFingerprint string
-	Policies      []X509DidPolicy
+	// Policies contain the fields that are included in the did:x509, which must be validated against the certificates.
+	Policies []X509DidPolicy
 }
 
 // Resolve resolves a DID document given its identifier and corresponding metadata.
--- a/vdr/didx509/resolver.go
+++ b/vdr/didx509/resolver.go
@@ -117,7 +117,7 @@ func (r Resolver) Resolve(id did.DID, metadata *resolver.ResolveMetadata) (*did.
 		return nil, nil, err
 	}
 	if bytes.Equal(caFingerprintCert.Raw, validationCert.Raw) {
-		return nil, nil, fmt.Errorf("did:509 ca-fingerprint refers to leaf certificate, must be either root or intermediate CA certificate")
+		return nil, nil, fmt.Errorf("did:x509 ca-fingerprint refers to leaf certificate, must be either root or intermediate CA certificate")
 	}
 
 	// Validate certificate chain, checking signatures and whether the chain is complete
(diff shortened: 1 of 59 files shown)


gerardsn:
`make install-tools && make gen-mocks`

vdr/didx509/resolver.go
```diff
@@ -107,14 +108,17 @@ func (r Resolver) Resolve(id did.DID, metadata *resolver.ResolveMetadata) (*did.
...
+	caFingerprintCert, err := findCertificateByHash(chain, ref.CAFingerprint, ref.Method)
 	if err != nil {
 		return nil, nil, err
 	}
 	validationCert, err := findValidationCertificate(metadata, chain)
 	if err != nil {
 		return nil, nil, err
 	}
+	if bytes.Equal(caFingerprintCert.Raw, validationCert.Raw) {
+		return nil, nil, fmt.Errorf("did:509 ca-fingerprint refers to leaf certificate, must be either root or intermediate CA certificate")
```
gerardsn, line 120:
```suggestion
		return nil, nil, fmt.Errorf("did:x509 ca-fingerprint refers to leaf certificate, must be either root or intermediate CA certificate")
```

vdr/didx509/resolver.go
```diff
@@ -107,14 +108,17 @@ func (r Resolver) Resolve(id did.DID, metadata *resolver.ResolveMetadata) (*did.
...
+	}
```
gerardsn, line 121: Is this requirement true for all did:x509 resolving, or specifically for our use-case?
If so, I think this means the policy file is incorrect: Can we detect that once somewhere else instead of during every resolving action?
(I can't find in our documentation which did:x509 spec we are using so I couldn't look it up)

gerardsn (changes requested):

vdr/didx509/resolver.go
```diff
@@ -78,9 +79,9 @@ type X509DidPolicy struct {
 
 // X509DidReference represents a reference for an X.509 Decentralized Identifier (DID) including method, root certificate, and policies.
 type X509DidReference struct {
-	Method      HashAlgorithm
-	RootCertRef string
-	Policies    []X509DidPolicy
+	Method        HashAlgorithm
+	CAFingerprint string
```
gerardsn, line 83: update godoc

vdr/didx509/resolver.go
reinkrul: https://trustoverip.github.io/tswg-did-x509-method-specification/#identifier-syntax states:

> ca-fingerprint is chain[i].fingerprint[ca-fingerprint-alg] with i > 0, that is, either an intermediate or root CA certificate.

Although it's kinda cryptic, given x5c is sorted leaf-to-root, I think what they mean (twice) is that it shouldn't be the leaf, but a CA cert.

gerardsn (approved):

Issue #3629, opened 2024-12-19
Scheduled CodeQL action fails for non-master branches


gerardsn:
see https://github.com/nuts-foundation/nuts-node/actions/runs/12401906128

Same for V5.4 branch:
> [Analyze (V6.0)](https://github.com/nuts-foundation/nuts-node/actions/runs/12401906128/job/34622275716#step:4:60)
Unable to validate code scanning workflow: error: getWorkflow() failed: Error: Expected to find a code scanning workflow file at /home/runner/work/nuts-node/nuts-node/.github/workflows/codeql-analysis-cron-schedule.yml, but no such file existed. This can happen if the currently running workflow checks out a branch that doesn't contain the corresponding workflow file.

Copying the file sounds like an easy fix, but the action should fail if it cannot run successfully.

---

action was added in #3522 with several other concerns

---

Proposed solution:
move to govulncheck and 

TODO:
- [ ] add documentation on release procedure for new major/minor releases (add branch to actions since pattern matching doesn't work) #3656
- [x] add `govulncheck` as a scheduled action #3642, #3645, #3646
- [x] add slack notification to the action if it fails #3652
- [ ] check if CodeQL can still be useful -> leaving it as is for now

Issue #3522, opened 2024-10-28
Periodically scan for vulnerabilities on all supported release branches


gerardsn:
We currently scan for vulnerabilities using `CodeQL` and `govulncheck` on the `master` and `V*` branches when a PR is opened/merged into these branches.

We should periodically scan for vulnerabilities on the same branches (or at least supported versions). I can't find a reliable way to [schedule github actions](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#schedule) for other branches because:
- ~each branch requires its own github action file since you manually need to specify the branch if it is not the main branch.~ using a matrix for branches in `actions/checkout` allows using 1 file. This means we need to manually add a new action for every released minor version.
- schedules are automatically deactivated after 60 days of inactivity on the repo. This is probably fine, but the schedule is not automatically reactivated if the repo has new activity. 
- schedules are deactivated if the user that made the last change to the 'cron' is removed from the organization. To reactivate the schedule, a change should be made to the 'cron' schedule (a change to the Github action file is insufficient).
- schedule outcomes are only sent to the user that last modified the 'cron' value

So the main concern is: scheduled actions are easily deactivated without any warning, and it is hard to (re-)activate them.

gerardsn:
CodeQL is now scheduled to run twice per week

gerardsn:
remaining concerns are limitations of the ecosystem and are accepted as is.

PR #3630, opened 2024-12-19
Add scheduled CodeQL action to v6.0

gerardsn:
part of #3629

### DIFF
--- a/.github/workflows/codeql-analysis-cron-schedule.yml
+++ b/.github/workflows/codeql-analysis-cron-schedule.yml
@@ -1,0 +1,69 @@
+# This is an alternative to the codeql-analysis.yml that only contains a scheduled evaluation of CodeQL
+# The action runs for all branches defined in jobs.analyze.strategy.matrix.branches.
+# Every new production branch (minor release branches) should be added to this list.
+
+name: "Scheduled CodeQL"
+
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql-analysis-cron-schedule.yml
```diff
@@ -0,0 +1,69 @@
...
+      security-events: write
+
+    strategy:
+      fail-fast: false
+      matrix:
+        # CodeQL runs on these branches
+        branches:
+          - 'master'
+          - 'V5.4'
+          - 'V6.0'
```
reinkrul, line 28: Why not "V6*"? And hopefully the future us don't forget to add V7... (THey will)

reinkrul (changes requested):

.github/workflows/codeql-analysis-cron-schedule.yml
gerardsn: pattern matching doesn't work here

gerardsn: so yes, this will have to be added to every future minor/major release...

gerardsn: see #3522

reinkrul (approved):

PR #3676, opened 2025-02-04
change system requirements for V6

woutslakhorst:
backport to V6 branch.

The v5 version will get an additional remark on CPU usage when using AuthCreds.

PR #3677, opened 2025-02-04
change on system requirements

woutslakhorst:


docs/_static/auth/v2-iam-combined.tmp.yaml
gerardsn, line 1: remove file

gerardsn (approved):
+ same comments as #3676

reinkrul (approved):

PR #3685, opened 2025-02-07
Backport v6.0: upgrade min go version to 1.23.6 to fix std lib crypto vulnerabilities

gerardsn:


Dockerfile
```diff
@@ -1,5 +1,5 @@
 # golang alpine
-FROM golang:1.23.5-alpine AS builder
+FROM golang:1.23.6-alpine as builder
```
gerardsn, line 2:
```suggestion
FROM golang:1.23.6-alpine AS builder
```

woutslakhorst (approved):

PR #3693, opened 2025-02-11
fix expires_in in cached tokenResult

### LINKED ISSUE #3688: Cached access token result has wrong expires_in
It's always 900 and to the remainder of the token validity.

woutslakhorst:
Fixes #3688


auth/api/iam/api.go
```diff
@@ -760,9 +763,10 @@ func (r Wrapper) RequestServiceAccessToken(ctx context.Context, request RequestS
 	if tokenResult.ExpiresIn != nil {
 		ttl = time.Second * time.Duration(*tokenResult.ExpiresIn)
 	}
+	expiresAt := time.Now().Add(ttl)
```
reinkrul, line 766: can't we just add expires_at to the API spec?

reinkrul (approved):

auth/api/iam/api.go
woutslakhorst: yeah why not.

reinkrul (approved):

PR #3714, opened 2025-02-26
fix holder credential autocorrect credentialsubject.id

### LINKED ISSUE #3696: VP builder should set correct credentialSubject.id on self-asserted credentials
for subjects with multiple DIDs an access token request can result in:
```
{ "detail": "invalid_request - not all VCs have the same credentialSubject.id", "status": 400, "title": "RequestServiceAccessToken failed" }
```

to reproduce with scope `test:HuisartsVVTInzage` that requires a `X509Credential` issued to a `did:web` and the `EmployeeCredential` (for a subject that also has a `did:nuts`):
```
POST https://example.com/internal/auth/v2/{serviceID}/request-service-access-token
Content-Type: application/json

{
  "authorization_server": "https://example.com/oauth2/{serviceID}",
  "scope": "test:HuisartsVVTInzage",
  "credentials": [
    {
      "@context": [
        "https://www.w3.org/2018/credentials/v1",
        "https://nuts.nl/credentials/v1"
      ],
      "type": [
        "VerifiableCredential",
        "EmployeeCredential"
      ],
      "credentialSubject": {
        "name": "John Doe",
        "roleName": "Janitor",
        "identifier": "123456"
      }
    }
  ],
  "token_type": "Bearer"
}
```

Manually setting the `credentialSubject.id` to the `did:web` fixes the problem.

I expect the node to set the correct subject on the `EmployeeCredential`

woutslakhorst:
fixes #3696 

backport to V6.0

### DIFF
--- a/vcr/credential/util.go
+++ b/vcr/credential/util.go
@@ -136,6 +136,7 @@ func AutoCorrectSelfAttestedCredential(credential vc.VerifiableCredential, reque
 		}
 		if _, ok := credentialSubject[0]["id"]; !ok {
 			credentialSubject[0]["id"] = requester.String()
+			credential.CredentialSubject = make([]interface{}, 1)
 			credential.CredentialSubject[0] = credentialSubject[0]
 		}
 	}
--- a/docs/pages/release_notes.rst
+++ b/docs/pages/release_notes.rst
@@ -3,6 +3,16 @@ Release notes
 #############
 
 ***************
+Peanut (v6.0.9)
+***************
+
+Release date: 2025-02-26
+
+- `#3696 <https://github.com/nuts-foundation/nuts-node/issues/3696>`_: Fix holder credential usage when a subject has multiple DIDs.
+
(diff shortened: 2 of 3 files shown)


reinkrul (approved):

vcr/credential/util.go
```diff
@@ -136,6 +136,7 @@ func AutoCorrectSelfAttestedCredential(credential vc.VerifiableCredential, reque
 		}
 		if _, ok := credentialSubject[0]["id"]; !ok {
 			credentialSubject[0]["id"] = requester.String()
+			credential.CredentialSubject = make([]interface{}, 1)
```
gerardsn, line 139: this is only a partial fix. If `didmetnods`config is `[web,nuts]` and the `credentialSubject.id` is already set by the user to the `did:nuts` address, it will still create a wallet with multiple DIDs as subject

gerardsn (dismissed):

vcr/credential/util.go
woutslakhorst: that was the idea: 3717

Issue #3719, opened 2025-02-28
Auth: allow requesting a fresh, non-cached access token


reinkrul:
When the resource server doesn't accept the access token, the client might want to use a fresh access token to retry the request. However, since the requesting Nuts node always returns a cached response if there is one, this isn't possible without restarting the Nuts node.

I propose introducing usage of `Cache-Control: no-cache` as HTTP request header, so so the client can bypass the cache.

See:
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control#request_directives 
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control#up-to-date_contents_always

PR #3720, opened 2025-02-28
Auth: allow client-side access token cache to be bypassed

reinkrul:
Fixes https://github.com/nuts-foundation/nuts-node/issues/3719


gerardsn (approved):
code changes are fine, but documentation is key here

woutslakhorst (approved):

PR #3722, opened 2025-02-28
Upgrade V5.4 to go 1.24

gerardsn:
added some more fixes relating to https://github.com/golang/go/issues/60529


reinkrul (approved):

gerardsn:
after also updating `github.com/chromedp/chromedp` to v0.13.0 this works when using the local chrome browser, but the headless test still fails

gerardsn:
The test doesn't work anymore for chrome versions > 133 (134 and 135 don't work). Pinning the version to 133.0.6943.142 seems to solve the issue. Not sure how to proceed.

gerardsn:
apparently V6.0+ is already pinned to chrome 125. Probably fine to merge this then

PR #3724, opened 2025-03-04
V5.4: renew expired test cert

gerardsn:
v6.0 and up will expire in december


gerardsn:
now part of #3722 since they depend on each other

PR #3732, opened 2025-03-11
Reject self-asserted credentials with an issuer or credentialSubject.id

### LINKED ISSUE #3717: Passed holder credentials may not contain credentialSubject.id
Due to the way the autocorrect feature currently works: it adds issuer and credentialSubject if not exists, a VC passed with credentialSubject.id will break.

holder credentials are a VC data model 2.0 feature.

I suggest that we disallow setting credentialSubject.id for holder credentials for now. So any call to request access token with a VC with credentialSubject.id should fail immediately.

gerardsn:
fixes #3717

### DIFF
--- a/auth/api/iam/api.go
+++ b/auth/api/iam/api.go
@@ -750,6 +750,24 @@ func (r Wrapper) RequestServiceAccessToken(ctx context.Context, request RequestS
 	if request.Body.Credentials != nil {
 		credentials = *request.Body.Credentials
 	}
+	// assert that self-asserted credentials do not contain an issuer or credentialSubject.id. These values must be set
+	// by the nuts-node to build the correct wallet for a DID. See https://github.com/nuts-foundation/nuts-node/issues/3696
+	// As a sideeffect it is no longer possible to pass signed credentials to this API.
+	for _, cred := range credentials {
+		var credentialSubject []map[string]interface{}
+		if err := cred.UnmarshalCredentialSubject(&credentialSubject); err != nil {
+			// extremely unlikely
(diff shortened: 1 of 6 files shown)


auth/api/iam/api.go
```diff
@@ -750,6 +750,24 @@ func (r Wrapper) RequestServiceAccessToken(ctx context.Context, request RequestS
...
 		credentials = *request.Body.Credentials
 	}
+	// assert that self-asserted credentials do not contain an issuer or credentialSubject.id. These values must be set
+	// by the nuts-node to build the correct wallet for a DID. See https://github.com/nuts-foundation/nuts-node/issues/3696
+	// As a sideeffect it is no longer possible to pass signed credentials to this API.
+	for _, cred := range credentials {
+		var credentialSubject []map[string]interface{}
+		if err := cred.UnmarshalCredentialSubject(&credentialSubject); err != nil {
+			// extremely unlikely
+			return nil, core.InvalidInputError("failed to parse credentialSubject.id: %w", err)
```
reinkrul, line 760: Not really needed to unmarshal the credential subject here into a map, since it's most probably already a slice of maps. If it's not, we don't need to check it since it's invalid anyways (e.g. if `credentialSubject` is a string or boolean).

So we could actually optimize this in go-did (check if we actually need to unmarshal, or maybe we can do some type checking and casting), or check types here. (for less JSON unmarshalling).

gerardsn: so where should this be fixed? Because the same logic is used in more places in the nuts-node. It does help during testing when the concrete type is set to something other than []map[string]interface{}

PR #3734, opened 2025-03-15
Replace JSON stdlib with Sonnet for better performance

reinkrul:


woutslakhorst:
I would prefer to remove (un)marshalling steps where not needed in the code before changing the library.

gerardsn:
If we are switching the library I'd also prefer to use json/v2. It's not the fastest, but comparable with the others and I prefer using the std lib over external deps. Or is there a reason to switch to a different lib?

reinkrul:
> If we are switching the library I'd also prefer to use json/v2. It's not the fastest, but comparable with the others and I prefer using the std lib over external deps. Or is there a reason to switch to a different lib?

It's unclear what happens to `json/v2`, they stated an outcome might be that they discontinue it if it doesn't provide the expected benefits. Although json/v2 is preferred if it's ultimately adopted in the stdlib. Although sonnet is faster, according to their own benchmarks.

But, I just wanted to test whether it's feasible to replace it and what the performance gains are. But I agree with Wout that we should simplify things first (less marshalling).

reinkrul:
Test `e2e-tests/nuts-network/private-transactions` is performed using https://github.com/nuts-foundation/nuts-node/pull/3753 on my local system. This test performs 40 iterations, and it sleeps 1 second in each iteration. So 40 seconds needs to be retracted for wait time.

- Baseline (`20bdc8f4725ff1073b93fec20aa2f1ea5345a0e4`) completes in 56~57 seconds
- This PR (`d8d226a3b18fb873a49201adc1db040d90115601`) completes in 57 seconds

No measurable benefit.

reinkrul:
Stale.

PR #3738, opened 2025-03-18
Cache loaded JSON-LD contexts

reinkrul:


woutslakhorst:
note: this only caches the file. The context will be parsed everytime.

reinkrul:
> note: this only caches the file. The context will be parsed every time.

Yes, although it does ache the unmarshalled JSON result. But indeed, the JSON-LD context will still be parsed (from this RemoteDocument, being a `map[string]any`) again and again.

reinkrul:
Stale.

Issue #3663, opened 2025-01-20
General performance issue


woutslakhorst:
While looking into some performance challenges I noticed some slow operations or at least operations that can cause peak load.

- JSON-LD processing is really slow. The json-gold library creates the context and validates VCs every time. The validating step does a lot of regexes.
- we do a lot of json (un)marshalling. For VC/JSON-LD processing, DID resolving and internals where we marshal/unmarshal to/from interface{}
- the memory allocations for json processing also causes GC cycles.

There are some quick fixes:
- cache DID resolve after unmarshal
- fork json-gold and disable some validation checks
- check functions/methods on `interface{}` arguments and remove some of the marshal/unmarshal operations where possible.

reinkrul:
I don't think we process DID documents as JSON-LD?

woutslakhorst:
> I don't think we process DID documents as JSON-LD? And we should cache did:web documents in the HTTP client, meaning if there's improvements to be made there, it's in the JSON (unmarshalling).?

yes did resolving performance is mainly about json unmarshal. Indeed nothing to do with JSON-LD

woutslakhorst:
For test I requested 100 access tokens in parallel every second for 60 seconds.

First is baseline.
Second has json-ld validating disabled
third has less marshalling

![Image](https://github.com/user-attachments/assets/1935119f-5f9e-4e41-b5a7-28b9d5d9a2e5)
![Image](https://github.com/user-attachments/assets/b12818ee-a319-415d-9955-dc86a6c32ba2)
![Image](https://github.com/user-attachments/assets/6fa97c93-06b8-415c-a031-9348f494017d)

reinkrul:
> `encoding/json` (un)marshaling is slow, which is why there are a bunch of different implementations. There is an ongoing [discussion](https://github.com/golang/go/discussions/63397) to replace the current implementation. There is an [experimental implementation](https://github.com/go-json-experiment/json) for `encoding/json/v2` from the go team? that also compares some [alternatives with benchmarks](https://github.com/go-json-experiment/jsonbench). We could also consider using one of the alternatives to reduce JSON processing times if that makes sense. Can you post some of the results so we can make a better estimate of where to improve performance?
> 
[...]

PR #3739, opened 2025-03-19
#3663: change credentialsubject to map, use casting instead of marshalling where possible

reinkrul:
This lower the amount of (now unnecessary) casting and JSON unmarshalling we do.

TODO:
- [ ] Release go-did if PR is OK


vcr/credential/util.go
```diff
@@ -160,22 +159,10 @@ outer:
...
-			if b.ID != "" {
-				// check credentialSubject
-				subjectDID, err := did.ParseDID(b.ID)
-				if err == nil {
-					if !slices.Contains(didMethods, subjectDID.Method) {
-						continue outer
-					}
-				}
-			}
-		}
```
gerardsn, line 178: this used to require only one subjectDID to be valid, but `credential.SubjectDID()` now requires all to be valid and have the same value.

reinkrul:
Baseline (`20bdc8f4725ff1073b93fec20aa2f1ea5345a0e4`) using https://github.com/nuts-foundation/nuts-node/pull/3753 on my local system is 57 seconds

woutslakhorst (commented):
Slightly better code, but unfortunately no real gain.

gerardsn:
I still cannot make any conclusions based on these results. Subtracting the wait time and looking at the difference based on the information I have the gain could ~6% (=1-16/17). 
This excludes several important factors:
- Time lost in communicating to external DBs
- Tests do not run in parallel, the condition that triggers issues in practice.
- Tests are performed on hardware that does not have the same resource constraints that typical production node runs in.

woutslakhorst:
> and pending performance test of course

even without performance gains, it cleans up the code (a bit)?

gerardsn:
> > and pending performance test of course
> 
> even without performance gains, it cleans up the code (a bit)?

Agreed, but other comments above still stand.

reinkrul:
Superseded by #3811

PR #3764, opened 2025-04-01
Expose /status on public HTTP interface as well

### LINKED ISSUE #3755: Azure Application Gateway probing causes a lot of access logging spam
Azure Application Gateway probes the upstream server for availability every X seconds. Since the Nuts node's internal HTTP interface isn't exposed outside the k8s/Container Apps cluster (because it's internal), Azure Application Gateway can't use the internal HTTP interface's `/status` endpoint.

This means it uses the Nuts node's public HTTP interfaces landing page (`/`), which is then logged in the access log every X seconds.

Options I came up with:

- Don't log calls to the landing page (`/`)
- Add `/status` to the public HTTP interface as well (it only serves a static response; `OK`).
- Expose the internal HTTP interface (but this increases risk, especially as there's no auth/authz)

reinkrul:
Fixes #3755

### DIFF
--- a/http/echo.go
+++ b/http/echo.go
@@ -83,10 +83,8 @@ type MultiEcho struct {
 	binds      map[string][]string
 }
 
-// Bind binds the given path (first part of the URL) to the given HTTP interface. Calling Bind for the same path twice
+// Bind binds the given path (first part of the URL) to the given HTTP interfaces. Calling Bind for the same path twice
 // results in an error being returned.
-// If address wasn't used for another bind and thus leads to creating a new Echo server, it returns true.
-// If an existing Echo server is returned, it returns false.
 func (c *MultiEcho) Bind(path string, addresses []string, creatorFn func(ipHeader string) (EchoServer, error), ipHeader string) error {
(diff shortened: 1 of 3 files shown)


http/echo.go
```diff
@@ -65,16 +80,21 @@ type MultiEcho struct {
...
 }
 
 // Bind binds the given path (first part of the URL) to the given HTTP interface. Calling Bind for the same path twice
 // results in an error being returned.
 // If address wasn't used for another bind and thus leads to creating a new Echo server, it returns true.
 // If an existing Echo server is returned, it returns false.
-func (c *MultiEcho) Bind(path string, address string, creatorFn func(ipHeader string) (EchoServer, error), ipHeader string) error {
-	if len(address) == 0 {
-		return errors.New("empty address")
+func (c *MultiEcho) Bind(path string, addresses []string, creatorFn func(ipHeader string) (EchoServer, error), ipHeader string) error {
```
woutslakhorst, line 90: godoc is no longer correct.

woutslakhorst (approved):

PR #3802, opened 2025-04-25
Add index on subject with org credential

### LINKED ISSUE #3801: Missing index on credential collection
When the credential store contains a substantial amound of credentials (33k on stable for example), calls to `internal/auth/v1/request-access-token` and `internal/auth/v1/signature/verify` become slow. It seems that there is an index missing on the credential collection for subject, legalname and city.

Calls on my macbook m1 take around 250ms to finish.

stevenvegt:
fixes #3801 

Before and after:

Screenshot 2025-04-25 at 17 35 57

### DIFF
--- a/vcr/assets/assets/default.index.yaml
+++ b/vcr/assets/assets/default.index.yaml
@@ -14,16 +14,12 @@ indices:
             "http://schema.org/organization",
             "http://schema.org/legalname",
           ]
-        tokenizer: whitespaceOrExact
-        transformer: cologne
       - iriPath:
           [
             "https://www.w3.org/2018/credentials#credentialSubject",
             "http://schema.org/organization",
             "http://schema.org/city",
           ]
-        tokenizer: whitespaceOrExact
-        transformer: cologne
   - name: index_organization
     parts:
       - iriPath:
--- a/vcr/assets/assets/default.index.yaml
+++ b/vcr/assets/assets/default.index.yaml
@@ -5,20 +5,63 @@ indices:
   - name: index_issuer
     parts:
       - iriPath: ["https://www.w3.org/2018/credentials#issuer"]
-  - name: index_organization
+  - name: index_subject_organization
     parts:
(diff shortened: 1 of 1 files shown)


vcr/assets/assets/default.index.yaml
```diff
@@ -5,20 +5,63 @@ indices:
...
+        transformer: cologne
+      - iriPath:
+          [
+            "https://www.w3.org/2018/credentials#credentialSubject",
+            "http://schema.org/organization",
+            "http://schema.org/city",
+          ]
+        tokenizer: whitespaceOrExact
+        transformer: cologne
   - name: index_organization
```
reinkrul, line 27: this indexes the same fields but in reverse order. This one isn't used?

Can't we sort the search fields for the available indexes?

reinkrul:
Which is before and which one is after?

vcr/assets/assets/default.index.yaml
```diff
@@ -5,20 +5,63 @@ indices:
...
+            "http://schema.org/legalname",
+          ]
+        tokenizer: whitespaceOrExact
+        transformer: cologne
+      - iriPath:
+          [
+            "https://www.w3.org/2018/credentials#credentialSubject",
+            "http://schema.org/organization",
+            "http://schema.org/city",
+          ]
```
woutslakhorst, line 24: This index is for checking, so it should do an exact match meaning that both the tokenizer and transformer are not needed.

vcr/assets/assets/default.index.yaml
woutslakhorst: This one has an added `credentialSubject.id`. In the index this must be the first field since the other fields are queried with `not_nil`. The other index only has name and city. Leia uses the longest, best match index it can find. Query fields are sorted/matched to the index.

woutslakhorst (changes requested):

stevenvegt:
> Which is before and which one is after?

Are you referring to the screenshot? The faster and most recent one is from the request after adding the index.

This test was performed on the stable network with both requester and authorizer on the same node. The node runs on my local Macbook M1.

PR #3885, opened 2025-07-26
Move Makefile install-tools to go mod tools

confiks:


gerardsn (commented):
I like the idea of adding the tools to the dependencies since it makes managing tool versions a bit easier and more transparent to users. However, this adds ~650 lines to go.mod/go.sum and nearly doubles their sizes. One of the reasons oapi-codegen was split into a runtime and a generator repo was to remove this dependency bloat.

I consider the current list of dependencies (just go.mod) sort of manageable by hand in terms of deciding when/how to make changes. After this change the list will be too long for me to keep track of what everything is and what it is used for. But that is just an opinion and others may disagree / not care about this.

reinkrul (approved):
I understand and somewhat share the concern of @gerardsn but I think the trade-off tips in favor of having one less tool to worry about.

PR #3886, opened 2025-07-27
Upgrade dev-image Dockerfile to stable-slim

confiks:


gerardsn (approved):
probably fine. We need to reconsider if this is useful though. Every time I tried to use this I had to fix something since MS DevTunnel keeps changing.

PR #3924, opened 2025-09-07
feat: add extraEnv to helm chart values

dmeijboom:
This PR adds an `extraEnv` Helm value which can be used to set arbitrary environment values on the nuts-node deployment. This is often used when values need to be injected dynamically. My use-case is setting the `NUTS_STORAGE_SQL_CONNECTION` environment variable automatically from an automatically generated secret (required for: https://github.com/nuts-foundation/nuts-knooppunt/issues/111).

Although it's not part of the default Helm chart starter template it's a common property that is available in many other Helm charts.


rolandgroen (approved):
Minor change, looks good.

PR #3958, opened 2025-12-18
Bug: fix VP JWT marshalling flattening single-entry arrays incorrectly

### LINKED ISSUE #3957: Verifiable Presentation `type` is marshalled incorrectly in JWT format
It follows JSON-LD semantics; if there's 1 (which usually is the case, `Verifiable Presentation`), it marshals `type` as string instead of an array of strings.

Make sure the fix is backwards compatible.

reinkrul:
We used JSON-LD marshalling semantics for JWT VPs, which is incorrect; single-entry arrays shouldn't be flattened to singular values.

Found during LSPxNuts hackaton.

This applied to the following fields:
- `type`
- `verifiableCredential`
- `@context`

Fixes #3957 

TODO:
- [x] Switch to release version of go-did after https://github.com/nuts-foundation/go-did/pull/141 is merged
- [ ] Fix https://github.com/nuts-foundation/nuts-node/issues/4126 as well for improved Presentation Definition compatibility (current PDs might make assumptions about credentialSubject being an object or array)


stevenvegt (approved):
Assuming the issuedAt field is set at the callers, this is ok. Otherwise, add back an equivalent of "claims[jwt.NotBeforeKey] = time.Now().Unix()".
Also, this will conflict with #4191, merged it so we can update the code here.

PR #3968, opened 2026-01-14
Support JWT bearer grant type

### LINKED ISSUE #3965: Support JWT-bearer client assertions grant type in Nuts node
GF AuthN specifies JWT bearer grant type, instead of Nuts' `vp_token-bearer` grant type (see https://www.rfc-editor.org/rfc/rfc7523#section-2.1).

The GF also separates care organization authentication from client authentication. Client authentication can be used to hold info about the client itself, e.g.:
- this software is audited to do use case X (e.g. eOverdracht)
- this software/vendor may act on behalf of care organization Y

Changes in Nuts node:
- [ ] Support `urn:ietf:params:oauth:grant-
   type:jwt-bearer` grant type
  - [ ] Support it in the client (take inspiration from LSPxNuts PoC: https://github.com/nuts-foundation/nuts-node/commit/19f5960d443696e715cf944c5d66230e42f6ecc2)
  - [ ] Support it in the server
    - [ ] Add it to the OAuth2 Authorization Server metadata
    - [ ] Authenticate it when an access token is issued
    - [ ] Add support for `client` credentials requirements to Nuts Policy Definition (we currently support `user` and `organization`).

reinkrul:
Fixes https://github.com/nuts-foundation/nuts-node/issues/3965

## TODO

- [ ] **Bind `client_id` to the assertion signer.** `auth/api/iam/api.go:241-247` forwards `ClientId` verbatim into `handleJWTBearerTokenRequest`, landing in `access_token.go:91` (`accessToken.ClientId = clientID`) without checking it matches the VP signer DID. A client holding a valid VP for subject A could request a token claiming `client_id=B`. Fix: after `validatePresentationSigner` in `bearer_token.go`, enforce `clientID == subjectToBaseURL(signer)` or reject with `invalid_client`.
- [ ] **Add explicit `iat`-near-now check / regression test.** `validateS2SPresentationMaxValidity` (`bearer_token.go:180-196`) only checks `exp - iat ≤ 5s`. In practice `VerifyVP` + the 5s cap catches stale assertions, but a regression test for "old `iat`, `exp` just in future" is missing.


reinkrul:
We decided to do this in smaller PRs.

PR #3981, opened 2026-02-02
#3980: Support of DeziUserCredential

### LINKED ISSUE #3980: Support validation of DeziIDTokenCredential

reinkrul:
I have chosen to simply add an `id_token` field in the token request, and the Nuts node (OAuth client side) creates a DeziUserCredential from it. This is then added as extra credential, so the same could be achieved by the EHR by making the credential itself and passing it in the existing `credentials` field.

TODO:
- [ ] Check test coverage
- [ ] Wet DIAZ door de kamer

Closes #3980


reinkrul:
Closing for now: re-open when working on this again.

PR #3982, opened 2026-02-02
V6.1: Address GO-2026-4340

reinkrul:


reinkrul:
> Why new certs?

broken tests due to expired certs

PR #3987, opened 2026-02-05
#3986: Pass correct wallet DID in OpenID4VCI proofs

### LINKED ISSUE #3986: OpenID4VCI: proof for calling Token Endpoint uses random DID instead of specified `wallet_did`

reinkrul:
Fixes #3986

### DIFF
--- a/auth/api/iam/openid4vp_test.go
+++ b/auth/api/iam/openid4vp_test.go
@@ -41,7 +41,6 @@ import (
 )
 
 var holderDID = did.MustParseDID("did:web:example.com:iam:holder")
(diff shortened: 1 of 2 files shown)


auth/api/iam/openid4vp_test.go
```diff
@@ -41,6 +41,7 @@ import (
 )
 
 var holderDID = did.MustParseDID("did:web:example.com:iam:holder")
+var holderDIDAlt = did.MustParseDID("did:web:example.com:iam:holder_alt")
```
rolandgroen: This value seems unused.

PR #4026, opened 2026-02-13
Fix Dependabot config

reinkrul:
https://github.com/nuts-foundation/nuts-node/pull/3958/checks?check_run_id=63496742987

```
The property '#/updates/2/ignore/0/versions' includes invalid version requirements for a docker ignore condition
```

No guarantees


gerardsn (approved):
Why does dependabot run on a PR?

I don't think this is going to work since GO is violating the semver syntax.
It should be x.y.z-RC, but go does x.yRC.z. 

I'm fine merging this to prevent blocking other PRs

PR #4057, opened 2026-03-09
feat(openid4vci): migrate to v1.0 spec

### LINKED ISSUE #3953: Update OpenID4VCI implementation to spec v1.0
Version 1.0 is final now: https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html

JorisHeadease:
## Summary

Migrate the user/browser OpenID4VCI flow in `auth/iam` to OpenID4VCI v1.0 (ID-1).

The internal node-to-node issuance flow in `vcr/openid4vci` is deliberately left on its existing draft-11-inspired wire format. That flow is used in production for Nuts node-to-node issuance and is not a standards-conformant Wallet implementation per OpenID4VCI v1.0 §2. The package godoc documents the divergence and points readers to the new `auth/openid4vci` package for the v1.0 user/browser flow.

## v1.0 Spec Alignment

- Add an isolated `auth/openid4vci` package for v1.0 protocol types, error handling, issuer metadata lookup, Nonce Endpoint calls, and Credential Requests.
- Use the Credential Issuer Metadata document from `/.well-known/openid-credential-issuer`, including `credential_endpoint`, `nonce_endpoint`, and `authorization_servers`.
- Fetch a fresh nonce from the Nonce Endpoint before building the proof JWT.
- Retry the Credential Request once on `invalid_nonce` by fetching a fresh nonce.
- Send Credential Requests with `proofs.jwt` and either:
  - `credential_identifier` when the Token Response contains matching `authorization_details[].credential_identifiers`, or
  - `credential_configuration_id` when the Authorization Server did not return `authorization_details`.
- Parse Credential Responses using the v1.0 `credentials[]` wrapper and support both JSON string credentials (JWT/SD-JWT) and JSON object credentials (JSON-LD).
- Use the Credential Issuer Identifier (`credential_issuer`) as the proof JWT `aud`, not the Authorization Server issuer.

## Implementation Improvements

- Move OpenID4VCI-specific client logic out of the OpenID4VP/IAM client and into `auth/openid4vci`.
- Type `authorization_details` in the auth OpenAPI schema and require `type=openid_credential` plus `credential_configuration_id`.
- Validate Credential Issuer metadata by checking that `credential_issuer` matches the requested issuer identifier.
- Reject Credential Issuer Identifiers with query or fragment components.
- Validate outbound OpenID4VCI target URLs before HTTP calls to satisfy the current CodeQL SSRF finding; this is intentionally local for now and can move into the shared HTTP client in a follow-up.
- Reuse `oauth.OAuth2Error` for OpenID4VCI endpoint errors instead of introducing a duplicate error type.

## Scope Notes

- `vcr/openid4vci` stays on the existing internal draft-11-style flow. This PR only adds documentation clarifying that package's purpose and limitations.
- Batch/deferred issuance, `credential_offer_uri`, notification endpoints, and issuer-side v1.0 support are out of scope.
- The auth/iam flow stores only the first returned credential, matching the existing single-credential behavior.


reinkrul (changes requested):

auth/api/iam/openid4vci.go
```diff
@@ -142,31 +143,68 @@ func (r Wrapper) handleOpenID4VCICallback(ctx context.Context, authorizationCode
...
+		}
+	}
+	if len(credentialResponse.Credentials) == 0 {
+		return nil, withCallbackURI(oauthError(oauth.ServerError, "credential response does not contain any credentials"), appCallbackURI)
+	}
+
+	// Per OpenID4VCI 1.0 §8.3, the credential field can be either a JSON string
+	// (JWT-VC, SD-JWT-VC) or a JSON object (JSON-LD). Try unmarshalling as a
+	// string first; if that fails, treat the raw bytes as the JSON-LD object.
+	rawCredential := credentialResponse.Credentials[0].Credential
```
reinkrul, line 206: `ParseVerifiableCredential` handles both forms

auth/client/iam/client.go
```diff
@@ -241,19 +240,6 @@ func (hb HTTPClient) PostAuthorizationResponse(ctx context.Context, vp vc.Verifi
 	return hb.postFormExpectRedirect(ctx, data, verifierResponseURI)
 }
 
-func (hb HTTPClient) OpenIdCredentialIssuerMetadata(ctx context.Context, oauthIssuerURI string) (*oauth.OpenIDCredentialIssuerMetadata, error) {
```
reinkrul, line 244: good to separate them!

auth/openid4vci/client.go
```diff
@@ -0,0 +1,225 @@
...
+
+type client struct {
+	httpClient core.HTTPRequestDoer
+	strictMode bool
+}
+
+// validateURL guards against SSRF by rejecting target URLs that fail
+// core.ParsePublicURL (in strict mode: HTTPS only, no IP/reserved hosts).
+// Called at the entry of every method that makes outbound HTTP.
+func (c *client) validateURL(name, target string) error {
```
reinkrul, line 103: I'm not sure whether we should do that here, but rather in the HTTPClient itself. I'd keep it out of scope.

auth/openid4vci/error.go
```diff
@@ -0,0 +1,46 @@
...
+const (
+	// InvalidNonce means at least one of the key proofs in the Credential
+	// Request contained an invalid c_nonce. Per §8.3.1.2 the wallet should
+	// retrieve a new c_nonce from the Nonce Endpoint (§7) and may retry.
+	InvalidNonce ErrorCode = "invalid_nonce"
+)
+
+// Error is a wire-format error returned by an OpenID4VCI endpoint.
+// Specified by §6.4 and §8.3.
+type Error struct {
```
reinkrul: we already have `oauth.OAuthError`, can't we reuse that one?

reinkrul (changes requested):

auth/api/iam/openid4vci.go
JorisHeadease: Clarified the comment.

`ParseVerifiableCredential` handles `JSON-LD` object strings and raw JWT strings, but not a JSON-encoded JWT string literal.

Per OpenID4VCI §8.3, the credential field may be a string or an object; for `JWT/SD-JWT` credentials the wire shape is `{"credentials":[{"credential":"eyJ..."}]}`. With credential typed as `json.RawMessage`, the field keeps the raw JSON encoding, so for a JWT `string(raw)` is `"\"eyJ...\""` including the surrounding quotes. Passing that to `ParseVerifiableCredential` fails because it tries to parse the quoted value as a JWT.

The try-unmarshal-as-string-with-fallback strips the JSON quotes for JWT/string credentials and leaves `JSON-LD` object credentials intact.

auth/openid4vci/client.go
JorisHeadease: I agree 👍 Added a TODO that I'll address in another PR. Leaving the logic in for now to keep CodeQL happy.

auth/openid4vci/client.go
```diff
@@ -0,0 +1,239 @@
...
+			JWT: []string{opts.ProofJWT},
+		},
+	}
+	// Per §8.2: CredentialIdentifier and CredentialConfigurationID are mutually
+	// exclusive. CredentialIdentifier wins when set.
+	if opts.CredentialIdentifier != "" {
+		body.CredentialIdentifier = opts.CredentialIdentifier
+	} else {
+		body.CredentialConfigurationID = opts.CredentialConfigurationID
+	}
```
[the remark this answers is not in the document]
JorisHeadease: The Client is package-public but the only caller today is the internal `auth/api/iam` handler, which constructs `RequestCredentialOpts` from validated session state plus a freshly-built proof JWT.

reinkrul (approved):

reinkrul:
Can you add OpenID4VCI to these docs: https://nuts-node.readthedocs.io/en/stable/pages/deployment/oauth.html

reinkrul (approved):

PR #4058, opened 2026-03-11
feat(openid4vci): client compliance (issue #3152)

### LINKED ISSUE #3152: Align OpenID4VCI client with implementers draft
An Implementers Draft has been released for [OpenID4VCI](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html). This is the version that will be supported for a longer time by implementing parties for interoperability and is in the DIIP profile. We should align our implementation with this version of the spec. Since the Nuts node only implement the client side for OpenID4VCI this is even more relevant.

The summary below contains the points where the Nuts node does not comply with the spec or where additional features could simplify things. This should be fixed when we have implemented and OpenID4VCI issuer, or see if we can collaborate with a party that has a DIIP compliant issuer.

---
# 5 Authorization Endpoint
Recommended to use PKCE (implemented) and PAR (not implemented. mostly useful for applications (mobile wallets) without a domain to host the `request_uri` of a JAR).  
Credentials can be requested using `authorization_details` and `scope`s.

## 5.1.1 request credentials using `authorization_details` RFC9396
Only supported option. Supported credentials are listed in the credential issuer metadata. We currently require the `authorization_details` to be provided by the user, meaning that the user needs to figure out the details using some out of bound method. It would be better to only provide the credential issuer and then present the user with available options, but this works for now.
- user input is not validated
- credential issuer metadata lists `authorization_servers`: "the authorization detail's locations common data field MUST be set to the Credential Issuer Identifier value."
- should check the credential issuer metadata if the supported methods match the capabilities of the nuts node and requester/holder. 
- `authorization_details` is currently sent as query param in the request. Browser redirects may fail if it contains too many chars. JAR/PAR may be needed.

## 5.1.2 request credentials using `scope`s
Not implemented, but is the simplest option. This may require a change to the API we provide for credential requests.

## 5.1.4 Pushed Authorization requests
Recommended, but not implemented.

# 6 Token Endpoint
For the `authorized_code` flow this the request is the default per RFC6749
- Should this use DPoP?

## 6.2 Token response
RFC6749 + some extra response params:
- missing `authorization_details`: REQUIRED when `authorization_details` parameter is used to request issuance of a certain Credential type as defined in Section 5.1.1
- `credential_identifiers`: using this makes for a simpler credential request when using `authorization_details`, but is OPTIONAL.

# 7 Credential Endpoint
- Should this use DPoP?

## 7.2 credential request
- missing some required parameters (`format` or `credential_identifier` MUST be present)
- credential issuers metadata `proof_types_supported` (we only support `jwt`) and `proof_signing_alg_values_supported` are not checked.
- `proof` issues
	- header `typ` MUST be `openid4vci-proof+jwt` (currently `jwt`)
	- missing REQUIRED `iat`
	- contains non-specced `jti`
	- `aud` == credential issuer identifier: "11.2.1 A Credential Issuer is identified by a case sensitive URL using the `https` scheme that contains scheme, host and, optionally, port number and path components, but no query or fragment components." Currently contains a DID, should be did:web URL.

## 7.3 Credential Response
- may contain `transaction_id` if issuer is using deferred flow. This is currently not supported, but should be detected and return an appropriate error.

# 11 Metadata

## 11.2 credential issuer metadata: `/.well-known/openid-credential-issuer`
For correct interpretation of the issuers's capabilities and intentions we need to add the following parameters
- `credential_identifiers_supported`, but only if we want to use this in §6.2
- `signed_metadata` OPTIONAL just for convenience as it eliminates one call in the flow if present
- `credential_configurations_supported` which is REQUIRED and shows the capabilities of the issuer

## 11.3 authorization server metadata: `/.well-known/authorization-server`
- `pre-authorized_grant_anonymous_access_supported==true` is incorrectly added to our metadata since we do not support `pre-authorized_code` flow

**4, 8-10 Credential Offer, Batch, Deferred, and Notification Endpoints are not supported**

JorisHeadease:
## Summary

Client-side OpenID4VCI v1.0 compliance improvements per issue #3152, building on the wire format migration in #4057.

### Planned commits

- [x] Validate `authorization_details` against issuer metadata (Section 5.1.1)
- [x] Validate `proof_signing_alg_values_supported` (Appendix F.1)
- [x] Detect deferred credential issuance (Section 7.3)
- [x] PAR support (Section 5.1.4)
- [x] `credential_identifiers` in token response (Section 6.2)
- [x] Scope-based credential requests (Section 5.1.2)
- [x] `signed_metadata` (Section 11.2)

### Skipped
- **DPoP (Section 6 + 7)** The credential request's proof JWT (Section 8.2) already covers the same threat surface: it's signed by the wallet's DID key, verified against the DID document, and includes a nonce. A stolen access token alone is useless without the wallet's private key. DPoP would add a second PoP layer with no practical security gain. Worth revisiting if an external issuer requires it.

- **`credential_identifiers_supported`** Not part of the v1 spec?

- **11.3 `pre-authorized_grant_anonymous_access_supported`** The issue notes this is "incorrectly added to our metadata since we do not support `pre-authorized_code` flow". This was already addressed in the v1 wire format migration (#4057). The field was restored because we do support the `pre-authorized` code flow on the issuer side.

- **Notification endpoint (Section 10)** Deferred, depends on credential lifecycle management.

Closes #3152


reinkrul:
I know we discussed to pick this up, but due to the time it's been lying around (I apologize), we don't want to merge this in this form;

- At time of development, there weren't any actual OpenID4VCI issuers, so no impact. Today, we have at least 1 issuer (AET SDK), meaning we can't simply change the implementation without proper analysis (what'll break/keep working).
- Too big to review, ideally we have a PR per item/feature
- We want to continue OpenID4VCI development change-request-driven, meaning if we run into something we miss (feature-wise), we implement/fix it. Chances are with this PR, we implement features we'll never use, leading to complexity we don't need.
- Given the state of AI, we want to introduce changes/feature by specifying it in a PRD first, have it reviewed, and then give the implementation task to the AI.

PR #4112, opened 2026-03-25
fix: nil pointer dereference DoS via invalid DPoP htu URL

reinkrul:
## Summary

- `crypto/dpop/dpop.go` — `url.Parse` returns `(nil, error)` for inputs with invalid percent-encoded sequences (e.g. `%zz`). `strip()` ignored the error and immediately wrote to the nil pointer, causing a server panic. Fixed by validating the `htu` claim in `Parse()` (reject early) and changing `strip()` to propagate the error through `Match()`.

### Attack vector (before fix)
An attacker with access to the internal API crafts a DPoP JWT with `htu: "%zz"`, signs it with their own key, and POSTs it to `POST /internal/auth/v2/dpop/validate`. `dpop.Parse()` accepted the token (the only `htu` guard was `!= ""`), and the subsequent call to `Match()` → `strip()` → `url.Parse("%zz")` → `nil` → **panic**. The connection was reset; the server survived due to `net/http`'s per-connection recover, but the endpoint could be degraded by repeated calls.

### DIFF
--- a/crypto/dpop/dpop.go
+++ b/crypto/dpop/dpop.go
@@ -251,7 +251,9 @@ func strip(raw string) (string, error) {
 		return "", err
 	}
 	u.Scheme = "https"
-	u.Host = strings.Split(u.Host, ":")[0]
+	if host := u.Hostname(); host != "" {
+		u.Host = host
+	}
 	u.RawQuery = ""
 	u.Fragment = ""
 	return u.String(), nil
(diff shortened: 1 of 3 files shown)


stevenvegt (changes requested):
I think we can now use the JWTProfiles for some of these checks right?

reinkrul:
Let's find out if this is still an issue.

reinkrul:
stale

PR #4163, opened 2026-04-09
Fix #4162: start notifiers in goroutine with retry to avoid BBolt lock contention

### LINKED ISSUE #4162: Network notifiers fail to start, causing node to shut down

```
```

Mitigations:

reinkrul:
## Summary

Partial fix for nuts-foundation/nuts-node#4162

- `connectToKnownNodes()` triggers transaction syncing from peers, which holds the BBolt write lock
- `notifier.Run()` needs a read lock and times out, causing `Start()` to return an error and the node to shut down
- Fix: run notifiers in a background goroutine (`startNotifiers`) that retries any failed notifier every 30 seconds until all have started successfully

## Test plan

- [x] `TestNetwork_startNotifiers/ok_-_succeeds_on_first_try`: notifier starts on the first attempt
- [x] `TestNetwork_startNotifiers/ok_-_retries_on_first_failure`: notifier fails once, is retried, and succeeds

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/network/network.go
+++ b/network/network.go
@@ -378,12 +378,31 @@ func (n *Network) Start() error {
 
 	// Resume all notifiers. Notifiers may access other components of the network stack.
 	// To prevent nil derefs run the notifiers last. https://github.com/nuts-foundation/nuts-node/issues/3155
-	for _, notifier := range n.state.Notifiers() {
-		if err = notifier.Run(); err != nil {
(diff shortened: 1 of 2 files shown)


reinkrul:
We chose to solve it using https://github.com/nuts-foundation/nuts-node/pull/4170

PR #4164, opened 2026-04-09
Fix #4162: limit concurrent outgoing TransactionListQuery conversations

### LINKED ISSUE #4162: Network notifiers fail to start, causing node to shut down

reinkrul:
## Summary

Second mitigation for nuts-foundation/nuts-node#4162.

- When many peers connect at startup, `sendTransactionListQuery`/`sendTransactionRangeQuery` is called for each peer that has missing transactions
- All peers respond simultaneously with `TransactionList` messages, causing a flood of BBolt writes
- Concurrent writes starve read operations (e.g. `notifier.Run()` during startup)

**Fix:** a token bucket rate-limiter shared by both `sendTransactionListQuery` and `sendTransactionRangeQuery`. Queries that find no token available are dropped — the gossip/state-comparison mechanism retries them naturally.

- The bucket (buffered channel) is pre-filled to capacity at startup, so queries are allowed immediately
- A goroutine refills one token per tick (`1s / rate`) up to the bucket capacity
- Default rate: **2/s** (configurable via `network.v2.transactionlistqueryrate`)
- Unlike a semaphore, a rate limiter is unaffected by slow or non-responsive peers

## Test plan

- [ ] `TestProtocol_sendTransactionListQuery/rate_limit_reached_-_query_is_dropped`: drains the bucket, verifies the next query is dropped without error and no conversation is left open
- [ ] `TestProtocolV2_Pagination`: integration test still passes
- [ ] All existing `network/...` tests pass

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/network/transport/v2/protocol.go
+++ b/network/transport/v2/protocol.go
@@ -54,18 +54,24 @@ type Config struct {
 	GossipInterval int `koanf:"gossipinterval"`
 	// DiagnosticsInterval specifies how often (in milliseconds) the node should broadcast its diagnostics message.
 	DiagnosticsInterval int `koanf:"diagnosticsinterval"`
+	// TransactionListQueryRate limits the rate (per second) at which outgoing TransactionListQuery and
+	// TransactionRangeQuery messages are sent. Limiting this reduces BBolt write lock contention during
+	// peer synchronization at startup. See https://github.com/nuts-foundation/nuts-node/issues/4162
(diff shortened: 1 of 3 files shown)


reinkrul:
Don't want to solve it this way, right now. See if we can improve performance first, before throttling.

PR #4168, opened 2026-04-09
Batch transaction writes in handleTransactionList to reduce fsync count

stevenvegt:
## Summary

Add `State.AddMany()` that writes multiple transactions from a `TransactionList` message in a single BBolt write transaction, requiring only one fsync instead of one per transaction.

### The problem

Previously, `handleTransactionList` called `state.Add()` for each transaction individually. Each `Add()` acquires a write lock, writes to BBolt, and triggers an fsync on commit. For a message with N transactions, that's N fsyncs.

On local SSD, fsync takes <1ms — barely noticeable. On network-attached storage (e.g. Azure premium SMB), fsync latency is 10-100ms per call. For 100 transactions, that's 1-10 seconds just in fsyncs per message.

This is compounded by the go-stoabs read lock issue ([go-stoabs#146](https://github.com/nuts-foundation/go-stoabs/pull/146)): slow fsyncs hold the write lock longer, and Go's `sync.RWMutex` writer-preference blocks all concurrent reads while a writer is pending. The result is a cascading slowdown where a bootup that takes 3 minutes on local storage takes 30+ minutes on SMB — not a proportional slowdown, but a multiplicative one due to lock contention amplifying the raw I/O penalty.

### The fix

`AddMany()` processes all transactions from a message in a single write transaction:
- Verification happens inside the write TX so that later transactions can reference earlier ones in the same batch (BBolt supports read-your-own-writes within a transaction)
- On the first failure, processing stops and all successfully added transactions are committed
- Returns `(added int, err error)` — the caller uses `added` to identify which transaction failed and `err` to determine the action (`ErrPreviousTransactionMissing` triggers state reconciliation, other errors are logged and recovered via gossip)
- Context cancellation is checked between transactions

### Relationship to go-stoabs#146

This PR and go-stoabs#146 are complementary:
- **go-stoabs#146** eliminates read lock contention (reads no longer blocked by writes)
- **This PR** reduces write lock duration (1 fsync instead of N)

Together they address both sides of the lock contention that causes slow bootup on network storage.

## Test plan

- [x] All existing `TestProtocol_handleTransactionList` tests updated and passing
- [x] Full test suite (106 packages) passes
- [ ] Test on Azure SMB to measure bootup time improvement

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

### DIFF
--- a/network/dag/interface.go
+++ b/network/dag/interface.go
@@ -55,6 +55,12 @@ type State interface {
 	// If the transaction already exists, nothing is added and no observers are notified.
 	// The payload may be passed as well. Allowing for better notification of observers
 	Add(ctx context.Context, transactions Transaction, payload []byte) error
+	// AddMany adds multiple transactions to the DAG in a single write transaction,
+	// requiring only a single fsync. Transactions are processed in order so that later
+	// transactions can reference earlier ones in the same batch.
+	// Returns the number of transactions successfully added and the first error encountered.
+	// Successfully added transactions are committed even when a later transaction fails.
+	AddMany(ctx context.Context, transactions []Transaction, payloads [][]byte) (int, error)
 	// FindBetweenLC finds all transactions which lamport clock value lies between startInclusive and endExclusive.
 	// They are returned in order: first sorted on lamport clock value, then on transaction reference (byte order).
 	FindBetweenLC(ctx context.Context, startInclusive uint32, endExclusive uint32) ([]Transaction, error)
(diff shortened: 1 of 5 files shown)


stevenvegt:
Not ready yet, see failing e2e tests. Keys from previous txs in the same batch don't become available for later txs because the notifier hasn't ran yet. Needs some more thinking.

reinkrul:
not doing this right now

PR #4169, opened 2026-04-10
Add configurable BBolt sync interval for improved performance on network mounts

### LINKED ISSUE #4162: Network notifiers fail to start, causing node to shut down

reinkrul:
## Summary

Adds a `storage.bbolt.syncinterval` configuration option that controls how often BBolt flushes committed transactions to disk.

- By default (`0`), every transaction is flushed immediately (existing behaviour, no change)
- Setting a positive interval (e.g. `1s`) batches flushes via a background goroutine, significantly improving write throughput on network mounts such as SMB/Azure Files

Performance tests on Azure ACI with an SMB volume mount showed ~4× improvement in transaction throughput (55 → 212 tx/s average) when combined with the read lock removal ([go-stoabs#146](https://github.com/nuts-foundation/go-stoabs/pull/146)). See the analysis in [#4162](https://github.com/nuts-foundation/nuts-node/issues/4162#issuecomment-4222410048) for full details.

The risk of data loss on a hard crash is low for network-replicated data, since any transactions not yet flushed will be re-synced from peers on restart.

Depends on [go-stoabs#4](https://github.com/nuts-foundation/go-stoabs/pull/4) (`WithSyncInterval`).

Fixes #4162

## TODO before merging

- [ ] Release go-stoabs with [go-stoabs#146](https://github.com/nuts-foundation/go-stoabs/pull/146) and [go-stoabs#4](https://github.com/nuts-foundation/go-stoabs/pull/4), then update the `go-stoabs` dependency from the pseudo-version to the release tag

## Test plan

- [ ] Verify default behaviour is unchanged (no `syncinterval` set)
- [ ] Set `storage.bbolt.syncinterval=1s`, confirm log line appears on startup
- [ ] Verify data survives graceful shutdown with sync interval set
- [ ] Performance test on SMB/network mount


reinkrul:
not doing this right now

Issue #4175, opened 2026-04-13
Consider circuit breaker for AuthZen PDP client


stevenvegt:
## Context

Issue #4144 introduces an AuthZen-compatible PDP integration for evaluating scopes when `scope_policy: "dynamic"` is configured. If the PDP endpoint becomes unavailable, every token request will wait for the full timeout before returning 503.

## Problem

Without a circuit breaker, a down PDP causes all dynamic-policy token requests to hang for the timeout duration. This degrades the node's responsiveness even though the outcome (503) is already known after the first few failures.

## Proposal

Add a circuit breaker to the AuthZen HTTP client so that after N consecutive failures, subsequent requests fail immediately until a probe request succeeds. This gives bounded, fast failure instead of repeated timeout waits.

## Considerations

- Configuration surface: failure threshold, recovery probe interval
- State management: tracking consecutive failures per endpoint
- Whether this should be a generic HTTP client feature or specific to the AuthZen client
- May not be needed if operators ensure PDP reliability through monitoring — assess based on real-world usage after #4144 ships

## Related

- #4144 (parent: mixed OAuth2 scopes with configurable scope policy)

PR #4176, opened 2026-04-13
Scope parsing, classification & policy config

stevenvegt:
## Parent PRD

#4144

## Summary

Foundation layer for mixed OAuth2 scope support. Renames `PDPBackend.PresentationDefinitions()` to `FindCredentialProfile()` and enriches its return type to include the credential profile scope, its policy, and any remaining ("other") scopes. Extends the policy JSON format with a `scope_policy` field and adds a `policy.authzen.endpoint` CLI flag validated at startup.

## What changed

**Policy interface (`policy/interface.go`, `policy/local.go`):**
- Method renamed `PresentationDefinitions` → `FindCredentialProfile`, returning `*CredentialProfileMatch` (scope, WalletOwnerMapping, scope policy, other scopes) instead of just `pe.WalletOwnerMapping`.
- Multi-scope parsing via `strings.Fields` (tolerant of tabs, newlines, consecutive spaces).
- New errors: `ErrAmbiguousScope` (multiple profile scopes in one request), distinct from `ErrNotFound`.

**Policy JSON format (`policy/local.go`, `policy/test/scope_policy/`):**
- Added `scope_policy` field per credential profile: `"profile-only"` (default), `"passthrough"`, or `"dynamic"`.
- Existing policy files without `scope_policy` continue to work (defaults to `profile-only`).
- Internal `credentialProfileConfig` replaces the map-based `validatingWalletOwnerMapping` with an explicit struct (organization/user fields + scope_policy).
- PD schema validation preserved via `validatingPresentationDefinition` — moved from mapping-level to per-PD level.

**Config & validation (`policy/config.go`, `policy/cmd.go`, `policy/local.go`):**
- New `policy.authzen.endpoint` CLI flag.
- Startup fails when any credential profile uses `dynamic` without an AuthZen endpoint configured.
- `passthrough` and `profile-only` continue to work without an endpoint.

**Callers updated (`auth/api/iam/`):**
- `validation.go` and `api.go` updated for the new return type, with `ErrAmbiguousScope` mapped to `invalid_scope`.
- Mock regenerated; existing test expectations updated to use `*CredentialProfileMatch`.

## How to review

**Start with `policy/interface.go`** — understand the new types (`CredentialProfileMatch`, `ScopePolicy`) and the renamed interface method.

**Then `policy/local.go`** — the core logic:
- `FindCredentialProfile` (scope parsing + classification)
- `credentialProfileConfig.UnmarshalJSON` (JSON format handling)
- `Configure` (startup validation for AuthZen endpoint)

**Tests to focus on** (`policy/local_test.go`):
- `TestLocalPDP_FindCredentialProfile` — multi-scope parsing edge cases (consecutive whitespace, empty strings, multiple profile scopes, zero profile scopes)
- `TestLocalPDP_ScopePolicyConfig` — JSON parsing, invalid value rejection
- `TestLocalPDP_Configure` — startup validation behavior

**Callers (`auth/api/iam/`)** are mechanical updates — they extract `.WalletOwnerMapping` from the result to preserve existing behavior.

## Deviations from spec

- **`credentialProfileConfig` uses an explicit struct with `Organization`/`User` fields**, not the map-based approach suggested in the spec. This avoided custom JSON parsing of the scope_policy field while preserving the existing flat JSON format. PD validation moved to per-PD level (still uses the v2 schema).
- **Added `ErrAmbiguousScope`** as a distinct error (not in spec). Wrapping `ErrNotFound` for "too many profile scopes" was semantically wrong.
- **Added nil-check** that credential profile JSON must define at least one of `organization` / `user` — catches typos like `organisations` that standard struct unmarshaling would silently accept.
- **Used `strings.Fields` instead of `strings.Split`** — handles tabs, newlines, and consecutive spaces robustly.

## Dependencies

None — this is the foundation PR. Subsequent PRs #4177–#4180 build on this interface.

## Design context

- PRD: #4144 (mixed OAuth2 scopes with configurable scope policy)
- Scope policy modes discussion: `passthrough` mode was added in-scope after initial PR split (see PRD comment 4235711103)
- Related issue: #4175 (circuit breaker for AuthZen client — future enhancement)

## Acceptance Criteria

- [x] `FindCredentialProfile()` accepts space-delimited multi-scope strings (uses `strings.Fields` for robust whitespace handling)
- [x] Returns exactly one credential profile scope or errors (`ErrNotFound` for zero, `ErrAmbiguousScope` for multiple)
- [x] `scope_policy` field parsed from policy config (defaults to `"profile-only"`)
- [x] Invalid `scope_policy` values rejected at load time
- [x] Credential profile must define at least one of `organization` or `user`
- [x] `policy.authzen.endpoint` config field and CLI flag added
- [x] Startup fails when `dynamic` scope policy is used without AuthZen endpoint
- [x] `passthrough` and `profile-only` work without AuthZen endpoint
- [x] All existing tests pass (callers and mocks updated for new interface)
- [x] New unit tests for scope parsing, config loading, and startup validation
- [x] PD validation preserved via `validatingPresentationDefinition` (validates against v2 JSON schema)

<details>

## Parent PRD

#4144

### Overview

Foundation layer for mixed OAuth2 scope support. This PR modifies the policy subsystem to:
1. Parse multi-scope strings (space-delimited) and classify each scope as a **credential profile scope** (has a PD mapping) or an **other scope** (no mapping).
2. Enforce exactly one credential profile scope per request.
3. Extend the policy configuration format with a `scope_policy` field per credential profile.
4. Add a global `policy.authzen.endpoint` configuration field for the AuthZen PDP URL.
5. Validate at startup that `scope_policy: "dynamic"` requires an AuthZen endpoint.

### Key files modified

- `policy/interface.go` — Renamed `PDPBackend.PresentationDefinitions()` to `FindCredentialProfile()`, added `CredentialProfileMatch`, `ScopePolicy`, `ErrAmbiguousScope` types
- `policy/local.go` — New `credentialProfileConfig` struct with explicit Organization/User fields, multi-scope parsing via `strings.Fields`, scope policy defaulting and validation
- `policy/config.go` — Added `AuthZenConfig` with endpoint field
- `policy/cmd.go` — Added `policy.authzen.endpoint` CLI flag
- `auth/api/iam/validation.go` — Updated caller to use `FindCredentialProfile`, handles `ErrAmbiguousScope`
- `auth/api/iam/api.go` — Updated caller to use `FindCredentialProfile`

#### Interface change

Renamed `PresentationDefinitions` to `FindCredentialProfile` to better reflect the method's responsibility: resolving a scope string against the policy configuration to find the matching credential profile.

```go
type PDPBackend interface {
    FindCredentialProfile(ctx context.Context, scope string) (*CredentialProfileMatch, error)
}

type CredentialProfileMatch struct {
    CredentialProfileScope string
    WalletOwnerMapping     pe.WalletOwnerMapping
    ScopePolicy            ScopePolicy
    OtherScopes            []string
}
```

#### Policy configuration format

The per-scope JSON structure now supports `scope_policy` alongside the existing `organization`/`user` keys:

```json
{
  "urn:nuts:medication-overview": {
    "organization": { /* PD */ },
    "scope_policy": "profile-only"
  }
}
```

Internally, `credentialProfileConfig` uses a struct with explicit fields (instead of the old map-based `validatingWalletOwnerMapping`), allowing standard JSON unmarshaling. Individual PDs are validated against the v2 JSON schema via `validatingPresentationDefinition`.

#### Error handling

- `ErrNotFound` — no credential profile scope matched
- `ErrAmbiguousScope` — multiple credential profile scopes found in the request

## Acceptance Criteria

- [x] `FindCredentialProfile()` accepts space-delimited multi-scope strings (uses `strings.Fields` for robust whitespace handling)
- [x] Returns exactly one credential profile scope or errors (`ErrNotFound` for zero, `ErrAmbiguousScope` for multiple)
- [x] `scope_policy` field parsed from policy config (defaults to `"profile-only"`)
- [x] Invalid `scope_policy` values rejected at load time
- [x] Credential profile must define at least one of `organization` or `user`
- [x] `policy.authzen.endpoint` config field and CLI flag added
- [x] Startup fails when `dynamic` scope policy is used without AuthZen endpoint
- [x] `passthrough` and `profile-only` work without AuthZen endpoint
- [x] All existing tests pass (callers and mocks updated for new interface)
- [x] New unit tests for scope parsing, config loading, and startup validation
- [x] PD validation preserved via `validatingPresentationDefinition` (validates against v2 JSON schema)

</details>

### DIFF
--- a/policy/interface.go
+++ b/policy/interface.go
@@ -24,10 +24,14 @@ import (
 	"github.com/nuts-foundation/nuts-node/vcr/pe"
 )
 
-// ModuleName is the name of the policy module
+// ModuleName is the name of the policy module.
 const ModuleName = "policy"
 
+// ErrNotFound is returned when no credential profile matches the requested scope.
 var ErrNotFound = errors.New("not found")
+
+// ErrAmbiguousScope is returned when multiple credential profile scopes are found in a single request.
+var ErrAmbiguousScope = errors.New("multiple credential profile scopes found")
 
 // ScopePolicy defines how extra scopes (beyond the credential profile scope) are handled.
 type ScopePolicy string
--- a/policy/interface.go
+++ b/policy/interface.go
@@ -29,10 +29,37 @@ const ModuleName = "policy"
 
 var ErrNotFound = errors.New("not found")
 
-// PDPBackend is the interface for the policy backend
-// Both the remote and local policy backend implement this interface
+// ScopePolicy defines how extra scopes (beyond the credential profile scope) are handled.
+type ScopePolicy string
+
+const (
+	// ScopePolicyProfileOnly only accepts the credential profile scope. Extra scopes cause an error.
+	ScopePolicyProfileOnly ScopePolicy = "profile-only"
+	// ScopePolicyPassthrough grants all requested scopes without evaluation.
+	ScopePolicyPassthrough ScopePolicy = "passthrough"
+	// ScopePolicyDynamic evaluates extra scopes via an external AuthZen PDP.
+	ScopePolicyDynamic ScopePolicy = "dynamic"
+)
+
+// CredentialProfileMatch is the result of matching a scope string against the policy configuration.
+// It contains the matched credential profile (WalletOwnerMapping + ScopePolicy) and the
+// remaining scopes that did not match any credential profile.
+type CredentialProfileMatch struct {
+	// CredentialProfileScope is the scope that matched a credential profile.
+	CredentialProfileScope string
+	// WalletOwnerMapping contains the PresentationDefinitions per wallet owner type for the matched credential profile.
+	WalletOwnerMapping pe.WalletOwnerMapping
+	// ScopePolicy is the configured scope policy for the matched credential profile.
+	ScopePolicy ScopePolicy
+	// OtherScopes contains the scopes from the request that did not match any credential profile.
+	OtherScopes []string
+}
+
+// PDPBackend is the interface for the policy backend.
+// Both the remote and local policy backend implement this interface.
 type PDPBackend interface {
-	// PresentationDefinitions returns the PresentationDefinitions (mapped to a WalletOwnerType) for the given scope
-	// scopes are space delimited. It's up to the backend to decide how to handle this
-	PresentationDefinitions(ctx context.Context, scope string) (pe.WalletOwnerMapping, error)
+	// FindCredentialProfile resolves a scope string against the policy configuration.
+	// It parses the space-delimited scope string, identifies exactly one credential profile scope,
+	// and returns the matched profile along with any remaining scopes.
+	FindCredentialProfile(ctx context.Context, scope string) (*CredentialProfileMatch, error)
 }
--- a/auth/api/iam/api.go
+++ b/auth/api/iam/api.go
@@ -703,7 +703,7 @@ func (r Wrapper) PresentationDefinition(ctx context.Context, request Presentatio
 		return PresentationDefinition200JSONResponse(PresentationDefinition{}), nil
 	}
 
-	mapping, err := r.policyBackend.PresentationDefinitions(ctx, request.Params.Scope)
+	match, err := r.policyBackend.FindCredentialProfile(ctx, request.Params.Scope)
 	if err != nil {
 		return nil, oauth.OAuth2Error{
 			Code:        oauth.InvalidScope,
@@ -715,7 +715,7 @@ func (r Wrapper) PresentationDefinition(ctx context.Context, request Presentatio
 	if request.Params.WalletOwnerType != nil {
 		walletOwnerType = *request.Params.WalletOwnerType
 	}
-	result, exists := mapping[walletOwnerType]
+	result, exists := match.WalletOwnerMapping[walletOwnerType]
 	if !exists {
 		return nil, oauthError(oauth.InvalidRequest, fmt.Sprintf("no presentation definition found for '%s' wallet", walletOwnerType))
 	}
--- a/auth/api/iam/api_test.go
+++ b/auth/api/iam/api_test.go
@@ -190,7 +190,7 @@ func TestWrapper_PresentationDefinition(t *testing.T) {
 
 	t.Run("ok", func(t *testing.T) {
 		test := newTestClient(t)
-		test.policy.EXPECT().PresentationDefinitions(gomock.Any(), "example-scope").Return(walletOwnerMapping, nil)
+		test.policy.EXPECT().FindCredentialProfile(gomock.Any(), "example-scope").Return(&policy.CredentialProfileMatch{WalletOwnerMapping: walletOwnerMapping, ScopePolicy: policy.ScopePolicyProfileOnly}, nil)
 
 		response, err := test.client.PresentationDefinition(ctx, PresentationDefinitionRequestObject{SubjectID: verifierSubject, Params: PresentationDefinitionParams{Scope: "example-scope"}})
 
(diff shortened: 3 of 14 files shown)


policy/interface.go
```diff
@@ -24,15 +24,46 @@ import (
...
+var ErrAmbiguousScope = errors.New("multiple credential profile scopes found")
+
+// ScopePolicy defines how extra scopes (beyond the credential profile scope) are handled.
+type ScopePolicy string
+
+const (
+	// ScopePolicyProfileOnly only accepts the credential profile scope. Extra scopes cause an error.
+	ScopePolicyProfileOnly ScopePolicy = "profile-only"
+	// ScopePolicyPassthrough grants all requested scopes without evaluation.
+	ScopePolicyPassthrough ScopePolicy = "passthrough"
```
reinkrul, line 43: maybe we should remove this after the PoCs and "dynamic" is implemented? Because it sounds quite unsafe.

policy/interface.go
```diff
@@ -24,15 +24,46 @@ import (
...
+	// ScopePolicyDynamic evaluates extra scopes via an external AuthZen PDP.
+	ScopePolicyDynamic ScopePolicy = "dynamic"
+)
+
+// CredentialProfileMatch is the result of matching a scope string against the policy configuration.
+// It contains the matched credential profile (WalletOwnerMapping + ScopePolicy) and the
+// remaining scopes that did not match any credential profile.
+type CredentialProfileMatch struct {
```
reinkrul, line 51: i still find this abstraction a bit weird

policy/interface.go
```diff
@@ -24,15 +24,46 @@ import (
...
+// PDPBackend is the interface for the policy backend.
+// Both the remote and local policy backend implement this interface.
 type PDPBackend interface {
-	// PresentationDefinitions returns the PresentationDefinitions (mapped to a WalletOwnerType) for the given scope
-	// scopes are space delimited. It's up to the backend to decide how to handle this
-	PresentationDefinitions(ctx context.Context, scope string) (pe.WalletOwnerMapping, error)
+	// FindCredentialProfile resolves a scope string against the policy configuration.
+	// It parses the space-delimited scope string, identifies exactly one credential profile scope,
+	// and returns the matched profile along with any remaining scopes.
+	FindCredentialProfile(ctx context.Context, scope string) (*CredentialProfileMatch, error)
```
reinkrul, line 68: Shouldn't it be named "MatchCredentialProfile" if it returns a Match?

policy/interface.go
```diff
@@ -24,15 +24,46 @@ import (
...
+	// ScopePolicyDynamic evaluates extra scopes via an external AuthZen PDP.
+	ScopePolicyDynamic ScopePolicy = "dynamic"
+)
+
+// CredentialProfileMatch is the result of matching a scope string against the policy configuration.
+// It contains the matched credential profile (WalletOwnerMapping + ScopePolicy) and the
+// remaining scopes that did not match any credential profile.
+type CredentialProfileMatch struct {
+	// CredentialProfileScope is the scope that matched a credential profile.
+	CredentialProfileScope string
```
reinkrul, line 53: Odd naming, it repeats its container, but just "Scope" might be too ambiguous. Abstraction/naming is off, but I can't think of a better one right now

policy/interface.go
stevenvegt: Let's discuss this in #4213 — I think there can be value in a passthrough policy. For now let's keep it in for the PoCs.

policy/interface.go
stevenvegt: Going to keep `FindCredentialProfile` — Go stdlib convention leans toward verbs like `Find`/`Lookup` for lookups rather than mirroring the result type (e.g. `regexp.Find*` returns matches). `MatchCredentialProfile` reads to me like a verb form of a noun. Open to alternatives if you've got a stronger verb in mind.

policy/interface.go
stevenvegt: Fair — it is a bit awkward. I landed on this shape because the caller needs four things back (profile scope, mapping, policy, other scopes) and a single struct beats a 5-value return. Happy to restructure if you have a concrete alternative in mind.

policy/interface.go
stevenvegt: Agreed it's imperfect. `CredentialProfileScope` was the least ambiguous I could find — plain `Scope` reads as "the scope" (which scope?) when referenced via `match.Scope`, given the struct already has `OtherScopes`. Leaving as-is for now; happy to rename if a better name surfaces.

PR #4177, opened 2026-04-13
AuthZen PDP client

stevenvegt:
## Parent PRD

#4144

## Summary

New HTTP client for the AuthZen Access Evaluations batch API (`POST /access/v1/evaluations`). Provides the types and the transport layer for dynamic scope policy evaluation. The server-side token flow in #4179 uses this client to evaluate requested scopes against an external PDP.

## What changed

**New package `policy/authzen`:**
- `types.go` — AuthZen request/response types matching the [AuthZen spec](https://openid.net/specs/authorization-api-1_0.html) batch format
- `client.go` — Thin HTTP client with a single `Evaluate` method returning `map[string]bool` (scope → decision)
- `client_test.go` — Unit tests using `httptest.NewServer`

**Batch format:** top-level `subject` / `action` / `context` are shared defaults, and an `evaluations` array contains per-scope `resource` overrides. This avoids repeating shared fields for every scope.

**Error handling:**
- HTTP non-2xx → error with status code, PDP error body truncated to 200 chars (prevents log injection)
- Network errors / cancelled context → wrapped error
- Malformed JSON / evaluation count mismatch → error
- Duplicate resource IDs in request → early error before HTTP call (client-side invariant check)

## How to review

**Start with `types.go`** — verify the AuthZen request/response shape against the spec (the batch format uses top-level defaults + per-evaluation overrides).

**Then `client.go`** — single `Evaluate` method:
- Sends JSON with `Content-Type` + `Accept` headers
- Reads the response through the caller-provided `HTTPRequestDoer` (StrictHTTPClient is used in wiring for response body limits / timeout)
- Validates evaluation count matches request
- Returns `map[string]bool` keyed by `Resource.ID`

**Tests** (`client_test.go`) — 8 scenarios covering the full surface: success, partial denial, HTTP 500, unreachable endpoint, cancelled context, count mismatch, malformed response, duplicate resource IDs.

## Deviations from spec

- **Request type changed from the original spec.** The original spec had `Resource []Resource`, implying a flat batch with multiple resources. During planning we verified this didn't match the AuthZen batch API. Corrected to `Evaluations []Evaluation` with per-evaluation overrides, matching the spec's top-level-defaults-plus-overrides pattern.
- **Claim extraction removed from this PR.** Originally the spec included an `ExtractSubjectProperties` helper. Moved to #4179 where the credential map's shape is concrete. This client is transport-only — it takes a pre-built `EvaluationsRequest` from the caller.
- **`NewClient` signature simplified** — no `timeout` parameter. Timeout is the responsibility of the injected `HTTPRequestDoer` (wired with `StrictHTTPClient` in #4179).
- **Added at self-review:** duplicate resource ID precondition check, error body truncation, `Accept: application/json` header.
- **`subject.type` in the PRD example** was updated from `"token_request"` to `"organization"` (see PRD comment). The AuthZen client itself is agnostic — the caller sets the type.

## Dependencies

**Depends on:** #4176 (foundation PR, provides the base branch).

**Review order:** Review #4176 first for the foundation; this PR can then be reviewed independently as it's a self-contained new package with no cross-cutting changes.

## Design context

- PRD: #4144
- AuthZen spec: https://openid.net/specs/authorization-api-1_0.html
- Request format discussion: determined during implementation planning — switched from "multi-resource single evaluation" to "evaluations array" per the spec
- Future enhancement: #4175 (circuit breaker for PDP calls)

## Acceptance Criteria

- [x] AuthZen client sends correctly formatted batch evaluation requests (per AuthZen spec)
- [x] Response parsed into scope → decision map
- [x] Timeout and error handling works correctly
- [x] Evaluation count mismatch detected and reported
- [x] Duplicate resource IDs in request detected (added during self-review)
- [x] Error response body truncated to prevent log injection (added during self-review)
- [x] Unit tests with HTTP mock cover success, partial denial, errors

<details>

## Parent PRD

#4144

### Overview

New HTTP client for the AuthZen batch Access Evaluations API (`POST /access/v1/evaluations`). This client is used by the server-side token flow (PR #4179) when `scope_policy: "dynamic"` to evaluate requested scopes against an external PDP.

### Key files to create/modify

- `policy/authzen/client.go` — New HTTP client
- `policy/authzen/types.go` — Request/response types
- `policy/authzen/client_test.go` — Unit tests

### Design decisions

- **Batch format follows the [AuthZen spec](https://openid.net/specs/authorization-api-1_0.html):** top-level subject/action/context are shared defaults, the `evaluations` array contains per-scope resource overrides. This avoids repeating the same subject/action/context for every scope.
- **Claim extraction deferred to PR #4179:** The `ExtractSubjectProperties` helper depends on how the credential map from PEX evaluation maps to the organization bucket. This becomes clear in the server-side flow where the actual credential data is available. The AuthZen client takes a pre-built `EvaluationsRequest` — it doesn't need to know about credentials.
- **`map[string]bool` return type:** Denial reasons from the AuthZen response are logged for debugging but not exposed in the return value. Rego-produced reasons are typically terse and few. A richer return type can be added later if operators need it.

### AuthZen request/response examples

**Request** (`POST /access/v1/evaluations`):

```json
{
  "subject": {
    "type": "token_request",
    "id": "did:web:hospital.example.com",
    "properties": {
      "organization": {
        "id": "did:web:hospital.example.com",
        "name": "Hospital B.V.",
        "ura": "12345678"
      }
    }
  },
  "action": { "name": "request_scope" },
  "context": { "policy": "urn:nuts:medication-overview" },
  "evaluations": [
    { "resource": { "type": "scope", "id": "urn:nuts:medication-overview" } },
    { "resource": { "type": "scope", "id": "patient/Observation.read" } },
    { "resource": { "type": "scope", "id": "launch/patient" } }
  ]
}
```

**Response**:

```json
{
  "evaluations": [
    { "decision": true },
    { "decision": true },
    { "decision": false, "context": { "reason": "scope not permitted by policy" } }
  ]
}
```

### Go types

```go
type EvaluationsRequest struct {
    Subject     Subject           `json:"subject"`
    Action      Action            `json:"action"`
    Context     EvaluationContext `json:"context"`
    Evaluations []Evaluation      `json:"evaluations"`
}

type Evaluation struct {
    Resource Resource `json:"resource"`
}

type Subject struct {
    Type       string            `json:"type"`
    ID         string            `json:"id"`
    Properties SubjectProperties `json:"properties"`
}

type SubjectProperties struct {
    Client       map[string]any `json:"client,omitempty"`
    Organization map[string]any `json:"organization,omitempty"`
    User         map[string]any `json:"user,omitempty"`
}

type Action struct {
    Name string `json:"name"`
}

type Resource struct {
    Type string `json:"type"`
    ID   string `json:"id"`
}

type EvaluationContext struct {
    Policy string `json:"policy"`
}

type EvaluationsResponse struct {
    Evaluations []EvaluationResult `json:"evaluations"`
}

type EvaluationResult struct {
    Decision bool                    `json:"decision"`
    Context  *EvaluationResultContext `json:"context,omitempty"`
}

type EvaluationResultContext struct {
    Reason string `json:"reason,omitempty"`
}
```

### Client implementation

```go
type Client struct {
    endpoint   string
    httpClient core.HTTPRequestDoer
}

func NewClient(endpoint string, httpClient core.HTTPRequestDoer) *Client

// Evaluate sends a batch evaluation request for the given scopes.
// Returns a map of scope → decision (true = granted, false = denied).
func (c *Client) Evaluate(ctx context.Context, req EvaluationsRequest) (map[string]bool, error)
```

The `Evaluate` method:
1. Serializes the request to JSON.
2. POSTs to `{endpoint}/access/v1/evaluations`.
3. Parses the response.
4. Maps each evaluation result back to the corresponding scope (by index).
5. Returns a `map[string]bool` for easy lookup.

### Error handling

- HTTP errors (non-2xx) → return error with status code
- Network errors / timeouts → return error (caller maps to 503)
- Malformed response → return error
- Evaluation count mismatch (response has fewer evaluations than scopes) → return error

### Testing

- **Successful evaluation**: batch request → scope→decision map
- **Partial denial**: some scopes approved, some denied
- **HTTP error**: PDP returns 500 → error
- **Timeout**: PDP unreachable → error
- **Evaluation count mismatch**: response has fewer evaluations than request → error
- **Malformed response**: invalid JSON → error
- Use `httptest.NewServer` for HTTP mocking

## Acceptance Criteria

- [x] AuthZen client sends correctly formatted batch evaluation requests (per AuthZen spec)
- [x] Response parsed into scope → decision map
- [x] Timeout and error handling works correctly
- [x] Evaluation count mismatch detected and reported
- [x] Duplicate resource IDs in request detected (added during self-review)
- [x] Error response body truncated to prevent log injection (added during self-review)
- [x] Unit tests with HTTP mock cover success, partial denial, errors

</details>

### DIFF
--- a/policy/authzen/client.go
+++ b/policy/authzen/client.go
@@ -1,0 +1,85 @@
+/*
+ * Copyright (C) 2026 Nuts community
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *
+ */
+
+package authzen
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+
+	"github.com/nuts-foundation/nuts-node/core"
+)
+
+const evaluationsPath = "/access/v1/evaluations"
+
+// Client is an HTTP client for the AuthZen Access Evaluations API.
+type Client struct {
+	endpoint   string
+	httpClient core.HTTPRequestDoer
+}
+
+// NewClient creates a new AuthZen client. The endpoint is the base URL of the PDP.
+// The httpClient handles timeouts and TLS configuration.
+func NewClient(endpoint string, httpClient core.HTTPRequestDoer) *Client {
+	return &Client{
+		endpoint:   endpoint,
+		httpClient: httpClient,
+	}
+}
+
+// Evaluate sends a batch evaluation request and returns a map of scope → decision.
+func (c *Client) Evaluate(ctx context.Context, req EvaluationsRequest) (map[string]bool, error) {
+	body, err := json.Marshal(req)
+	if err != nil {
+		return nil, fmt.Errorf("authzen: marshal request: %w", err)
+	}
+	httpReq, err := http.NewRequestWithContext(ctx, http.MethodPost, c.endpoint+evaluationsPath, bytes.NewReader(body))
+	if err != nil {
+		return nil, fmt.Errorf("authzen: create request: %w", err)
+	}
+	httpReq.Header.Set("Content-Type", "application/json")
+
+	httpResp, err := c.httpClient.Do(httpReq)
+	if err != nil {
+		return nil, fmt.Errorf("authzen: execute request: %w", err)
+	}
+	defer httpResp.Body.Close()
+
+	if httpResp.StatusCode != http.StatusOK {
+		respBody, _ := io.ReadAll(httpResp.Body)
+		return nil, fmt.Errorf("authzen: PDP returned HTTP %d: %s", httpResp.StatusCode, string(respBody))
+	}
+
+	var resp EvaluationsResponse
+	if err := json.NewDecoder(httpResp.Body).Decode(&resp); err != nil {
+		return nil, fmt.Errorf("authzen: decode response: %w", err)
+	}
+	if len(resp.Evaluations) != len(req.Evaluations) {
+		return nil, fmt.Errorf("authzen: expected %d evaluations, got %d", len(req.Evaluations), len(resp.Evaluations))
+	}
+
+	decisions := make(map[string]bool, len(req.Evaluations))
+	for i, eval := range resp.Evaluations {
+		decisions[req.Evaluations[i].Resource.ID] = eval.Decision
+	}
+	return decisions, nil
+}
--- a/policy/authzen/client.go
+++ b/policy/authzen/client.go
@@ -16,6 +16,7 @@
  *
  */
 
+// Package authzen implements an HTTP client for the AuthZen Access Evaluations API.
 package authzen
 
 import (
@@ -57,6 +58,16 @@ func (c *Client) Evaluate(ctx context.Context, req EvaluationsRequest) (map[stri
 		return nil, fmt.Errorf("authzen: create request: %w", err)
 	}
 	httpReq.Header.Set("Content-Type", "application/json")
+	httpReq.Header.Set("Accept", "application/json")
+
+	// Check for duplicate resource IDs before sending the request
+	seen := make(map[string]bool, len(req.Evaluations))
+	for _, eval := range req.Evaluations {
+		if seen[eval.Resource.ID] {
+			return nil, fmt.Errorf("authzen: duplicate resource ID in request: %s", eval.Resource.ID)
+		}
+		seen[eval.Resource.ID] = true
+	}
 
 	httpResp, err := c.httpClient.Do(httpReq)
 	if err != nil {
@@ -66,7 +77,13 @@ func (c *Client) Evaluate(ctx context.Context, req EvaluationsRequest) (map[stri
 
 	if httpResp.StatusCode != http.StatusOK {
 		respBody, _ := io.ReadAll(httpResp.Body)
-		return nil, fmt.Errorf("authzen: PDP returned HTTP %d: %s", httpResp.StatusCode, string(respBody))
+		// Truncate to keep error messages reasonable and prevent log injection
+		const maxErrorLen = 200
+		bodyStr := string(respBody)
+		if len(bodyStr) > maxErrorLen {
+			bodyStr = bodyStr[:maxErrorLen] + "..."
+		}
+		return nil, fmt.Errorf("authzen: PDP returned HTTP %d: %s", httpResp.StatusCode, bodyStr)
 	}
 
 	var resp EvaluationsResponse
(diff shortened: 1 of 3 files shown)


policy/authzen/client.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+}
+
+// Evaluate sends a batch evaluation request and returns a map of scope → decision.
+func (c *Client) Evaluate(ctx context.Context, req EvaluationsRequest) (map[string]bool, error) {
+	body, err := json.Marshal(req)
+	if err != nil {
+		return nil, fmt.Errorf("authzen: marshal request: %w", err)
+	}
+	httpReq, err := http.NewRequestWithContext(ctx, http.MethodPost, c.endpoint+evaluationsPath, bytes.NewReader(body))
```
reinkrul, line 56: timeout?

policy/authzen/client.go
```diff
@@ -0,0 +1,102 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("authzen: create request: %w", err)
+	}
+	httpReq.Header.Set("Content-Type", "application/json")
+	httpReq.Header.Set("Accept", "application/json")
+
+	// Check for duplicate resource IDs before sending the request
```
reinkrul, line 63: why?

policy/authzen/client.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+
+	httpResp, err := c.httpClient.Do(httpReq)
+	if err != nil {
+		return nil, fmt.Errorf("authzen: execute request: %w", err)
+	}
+	defer httpResp.Body.Close()
+
+	if httpResp.StatusCode != http.StatusOK {
+		respBody, _ := io.ReadAll(httpResp.Body)
```
reinkrul, line 79: use limitreader to prevent DoS

policy/authzen/client.go
```diff
@@ -0,0 +1,102 @@
...
+		seen[eval.Resource.ID] = true
+	}
+
+	httpResp, err := c.httpClient.Do(httpReq)
+	if err != nil {
+		return nil, fmt.Errorf("authzen: execute request: %w", err)
+	}
+	defer httpResp.Body.Close()
+
+	if httpResp.StatusCode != http.StatusOK {
```
reinkrul, line 78: we have a TestResponseCode func

policy/authzen/client.go
stevenvegt: `HTTPRequestDoer` is expected to enforce timeout — in practice this is always `StrictHTTPClient`, the project convention for outbound HTTP (handles timeout, TLS, body size limits). Wiring in #4179 uses `StrictHTTPClient`. Updated the `NewClient` godoc to make the contract explicit.

policy/authzen/client.go
stevenvegt: The AuthZen response format correlates request to response by index, not by resource ID. That's fragile — `map[string]bool` would collapse duplicates silently, and upstream scope parsing could in theory let dupes through. Keeping this as a boundary check to make the positional assumption safe. Replaced the comment with the actual rationale.

policy/authzen/client.go
stevenvegt: Switched to `core.TestResponseCode` (see reply on the status-check comment). Body-size protection is handled by `StrictHTTPClient` — the project-wide `HTTPRequestDoer` implementation caps response bodies at 1 MiB via `limitedReadAll`. Not duplicating that here to avoid diverging defenses across the codebase.

policy/authzen/client.go
stevenvegt: Fixed in latest push — switched to `core.TestResponseCode`. Wrapped with "authzen: PDP call failed" to disambiguate from the AS server in logs. Also drops the bespoke body-truncation code since `HttpError` doesn't include the body in its message (no log-injection risk).

PR #4178, opened 2026-04-13
Client-side multi-scope token flow

stevenvegt:
## Parent PRD

#4144

## Summary

Modifies the client-side S2S token request flow to support mixed OAuth2 scopes. Introduces a `PresentationDefinitionResolver` that decides whether to fetch the PD from the remote AS (Nuts-to-Nuts, trust the server) or fall back to local policy resolution (integration with non-Nuts AS), and enforces scope policy when resolving locally.

## What changed

**New `PresentationDefinitionResolver`** (`auth/client/iam/pd_resolver.go`):
- Resolves a PD + the scope string to forward in the token request
- Remote path: metadata has a PD endpoint → fetch from remote, forward all scopes unchanged
- Local path: no PD endpoint → call `FindCredentialProfile` locally; enforce scope policy (profile-only rejects extras; passthrough/dynamic forward all)
- Nil guard: resolver returns a clear error if no policy backend is configured but the remote has no PD endpoint

**`OpenID4VPClient` wiring** (`auth/client/iam/openid4vp.go`):
- `RequestRFC021AccessToken` replaces its direct PD fetch with `c.pdResolver.Resolve(...)` and uses `resolved.Scope` for the token request
- `OpenID4VPClient` gets a `pdResolver` field; `NewClient` takes a `policy.PDPBackend` parameter and wires the resolver internally

**Propagation through `Auth`** (`auth/auth.go`, `cmd/root.go`):
- `Auth` struct gets a `policyBackend` field
- `NewAuthInstance` takes `policy.PDPBackend` as a new parameter
- `cmd/root.go` reorders: `policy.New()` moves before `auth.NewAuthInstance` so it can be passed in

**Dynamic = passthrough on client:** the client doesn't call the AuthZen PDP. The server enforces dynamic at token-grant time (#4179).

## How to review

**Start with `pd_resolver.go`** — the core logic:
- `Resolve` dispatches to remote or local
- `resolveLocal` enforces scope policy and selects the organization PD (TODO comment for #4080 two-VP flow)
- Tests in `pd_resolver_test.go` cover both paths with all scope policies, nil guards, and missing-PD / unknown-scope edge cases (10 cases total)

**Then `openid4vp.go`** — the call site change is minimal (replacing the direct PD fetch with `pdResolver.Resolve`). The important detail is that `data.Set(oauth.ScopeParam, resolved.Scope)` uses the resolver's output, not the raw input.

**Wiring review** — `auth.go` and `cmd/root.go` add the policy backend dependency. Note the line reordering in `cmd/root.go` (policyInstance moved up).

## Deviations from spec

- **Spec targeted `presentationDefinitionForScope`, but that's server-side only.** The actual client-side flow goes through `RequestRFC021AccessToken`. The resolver replaces the direct PD fetch there.
- **Added remote-vs-local PD resolution** (not in original spec). The spec assumed the client always resolves PD locally. In practice the client-side flow fetches from the remote PD endpoint. Added local fallback explicitly for non-Nuts AS integrations — this is a meaningful new capability.
- **`PresentationDefinitionResolver` was added as an abstraction** instead of inlining the logic in `RequestRFC021AccessToken`. Keeps the OAuth client focused on the OAuth flow and makes PD resolution independently testable.
- **Profile-only returns canonical `CredentialProfileScope`** (not raw input) — caught during self-review to avoid echoing whitespace/formatting artifacts.

## Dependencies

**Depends on:** #4176 (base branch, provides `FindCredentialProfile` and scope policy types).

Independent of #4177 (AuthZen client) — client-side doesn't call the PDP. Can be reviewed in parallel with #4177.

## Design context

- PRD: #4144
- Fallback design: the fallback ordering (metadata-first, local fallback if no remote PD endpoint) was discussed during implementation — simpler than always consulting both.
- Resolver abstraction: decided to extract rather than inline, to keep the OAuth client focused
- TODO in code for #4080: resolver currently selects organization PD; when two-VP flow lands, resolver returns both

## Acceptance Criteria

- [x] `PresentationDefinitionResolver` resolves PD from remote when PD endpoint exists
- [x] Resolver falls back to local PD via `FindCredentialProfile` when no remote PD endpoint
- [x] Local fallback: profile-only rejects extra scopes
- [x] Local fallback: passthrough and dynamic forward all scopes
- [x] Remote path: all scopes forwarded, scope policy not enforced locally
- [x] `RequestRFC021AccessToken` uses resolver instead of direct PD fetch
- [x] Single-scope requests work unchanged (backwards compatible)
- [x] Unit tests cover both resolver paths with all scope policies
- [x] Nil policy backend guard (added during self-review)
- [x] Profile-only returns canonical scope, not raw input (added during self-review)

<details>

## Parent PRD

#4144

### Overview

Modify the client-side token request flow to support mixed OAuth2 scopes. Introduces a `PresentationDefinitionResolver` that abstracts the decision of whether to fetch a PD from the remote AS or fall back to local policy resolution. When using local resolution, scope policy is enforced.

### Design decisions

- **`PresentationDefinitionResolver` abstraction**: The client (`RequestRFC021AccessToken`) should not decide where the PD comes from. A resolver encapsulates the remote-vs-local decision and scope policy enforcement, keeping the client focused on the OAuth flow.
- **Remote PD endpoint → trust remote AS**: When the remote AS metadata advertises a PD endpoint, the resolver fetches the PD from there and returns the full scope string. The remote server enforces scope policy (covered by PR #4179).
- **No remote PD endpoint → local fallback**: When no PD endpoint exists, the resolver calls `FindCredentialProfile` locally to get the PD and scope classification. Scope policy is enforced locally: profile-only rejects extra scopes, passthrough/dynamic forward all scopes.
- **Dynamic same as passthrough on client side**: The client does not call the AuthZen PDP — the server handles dynamic scope evaluation at token-grant time (PR #4179).
- **Resolver on `OpenID4VPClient`**: The resolver is a struct dependency, wired through `Auth` → `NewClient`. Policy backend passed through `Auth` to enable local PD fallback.

### Deviations from original spec

- **Spec assumed `presentationDefinitionForScope` was the modification point**: That helper is server-side only. The actual client-side flow goes through `RequestRFC021AccessToken` → remote PD endpoint. The resolver replaces the direct PD fetch.
- **Spec didn't account for remote-vs-local PD resolution**: The client fetches PDs from the remote AS when possible, falling back to local only when no PD endpoint exists. This enables integration with external (non-Nuts) authorization servers.
- **Policy backend wired through `Auth` struct**: Added `policyBackend` parameter to `NewAuthInstance` and `iam.NewClient` to propagate the policy backend to the resolver.

### Modified flow

```
RequestRFC021AccessToken(ctx, ..., scope, ...)
    1. Fetch remote AS metadata
    2. resolved, err := c.pdResolver.Resolve(ctx, scope, metadata)
       Internally:
       a. If metadata has PD endpoint → fetch from remote, return full scope
       b. If no PD endpoint → FindCredentialProfile locally, enforce scope policy
    3. Build VP using resolved.PresentationDefinition
    4. Send token request with resolved.Scope
```

## Acceptance Criteria

- [x] `PresentationDefinitionResolver` resolves PD from remote when PD endpoint exists
- [x] Resolver falls back to local PD via `FindCredentialProfile` when no remote PD endpoint
- [x] Local fallback: profile-only rejects extra scopes
- [x] Local fallback: passthrough and dynamic forward all scopes
- [x] Remote path: all scopes forwarded, scope policy not enforced locally
- [x] `RequestRFC021AccessToken` uses resolver instead of direct PD fetch
- [x] Single-scope requests work unchanged (backwards compatible)
- [x] Unit tests cover both resolver paths with all scope policies
- [x] Nil policy backend guard (added during self-review)
- [x] Profile-only returns canonical scope, not raw input (added during self-review)

</details>

### DIFF
--- a/auth/client/iam/pd_resolver.go
+++ b/auth/client/iam/pd_resolver.go
@@ -1,0 +1,99 @@
+/*
+ * Copyright (C) 2026 Nuts community
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *
+ */
+
+package iam
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/nuts-foundation/nuts-node/auth/oauth"
+	"github.com/nuts-foundation/nuts-node/core"
+	nutsHttp "github.com/nuts-foundation/nuts-node/http"
+	"github.com/nuts-foundation/nuts-node/policy"
+	"github.com/nuts-foundation/nuts-node/vcr/pe"
+)
+
+// ResolvedPresentationDefinition contains a resolved PD and the scope to use in the token request.
+type ResolvedPresentationDefinition struct {
+	PresentationDefinition pe.PresentationDefinition
+	// Scope is the scope string to include in the token request.
+	// When resolved remotely, this is the original scope string (remote AS handles scope policy).
+	// When resolved locally, this depends on the configured scope policy.
+	Scope string
+}
+
+// PresentationDefinitionResolver resolves a PresentationDefinition for a given scope string.
+// It uses the remote AS's PD endpoint when available, falling back to local policy resolution.
+type PresentationDefinitionResolver struct {
+	httpClient    HTTPClient
+	policyBackend policy.PDPBackend
+	strictMode    bool
+}
+
+// Resolve resolves a PresentationDefinition for the given scope string.
+// If the remote AS metadata advertises a PD endpoint, the PD is fetched remotely
+// and the full scope string is returned (remote AS handles scope policy).
+// If no PD endpoint is available, the local policy backend is used and scope policy is enforced.
+func (r *PresentationDefinitionResolver) Resolve(ctx context.Context, scope string, metadata oauth.AuthorizationServerMetadata) (*ResolvedPresentationDefinition, error) {
+	if metadata.PresentationDefinitionEndpoint != "" {
+		return r.resolveRemote(ctx, scope, metadata)
+	}
+	return r.resolveLocal(ctx, scope)
+}
+
+func (r *PresentationDefinitionResolver) resolveRemote(ctx context.Context, scope string, metadata oauth.AuthorizationServerMetadata) (*ResolvedPresentationDefinition, error) {
+	parsedURL, err := core.ParsePublicURL(metadata.PresentationDefinitionEndpoint, r.strictMode)
+	if err != nil {
+		return nil, err
+	}
+	pdURL := nutsHttp.AddQueryParams(*parsedURL, map[string]string{
+		"scope": scope,
+	})
+	pd, err := r.httpClient.PresentationDefinition(ctx, pdURL)
+	if err != nil {
+		return nil, err
+	}
+	return &ResolvedPresentationDefinition{
+		PresentationDefinition: *pd,
+		Scope:                  scope,
+	}, nil
+}
+
+func (r *PresentationDefinitionResolver) resolveLocal(ctx context.Context, scope string) (*ResolvedPresentationDefinition, error) {
+	match, err := r.policyBackend.FindCredentialProfile(ctx, scope)
+	if err != nil {
+		return nil, fmt.Errorf("local PD resolution failed: %w", err)
+	}
+	if match.ScopePolicy == policy.ScopePolicyProfileOnly && len(match.OtherScopes) > 0 {
+		return nil, oauth.OAuth2Error{
+			Code:        oauth.InvalidScope,
+			Description: "scope policy 'profile-only' does not allow additional scopes",
+		}
+	}
+	// Select the organization PD (default for current single-VP flow).
+	// TODO: When #4080 adds two-VP support, this resolver will need to return multiple PDs.
+	pd, ok := match.WalletOwnerMapping[pe.WalletOwnerOrganization]
+	if !ok {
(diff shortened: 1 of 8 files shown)


auth/client/iam/pd_resolver.go
```diff
@@ -0,0 +1,108 @@
...
+	PresentationDefinition pe.PresentationDefinition
+	// Scope is the scope string to include in the token request.
+	// When resolved remotely, this is the original scope string (remote AS handles scope policy).
+	// When resolved locally, this depends on the configured scope policy.
+	Scope string
+}
+
+// PresentationDefinitionResolver resolves a PresentationDefinition for a given scope string.
+// It uses the remote AS's PD endpoint when available, falling back to local policy resolution.
+type PresentationDefinitionResolver struct {
```
reinkrul, line 43: I think this is what the PDPBackend is for, don't need another abstraction layer (just implement another PDPBackend?)

stevenvegt: Checked the history — the `resolveRemote` logic already existed in `openid4vp.go` on master (direct call to the remote PD endpoint inside `RequestRFC021AccessToken`); `PDPBackend` was never wired into the client-side flow. The resolver was added in this PR specifically to introduce the **local fallback** when the remote AS has no PD endpoint. The pre-existing remote logic moved into the resolver for cleanliness, not as a new abstraction layer.

`PDPBackend` is a server-side policy abstraction — `FindCredentialProfile` returns a full `CredentialProfileMatch` (credential profile scope, scope policy, other-scope classification). A remote AS doesn't give us any of that via its PD endpoint; forcing it into `PDPBackend` means faking three of four fields and pulling HTTP dependencies into the `policy` package.

So `PDPBackend` stays as "local policy decisions" with `LocalPDP` as its only logical implementation. The resolver does the distinct job of client-side dispatch (remote-if-available / local-fallback) + scope policy enforcement.

stevenvegt: Small follow-up: while reviewing this discussion, noticed `resolveRemote` duplicated the parse + strictmode-check + HTTP call that already lived in `OpenID4VPClient.PresentationDefinition`. Refactored in 4a920e32 — resolver now takes a narrow `pdFetcher` interface (satisfied by `OpenID4VPClient`), composes the URL with the scope query param, and delegates the fetch. No more `strictMode` field or low-level `HTTPClient` in the resolver. Tests swap `httptest.NewServer` for a tiny fake fetcher that also lets us assert the URL.

PR #4181, opened 2026-04-13
Move Jaeger e2e test to standalone single-node test

reinkrul:
## Summary
- Extracts Jaeger trace verification out of the rfc021 OAuth flow test into a dedicated, simpler single-node test at `e2e-tests/ops/tracing/`.
- The rfc021 test no longer depends on Jaeger; removed the service, tracing env vars, `tracing:` config block, and the `traceparent` / `assertJaegerTrace` logic from the flow.
- New test boots one `nuts-node` + `jaeger`, hits `/status` with a `traceparent` header, then asserts the trace appears in Jaeger.

## Test plan
- [ ] Run `e2e-tests/ops/tracing/run-test.sh` and verify it passes
- [ ] Run `e2e-tests/oauth-flow/rfc021/run-test.sh` and verify the OAuth flow still passes without the Jaeger bits

### DIFF
--- a/e2e-tests/oauth-flow/rfc021/do-test.sh
+++ b/e2e-tests/oauth-flow/rfc021/do-test.sh
@@ -140,7 +140,7 @@ cat << EOF
 EOF
 )
 # Request access token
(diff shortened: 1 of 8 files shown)


stevenvegt (approved):
No node spanning tracing test anymore, but more stable test and separation of concerns is good.

PR #4191, opened 2026-04-15
Introduce JWTProfile for declarative JWT validation

stevenvegt:
## Summary

JWT validation was scattered across the codebase with ad-hoc checks after `ParseJWT` calls. Each JWT type (access token, bearer token, JAR, OpenID4VCI proof, VC, VP) had its own validation pattern — some re-parsed the JWS to read headers, some duplicated iss-to-kid binding, and critically the v1 access token introspection did not call `jwt.Validate` at all, so tokens without `exp`/`iat` were silently accepted with no max-lifetime enforcement. Clock skew tolerance was also inconsistent: the operator-configured `auth.clockskew` only reached two of the five ParseJWT call sites; the rest defaulted to 0.

This PR introduces a declarative `JWTProfile` type and threads it through `ParseJWT`. All caller sites now share one validation shape, one skew policy, and one place to reason about what each JWT type requires.

### Core: `crypto/jwt_profile.go` + `crypto/jwx.go`

`JWTProfile` declares:

- `Typ` — required `typ` header value
- `RequiredClaims` — must be present and non-empty
- `MaxValidity` — max allowed `exp - iat` delta (auto-requires exp/iat via jwx's `WithMaxDelta`)
- `ClockSkew` — acceptable clock skew; zero falls back to `DefaultJWTClockSkew` (5s)
- `Validators` — additional checks (e.g. `IssuerKidValidator`)

Fluent overrides `WithMaxValidity(d)` and `WithClockSkew(d)` let callers override per-call without mutating the shared profile globals.

`ParseJWT`'s signature becomes `(tokenString, keyFunc, profile, at *time.Time)`:

- `Typ` is checked early, before signature verification.
- `MaxValidity` and `ClockSkew` flow into jwx's validation pass.
- Presence + non-empty is checked post-parse; errors mirror the jwx format (`"<claim>" not satisfied: ...`) for consistency.
- Custom validators run last.
- The variadic `jwt.ParseOption` is gone; historical verification uses the explicit `at *time.Time` (only VC/VP signature verification passes non-nil).

### Profiles

- `v1AccessTokenProfile` (auth/services/oauth) — `typ: "at+jwt"`, requires iss/sub/service/exp/iat, iss-to-kid binding, max lifetime tracks configured `accessTokenLifeSpan`
- `v1BearerTokenProfile` (auth/services/oauth) — requires iss/sub/exp/iat/aud, max 5s (aud set by `claimsFromRequest`, required because the builder always sets it)
- `jarProfile` (auth/api/iam) — requires iss/iat/exp, max 5m; `iat`/`exp` are set in `Sign()` (not `Create()`) because the jarRequest is persisted between two HTTP requests — setting them at Create would eat into the validity window
- `openID4VCIProofProfile` (vcr/issuer) — `typ: "openid4vci-proof+jwt"`, requires iat/aud (aud comparison against `i.issuerIdentifierURL` still runs post-parse)
- `vcJWTProfile` (vcr/verifier) — iss-to-kid binding for JWT VCs
- `vpJWTProfile` (vcr/verifier) — requires `nbf` as issuance anchor

### Clock skew unification

- `auth.clockskew` config now reaches every ParseJWT call site via `profile.WithClockSkew(s.clockSkew)` (previously only two call sites honored it).
- `DefaultJWTClockSkew` (5s) applies whenever a profile doesn't set its own — no more silent zero-skew defaults for JAR/VC/VP paths.
- OpenID4VCI proof's hardcoded `WithAcceptableSkew(5s)` is gone (covered by the default).

### VP hardening

The VP builder (`vcr/holder/presenter.go`) now always sets `iat` and `exp` (default 15m when caller doesn't pass `Expires`). Previously VPs created via `/internal/vcr/v2/holder/vp` could lack an expiration.

### Removed

- Ad-hoc `jwt.Validate` call and manual iss/sub/service checks in `IntrospectAccessToken`
- Manual `JWT validity too long` check in `validateAccessTokenRequest`
- `jws.ParseString` re-parse in `vcr/issuer/openid.go` that existed only to check `typ`
- Redundant `strings.Split(keyID, "#")[0] != issuer` check in the VC signature path (covered by `IssuerKidValidator`)
- Redundant `jwt.WithRequiredClaim` loop in `ParseJWT` (presence check is already done post-parse; jwx's `WithMaxDelta` auto-requires its time claims)

## Test plan

- [x] `go test ./...` passes
- [x] `TestService_IntrospectAccessToken` covers: missing/wrong typ, VP JWT replayed as access token, missing service/iss/exp/iat, iss/kid DID mismatch, excessive lifetime (both the strict 60s and configurable `accessTokenLifeSpan` paths), expired token
- [x] New `TestParseJWT` cases for default clock skew and profile-level override
- [x] JAR `Create`/`Sign`/`Parse` tests verify iat/exp are set at sign time, caller's map is not mutated, and the 5m max is enforced on receive
- [x] Existing test assertions updated to the jwx-style error format (`"<claim>" not satisfied: ...`)


auth/api/iam/jar.go
```diff
@@ -42,6 +42,11 @@ type jarRequest struct {
 	RequestURIMethod string          `json:"request_uri_method"`
 }
 
+// jarProfile defines JWT validation rules for JWT Authorization Requests (JAR).
+var jarProfile = &cryptoNuts.JWTProfile{
+	RequiredClaims: []string{"iss"},
```
reinkrul: I think we can safely add max validity, JAR is the signed request object for OpenID4VP, which we don't actually use yet

auth/services/oauth/authz_server.go
```diff
@@ -170,6 +170,20 @@ func NewAuthorizationServer(
 // BearerTokenMaxValidity is the number of seconds that a bearer token is valid
 const BearerTokenMaxValidity = 5
 
+// v1AccessTokenProfile defines JWT validation rules for v1 access tokens.
+var v1AccessTokenProfile = &nutsCrypto.JWTProfile{
+	Typ:            "at+jwt",
+	RequiredClaims: []string{jwt.ExpirationKey, jwt.IssuedAtKey, jwt.IssuerKey, jwt.SubjectKey, "service"},
+	MaxValidity:    secureAccessTokenLifeSpan,
```
reinkrul, line 178: This is configurable (used when creating ATs), I think we should honor that (otherwise, setting it higher effecitvely won't work).

(https://nuts-node.readthedocs.io/en/v5.4/pages/deployment/configuration.html states `auth.accesstokenlifespan` is 60 seconds by default)

auth/services/oauth/authz_server.go
```diff
@@ -170,6 +170,20 @@ func NewAuthorizationServer(
...
+	Validators:     []nutsCrypto.JWTValidator{nutsCrypto.IssuerKidValidator},
+}
+
+// v1BearerTokenProfile defines JWT validation rules for v1 JWT bearer tokens (RFC003 §5.2.1).
+var v1BearerTokenProfile = &nutsCrypto.JWTProfile{
+	RequiredClaims: []string{jwt.ExpirationKey, jwt.IssuedAtKey, jwt.IssuerKey, jwt.SubjectKey},
```
reinkrul: isn't scope or purposeOfUse also required? (although it might be already checked in the HTTP handler)

auth/services/oauth/authz_server.go
```diff
@@ -476,36 +486,21 @@ func (s *authzServer) validateAuthorizationCredentials(context *validationContex
 
 // parseAndValidateJwtBearerToken validates the jwt signature and returns the containing claims
 func (s *authzServer) parseAndValidateJwtBearerToken(context *validationContext) error {
-	var kidHdr string
 	token, err := nutsCrypto.ParseJWT(context.rawJwtBearerToken, func(kid string) (crypto.PublicKey, error) {
-		kidHdr = kid
+		context.kid = kid
 		return s.keyResolver.ResolveKeyByID(kid, nil, resolver.NutsSigningKeyType)
-	}, jwt.WithAcceptableSkew(s.clockSkew))
+	}, v1BearerTokenProfile, jwt.WithAcceptableSkew(s.clockSkew))
```
reinkrul: shouldn't ClockSkew be handled by ParseJWT and be default?

When validating self-issued tokens, there's no expected clockskew, but for tokens issued by other systems we should always allow (the same) clock skew?

crypto/jwt_profile.go
```diff
@@ -0,0 +1,59 @@
+/*
+ * Copyright (C) 2024 Nuts community
```
reinkrul: AI fail (incorrect year)

auth/services/oauth/authz_server.go
```diff
@@ -170,6 +170,20 @@ func NewAuthorizationServer(
...
+// v1AccessTokenProfile defines JWT validation rules for v1 access tokens.
+var v1AccessTokenProfile = &nutsCrypto.JWTProfile{
+	Typ:            "at+jwt",
+	RequiredClaims: []string{jwt.ExpirationKey, jwt.IssuedAtKey, jwt.IssuerKey, jwt.SubjectKey, "service"},
+	MaxValidity:    secureAccessTokenLifeSpan,
+	Validators:     []nutsCrypto.JWTValidator{nutsCrypto.IssuerKidValidator},
+}
+
+// v1BearerTokenProfile defines JWT validation rules for v1 JWT bearer tokens (RFC003 §5.2.1).
+var v1BearerTokenProfile = &nutsCrypto.JWTProfile{
```
reinkrul, line 185: what about audience?

crypto/jwx.go
```diff
@@ -184,10 +202,49 @@ func ParseJWT(tokenString string, f PublicKeyFunc, options ...jwt.ParseOption) (
...
+	token, err := jwt.ParseString(tokenString, options...)
+	if err != nil {
+		return nil, err
+	}
+
+	if profile == nil {
+		return token, nil
+	}
+
+	// Check required claims are non-empty (WithRequiredClaim only checks existence)
```
reinkrul: then why use "WithRequiredClaim" if it will also fail on this?

vcr/verifier/signature_verifier.go
```diff
@@ -43,13 +43,32 @@ type signatureVerifier struct {
...
+// vpJWTProfile is the default JWT validation profile for Verifiable Presentations.
+// Requires nbf as an issuance-time anchor (the VP builder has always set it). exp is NOT
+// required because other nodes may produce VPs without it. Callers that want to enforce
+// a max age do so via their own checks after VerifyVP returns (see
+// validateS2SPresentationMaxValidity in auth/api/iam/s2s_vptoken.go and the discovery
+// module's PresentationMaxValidity check).
+// No IssuerKidValidator here: in this codebase VP JWTs do not set iss (spec deviation);
+// the signer identity comes from the kid header instead.
+var vpJWTProfile = &crypto.JWTProfile{
+	RequiredClaims: []string{jwt.NotBeforeKey},
```
reinkrul, line 62: we don't require expiration? VPs can't be revoked..

reinkrul (changes requested):

vcr/verifier/signature_verifier.go
stevenvegt: It is possible to create VPs without exp in our nodes by using the API. If we require it here, we potentially break that. Callers of the verify function do now enforce their own exp. See the comment.

crypto/jwx.go
stevenvegt: It gives us consistent error messages, and I like the jwt lib to do as much as possible, however, you are right and it is redundant. Will remove.

auth/services/oauth/authz_server.go
stevenvegt: Hah, yes the exp could be configured in non-strict mode, but the check was always set on max 60s. So this was a bug. I will fix it and use the same value for issuance and the max validity on check.

auth/services/oauth/authz_server.go
stevenvegt: purpose of use IS the service.

auth/services/oauth/authz_server.go
stevenvegt: They are different for some reason on the different callers

crypto/jwt_profile.go
stevenvegt: 🙃

auth/api/iam/jar.go
stevenvegt: done

auth/services/oauth/authz_server.go
stevenvegt: I've set it to 5s default and in auth the configured value is used.

auth/services/oauth/authz_server.go
stevenvegt: added in places were it is consistently set.

reinkrul (approved):

PR #4214, opened 2026-04-22
Accept explicit token endpoint on /internal/auth/v1/request-access-token

stevenvegt:
## Summary

Adds an optional `authorization_server_endpoint` field to `/internal/auth/v1/jwt-grant` and `/internal/auth/v1/request-access-token`. When provided, it is used directly as the JWT `aud` claim, bypassing the compound-service lookup on the authorizer's DID document.

  - `service` remains required and still drives the `purposeOfUse` JWT claim — this only overrides endpoint resolution.
  - No metadata discovery added in the node; the client hands over a final token endpoint URL.
  - Request schemas `CreateJwtGrantRequest` and `RequestAccessTokenRequest` are refactored onto a shared `JwtGrantBaseRequest` via `allOf` so the new field lives in one place.
  - Existing response field `authorization_server_endpoint` kept as-is (description clarified, despite the name, it's the RFC 8414 `token_endpoint`).

## Why

Preparing for the FHIR mCSD transition. Clients already perform mCSD lookups for data-access endpoints, so letting them pass the OAuth token endpoint explicitly keeps nuts-node agnostic of the discovery mechanism.

### DIFF
--- a/docs/_static/auth/v1.yaml
+++ b/docs/_static/auth/v1.yaml
@@ -503,45 +503,8 @@ components:
     #
     # Everything related to JWT Grants
     #
-    CreateJwtGrantRequest:
-      description: Request for a JWT Grant. The grant can be used during a Access Token Request in the assertion field
-      required:
-        - authorizer
-        - requester
-        - service
-        - credentials
-      properties:
-        authorizer:
-          type: string
-        requester:
-          type: string
-        identity:
-          $ref: "#/components/schemas/VerifiablePresentation"
-        service:
-          type: string
-          description: The service for which this access token can be used. The right oauth endpoint is selected based on the service.
-          example: nuts-patient-transfer
-        credentials:
-          type: array
-          items:
-            $ref: '#/components/schemas/VerifiableCredential'
-    JwtGrantResponse:
-      description: Response with a JWT Grant. It contains a JWT, signed with the private key of the requestor software vendor.
-      required:
-        - bearer_token
-        - authorization_server_endpoint
-      properties:
-        bearer_token:
-          type: string
-        authorization_server_endpoint:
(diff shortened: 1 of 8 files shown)


docs/_static/auth/v1.yaml
```diff
@@ -516,15 +516,30 @@ components:
...
+        authorization_server_endpoint:
+          type: string
+          format: uri
+          description: |
+            Optional. OAuth token endpoint URL of the authorizer (RFC 8414 `token_endpoint`).
+            When provided, this URL is used directly as the JWT `aud` claim, bypassing the
+            compound-service lookup on the authorizer's DID document. `service` remains
+            required and is still used as the `purposeOfUse` JWT claim. Intended for
+            clients that discover the endpoint out-of-band (e.g. via FHIR mCSD).
+          example: https://as.example.nl/ura/12345678/token
```
reinkrul, line 533: this is ambiguous, especially when compared to our v2/IAM API;
- in v2/IAM we have a request field `authorization_server`, which holds the Authorization Server URL (base URL, so to say), which is NOT the token endpoint. This new endpoint has a postfix `_endpoint`, so it's inconsistent
- In this v1 API it's used as Token Endpoint, but it sounds like it's the Authorization Server URL

I suggest to rename it to `token_endpoint` (or `authorizer_token_endpoint` to stay in line with the rest of the API)

stevenvegt: Well, the problem was already there in the `JwtGrantResponse`. That field contains exactly this value (as far as I understand correctly). So instead of either renaming that field which would cause a breaking change, use a different name here, which would cause an inconsistency, I opted for fixing it in docs by explicitly explaining this is the RFC 8414 token_endpoint.
Since the authorization_server field in v2 is named differently, I don't see a hard clash.

PR #4219, opened 2026-04-24
Feature/4144 add support for more scope policies to project-gf

stevenvegt:
Implements #4144


stevenvegt:
Is already merged into master and project-gf is updates with master.

PR #4220, opened 2026-04-24
Feature/4144 add support for more scope policies

stevenvegt:
Implements #4144 

Allows policies to configure more scope policies, including pass-through and dynamic. Dynamic uses a configured PDP server with an Authzen (openid auth api) API which evaluates the AT request.


auth/api/iam/s2s_vptoken.go
```diff
@@ -73,11 +76,17 @@ func (r Wrapper) handleS2SAccessTokenRequest(ctx context.Context, clientID strin
 			return nil, err
 		}
 	}
-	walletOwnerMapping, err := r.presentationDefinitionForScope(ctx, scope)
+	match, err := r.findCredentialProfile(ctx, scope)
```
reinkrul: shouldn't `match` this be named `credentialProfile`?

auth/api/iam/s2s_vptoken.go
```diff
@@ -107,15 +116,108 @@ func (r Wrapper) handleS2SAccessTokenRequest(ctx context.Context, clientID strin
...
+			Code:        oauth.ServerError,
+			Description: fmt.Sprintf("unsupported scope policy: %s", match.ScopePolicy),
+		}
+	}
+}
+
+// evaluateDynamicScopes calls the AuthZen PDP to evaluate each requested scope.
+// Returns the space-joined granted scopes. If the PDP denies the credential profile scope,
+// the request is rejected. Other denied scopes are simply excluded from the granted set.
+func (r Wrapper) evaluateDynamicScopes(ctx context.Context, match *policy.CredentialProfileMatch, subjectDID did.DID, pexState PEXConsumer) (string, error) {
```
reinkrul: I don't think this should all in the API layer, maybe have `GrantScopes(...)` as function on a `CredentialScope` interface, and implement them for every policy?

core/echo_errors.go
```diff
@@ -76,19 +85,12 @@ func CreateHTTPErrorHandler() echo.HTTPErrorHandler {
...
-		} else {
-			logger.
-				WithError(err).
-				Warn("Unable to send error back to client, response already committed")
+		errorWriter, _ := ctx.Get(ErrorWriterContextKey).(ErrorWriter)
+		if errorWriter == nil {
+			errorWriter = &problemErrorWriter{}
+		}
+		if writeError := errorWriter.Write(ctx, statusCode, title, err); writeError != nil {
+			logger.Error(err)
```
reinkrul, line 93: not part of PR? If this is a bugfix of an existing bug, make separate PR

policy/cmd.go
```diff
@@ -27,5 +27,6 @@ func FlagSet() *pflag.FlagSet {
 	defCfg := defaultConfig()
 	flagSet := pflag.NewFlagSet("policy", pflag.ContinueOnError)
 	flagSet.String("policy.directory", defCfg.Directory, "Directory to read policy files from. Policy files are JSON files that contain a scope to PresentationDefinition mapping.")
+	flagSet.String("policy.authzen.endpoint", defCfg.AuthZen.Endpoint, "Base URL of the AuthZen PDP endpoint. Required when any credential profile uses scope_policy 'dynamic'.")
```
reinkrul: maybe note what happens if it's not set, but a credential profile requires it.

reinkrul (changes requested):

auth/api/iam/s2s_vptoken.go
stevenvegt: Agreed — renamed throughout `s2s_vptoken.go` and the new granter code in bd2ccb45.

auth/api/iam/s2s_vptoken.go
stevenvegt: Done in bd2ccb45. Moved scope-policy evaluation into the `policy` package as a `ScopeGranter` interface with one implementation per mode (`profileOnlyGranter`, `passthroughGranter`, `dynamicGranter`).

Two notes on the shape:

1. **AuthZen is now an adapter, not a direct dependency.** I added a generic `policy.ScopeEvaluator` interface — `EvaluateScopes(ctx, ScopeEvaluationInput) (map[string]bool, error)` — so the dynamic granter no longer knows about AuthZen wire types. `policy.NewAuthZenScopeEvaluator` adapts an `AuthZenEvaluator` to satisfy `ScopeEvaluator`, and `LocalPDP.ScopeEvaluator()` returns the adapter (or `nil` when no PDP is configured). Future PDP backends (Rego, etc.) can ship as additional `ScopeEvaluator` implementations without touching the granter.

2. **Construction-time validation, fail-fast.** `NewScopeGranter` rejects `profile-only + extra scopes`, `dynamic without evaluator`, and unknown policies *before* the handler does VP cryptographic verification, so misconfigured requests fail without burning crypto. The handler call site shrinks to ~3 lines and no longer imports `policy/authzen`.

AuthZen request-shape coverage moved into `policy/scope_granter_test.go::TestAuthZenScopeEvaluator` (unit) and is still exercised end-to-end in `auth/api/iam/integration_test.go` against an `httptest` PDP.

Worth noting: the PRD's "Out of Scope" entry says "two modes don't warrant an interface" but we ended up with three (`passthrough` was promoted in-scope mid-design), and the layering concern stood independently. I'll update the PRD note in a follow-up.

core/echo_errors.go
stevenvegt: You're right, sorry — it slipped in because I needed the idempotent behavior for some of the integration test wiring, but it's a standalone bugfix. Dropped from this branch (force-pushed to remove commit 25c35af2). I'll open a separate PR off `master` for it.

policy/cmd.go
stevenvegt: Done in bd2ccb45. Extended the flag help to:

> "Base URL of the AuthZen PDP endpoint. Required when any credential profile uses scope_policy 'dynamic'; the node refuses to start if such a profile is configured but this flag is empty."

core/echo_errors.go
stevenvegt: Opened as #4243.

reinkrul (approved):

PR #4225, opened 2026-04-30
Add debug logging to PD matching

### LINKED ISSUE #4218: Facilitate developers with their policy development
During development process, often policies don't match with credentials in a wallet. The current debugging process is trimming the policy file until there is a match and they know which parts was causing the mismatch. Can we add tooling which allows developers to get better insights in the workings of the policy and their current wallets credentials?

reinkrul:
Closes #4218.

## Summary

When debug-level logging is enabled, every (input descriptor, credential) pair the Presentation Exchange matcher considers is logged with a human-readable reason for rejection — which field/path/filter rejected, what value was found, what the filter expected. Plus a per-input-descriptor summary line showing how many credentials were considered, how many matched, and which one was selected.

Production hot path is untouched: gated on \`logrus.IsLevelEnabled(DebugLevel)\` and only does the extra reason-building work when debug is on.

Example debug output for a non-matching VC:

\`\`\`
DEBU[…] PE: credential rejected  credential=urn:test:non-matching input_descriptor=as_jsonld module=VCR reason=\"no value found at any of paths [\$.credentialSubject.organization.city]\"
DEBU[…] PE: input descriptor evaluated  considered=2 input_descriptor=as_jsonld matched=1 module=VCR selected=\"did:nuts:CuE...#ec8af8cf-…\"
\`\`\`

A debug-only endpoint exposing the same info as a structured \`MatchReport\` was [proposed](https://github.com/nuts-foundation/nuts-node/issues/4218) and prototyped, but dropped from this PR — too much public-API surface for what is essentially a development aid. The pe-internal helpers that build the reasons stay, and we can revisit the endpoint later if logs prove insufficient.

## Drive-by fix

The existing \`log.Logger().Level >= logrus.DebugLevel\` idiom (e.g. \`auth/services/irma/signer.go\`, \`network/dag/notifier.go\`) is broken: \`Level\` on a \`*logrus.Entry\` is the per-call level (zero by default) rather than the logger's threshold, so the guard never trips. New code uses \`logrus.IsLevelEnabled(...)\`, matching the already-correct usage in \`vcr/api/vcr/v2/registry.go\`. Existing call sites are untouched — narrow scope.

## Out of scope

- The \"PD-from-VC suggester\" half of #4218 — separate follow-up.
- Touching production \`Match\` callers in \`discovery/\` or \`auth/\`.

## Test plan

- [x] \`vcr/pe\`: tests for \`explainFieldMismatch\` (4 cases) and a debug-logging test using a logrus test hook to assert the rejection + summary log entries
- [x] \`go build ./...\`, \`go test ./vcr/...\`

Assisted by AI


reinkrul:
This will be superseded by #4253

PR #4227, opened 2026-05-01
Two-VP token request flow (4078-2)

stevenvegt:
## Parent PRD

#4078

## Summary

Adds the RFC 7523 jwt-bearer two-VP token request flow inside the access-token client, gated behind `auth.experimental.jwt_bearer_client` (default off). The entry point — formerly `RequestRFC021AccessToken`, renamed to `RequestServiceAccessToken` — is now a thin dispatcher: when an SP subject identifier is supplied it builds VP1 from the HCP wallet (organization PD) and VP2 from the SP wallet (service_provider PD), then POSTs them as `assertion` and `client_assertion`. When no SP identifier is supplied, the existing single-VP `vp_token-bearer` flow runs unchanged. No public API surface changes — that lands in PR #4228.

### Entry-point reshape (`auth/client/iam/openid4vp.go`)

- `RequestRFC021AccessToken` renamed to `RequestServiceAccessToken` (matches the public HTTP endpoint name, no longer claims to do only RFC021).
- Becomes a thin dispatcher: gate the experimental flag, fetch AS metadata once, then call `requestVPTokenAccessToken` (existing single-VP) or `requestJwtBearerAccessToken` (new two-VP).
- Both grant-specific paths share `buildSubmissionForSubject` (DID listing + AS-method filtering + per-DID `additionalCredentials` expansion + wallet `BuildSubmission`) and the new `signDPoPHeader` helper.

### Two-VP path (`requestJwtBearerAccessToken`)

- Resolves both PDs from the local policy backend via the new `loadAndValidateProfile` helper. The two-VP flow is local-only — there is no AS-side mechanism to advertise a service_provider PD.
- Builds VP1 from HCP wallet → resolves the submission against VP1 → extracts id-bearing constraint field values via `pe.PresentationDefinition.ResolveConstraintsFields` → merges captured values into `credential_selection` → builds VP2 from SP wallet using the merged selection.
- Assembles the form body with `grant_type=urn:ietf:params:oauth:grant-type:jwt-bearer`, `assertion=<VP1 raw>`, `client_assertion_type`, `client_assertion=<VP2 raw>`, `scope=<resolved scope>`. **No** `presentation_submission` and **no** `client_id` (per RFC 7521 §4.2 the client is authenticated by the assertion).
- DPoP, when requested, signs the proof with VP2's holder key (the SP DID) — the SP is the OAuth client and will present the issued token.
- Scope policy mirrors `pdResolver.resolveLocal`: `profile-only` with extra scopes returns `oauth.InvalidScope`; `passthrough` forwards the full input.

### Shared helpers

- `loadAndValidateProfile` (`auth/client/iam/profile_validation.go`) — `FindCredentialProfile` + scope-policy + organization-PD-required checks. Called from both `pdResolver.resolveLocal` and the two-VP path; the rules and error messages are now defined once.
- `applyCapturedFieldsToSelection` — pure function, takes the existing selection map and the captured field values, returns a fresh map (never mutates caller's input). String-only; non-string captured values are dropped.
- `pe.PresentationSubmission.ResolveVP` (`vcr/pe/presentation_submission.go`) — convenience wrapper around `Resolve` for callers that already hold a parsed VP, so they don't need the `Raw()` + `ParseEnvelope` round-trip.
- `signDPoPHeader` — returns zero values when `useDPoP=false` so both grant paths can call it unconditionally.

### Configuration

- `auth.Config.Experimental.JwtBearerClient` (new nested `Experimental` block).
- Flag: `auth.experimental.jwt_bearer_client`, default `false`. Description carries "Subject to change without notice".

### OAuth constants (`auth/oauth/types.go`)

- `JwtBearerGrantType`, `JwtBearerClientAssertionType`, `ClientAssertionTypeParam`, `ClientAssertionParam`. The client-assertion-type lives in its own const block; it isn't a grant type.

### Errors

- The three jwt-bearer rejections (feature flag off, AS doesn't advertise, no service_provider PD) return typed `oauth.OAuth2Error` (`UnsupportedGrantType` / `InvalidScope`) so the API layer maps them to 4xx, not generic 500.

### Tests

- `TestRelyingParty_RequestServiceAccessToken_TwoVP` (orchestration) — gate, AS-advertise, missing SP-PD, happy-path form body, cross-VP binding e2e, DPoP-with-SP-key.
- `TestLoadAndValidateProfile` (unit) — every scope-policy + missing-PD scenario with just a policyBackend mock; no HTTP, wallet, or subject-manager mocks.
- `TestApplyCapturedFieldsToSelection` (unit) — merge invariants on plain maps, including the new no-mutation guarantee.
- `TestPresentationSubmission_ResolveVP` (in `vcr/pe`) — guards the new convenience method.
- All existing `TestRelyingParty_RequestServiceAccessToken` (single-VP) tests preserved and still passing.

### Bonus fix

- Both flows now return `resolvedScope` on `TokenResponse.Scope` instead of the raw input. Pre-existing bug in single-VP, mirrored across.

## How to review

1. **Start with `auth/client/iam/openid4vp.go:250-266`** — the dispatcher. ~15 lines, makes the structural shape obvious.
2. **`auth/client/iam/openid4vp.go:330-405`** — `requestJwtBearerAccessToken`. Read alongside the inline comment about cross-VP binding (the four lines where parse → resolve descriptors → extract constraint fields → merge). The `signDPoPHeader` derivation from `vp2.Holder` is the SP-as-OAuth-client decision in code form.
3. **`auth/client/iam/profile_validation.go`** — the shared profile/scope helper. Trivial to read top-to-bottom; `pdResolver.resolveLocal` becomes a 6-line caller of it.
4. **`vcr/pe/presentation_submission.go:166-179`** — the `ResolveVP` addition. Single new method, delegates to `Resolve`; fixes a previously-latent panic where the wrapped `Envelope`'s `raw` field was nil.
5. **Tests** — `TestRelyingParty_RequestServiceAccessToken_TwoVP/captured_VP1_field-id_values_flow_into_VP2_credential_selection_end-to-end` is the hardest one to read; it has line-by-line comments explaining the cross-VP capture chain.
6. **Decisions to look at**:
   - Two-VP bypasses `pdResolver` (always local — no AS advertisement standard exists). Documented in the `requestJwtBearerAccessToken` doc comment.
   - DPoP signs with the SP DID's key, not HCP. Per OAuth: the OAuth client (SP) presents the access token, so the proof must be SP-bound.
   - No OAuth `client_id` form param in jwt-bearer mode. Per RFC 7521 §4.2.
   - `additionalCredentials` flow into both wallets; each PD selects what matches. Documented on the interface.

## Deviations from spec

- **Function rename: `RequestRFC021AccessToken` → `RequestServiceAccessToken`.** The function dispatches between RFC021 and RFC 7523 now, so the old RFC021 name no longer fits. The new name matches the public API endpoint.
- **Wallet-owner-type `client` → `service_provider`** (landed in PR #4226). Avoids overloading `httpClient`/`authzenClient` and the OAuth `client_id` form parameter; matches the PRD's prose ("service provider acting on behalf of an organization").
- **DPoP wired through the two-VP path.** The PRD listed DPoP as out of scope (already implemented for single-VP), but the dispatcher accepts `useDPoP`, so silently dropping it on the two-VP path would be a footgun. The proof is signed with the SP DID's key (`vp2.Holder`).
- **No OAuth `client_id` form parameter on the jwt-bearer path.** Per RFC 7521 §4.2 the `client_assertion` authenticates the client, so `client_id` is optional. Omitting it removes the need to plumb a service-provider base URL through this PR (or PR #4228) just to fill a form field.
- **`loadAndValidateProfile` extracted as a shared helper.** Both `pdResolver.resolveLocal` and `requestJwtBearerAccessToken` had the same FindCredentialProfile + scope-policy + organization-PD checks. Extracted; the rules are now unit-tested directly instead of through orchestration in both flows.
- **`pe.PresentationSubmission.ResolveVP` added to vcr/pe.** Not in the spec; introduced because the cross-VP capture step had a parsed VP in memory and would otherwise round-trip through `Raw()` + `ParseEnvelope`. Also fixes a latent panic where the constructed `Envelope`'s `raw` was nil.
- **`TokenResponse.Scope` returns the resolved scope** (collapsed under `profile-only`, full input under passthrough) for both flows. Pre-existing bug in single-VP, fixed in passing.
- **Scope policy enforced in the two-VP path.** Mirrors `pdResolver.resolveLocal` exactly.

## Dependencies

Stacked on **#4226** (the policy config loader for the `service_provider` PD block). Should be reviewed and merged in order.

PR #4228 (API binding) and PR #4229 (integration test) depend on this PR.

## Carry-overs to PR #4228

Recorded in PR #4228's body so they aren't forgotten when implementation starts:

- Public-interface signature cleanup — drop the `*string` sentinel on `Client.RequestServiceAccessToken` (split into two methods, or reshape).
- `iam.NewClient` constructor — 9 positional args including two booleans; functional options would be cheap once the API binding touches the wiring.
- Coverage gaps — SP-side `ListDIDs` empty, second `BuildSubmission` `ErrNoCredentials`.
- Test ergonomics — TwoVP failure-mode tests still spin up a TLS server they don't need.

## Design context

- PRD #4078 — overall design including the cross-VP `field.id` binding convention and the failure-loud rule (no silent fallback when the caller opts in to two-VP).
- PRD §"Cross-VP field binding" — describes the shared-`field.id` mechanism realized here via `ResolveConstraintsFields` + `applyCapturedFieldsToSelection`.
- PRD §"Token request parameters (jwt-bearer flow, PSA 10.10.6)" — the wire format this PR's happy-path test asserts.
- PRD §"Out of Scope" — DPoP was listed there because it was already implemented for single-VP; this PR extends it to two-VP rather than letting the explicit `useDPoP` parameter silently downgrade.

<details>

Internal client-side change. Adds the two-VP `jwt-bearer` flow but does not yet expose it on the public API (no caller can reach it without PR 3). Gated behind an experimental feature flag.

#### Feature flag

- Register new config key `auth.experimental.jwt_bearer_client` (default `false`).
- The flag gates the new code path inside `RequestRFC021AccessToken`.

#### Two-VP flow inside `RequestRFC021AccessToken`

Located in `auth/client/iam/`. Extend its signature to accept an optional SP subject identifier (e.g. `clientID *string` or similar — internal API, called from API binding in PR 3).

When the SP subject identifier is **absent**: existing single-VP `vp_token-bearer` path runs unchanged.

When the SP subject identifier is **present**:
1. Fetch AS metadata. If `urn:ietf:params:oauth:grant-type:jwt-bearer` is not in `grant_types_supported` → return error.
2. Look up the `client` PD for the requested credential profile (loaded by PR 1). If absent → return error.
3. Build VP1 from the HCP wallet (existing path-param `subjectID`) using the `organization` PD. Existing `BuildSubmission` call.
4. Call the existing `resolveInputDescriptorValues` (`auth/api/iam/s2s_vptoken.go`) on VP1's selected credentials to extract `{field.id → matched value}`.
5. Additively merge captured values into the `credential_selection` map.
6. Build VP2 by calling `BuildSubmission` again with the SP subject (from `clientID`) and the `client` PD, passing the merged `credential_selection`.
7. Assemble the token request form body with grant_type, assertion, client_assertion_type, client_assertion, scope. No presentation_submission.
8. POST to the AS token endpoint and parse the token response as today.

### Modules touched

- `auth/client/iam/`
- `core/` config registration
- Reuse: `resolveInputDescriptorValues`, `NewFieldSelector` (from #4067).

## Acceptance Criteria

- [x] `auth.experimental.jwt_bearer_client` feature flag registered.
- [x] `RequestServiceAccessToken` accepts an optional SP subject identifier (renamed — see Deviations).
- [x] Two-VP flow constructs the correct jwt-bearer form body when SP identifier is present and AS supports jwt-bearer.
- [x] Single-VP flow runs unchanged when SP identifier is absent.
- [x] Cross-VP binding (shared `field.id`) works without EHR involvement.
- [x] All failure modes return clear errors (typed `oauth.OAuth2Error`).
- [x] Unit tests cover all the scenarios listed above.

</details>


auth/client/iam/openid4vp.go
```diff
@@ -54,32 +54,37 @@ var ErrPreconditionFailed = errors.New("precondition failed")
...
+	httpClient                  HTTPClient
+	jwtSigner                   nutsCrypto.JWTSigner
+	keyResolver                 resolver.KeyResolver
+	strictMode                  bool
+	wallet                      holder.Wallet
+	ldDocumentLoader            ld.DocumentLoader
+	subjectManager              didsubject.Manager
+	pdResolver                  PresentationDefinitionResolver
+	policyBackend               policy.PDPBackend
+	experimentalJwtBearerClient bool
```
reinkrul, line 70: why experimental? if it's good enough to be merged and it works, it should be fine.

If we really want to discourage people from actively using it in the Nuts node API, we could add it as warning/note in the OpenAPI spec?

auth/cmd/cmd.go
```diff
@@ -47,6 +47,9 @@ const ConfAccessTokenLifeSpan = "auth.accesstokenlifespan"
 // ConfAuthEndpointEnabled is the config key for enabling the Auth v2 API's Authorization Endpoint
 const ConfAuthEndpointEnabled = "auth.authorizationendpoint.enabled"
 
+// ConfExperimentalJwtBearerClient toggles the RFC 7523 jwt-bearer two-VP token request flow.
+const ConfExperimentalJwtBearerClient = "auth.experimental.jwt_bearer_client"
```
reinkrul: I'd do this by introducing a "granttypes" (supported grant types) config. Then you can also use that (later) to enable/disable authorized code flow, openid4vp, vp_bearer-token, etc

vcr/pe/presentation_submission.go
```diff
@@ -163,6 +163,22 @@ func (b *PresentationSubmissionBuilder) Build(format string) (PresentationSubmis
 	return presentationSubmission, signInstruction, nil
 }
 
+// ResolveVP is a convenience wrapper around Resolve for callers that already hold a single parsed
+// VerifiablePresentation in memory (e.g. a freshly built submission from the wallet) and would otherwise
+// need to round-trip through Raw() + ParseEnvelope just to call Resolve. Returns a map keyed by
+// InputDescriptor.Id; see Resolve for the underlying matching rules.
+func (s PresentationSubmission) ResolveVP(presentation vc.VerifiablePresentation) (map[string]vc.VerifiableCredential, error) {
```
reinkrul: I don't 1 invocation warrants a syntactic sugar func in this otherwise very clean API (of PresentationSubmission), since it doesn't really add functionality. Maybe better to export `vpAsInterface()` instead.

auth/client/iam/openid4vp.go
```diff
@@ -242,14 +247,36 @@ func (c *OpenID4VPClient) AccessToken(ctx context.Context, code string, tokenEnd
...
+	}
 	metadata, err := c.AuthorizationServerMetadata(ctx, authServerURL)
 	if err != nil {
 		return nil, err
 	}
+	if serviceProviderSubjectID != nil {
+		if !slices.Contains(metadata.GrantTypesSupported, oauth.JwtBearerGrantType) {
+			return nil, oauth.OAuth2Error{
+				Code:        oauth.UnsupportedGrantType,
+				Description: "authorization server does not advertise jwt-bearer support",
```
reinkrul: use spec'd grant type URI

auth/client/iam/interface.go
```diff
@@ -43,11 +43,17 @@ type Client interface {
...
+	// When serviceProviderSubjectID is non-nil it identifies a service-provider Nuts subject and triggers the RFC 7523
+	// jwt-bearer two-VP flow; that flow is only honored when the experimental jwt-bearer client feature is enabled and
+	// the AS advertises jwt-bearer.
+	// credentials are additional VCs to include alongside wallet-stored credentials. In the two-VP flow they are
+	// offered to both wallets; each PD selects what matches its input descriptors. Signed VCs flow through unchanged;
+	// unsigned self-attested credentials are auto-issued per holder DID by AutoCorrectSelfAttestedCredential.
 	// credentialSelection maps PD field IDs to expected values to disambiguate when multiple credentials match an input descriptor.
-	RequestRFC021AccessToken(ctx context.Context, clientID string, subjectDID string, authServerURL string, scopes string, useDPoP bool,
-		credentials []vc.VerifiableCredential, credentialSelection map[string]string) (*oauth.TokenResponse, error)
+	RequestServiceAccessToken(ctx context.Context, clientID string, subjectDID string, authServerURL string, scopes string, useDPoP bool,
```
reinkrul: should we rename;
-  `clientID` to `organizationSubjectID`
- `subjectDID` to `organizationSubjectDID` (or `organizationDID`)?

auth/client/iam/openid4vp.go
```diff
@@ -340,14 +325,165 @@ func (c *OpenID4VPClient) RequestRFC021AccessToken(ctx context.Context, clientID
...
 
+// requestJwtBearerAccessToken implements the RFC 7523 jwt-bearer two-VP token request flow.
+// It builds VP1 from the healthcare-provider (HCP) wallet using the organization PD, and VP2 from the
+// service-provider (SP) wallet using the service_provider PD, assembles them as `assertion` and
+// `client_assertion`, and POSTs the token request.
+// Per RFC 7521 §4.2 the client is authenticated by the client_assertion, so no OAuth client_id form
+// parameter is sent on this path.
+// Both PDs are resolved from the local policy backend; the AS's remote presentation_definition endpoint
+// is not consulted (no standardised mechanism exists today for the AS to advertise a service_provider PD).
+func (c *OpenID4VPClient) requestJwtBearerAccessToken(ctx context.Context, subjectID string, serviceProviderSubjectID string,
```
reinkrul, line 348: `subjectID` -> `organizationSubjectID`

auth/client/iam/openid4vp.go
```diff
@@ -340,14 +325,165 @@ func (c *OpenID4VPClient) RequestRFC021AccessToken(ctx context.Context, clientID
...
+	if !hasSP {
+		return nil, oauth.OAuth2Error{
+			Code:        oauth.InvalidScope,
+			Description: fmt.Sprintf("no service_provider presentation definition for scope %q", profile.CredentialProfileScope),
+		}
+	}
+	params := holder.BuildParams{
+		Audience:   authServerURL,
+		DIDMethods: metadata.DIDMethodsSupported,
+		Expires:    time.Now().Add(time.Second * 5),
```
reinkrul: don't we have a constant for this?

auth/client/iam/openid4vp.go
```diff
@@ -340,14 +325,165 @@ func (c *OpenID4VPClient) RequestRFC021AccessToken(ctx context.Context, clientID
...
+		Format:     metadata.VPFormatsSupported,
+		Nonce:      nutsCrypto.GenerateNonce(),
+	}
+	vp1, vp1Submission, err := c.buildSubmissionForSubject(ctx, subjectID, orgPD, additionalCredentials, credentialSelection, params)
```
reinkrul: can we use "clientVP" and "subjectVP", or "serviceProviderVP" and "organizationVP"? VP1 and VP2 are really non-descriptive

auth/client/iam/openid4vp.go
```diff
@@ -340,14 +325,165 @@ func (c *OpenID4VPClient) RequestRFC021AccessToken(ctx context.Context, clientID
...
+	// Cross-VP binding: capture id-bearing constraint field values resolved against VP1 and additively merge
+	// them into the credential_selection map for VP2. The submission tells us which credential satisfied each
+	// input descriptor; we use that to walk the PD's id-bearing fields and extract their matched values.
+	credentialMap, err := vp1Submission.ResolveVP(*vp1)
+	if err != nil {
+		return nil, fmt.Errorf("resolve VP1 submission for cross-VP binding: %w", err)
+	}
+	captured, err := orgPD.ResolveConstraintsFields(credentialMap)
+	if err != nil {
+		return nil, fmt.Errorf("resolve VP1 constraint fields for cross-VP binding: %w", err)
```
reinkrul: EHR developers will have no idea what they're doing wrong here

reinkrul (changes requested):

auth/client/iam/openid4vp.go
stevenvegt: I want this in master, but want to restrict it to lsp x nuts use cases. This also makes it very clear that we can change the api at any moment.

stevenvegt: Or, rewrite the flag to a list of support grant type (like we have for did methods and auth means).

auth/client/iam/openid4vp.go
stevenvegt: omschrijven naar `urn:ietf:params:oauth:grant-type:jwt-bearer`

vcr/pe/presentation_submission.go
stevenvegt: so, instead of calling resolveVP on the PS, create a newEnvelopFromVP and keep the resolve the same.

auth/client/iam/interface.go
stevenvegt: subjectDID already renamed to subjectID in the implementation, since we don't use did's in the API's anymore.

stevenvegt: the rest is still accurate, I propose to keep it the same. They come from the request param subjectID.

auth/cmd/cmd.go
stevenvegt: Agreed, but threading this through properly means also driving `grant_types_supported` in the AS metadata from the same config — otherwise enabling/disabling here doesn't match what we advertise to peers. That reaches further than this PR, so I've filed #4231 to do it for all grant types in one go and we'll drop the experimental flag there.

auth/client/iam/openid4vp.go
stevenvegt: Tracked in #4231 alongside the granttypes refactor — once that lands the flag goes away and operators opt in by adding the grant type to the enabled list.

vcr/pe/presentation_submission.go
stevenvegt: Agreed — dropped `ResolveVP` and added `pe.NewEnvelopeFromVP` instead, so the call site does `NewEnvelopeFromVP(vp).Resolve(submission)` against the existing API. (c80a2ee1)

auth/client/iam/openid4vp.go
stevenvegt: Done — error message now includes the full `urn:ietf:params:oauth:grant-type:jwt-bearer` URN. (4f592ca6)

auth/client/iam/openid4vp.go
stevenvegt: Hoisted into a named `vpAssertionLifetime = 5 * time.Second` const next to the helper. (d88dc8f9)

auth/client/iam/openid4vp.go
stevenvegt: Renamed throughout: `vp1` → `organizationVP`, `vp2` → `serviceProviderVP` (same for the corresponding `*Submission` locals). (7348a9c6)

auth/client/iam/openid4vp.go
stevenvegt: Made the cross-VP binding errors specific (which step failed: envelope wrap, submission resolve, or constraint extraction) and `matchField` now wraps mismatch errors with the JSONPath, so the chained message points at the field that didn't match. (410e67ae)

auth/client/iam/interface.go
stevenvegt: Partly addressed: renamed `subjectDID` → `subjectID` (the parameter has always been a subject identifier, not a DID). (e898ff3c)

The rest of the rename (`clientID` → `organizationSubjectID`, plus dropping the `*string` sentinel param) is deferred to the public-interface cleanup tracked under #4228 — the API binding lands there and that's the natural moment to reshape the signature, since this PR has no caller that exercises both the single-VP and two-VP shapes.

auth/client/iam/openid4vp.go
stevenvegt: Same as #discussion_r3181443225 — the broader `organizationSubjectID` rename is part of the public-interface cleanup deferred to #4228. The interface-level `subjectDID` was the misleading one and is fixed in e898ff3c.

reinkrul (approved):

PR #4236, opened 2026-05-07
auth: forward credential_request_params as overlay on OpenID4VCI Credential Request

### LINKED ISSUE #4233: OpenID4VCI: Add credential_details passthrough on the credential-issuance API
## Context

We want the Nuts node wallet to fetch credentials from the AET SDK's OpenID4VCI Credential Endpoint as part of the LSPxNuts flows tracked in #4188 (`HealthcareProfessionalDelegationCredential`, `PatientEnrollmentCredential` issued via UZI-pas through AET).

AET is spec-compliant but accepts additional fields in the Credential Request body beyond those OpenID4VCI defines. Example body:

```http
POST https://aet.example.com/credential
Authorization: Bearer 72e73cff-...
Content-Type: application/json

{
  "credential_identifier": "HealthCareProfessionalDelegationCredential",
  "did": "did:web:example.com:iam:d8e49bf4-9b89-42b5-bd5f-05b720eb1e74",
  "bsn": "900184590",
  "ura": "900030757"
}
```

The Nuts holder client (`auth/client/iam/client.go:VerifiableCredentials`) builds a closed `CredentialRequest` struct containing only `proof`. There's no surface for `credential_identifier` or for issuer-specific extras like `did` / `bsn` / `ura`.

We don't want to encode AET-specifics in the node, and we don't want every future issuer that accepts custom fields to land as a Go-coded "profile". The cleaner shape is to mirror the spirit of RFC 9396's `authorization_details`: a JSON object the EHR supplies on the internal API, which the node uses as the **base body** of the Credential Request.

## Problem

The internal credential-issuance API can't carry issuer-specific request fields, and the OpenID4VCI client can't emit a body that deviates from the closed `CredentialRequest` shape.

- `auth/api/iam/openid4vci.go:RequestOpenid4VCICredentialIssuance` has no extension point for the credential request body.
- `auth/client/iam/client.go:328` marshals a fixed `CredentialRequest{Proof: ...}`. No way to add fields.
- Result: AET integration is impossible without code changes, and every future issuer with custom fields would need its own.

## Proposal

Add an optional `credential_details` field (JSON object) to the credential-issuance API. The EHR populates it with whatever fields the target Credential Endpoint expects. The node uses it as the **base** body of the outgoing Credential Request, layers the node-built `proof` on top, and POSTs the result.

### API surface

`POST /internal/auth/v2/{subjectID}/request-credential` gains an optional `credential_details`:

```json
{
  "issuer": "https://aet.example.com",
  "credential_details": {
    "credential_identifier": "HealthCareProfessionalDelegationCredential",
    "did": "did:web:nuts.reinkrul.nl:iam:d8e49bf4-9b89-42b5-bd5f-05b720eb1e74",
    "bsn": "900184590",
    "ura": "900030757"
  }
}
```

When the issuer needs no fields beyond the JWT `proof`, the EHR omits `credential_details` and the node behaves as today. When the issuer expects spec-defined fields like `credential_identifier`, or accepts extras like AET's `did` / `bsn` / `ura`, the EHR supplies them via `credential_details`. The node treats the object as opaque from a schema standpoint.

`wallet_did` handling is unchanged — out of scope for this issue.

### Wire body construction

Final body sent on the wire =
**`credential_details` (EHR-provided) ⊕ `{ "proof": <node-built JWT> }`**.

Rules:

1. The node always injects its own `proof` (OpenID4VCI v1.0 §8.2). EHR-supplied `proof` is overwritten.
2. Otherwise `credential_details` is forwarded verbatim.
3. The node never auto-injects the holder DID. If an issuer needs it in-body, the EHR provides it explicitly via the field the issuer expects (e.g. `credential_details.did` for AET).

### Client extension

`auth/client/iam/client.go:VerifiableCredentials` changes its body construction from a closed `CredentialRequest` struct to a `map[string]any` seeded from `credential_details`, with `proof` set last. Signature gains `credentialDetails map[string]any`.

`auth/api/iam/openid4vci.go:RequestOpenid4VCICredentialIssuance` threads `credential_details` from the API into the OAuth session (`auth/api/iam/session.go:OAuthSession`) so it's available at the credential-fetch step after the redirect.

### Failure modes (fail loud)

- `credential_details` is not a JSON object → 400 at the API.
- Issuer Credential Endpoint returns non-2xx → propagate the body verbatim in the error. No retry with a different shape.

### Security considerations

- `credential_details` is opaque to the node and may contain PII (e.g. BSN). The node does not log the Credential Request body or `credential_details` contents. **Trade-off:** this is itself a security concern — operators lose forensic visibility when issuance fails or returns unexpected results, and the node-side audit trail is reduced to "a request was sent to issuer X". The EHR remains the authoritative source for what was requested.
- BSN forwarding: BSNs supplied via `credential_details` go out over TLS (existing strict client) and are stored in the `OAuthSession` between the redirect and credential-fetch steps. The session must be cleared on completion or expiry — verify against existing session GC, don't add new persistence.
- No new outbound surface. The request still goes to the same Credential Endpoint URL the issuer's metadata advertises.

### Why not a per-issuer "profile" field

Considered: `issuer_profile: { name: "aet", params: {...} }` with an `IssuerProfile` interface in code. Rejected — it encodes AET-specific knowledge in the node and needs a new code path for every future issuer with custom fields. `credential_details` keeps issuer-specific knowledge in the EHR.

## Scope

- API spec (`docs/_static/auth/v2.yaml` or wherever `request-credential` lives): add optional `credential_details`.
- `auth/api/iam/openid4vci.go`: accept `credential_details`, validate it's a JSON object, persist into `OAuthSession`.
- `auth/api/iam/session.go`: extend `OAuthSession` with `CredentialRequestDetails`.
- `auth/client/iam/client.go:VerifiableCredentials`: rebuild request body as `details ⊕ proof`; update interface, mock, callers.
- `auth/client/iam/interface.go`, `auth/client/iam/mock.go`: signature change.
- Tests: existing default behavior unchanged, `credential_details` passthrough end-to-end, AET wire-format assertion.
- Docs and release notes.

## Considerations

- **Field name `credential_details` vs `credential_request_extras`?** Position: keep `credential_details` — the analogy to `authorization_details` is the strongest mental model and no OpenID4VCI spec field collides with it. Revisit if review surfaces a clearer name.
- **Validate `credential_details` against issuer metadata?** Position: no. The whole point is passthrough; validation belongs at the issuer.

## Related

- Parent context: #4188 (LSP/AORTA implementation guides)

reinkrul:
Closes #4233.

## Summary

- Adds optional ``credential_request_params`` (JSON object) to ``POST /internal/auth/v2/{subjectID}/request-credential``.
- The object is overlaid on top of the node-built OpenID4VCI Credential Request body — any field set here (including ``credential_configuration_id`` / ``credential_identifier`` / ``proofs``) overrides the node's default. The caller takes responsibility for the resulting wire shape.
- Persisted on ``OAuthSession.CredentialRequestParams`` between the redirect and credential-fetch steps.
- ``RequestCredentialOpts`` in ``auth/openid4vci`` gains a ``CredentialRequestParams`` field; ``RequestCredential`` builds the body as ``map[string]any`` and overlays the caller's values last.

Lets the wallet talk to issuers that accept additional fields, or override the credential request entirely.

## Wire body

Final body sent to the issuer's Credential Endpoint:

```
node defaults (credential_*, proofs) ⊕ credential_request_params (EHR-provided)
```

## Test plan

- [x] ``ok - credential_request_params persisted into session`` asserts the field round-trips through the session store
- [x] ``ok - credential_request_params from session forwarded to credential endpoint`` asserts the value reaches the credential client
- [x] ``CredentialRequestParams overrides node-built defaults`` asserts the wire body: caller values win over node defaults for spec-defined fields
- [x] ``go build ./...``, ``go test ./auth/...``

Assisted by AI

### DIFF
--- a/auth/api/iam/generated.go
+++ b/auth/api/iam/generated.go
@@ -248,6 +248,13 @@ type RequestOpenid4VCICredentialIssuanceJSONBody struct {
 	// issuance per call and only consumes the first entry.
 	AuthorizationDetails []AuthorizationDetail `json:"authorization_details"`
 
+	// CredentialRequestParams Optional JSON object overlaid on top of the OpenID4VCI Credential Request body sent to
+	// the issuer's credential endpoint. Any field supplied here overrides the node's default —
+	// including credential_configuration_id, credential_identifier and proofs. Use this for
+	// issuers that diverge from the OpenID4VCI 1.0 Credential Request shape; the caller is
+	// responsible for the resulting wire shape (§8.2 mutual exclusivity, proof binding, etc.).
+	CredentialRequestParams *map[string]interface{} `json:"credential_request_params,omitempty"`
+
 	// Issuer The OAuth Authorization Server's identifier, that issues the Verifiable Credentials, as specified in RFC 8414 (section 2),
 	// used to locate the OAuth2 Authorization Server metadata.
 	Issuer string `json:"issuer"`
--- a/auth/api/iam/openid4vci.go
+++ b/auth/api/iam/openid4vci.go
@@ -91,6 +91,11 @@ func (r Wrapper) RequestOpenid4VCICredentialIssuance(ctx context.Context, reques
 	// metadata fetches.
 	authorizationDetails, _ := json.Marshal(request.Body.AuthorizationDetails)
 	credentialConfigID := request.Body.AuthorizationDetails[0].CredentialConfigurationId
+	// Capture optional credential_request_params, used as the base body of the Credential Request later in the flow.
+	var credentialRequestParams map[string]any
+	if request.Body.CredentialRequestParams != nil {
+		credentialRequestParams = *request.Body.CredentialRequestParams
+	}
 	// Generate the state and PKCE
 	state := crypto.GenerateNonce()
 	pkceParams := generatePKCEParams()
@@ -113,6 +118,7 @@ func (r Wrapper) RequestOpenid4VCICredentialIssuance(ctx context.Context, reques
 		IssuerNonceEndpoint:             credentialIssuerMetadata.NonceEndpoint,
 		IssuerCredentialConfigurationID: credentialConfigID,
 		IssuerCredentialIssuer:          credentialIssuerMetadata.CredentialIssuer,
+		CredentialRequestParams:         credentialRequestParams,
 	})
 	if err != nil {
 		return nil, fmt.Errorf("failed to store session: %w", err)
@@ -238,6 +244,7 @@ func (r Wrapper) requestCredentialWithProof(ctx context.Context, oauthSession *O
 		CredentialConfigurationID: oauthSession.IssuerCredentialConfigurationID,
 		CredentialIdentifier:      credentialIdentifier,
 		ProofJWT:                  proofJWT,
+		CredentialRequestParams:   oauthSession.CredentialRequestParams,
 	})
 }
 
--- a/auth/api/iam/openid4vci_test.go
+++ b/auth/api/iam/openid4vci_test.go
@@ -78,6 +78,22 @@ func TestWrapper_RequestOpenid4VCICredentialIssuance(t *testing.T) {
 		assert.Equal(t, "code", redirectUri.Query().Get("response_type"))
 		assert.Equal(t, `[{"credential_configuration_id":"UniversityDegreeCredential","format":"vc+sd-jwt","type":"openid_credential"}]`, redirectUri.Query().Get("authorization_details"))
 	})
+	t.Run("ok - credential_request_params persisted into session", func(t *testing.T) {
+		ctx := newTestClient(t)
+		ctx.openid4vciClient.EXPECT().OpenIDCredentialIssuerMetadata(nil, issuerClientID).Return(&metadata, nil)
+		ctx.iamClient.EXPECT().AuthorizationServerMetadata(nil, authServer).Return(&authzMetadata, nil)
+		params := map[string]interface{}{"bsn": "900184590"}
+		req := requestCredentials(holderSubjectID, issuerClientID, redirectURI)
+		req.Body.CredentialRequestParams = &params
+
+		response, err := ctx.client.RequestOpenid4VCICredentialIssuance(nil, req)
+
+		require.NoError(t, err)
+		redirectUri, _ := url.Parse(response.(RequestOpenid4VCICredentialIssuance200JSONResponse).RedirectURI)
+		var stored OAuthSession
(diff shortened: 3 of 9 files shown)


stevenvegt:
The LSPxNuts spec called this the `credential_subject_data` field and the contents are limited to information which could/should be put into the subject. Example request to the `credential` endpoint would look like this:

```json
{
  "credential_configuration_id": "PatientEnrollmentCredential",
  "proof": {
    "proof_type": "jwt",
    "jwt": "eyJ0eXAiOiJvcGVuaWQ0dmNpLXByb29mK2p3dCIsImFsZyI6IkVTMjU2Iiwia2lkIjoiZGlkOndlYjpodWlzYXJ0cy1kZWxpbmRlbi5ubCNhdXRoLWtleS0xIn0.eyJhdWQiOiJodHRwczovL3Nkay1zZXJ2aWNlLmludGVybi56b3JnaW5zdGVsbGluZy5ubCIsImlhdCI6MTc0MDAwMDAwMCwibm9uY2UiOiJhYmMxMjMifQ.kort-handtekening"
  },
  "credential_subject_data": {
    "@context": "http://gis-nl.example/context/v1",
    "@type": "HealthcareProvider",
    "hasEnrollment": {
      "patient": {
        "@type": "Patient",
        "identifier": {
          "@type": "Identifier",
          "system": "http://fhir.nl/fhir/NamingSystem/bsn",
          "value": "999911234"
        }
      }
    }
  }
}
```
I see why this currently works, since the SDK doesn't implement this API yet, but I'm reluctant to merge this into master just because we need it now, knowing it shape is sub optimal and will change any moment.

stevenvegt (commented):
Needs docs for the nuts-node.readthedocs.

You dropped the EXPERIMENTAL from the endpoint description, why?

Do I understand correctly that in order to issue the LSP credentials, I can just put this

```json
"credential_details": {
  "credential_subject_data": {
    "@context": "http://gis-nl.example/context/v1",
    "@type": "HealthcareProvider",
    "hasEnrollment": {
      "patient": {
        "@type": "Patient",
        "identifier": {
          "@type": "Identifier",
          "system": "http://fhir.nl/fhir/NamingSystem/bsn",
          "value": "999911234"
        }
      }
    }
  }
}
```
in the body, and that this will be added to the credential request?

Since this is overwriting any value in the credential request, shouldn't this be named something like `credential_request_overwrites`?

stevenvegt (approved):

PR #4246, opened 2026-05-11
Centralise outbound URL validation in StrictHTTPClient

### LINKED ISSUE #4244: Centralise outbound URL validation in StrictHTTPClient
## Background

PR #4057 introduced a new `auth/openid4vci.Client` that validates outbound URLs via `core.ParsePublicURL` to address a CodeQL SSRF finding (see [thread](https://github.com/nuts-foundation/nuts-node/pull/4057#discussion_r3209975083)). Rein noted the validation conceptually belongs on the shared HTTP transport so every outbound call benefits, not just OpenID4VCI.

## Goal

Centralise URL validation in `http/client.StrictHTTPClient.Do` so all outbound HTTP requests are checked against `core.ParsePublicURL` (HTTPS-only, no IP, no RFC 2606 reserved hosts when `httpclient.StrictMode` is true). Remove the per-caller validation that duplicates this today.

## Current state

Existing callers of `StrictHTTPClient` and what they validate today:

| Caller | Today | After centralisation |
| --- | --- | --- |
| `auth/openid4vci.Client.validateURL` | Full `core.ParsePublicURL` on every call | Remove (dedupe). Drop `strictMode` from `NewClient`. |
| `auth/client/iam.HTTPClient` metadata methods (`OAuthAuthorizationServerMetadata`, `ClientMetadata`, `OpenIDConfiguration`) | Validate via `oauth.IssuerIdToWellKnown` → `core.ParsePublicURL` | Could keep belt-and-suspenders or remove. |
| `auth/client/iam.HTTPClient` lower-level methods (`RequestObjectByGet`/`Post`, `AccessToken`, `PresentationDefinition`, redirect posts) | None at this layer; rely on `OpenID4VPClient` callers to validate first | Inherit consistent validation. |
| `auth/services/oauth/relying_party.RequestRFC003AccessToken` | Scheme-only check (`strictMode && Scheme != https`) | Inherits IP/reserved blocking. |
| `discovery/api/server/client/http.go`, `discovery/module.go` | None | Inherit. |
| `vcr/openid4vci/identifiers.go`, `vcr/vcr.go` (issuer/wallet/StatusList) | None | Inherit. |
| `vdr/didweb.Resolver` | None (constructs HTTPS URLs from `did:web` by convention) | Inherit (no functional impact expected). |

## Implementation sketch

```go
// http/client/client.go
func (s *StrictHTTPClient) Do(req *http.Request) (*http.Response, error) {
    if _, err := core.ParsePublicURL(req.URL.String(), StrictMode); err != nil {
        return nil, fmt.Errorf("httpclient: invalid target URL: %w", err)
    }
    req.Header.Set("User-Agent", core.UserAgent())
    // ... rest unchanged
}
```

Redirects are followed by `http.Client` up to 10 times without re-running `Do`, so the initial-URL check alone leaves a redirect-based SSRF gap (302 → `169.254.169.254/...`). Set `Client.CheckRedirect` to re-run `core.ParsePublicURL` on each redirect target.

Plus follow-up cleanup in `auth/openid4vci.Client` (drop `strictMode` + `validateURL`), the relying-party scheme check, and the `oauth.IssuerIdToWellKnown` callers if dedupe is preferred.

## Compatibility notes

- **Non-strict mode**: `core.ParsePublicURL(_, false)` calls `ParsePublicURLWithScheme(input, true, "http", "https")`, so it still allows IPs and reserved TLDs but rejects non-http(s) schemes. Grep confirms no `file://`/`ftp://` outbound calls exist, so this is a no-op in practice.
- **Tests**: most tests use `httptest.NewServer` (127.0.0.1) under `StrictMode=false`. With `allowReserved=true` in non-strict mode IPs are accepted, so existing tests should stay green. Verify per-package after the change.
- **Strict-mode tightening**: today strict mode only requires HTTPS; after this change it also rejects IP literals and RFC 2606 reserved TLDs. Whether any production config (RFC003 relying party, Discovery, VCR node-to-node) relies on such hosts is best raised on the PR.

## TODO marker in code

`auth/openid4vci/client.go:validateURL` has a TODO referencing this work.

JorisHeadease:
Closes #4244.

## What changed

`http/client.StrictHTTPClient.Do` and its `CheckRedirect` policy now validate every outbound URL and every redirect target via `core.ParsePublicURLAllowIP` (HTTPS-only + no RFC 2606 reserved hostnames in strict mode), instead of each caller duplicating that check.

Per-caller URL validation removed from:
- `auth/openid4vci.Client`
- `auth/client/iam.HTTPClient.ClientMetadata`
- `auth/client/iam.OpenID4VPClient` (6 call sites: `PostError`, `PostAuthorizationResponse`, `PresentationDefinition`, `RequestObjectByGet`, `RequestObjectByPost`, `AccessToken`)
- `auth/oauth.FetchMetadata` / `wellKnownCandidates` / `IssuerIdToWellKnown`
- `auth/services/oauth.relyingParty.RequestRFC003AccessToken` (scheme-only check)

All of these routed through `*StrictHTTPClient` already, so the checks were purely duplicated.

## IP-literal hosts are deliberately exempt from this check

`core.ParsePublicURL`'s "hostname is IP" rejection is **not** part of the centralized check. Blocking IP-literal URLs at this layer would make `http.client.allowedinternalcidrs` (#4420) unreachable for IP-literal targets — that config option exists precisely to let internal flows target a private address, and the guard that honors it is the dial-time SSRF guard (`denyNonPublicAddr`), not a URL-string check. So IP addresses are left entirely to that guard, which already runs on every connection (including redirect hops) and is allowlist/denylist-aware.

A new `core.ParsePublicURLAllowIP` behaves like `core.ParsePublicURL` except it never rejects an IP-literal host, for this reason.

## Breaking API change

`oauth.NewRelyingParty`, `openid4vci.NewClient`, `oauth.FetchMetadata`, `oauth.IssuerIdToWellKnown` no longer accept a `strictMode`/`strictmode` parameter. `iam.ClientConfig.StrictMode` and `iam.OpenID4VPClient`'s equivalent field are removed. Strict-mode behavior is controlled exclusively by `http/client.StrictMode`, set at startup from `core.ServerConfig.Strictmode`.

## Side effect: fixes a live gap

`auth.Auth.strictMode` was a dead field — declared, threaded into `iam.ClientConfig.StrictMode` and `openid4vci.NewClient`, but never assigned, so it was always `false`. That meant the `openid4vci` and IAM client-metadata URL-validation paths never actually enforced strict-mode SSRF rules regardless of the real `strictmode` config value. Removing it in favor of the correctly-wired `http/client.StrictMode` global fixes that.

## Out of scope

DNS rebinding and IP-literal SSRF are handled by #4420's dial guard, already merged. This PR only removes the now-redundant string-level validation duplicated across callers.


reinkrul:
We need to check: I think CRLs are sometimes published on plain HTTP endpoints (instead of HTTPS).

@JorisHeadease could you check that?

reinkrul:
CRL fetching (`pki/validator.go`) doesn't go through `StrictHTTPClient` at all — it uses `http.DefaultTransport` directly (see #4422), so it's unaffected by this PR's centralized URL validation. #4422 tracks bringing it onto the shared client with an HTTP-allowed variant; not addressed here to keep this PR to pure deduplication.

reinkrul (approved):

PR #4247, opened 2026-05-12
fix(vcr/verifier): handle unparseable issuer DID instead of nil deref

### LINKED ISSUE #4235: vcr/verifier: nil-pointer panic on unparseable issuer DID
## Description

When verifying a credential whose `issuer` field can't be parsed as a DID, the verifier discards the parse error and dereferences a nil pointer on the next line.

## Location

`vcr/verifier/verifier.go` (lines around 156–167):

```go
issuerDID, _ := did.ParseDID(credentialToVerify.Issuer.String())  // err discarded
// ... a few lines ...
_, _, err = v.didResolver.Resolve(*issuerDID, &metadata)           // panic if issuerDID == nil
```

The `_` on `did.ParseDID` swallows any parse error; if the input isn't a valid DID URI per RFC 3986, `issuerDID` is `nil` and the dereference at the `Resolve` call crashes the whole node (empty connection reply on the originating HTTP request, `runtime error: invalid memory address or nil pointer dereference` in logs).

## Reproduction

Any external issuer that emits a non-RFC3986 `did:x509`. The original case we hit:

The AET ZORG-ID SDK, signing with the shipped `test-nuts.zorginstelling-01.nl` soft cert, emits this `iss`:

```
did:x509:0:sha256:<hash>::subject:O:Tést Zorginstelling 01::san:otherName:2.16.528.1.1007.99.2110-1-…, othername: Permanent Identifier:<unsupported>
```

The trailing `…, othername: Permanent Identifier:<unsupported>` (literal spaces, comma, angle brackets) makes `did.ParseDID` return `(nil, err)`. Posting that JWT VC to `/internal/vcr/v2/holder/{subject}/vc` panics the verifier.

A bug on the issuer's side (separately reported to AET) — but a malformed-input panic is a nuts-side defect regardless.

## Proposed fix

Surface the parse error instead of dropping it:

```diff
-    issuerDID, _ := did.ParseDID(credentialToVerify.Issuer.String())
+    issuerDID, err := did.ParseDID(credentialToVerify.Issuer.String())
+    if err != nil {
+        return fmt.Errorf("could not parse issuer DID: %w", err)
+    }
```

Turns the panic into a clean validation error.

SAY-5:
Closes #4235

`Verify` discarded the error from `did.ParseDID(credentialToVerify.Issuer.String())` and then dereferenced the resulting (possibly nil) `*did.DID` at `v.didResolver.Resolve(*issuerDID, &metadata)`, panicking the node when an external issuer emitted a non-RFC3986 `did:x509`.

Surface the parse error as a regular validation failure (`"could not validate issuer: ..."`) so the request fails cleanly with a 4xx instead of crashing the verifier goroutine.

Added a regression test under `TestVerifier_Verify/with_signature_check` that wraps the call in `require.NotPanics` and asserts the error is surfaced. The test panics on `master` and passes with this change.

### DIFF
--- a/vcr/verifier/verifier.go
+++ b/vcr/verifier/verifier.go
@@ -155,7 +155,7 @@ func (v verifier) Verify(credentialToVerify vc.VerifiableCredential, allowUntrus
 	if checkSignature {
 		issuerDID, err := did.ParseDID(credentialToVerify.Issuer.String())
 		if err != nil {
-			return fmt.Errorf("could not validate issuer: %w", err)
+			return fmt.Errorf("could not parse issuer: %w", err)
 		}
 		metadata := resolver.ResolveMetadata{ResolveTime: validAt, AllowDeactivated: false}
 		rawJwt := credentialToVerify.Raw()
--- a/vcr/verifier/verifier.go
+++ b/vcr/verifier/verifier.go
@@ -153,7 +153,10 @@ func (v verifier) Verify(credentialToVerify vc.VerifiableCredential, allowUntrus
 
 	// Check signature
 	if checkSignature {
-		issuerDID, _ := did.ParseDID(credentialToVerify.Issuer.String())
+		issuerDID, err := did.ParseDID(credentialToVerify.Issuer.String())
+		if err != nil {
+			return fmt.Errorf("could not validate issuer: %w", err)
+		}
 		metadata := resolver.ResolveMetadata{ResolveTime: validAt, AllowDeactivated: false}
 		rawJwt := credentialToVerify.Raw()
 		if rawJwt != "" {
--- a/vcr/verifier/verifier_test.go
+++ b/vcr/verifier/verifier_test.go
@@ -147,7 +147,7 @@ func TestVerifier_Verify(t *testing.T) {
 			require.NotPanics(t, func() {
 				validationErr := ctx.verifier.Verify(malformed, true, true, nil)
 				assert.Error(t, validationErr)
-				assert.Contains(t, validationErr.Error(), "could not validate issuer")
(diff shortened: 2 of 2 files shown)


reinkrul (changes requested):
@SAY-5 this repository requires signed commits, I can't merge otherwise

vcr/verifier/verifier.go
```diff
@@ -153,7 +153,10 @@ func (v verifier) Verify(credentialToVerify vc.VerifiableCredential, allowUntrus
 
 	// Check signature
 	if checkSignature {
-		issuerDID, _ := did.ParseDID(credentialToVerify.Issuer.String())
+		issuerDID, err := did.ParseDID(credentialToVerify.Issuer.String())
+		if err != nil {
+			return fmt.Errorf("could not validate issuer: %w", err)
```
reinkrul:
```suggestion
			return fmt.Errorf("could not parse issuer: %w", err)
```

vcr/verifier/verifier_test.go
```diff
@@ -120,6 +120,36 @@ func TestVerifier_Verify(t *testing.T) {
...
+				if tp.String() == "VerifiableCredential" {
+					malformed.Type = append(malformed.Type, tp)
+				}
+			}
+			ctx.store.EXPECT().GetRevocations(*malformed.ID).Return(nil, ErrNotFound)
+
+			require.NotPanics(t, func() {
+				validationErr := ctx.verifier.Verify(malformed, true, true, nil)
+				assert.Error(t, validationErr)
+				assert.Contains(t, validationErr.Error(), "could not validate issuer")
```
reinkrul:
```suggestion
				assert.Contains(t, validationErr.Error(), "could not parse issuer")
```

PR #4250, opened 2026-05-12
ci: move Go test pipeline from CircleCI to GitHub Actions

### LINKED ISSUE #4249: CI: switch from CircleCI to GitHub Actions
The unit-test / coverage pipeline (`.circleci/config.yml`) is the last thing on CircleCI — everything else already runs on GitHub Actions. Consolidate.

## Scope

- Replace `.circleci/config.yml` with a GHA workflow running `go test ./... -race` on PRs and `master` pushes.
- Upload coverage to Qlty via `qltysh/qlty-action/coverage` (supports OIDC, no token secret needed).
- Delete `.circleci/config.yml`; update CircleCI build badge in `README.rst`; disable the CircleCI project once green.

## Reference

- `nuts-admin/.github/workflows/go-test.yaml` for a minimal Go-on-GHA template.
- Single-runner port is fine; the current 8-way `circleci tests split` and coverage merge drop out. Shard via matrix later only if needed.

reinkrul:
## Summary

- Replaces `.circleci/config.yml` with `.github/workflows/go-test.yaml`: `go test ./... -race` on PRs and pushes to `master` / `V*`.
- Uploads filtered coverage to Qlty Cloud via `qltysh/qlty-action/coverage` (SHA-pinned to v2.2.0) with OIDC — no token secret.
- `actions/checkout` runs with `persist-credentials: false` (zizmor finding).
- Updates the build-status badge in `README.rst` / `README_template.rst` to point at the new workflow.

Single-runner port — drops the 8-way `circleci tests split` + coverage merge. Shard via matrix later only if wall-clock becomes a problem.

## Drive-by fixes required to keep the new pipeline green

The GHA `ubuntu-latest` runner has different filesystem readdir order and uses systemd-resolved (different DNS error strings) than CircleCI's `cimg/go` image, which exposed two existing brittlenesses. Happy to split these into separate PRs if preferred.

- `policy/local.go`: `dir.Readdir(0)` → `os.ReadDir(directory)`. Directory entries now load in lexical order, so duplicate-scope detection reports the same file deterministically across platforms.
- `crypto/storage/vault/vault_test.go`, `didman/api/v1/client_test.go`: drop `assert.Regexp` against `no such host|Temporary failure in name resolution`. These matched on Go's `net` package error text and verified nothing about the code under test; `assert.Error(t, err)` + nil-result checks remain. Matches the pattern other "wrong address" subtests in the same didman file already use.

Closes #4249.

## Test plan

- [ ] `Go test` workflow green on this PR.
- [x] `qlty check` (zizmor) green.
- [ ] After merge to `master`: coverage shows up on the Qlty project (requires Qlty GitHub App + OIDC trust; fallback is a `QLTY_COVERAGE_TOKEN` secret with `token: \${{ secrets.QLTY_COVERAGE_TOKEN }}`).
- [ ] Disable CircleCI for the repo once master build is green.

Assisted by AI


.github/workflows/go-test.yaml
```diff
@@ -0,0 +1,41 @@
+name: 'Go test'
+
+on:
+  push:
+    branches:
+      - 'master'
+      - 'V*'
+  pull_request:
+    branches:
```
stevenvegt: why only on these?

stevenvegt (approved):
I'd like the test to run on every PR, so we have tests on PRs on feature branches as well.

.github/workflows/go-test.yaml
reinkrul: Dropped the `pull_request.branches` filter — PRs targeting any base now trigger the workflow. Kept the `push` filter on `master`/`V*` so we don't double-run on feature-branch pushes (the `pull_request` event covers those).

Issue #4258, opened 2026-05-18
ci: speed up PR feedback (remove -p 1, shard e2e, cache, drop multi-arch on PR)


reinkrul:
## Context

PR wall-clock to merge readiness is ~13 min, dominated by the `Go test` workflow (12m15s test step) and `E2E tests` (~10m45s). Audit of the workflows + a local empirical pass shows the bulk is achievable without test-code refactor and without sharding. This issue lists the prioritized findings as discrete deliverables.

Related: PR #4250 (CircleCI → GHA migration), commit 05a70569.

### 1. Drop `-p 1` from `go-test.yaml` (highest impact, lowest effort)

`.github/workflows/go-test.yaml:28` runs `go test -p 1 ./... -race -coverprofile=c_raw.out`. The `-p 1` was inherited verbatim from `.circleci/config.yml` where it sat *inside* `circleci tests split` (8-way shard). The GHA migration dropped the shard but kept the flag — its original purpose (per-shard serialization) no longer applies.

Audit for inter-package parallelism hazards on the current codebase:

| Hazard                              | Found              |
|-------------------------------------|--------------------|
| `net.Listen` to fixed port          | 0 (2 files call `net.Listen`: `storage/session_memcached_test.go:139` uses `:0`, `network/transport/grpc/connection_manager_test.go:63` uses `bufconn`) |
| `os.Chdir` in tests                 | 0                  |
| `TestMain` setting up shared infra  | 0                  |
| Shared filesystem paths             | 0 (the `./test/*` reads in `auth/services/selfsigned/` are package-local; Go sets cwd per package) |
| `os.MkdirTemp` to fixed path        | 0                  |

`go test -p N` runs N test *binaries* in parallel, each in its own process — so the 49 `os.Setenv`/`t.Setenv` call sites are isolated by process boundary, not a hazard either.

Empirical local run on 12-core machine:

| Config                       | Wall-clock | Outcome |
|------------------------------|-----------|---------|
| `go test -p 1 -race ./...`   | 7m 04s    | PASS    |
| `go test -p 2 -race ./...`   | 3m 31s    | PASS    |
| `go test -race ./...` (default) | 2m 06s | PASS    |

GHA `ubuntu-latest` is 2 vCPUs, so dropping `-p 1` means the default `GOMAXPROCS=2` parallelism kicks in. Expected CI saving: **~6 min/PR** (12m15s → ~6m).

**Change**: drop `-p 1`. One line.

[...]

PR #4254, opened 2026-05-18
docs: add COLLABORATORS.md

reinkrul:
## Summary
- Adds `COLLABORATORS.md` describing how to contribute: when to open an issue first, keep PRs small, signed commits, Conventional Commits, matching surrounding code conventions, tests, codegen via the `makefile`, and review etiquette (no force-push after review starts).
- Points the existing contribution touchpoints at the new file instead of duplicating content:
  - `README.rst` — adds a "Contributing" section linking to it.
  - `.github/README.md` — adds a bullet in "For Contributors".
  - `docs/pages/contribute.rst` — slimmed to a pointer (was a one-liner already).

## Test plan
- [ ] Render `COLLABORATORS.md` on GitHub and check links resolve.
- [ ] Render `docs/pages/contribute.rst` on Read the Docs and check the link resolves.

Assisted by AI

PR #4259, opened 2026-05-18
ci: speed up PR feedback

reinkrul:
## Summary

Implements findings 1, 2, 3 from #4258. Three independent CI tweaks targeting the ~13 min PR wall-clock:

1. **`go-test.yaml`**: drop `-p 1`. Inherited from CircleCI's per-shard config during PR #4250 migration; the new GHA pipeline isn't sharded so the flag just serialized 137 packages on a single 2-core runner. Audit shows no inter-package parallelism hazards in the current codebase (no fixed-port `net.Listen`, no `os.Chdir`, no `TestMain`, no shared on-disk paths). Local empirical: `-p 1` takes 7m04s, `-p 2` takes 3m31s, both pass. Expected CI: 12m15s → ~6m on PRs that bust the Go test cache.
2. **`e2e-tests.yaml`**: add gha cache to both `Build and push` steps. Dockerfile already orders go.mod/go.sum download before `COPY .`, so the cache is effective on PRs that don't touch go.sum. Saves ~1-2 min after warm-up.
3. **`build-images.yaml`**: build amd64 only on PRs, multi-arch on push/tag. arm64 under QEMU is ~5-8x slower; PRs are `push: false` anyway, so we only verify build doesn't break - arm64 adds no signal. Saves ~20 runner-min/PR (not on critical path but reduces resource use).

Drive-by: add `.scratch/` and `.claude/scheduled_tasks.lock` to `.gitignore` so local agent state doesn't land in commits.

## Measurement

Compared against #4254 (most recent docs-only PR before this change, same cache profile shape):

| Step | #4254 (baseline) | #4259 (this PR) | Delta |
|---|---|---|---|
| `go-test` - `Test` | 12m 05s | **1m 01s** | -91% (see caveat) |
| `e2e-tests` - `Build and push` | 1m 45s | 3m 06s | +1m 21s (first-run cache write) |
| `e2e-tests` - `Run E2E tests` | 8m 25s | 8m 10s | -15s (noise) |
| `build-images` - `Build and push` | 23m 03s | **1m 38s** | **-93%** |

**Verdict per change:**

- **(1) Drop `-p 1`** - works as intended. The 1m 01s figure looks too good because Go's build/test cache is fully hot on this run (most packages show `(cached)` in the log). On a PR that *does* change Go code (the realistic case), expect the local empirical projection: ~6 min instead of ~12 min, based on the 7m04s → 3m31s drop measured locally on a 2-core-equivalent run.
- **(2) E2E gha cache** - neutral on this first run (+1m 21s as the cache writes for the first time). Will pay off on subsequent PRs once the cache exists. The `Run E2E tests` step itself is unaffected as designed (sharding from #4258 finding 4 is a separate change).
- **(3) Multi-arch off for PRs** - clear win. Build went from 23m 03s to 1m 38s, removing the arm64-under-QEMU bulk. No coverage lost on PRs since they don't publish.

**Combined PR wall-clock effect**:

Before this PR (last comparable run #4254): max(go-test 12m, e2e 10m, build-images 23m) = ~23 min total elapsed. After: max(~6m worst-case go-test, ~11m e2e, ~2m build-images) = **~11 min** elapsed. `build-images` no longer the long pole on PRs.

## Test plan
- [x] `Go test` workflow green; `Test` step substantially faster.
- [x] `E2E tests` workflow green (no regression).
- [x] `Build Docker images` workflow green and ~93% faster.
- [ ] Confirm subsequent PRs (different go.sum, e2e cache pre-populated) show the projected wall-clocks.

Refs #4258.

Assisted by AI

### DIFF
--- a/.github/workflows/build-images.yaml
+++ b/.github/workflows/build-images.yaml
@@ -75,7 +75,9 @@ jobs:
         with:
           context: .
           file: Dockerfile
-          platforms: linux/amd64,linux/arm64
+          # On PRs we only verify the build doesn't break; arm64 under QEMU
+          # is ~5-8x slower and adds no signal there. Multi-arch on push/tag.
+          platforms: ${{ github.event_name == 'pull_request' && 'linux/amd64' || 'linux/amd64,linux/arm64' }}
           push: ${{ github.event_name != 'pull_request' }}
           tags: ${{ steps.docker_meta.outputs.tags }}
           build-args: |
--- a/.github/workflows/e2e-tests.yaml
+++ b/.github/workflows/e2e-tests.yaml
@@ -60,6 +60,8 @@ jobs:
           platforms: linux/amd64
           push: ${{ github.event_name == 'pull_request' || github.event_name == 'push' }}
           tags: ghcr.io/nuts-foundation/nuts-node-ci:${{ env.SHA }}
+          cache-from: type=gha,scope=e2e
+          cache-to: type=gha,scope=e2e,mode=max
           secrets: |
             GIT_AUTH_TOKEN=${{ secrets.PACKAGE_SECRET }}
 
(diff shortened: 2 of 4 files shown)


stevenvegt (approved):
Impressive what a few config options can do. Thanks for looking in to this!

Issue #4253, opened 2026-05-12
Fix credential_selection field-id ambiguity in two-VP and intra-PD binding


stevenvegt:
## Problem Statement

Two interacting issues in the current Presentation Definition (PD) field-id handling surface as opaque token-request failures and silently inconsistent credential selections.

1. **Cross-VP `credential_selection` is over-strict.** In the two-VP RFC 7523 flow, the wallet automatically captures every id-bearing field value from the organization VP and merges them into the `credential_selection` used to build the service-provider VP. The SP-side selector then validates that every key in the merged selection must correspond to a field id in the SP PD. Captured keys missing from the SP side fail loudly. Declaring useful selection ids on the org PD (e.g. `patient_bsn`, `delegating_uzi`) breaks any caller that doesn't also declare those ids on the SP PD, even though the caller never asked for SP-side filtering on those fields.

2. **Same-`id` fields across descriptors don't constrain values.** Within a single PD, when the same `id` name appears on multiple field-objects (e.g. `org_ura` on the healthcare-provider descriptor and on the professional-delegation descriptor), a PD author naturally expects "the chosen credentials must agree on this id's value". The current selector treats each descriptor independently, so it can select an HCP VC with `org_ura=A` and a delegation VC with `org_ura=B`. The id is effectively decorative within one VP.

A third concern: the matching logic is currently spread across `vcr/pe`, `vcr/holder`, and `auth/client/iam`. Reasoning about the combined behavior requires reading multiple packages and several layers of indirection. The rules cannot be exercised as a single unit.

## Solution

- A PD author can declare id-bearing fields freely on either side of a two-VP flow. Cross-VP capture stays internal and is silently filtered down to ids that exist on the target PD before being applied.
- When the same `id` appears across multiple descriptors in one PD, the wallet finds an assignment where every shared id resolves to the same value across the chosen credentials. The search uses DFS with forward checking on id-bearing fields, falling back to skipping optional descriptors only after all candidate combinations are exhausted.
- A PD whose same-`id` fields have incompatible filters fails to load, so the PD author learns about the inconsistency at boot time rather than at request time.
- PDs that admit more than one consistent assignment are reported as ambiguous (not silently resolved), and the caller is told which descriptors carry the ambiguity so they can add disambiguating selection keys.
[...]

PR #4280, opened 2026-05-26
[#4253 1/8] feat(pe): core selection engine Select + MatchReport

stevenvegt:
## Parent PRD

#4253

## Summary

This PR introduces `pe.Select`, the single credential-matching engine that will replace the matcher logic currently spread across `vcr/pe`, `vcr/holder`, and `auth/client/iam`. It implements same-`id` binding agreement across input descriptors (the headline fix of #4253) via a backtracking search, adds an opt-in `Strict` ambiguity strategy, and produces a structured `MatchReport` explaining why each candidate was or wasn't selected (folds in #4218, supersedes closed PR #4225). **No existing caller changes**: the engine is exercised only by its new black-box test suite; item 3 (#4284) wires the public API onto it.

## What changed

- **`vcr/pe/doc.go`** (new): package documentation with the vocabulary the engine docs use (binding tuple, caller-bound, interchangeable, decisive assignment) and the central design principle: the PD is the data contract, since only declared fields reach token introspection.
- **`vcr/pe/select.go`** (new): the engine. Step 1 evaluates per-descriptor eligibility (reusing `matchConstraint`/`matchFormat`) into pools indexed by field-id values; step 2 is a recursive backtracking search threading a bindings map (same-id agreement, prefer-fill-over-skip, optional-skip-after-exhaustion); step 3 applies the existing submission-requirement rule engine (`groups()`/`apply()`/`deduplicate()`, unchanged) with the selection consumed first-descriptor-wins. Functional options: `WithInitialBindings` (credential_selection semantics, strict resolution), `WithStrategy(Strict)` (rival assignments counted per binding tuple), `WithSelectionTrace`. A node-visit cap (1e6, `ErrSearchLimitReached`) guards against crafted counterparty PDs.
- **`vcr/pe/match_report.go`** (new): the `MatchReport` diagnostic, built post-hoc from the same step-1 primitive the selection uses, so it cannot drift; includes the `DivergingAlternatives` wallet-hygiene flag.
- **`vcr/pe/presentation_definition.go`**: the credential-to-JSON-map conversion inside `matchConstraint` extracted into `credentialAsMap` (shared with the report builder), keeping the fail-closed behavior for unknown credential formats.
- **Tests**: ~60 black-box cases in `select_test.go` (characterization of the legacy matcher plus every new policy, including both consent scenarios and the optionality-vs-multiplicity edges) and five runnable godoc examples in `example_select_test.go`.

## How to review

1. Start with `doc.go` (vocabulary, 5 minutes) and the `Select` godoc rule list; every other comment builds on those terms.
2. Read `select.go` top to bottom: options, `Select` orchestration, then the search core (`searcher.search`/`emit`), then the pool machinery (`buildPool`/`consistentGroups`).
3. `match_report.go` is consumable independently; its type godoc states exactly what the report does and does not promise (the ambiguity list covers only the first rival assignment found; explanations on failed runs are best-effort).
4. The tests are organized per concern (`TestSelect`, `_Eligibility`, `_InitialBindings`, `_SameIDBinding`, `_Strategy`, `_Bindings`, `_Trace`, `_SubmissionRequirements`, `_NodeLimit`); the worked examples from the PRD appear under their rule names.

Key decisions to weigh (all recorded in PRD #4253):
- **When are two credentials "the same" for selection?** When they have equal values for every field id the PD declares. Anything the PD does not declare is invisible to the relying party (it never reaches token introspection), so it cannot distinguish credentials. Consequence, and the one deliberate behavior change: a caller-bound request no longer fails when the wallet holds a reissued duplicate; it fails only when matches differ on a declared field, and then the error names the field to bind.
- **A bound field must resolve.** If the caller pins `patient_bsn=X`, a credential without a readable `patient_bsn` is rejected for that descriptor, exactly like the old field selector.
- **Two separate multiplicity checks, on purpose.** The per-descriptor check for caller-bound fields (the old `NewFieldSelector` contract) and the whole-assignment ambiguity check under `Strict` have different triggers and deliberately stay separate code paths.
- **The search is capped** at one million node visits, because presentation definitions come from the counterparty: a crafted PD must not be able to burn unbounded CPU.

## Deviations from spec

Additions beyond the original spec, from the self-review and design discussions (PRD updated in place for each):
- `DescriptorReport.DivergingAlternatives`: flags interchangeable alternatives whose credentialSubjects differ in undeclared fields (wallet hygiene, not an error).
- `ErrSearchLimitReached` exported sentinel so callers can classify the limit-abort failure class.
- `AmbiguousDescriptors` is documented as covering only the first rival assignment found (the search stops there; a complete list is not promised).
- A credential that fills two descriptors of a pick group now counts once against the rule and is kept on the first descriptor, matching the legacy descriptor-map behavior.
- Package vocabulary in `doc.go` and runnable godoc examples were added for reviewability.

Known, documented limitation (unchanged from spec): the coarse skip rule can disagree with `pick min>=1` floor satisfaction; floor-aware search cooperation is explicitly deferred (PRD "Out of Scope").

## Dependencies

None — first item of the chain, based directly on `feature/4253-credential-selection`. Items 3-5 stack on this PR; item 2 (#4281) is independent.

## Design context

- PRD #4253: Behavior Specification (Policies 1-7 with worked examples), Implementation Decisions, and the strategy note under the Implementation Plan.
- The 2026-07-14/15 decision trail (binding-tuple interchangeability, strict initial bindings, scale design for 20k-credential wallets, the consent-scenario analysis) is recorded in the PRD body sections "Policy 5", "Policy 6", and "Implementation Decisions".
- #4218 (developer matching diagnostics) is resolved by `MatchReport`; closed PR #4225 shows the rejected logging approach.

<details>

## Parent PRD

#4253

Item **1 of 8**. Foundation for the credential-selection rewrite. Also folds in the developer-diagnostics work from #4218 (this `MatchReport` supersedes the logging approach in closed PR #4225).

## Summary

Introduce the new pure matching engine `pe.Select` in `vcr/pe`, with structured diagnostics (`MatchReport`). `Select` is **consolidated**: it owns the whole matching pipeline in three steps: (1) find which credentials match each descriptor's constraints and format; (2) search for a combination of credentials whose shared-`id` field values agree; (3) apply the PD's submission-requirement rules (`pick`/`all`/`min`/`max`) to that combination, reusing the existing `groups()`/`apply()`/`deduplicate()` helpers. **No existing caller is changed in this PR**; the engine is exercised entirely by a new black-box test suite. Item #3 then rewires `Match` onto `Select` and deletes `matchConstraints`/`matchBasic`/`matchSubmissionRequirements`/`MatchWithSelector` and the `CredentialSelector` abstraction.

> Design note (decided during implementation): the engine is consolidated, not layered. Once `Select` receives the whole PD it can derive everything it needs from the PD itself, including which descriptors are allowed to go unfilled, so there is no reason for the caller to compute and pass that. Floor-aware `pick min>=1` cooperation stays out of scope (PRD #4253 "Out of Scope"); `Select` uses a simple skip rule.

### Surface (all new, in `vcr/pe`)

```go
type SelectionStrategy int
const (
    FirstMatch SelectionStrategy = iota // default ("loose"): take the first consistent assignment
    Strict                              // ErrMultipleCredentials on a second consistent assignment
)

type Option func(*selectOptions)
func WithInitialBindings(b map[string]string) Option // seed id->value (credential_selection)
func WithStrategy(s SelectionStrategy) Option
func WithSelectionTrace() Option                     // populate Result.Report; default off

func Select(pd PresentationDefinition, candidates []vc.VerifiableCredential, opts ...Option) (Result, error)

type Result struct {
    Candidates []Candidate        // reuse existing type; VC == nil => descriptor unfilled (skipped, or dropped by a submission-requirement rule)
    Bindings   map[string]string  // resolved id->value of the chosen assignment
    Report     *MatchReport       // non-nil only when WithSelectionTrace() is set
}
```

`Bindings` is `map[string]string`, reusing the value stringification already in `matchesSelections` (selector.go).

### Algorithm (three steps)

**Step 1: match each descriptor on its own.** For each input descriptor, evaluate every candidate credential via the existing `matchConstraint` (presentation_definition.go:400) + `matchFormat`, with no cross-descriptor binding yet. Record pass/fail and, on fail, the reason (no value / filter / format). No new field-matching logic.

**Step 2: search for a consistent combination** (depth-first over the matching credentials, in PD order):
- **Binding consistency (P3, always on):** a candidate is consistent only if its id-bearing resolved values agree with the running bindings. Seed from `WithInitialBindings`; keys not present as field ids on this PD are dropped.
- **Backtracking + optional skip (P4):** try every consistent candidate; a required descriptor with no consistent candidate fails the branch and backtracks; an optional descriptor is skipped (VC=nil) only after its candidates are exhausted. Whether a descriptor may be left unfilled is derived from `pd.SubmissionRequirements` with a simple rule: with no submission requirements every descriptor is required; otherwise a descriptor is optional when its group can be satisfied without it.
- **Unresolved optional doesn't bind (P6):** a candidate's `optional:true` field that resolves to no value contributes no binding. Initial bindings are stricter (decided 2026-07-14): a candidate for a descriptor whose field id appears in the initial bindings must resolve that field to the bound value; unresolved optional fields are rejected for that descriptor (preserves legacy `matchesSelections` semantics).
- **Ambiguity (P5):** see strategies below.
- **Scale (decided 2026-07-14):** eligible pools are hash-indexed by their id-bearing values, so binding-consistency filtering is a lookup, not a scan; large homogeneous wallets (20k per-patient credentials) cost a hash join, linear in wallet size. The search carries a node-visit cap (order 1e6) and aborts with an error naming the PD when hit, because PDs are counterparty-supplied in wallet flows and an unbounded search would be remotely triggerable CPU exhaustion.

**Step 3: apply the submission-requirement rules.** `Select` passes the step-2 combination to the existing `groups()`/`apply()`/`deduplicate()` helpers (rule logic unchanged) to enforce the PD's `pick`/`all`/`count`/`min`/`max`/nested rules and produce the final selection; a descriptor dropped by a rule ends with `VC=nil`. The descriptor-map / JSON-path formatting that `matchBasic` and `matchSubmissionRequirements` did is **not** part of `Select`; it moves to the thin `Match` wrapper in item #3. The simple skip rule from step 2 can disagree with exact rule satisfaction in the deferred `pick min>=1` case; that limitation is documented and now lives in one place.

### Selection strategies

The engine always enforces binding consistency (P3), backtracking/optional-skip (P4), and unresolved-optional-doesn't-bind (P6). The strategy controls **only** what happens when more than one complete, binding-consistent assignment exists:

- `FirstMatch` (default): return the first assignment. This is the lenient, backward-compatible behavior. For any PD that does not reuse an `id` across descriptors it reproduces today's matcher, including `NewFieldSelector`'s `ErrMultipleCredentials` when a caller-bound descriptor still resolves to more than one **binding tuple**. Matches identical on every declared field id are interchangeable and first-picked (decided 2026-07-15: the PD is the data contract; only declared fields reach token introspection, and multi-value filters such as issuer patterns are deliberate equivalence declarations). The one deliberate loosening vs. today: a reissued duplicate no longer fails a caller-bound request; diverging credentialSubjects among interchangeable matches are flagged in the `MatchReport` instead.
- `Strict`: a second complete consistent assignment is `ErrMultipleCredentials` (Policy 5), naming the ambiguous descriptors so the caller can add disambiguating selection keys. `Strict` supersedes `FirstMatch`. Rivals are counted per distinct **binding tuple**, not per credential tuple (decided 2026-07-14): credentials identical in every id-bearing value are interchangeable, never ambiguous (no `credential_selection` key could separate them; the engine picks the first). This also keeps Strict uniqueness proofs cheap on wallets holding many interchangeable credentials.

**Policy 5 (ambiguity-as-error) is the behavior-risk to be aware of.** It is the only rule here that can turn a previously-succeeding match into an error: a descriptor with several equally-valid credentials that the caller did not disambiguate is silently first-picked today but errors under `Strict`. That is exactly why it is gated behind the strategy rather than always on. By contrast Policy 3 (same-`id` agreement) is always on but safe — it only affects PDs that reuse an `id` across descriptors, which is a new pattern with no deployed users.

### Strategy is an input, not parsed here (scope)

`Select` takes a `SelectionStrategy` value and honors it. **How that value is chosen is out of scope for this PR.** The intent is to declare it per-PD at the config layer (policy `credentialProfileConfig` / discovery `ServiceDefinition`), defaulting to `FirstMatch` ("loose") and flipped to `Strict` in a future major release — but parsing and registration of that config land in items #5/#7. This PR only defines the enum and makes the engine behave correctly for each value.

### MatchReport (diagnostics; supersedes closed PR #4225)

```go
type MatchReport struct {
    Descriptors          []DescriptorReport // one per input descriptor, in PD order
    Outcome              Outcome            // matched | no_credentials | multiple_credentials
    AmbiguousDescriptors []string           // descriptor ids carrying >1 choice (multiple_credentials only)
}
type DescriptorReport struct {
    DescriptorID string
    Optional     bool
    Considered   []CandidateReport // every candidate VC evaluated against this descriptor
    SelectedID   string            // chosen credential id; "" if skipped/unfilled
    Skipped      bool
}
type CandidateReport struct {
    CredentialID string
    Eligible     bool       // passed this descriptor's constraints + format
    Dismissal    *Dismissal // why it wasn't used; nil when selected
}
type Dismissal struct {
    Reason   DismissalReason
    FieldID  string // constraint/binding reasons, when known
    Path     string // JSON path evaluated (constraint reasons)
    Expected string // filter const/type, or the bound value (binding conflict)
    Found    string // value found; "" = none
    Message  string // pre-rendered human-readable line (lazy, only under trace)
}
type DismissalReason string
const (
    ReasonNoValue         DismissalReason = "constraint_no_value"
    ReasonFilter          DismissalReason = "constraint_filter"
    ReasonFormat          DismissalReason = "format_mismatch"
    ReasonBindingConflict DismissalReason = "binding_conflict"
    ReasonNotSelected     DismissalReason = "not_selected"
)
```

Populated only under `WithSelectionTrace()` (`Message` rendered lazily, so a non-trace run pays nothing). Trace depth: step-1 (per-descriptor) dismissals recorded fully; binding conflicts recorded on the **decisive path** (the chosen assignment, or the conflicts that caused failure/ambiguity), not every backtracking visit. No surface (endpoint/UI) in this PR — consuming the report is a follow-up to #4218.

## Testing

Black-box `select_test.go`, no mocks or on-disk fixtures. Two buckets:

- **Characterization** (proves equivalence before #3 rewires callers): `FirstMatch`+bindings reproduces the `TestNewFieldSelector` matrix (single/multi-field AND, multi-descriptor, type conversions string/float64/bool, zero-match -> unfilled); `FirstMatch` with no bindings reproduces `FirstMatchSelector`; an all-required PD reproduces `matchBasic` (every descriptor must fill, else `ErrNoCredentials`); a representative submission-requirements PD reproduces `matchSubmissionRequirements` (`all`, and `pick` with `min`/`max`); `ErrNoCredentials` cases.
- **New policies:** P3, P4, P5 (both strategies), P6 worked examples from the PRD; trace on/off asserting `MatchReport` contents for a dismissed VC and for an ambiguity; the two-VP composition pattern (two `Select` calls chained through a bindings filter).
- **Scale and contract edges (added 2026-07-14):** interchangeable credentials (identical binding tuples) are not ambiguous under `Strict` while distinct binding tuples are; the node-visit cap aborts a crafted exponential PD with a clear error; `Result.Candidates` keeps one entry per input descriptor, in PD order, on every error path; format gating (PD-level and descriptor-level) and nil-`Constraints` descriptors are covered.

Exhaustive submission-requirement rule cases stay in `submission_requirement_test.go` (the `apply()` logic is unchanged) and the integration-level `TestMatch` suites; #1 adds only representative satisfaction characterization, not a full duplication.

## Acceptance Criteria

- [x] `Select` with functional options as specified; `Result` reuses `Candidate`.
- [x] P3/P4/P6 always on; P5 gated by `Strict`; `FirstMatch` is the default.
- [x] Characterization tests prove `FirstMatch` reproduces both `FirstMatchSelector` and `NewFieldSelector` behavior.
- [x] `Select` applies submission-requirement rules by reusing `groups()`/`apply()`/`deduplicate()`; whether a descriptor may go unfilled is derived from `pd.SubmissionRequirements`.
- [x] Characterization tests prove `FirstMatch` also reproduces `matchBasic` (all-required) and `matchSubmissionRequirements` (all/pick/min/max).
- [x] `MatchReport` populated only under `WithSelectionTrace()`, with the specified reasons and decisive-path conflict narration.
- [x] Ambiguity counted per distinct binding tuple; interchangeable credentials never error.
- [x] Caller-bound multiplicity counted per binding tuple; both consent scenarios pinned (declared field errors, undeclared field first-picks); diverging interchangeable alternatives flagged in `MatchReport`.
- [x] Initial bindings require resolution (legacy field-selector semantics); P6 leniency covers only search-accumulated bindings.
- [x] Eligible pools are hash-indexed; node-visit cap aborts pathological searches with a clear error.
- [x] `Result.Candidates` is full length (one entry per descriptor, PD order) on every path, including errors.
- [x] No caller or behavior changes outside `vcr/pe`; existing `vcr/pe` tests still green.
- [x] `go build ./...` and `go test ./vcr/pe/...` pass.

</details>

### DIFF
--- a/vcr/pe/doc.go
+++ b/vcr/pe/doc.go
@@ -1,0 +1,74 @@
+/*
+ * Copyright (C) 2026 Nuts community
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *
+ */
+
+// Package pe implements Presentation Exchange (PEX): matching the Verifiable Credentials in a
+// wallet against a Presentation Definition, and building the Presentation Submission that
+// presents the chosen ones. Select is the single matching engine; Match and the submission
+// builder are layered on top of it.
+//
+// # Vocabulary
+//
+// The documentation in this package uses these terms consistently:
+//
+//   - Presentation definition (PD): the verifier's description of the credentials it wants: a
+//     set of input descriptors, plus optional submission requirements. In this system the PD is
+//     the data contract: only fields the PD declares are mapped to token introspection, so a
+//     field the PD does not declare plays no role in credential selection either. A filter that
+//     admits several values, such as an issuer pattern, is a deliberate equivalence declaration
+//     by the PD author.
+//   - Input descriptor: one slot of the PD, filled by at most one credential. Its constraint
+//     fields (JSONPaths with optional filters) and format designations decide which credentials
+//     fit the slot.
+//   - Field id / binding name: the optional id on a constraint field. A field id names the
+//     resolved value in two directions: outward, as the key under which the value appears in
+//     token introspection and is addressed by the credential_selection parameter; and inward,
+//     as a binding name, meaning that wherever the same id appears in one selection, the
+//     resolved values must agree.
+//   - Binding: an id-to-value pair the selection has committed to. Bindings accumulate while
+//     the engine fills descriptors, and constrain every later choice. A credential's resolved
+//     id-to-value pairs for a descriptor are its binding tuple.
+//   - Initial bindings: bindings supplied before the search starts (WithInitialBindings): the
+//     caller's credential_selection, or values captured from an earlier presentation in a
+//     two-VP flow. A descriptor is caller-bound when one of its field ids appears in the
+//     initial bindings. A caller-bound descriptor must resolve the bound field to the bound
+//     value (an unresolved optional field does not qualify) and must resolve to exactly one
+//     interchangeable set of credentials.
+//   - Eligible credential: a credential that satisfies one input descriptor on its own,
+//     constraints and formats, before any cross-descriptor consistency is considered.
+//   - Interchangeable credentials: eligible credentials with identical binding tuples. By the
+//     data-contract principle nothing the PD declares tells them apart, and no
+//     credential_selection key could, so the engine treats them as a single choice and uses the
+//     first in candidate order. Interchangeable credentials whose credentialSubjects
+//     nevertheless differ are flagged in the MatchReport (DivergingAlternatives).
+//   - Assignment: a choice of credential, or none, per input descriptor. The engine searches
+//     for a complete assignment with consistent bindings. The assignment it settles on, also on
+//     failure, is the decisive assignment: it is what Result reports and what the MatchReport
+//     explains dismissals against.
+//   - Optional descriptor: a descriptor the submission requirements allow to go unfilled, for
+//     example a member of a pick group. Optionality governs whether zero matches is acceptable;
+//     it never relaxes the multiplicity or consistency rules (zero-vs-some, never one-vs-many).
+//   - Selection strategy: what to do when more than one materially different complete
+//     assignment exists. FirstMatch, the default, takes the first found; Strict reports
+//     ambiguity as ErrMultipleCredentials, naming the descriptors to disambiguate with
+//     credential_selection keys.
+//   - Selection trace: the opt-in diagnostic (WithSelectionTrace) that explains, per descriptor
+//     and candidate, why each credential was or wasn't used. See MatchReport.
+//
+// These concepts originate as the numbered policies of the design in
+// https://github.com/nuts-foundation/nuts-node/issues/4253.
+package pe
--- a/vcr/pe/example_select_test.go
+++ b/vcr/pe/example_select_test.go
@@ -1,0 +1,201 @@
+/*
+ * Copyright (C) 2026 Nuts community
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *
+ */
+
+package pe_test
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/nuts-foundation/go-did/vc"
+	"github.com/nuts-foundation/nuts-node/vcr/pe"
+)
+
+func mustParsePD(jsonStr string) pe.PresentationDefinition {
+	var pd pe.PresentationDefinition
+	if err := json.Unmarshal([]byte(jsonStr), &pd); err != nil {
+		panic(err)
+	}
+	return pd
+}
+
+func mustParseVC(jsonStr string) vc.VerifiableCredential {
+	var cred vc.VerifiableCredential
+	if err := json.Unmarshal([]byte(jsonStr), &cred); err != nil {
+		panic(err)
+	}
+	return cred
+}
+
+// The engine picks a credential per input descriptor. A field id (patient_id) both names the
+// value for token introspection and acts as a binding name.
+func ExampleSelect() {
+	pd := mustParsePD(`{
+		"id": "example-pd",
+		"input_descriptors": [{
+			"id": "patient_credential",
+			"constraints": {"fields": [{"id": "patient_id", "path": ["$.credentialSubject.patientId"]}]}
+		}]
+	}`)
+	wallet := []vc.VerifiableCredential{
+		mustParseVC(`{"id": "vc-1", "credentialSubject": {"patientId": "123"}}`),
+	}
+
+	result, err := pe.Select(pd, wallet)
+	if err != nil {
+		panic(err)
+	}
+
+	fmt.Println(result.Candidates[0].VC.ID.String())
+	fmt.Println(result.Bindings["patient_id"])
+	// Output:
+	// vc-1
+	// 123
+}
+
+// Initial bindings (the credential_selection parameter) pin a field id to a value, selecting
+// the credential that carries it.
+func ExampleWithInitialBindings() {
+	pd := mustParsePD(`{
+		"id": "example-pd",
+		"input_descriptors": [{
+			"id": "patient_credential",
+			"constraints": {"fields": [{"id": "patient_id", "path": ["$.credentialSubject.patientId"]}]}
+		}]
+	}`)
+	wallet := []vc.VerifiableCredential{
+		mustParseVC(`{"id": "vc-1", "credentialSubject": {"patientId": "123"}}`),
+		mustParseVC(`{"id": "vc-2", "credentialSubject": {"patientId": "456"}}`),
+	}
+
+	result, err := pe.Select(pd, wallet, pe.WithInitialBindings(map[string]string{"patient_id": "456"}))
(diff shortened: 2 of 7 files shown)


reinkrul:

PR #4286, opened 2026-05-27
[#4253 4/8] feat(holder): thread initialBindings through BuildSubmission

stevenvegt:
## Parent PRD

#4253

Item **4 of 8**. Depends on item #3 (#4284) — branched off `4253-3-rewire-select`, retargeted up the stack as dependencies merge.

## Summary

Add an `initialBindings map[string]string` parameter to `Wallet.BuildSubmission` for the **captured cross-VP** bindings (Policy 2), kept distinct from the user's `credentialSelection`. **Pure, behavior-neutral plumbing** — every caller passes `nil` in this PR, so nothing changes until #5 wires the two-VP caller to pass the filtered captured bindings.

## Why a separate parameter (not reuse `credentialSelection`)

- `credentialSelection` is the **user's** explicit selection, validated as user input at the API layer (#6).
- `initialBindings` is the **captured org-side** binding from the two-VP flow, internal plumbing.

They must stay separate at the `BuildSubmission` boundary. Merging captured values *into* `credentialSelection` at the caller is exactly the `applyCapturedFieldsToSelection` behavior the PRD removes (PRD problem #1 / Policy 2). They travel separately and merge only inside the presenter when it builds `Select`'s single bindings map. Per Policy 2 they cannot conflict (user values already filtered the org selection, so captured values agree); the merge gives the user `credentialSelection` precedence defensively.

## Implementation Spec

Signature gains `initialBindings map[string]string` after `credentialSelection`, before `params`:

- `vcr/holder/interface.go:56` — interface declaration.
- `vcr/holder/sql_wallet.go:83` — implementation; thread the param to `presenter.buildSubmission`.
- `vcr/holder/memory_wallet.go:63` — second implementation; same change.
- `vcr/holder/presenter.go` `buildSubmission` — add the param; merge `credentialSelection` + `initialBindings` (user precedence) and pass the merged map to `builder.SetInitialBindings`. With all callers passing nil, the merge is inert in this PR.
- `vcr/holder/mock.go` — regenerate via `go generate ./vcr/holder/...` (mockgen); update `MockWallet.BuildSubmission` and its recorder.

Callers pass `nil` for the new parameter:
- `auth/api/iam/openid4vp.go:321` → `BuildSubmission(ctx, []did.DID{walletDID}, nil, *presentationDefinition, nil, nil, buildParams)`.
- `auth/client/iam/openid4vp.go:465` → add `nil` for `initialBindings`. (#5 replaces this nil with the captured org bindings filtered to the SP PD's field ids.)

## Testing

- Existing `vcr/holder` and `auth/...` tests pass after the signature change (call sites updated, mock regenerated). No new behavior to assert here — #5 adds the behavioral test for captured bindings.
- A presenter-level unit test that a non-nil `initialBindings` is merged into the bindings handed to the match (so #5 can rely on it) — minimal, can also live in #5.

## Acceptance Criteria

- [ ] `BuildSubmission` (interface + `sql_wallet` + `memory_wallet`) takes `initialBindings map[string]string`; mock regenerated.
- [ ] Presenter merges `credentialSelection` + `initialBindings` (user precedence) into the builder's initial bindings.
- [ ] Both callers updated to pass `nil`; no behavior change in this PR.
- [ ] `go build ./...` and `go test ./vcr/holder/... ./auth/...` pass.


reinkrul:

PR #4301, opened 2026-06-02
test(storage): fix flaky minimemcached nil-pointer panic

reinkrul:
## Problem
The `storage` test job intermittently fails with a panic, crashing the whole test binary:

```
panic: runtime error: invalid memory address or nil pointer dereference
  github.com/daangn/minimemcached.(*MiniMemcached).serve()  minimemcached.go:125
```

## Root cause
`minimemcached.Run` synchronously creates the listener, then spawns a goroutine that loops on `m.l.Accept()`. `Close()` (called from `t.Cleanup`) sets that listener to `nil`. On very short tests the cleanup's `Close()` can run *before* the goroutine reaches `Accept()`, so `m.l.Accept()` dereferences a nil listener. The panic happens in a dependency-spawned goroutine, so no test can recover it — the binary dies.

This is an upstream bug; v1.2.0 is the latest release, so there's no version to bump to.

## Fix
Block in the test helper until the server actually answers a probe connection (`version\r\n`). That guarantees `serve()` is parked inside `Accept()` before the test proceeds, so the cleanup `Close()` unblocks it via `listener.Close()` cleanly instead of racing goroutine startup.

## Verification
`go test ./storage/ -run Memcached -count=10 -race` passes.

### DIFF
--- a/storage/session_memcached_test.go
+++ b/storage/session_memcached_test.go
@@ -170,11 +170,13 @@ func memcachedTestServer(t *testing.T) *minimemcached.MiniMemcached {
 	if err != nil {
 		t.Fatal(err)
 	}
+	// Workaround for https://github.com/daangn/minimemcached/issues/14 (fix proposed in PR #13).
 	// Wait until the server's accept loop is actually serving connections before returning.
 	// minimemcached.Run spawns a goroutine that calls Accept() on the listener; Close() nils that
 	// listener. On really short tests Close() (in the cleanup below) could run before the goroutine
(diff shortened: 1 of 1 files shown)


stevenvegt (approved):
>Remove this workaround once a minimemcached release includes the upstream fix.

How do we know this?

reinkrul:
> > Remove this workaround once a minimemcached release includes the upstream fix.
> 
> How do we know this?

We don't, I'm subscribed to the issue, so I'll get a notification. And we link to it in the source code, + we get dependency updates through dependabot (so check changelog).

PR #4309, opened 2026-06-03
fix(vcr): check ParseDID error to prevent nil pointer panic

1795771535y-cell:
Fixes 4235


reinkrul:
Thanks for the fix! This was addressed in #4302 (now merged), which resolves the same nil-pointer panic from an unparseable issuer DID (#4235). Closing as a duplicate — appreciate your contribution.

reinkrul:
Note that we couldn't merge your PR due to missing commit signature and unit test.

Issue #4316, opened 2026-06-04
Configurable OAuth client authentication to external authorization servers (OpenID4VCI auth-code flow)


reinkrul:
Related: #4310, #3197

> **Experimental.** This feature may change or be removed without further notice. Must be flagged as experimental in both code (doc comments) and documentation, and lives under `auth.experimental`.

## Problem

As an OpenID4VCI wallet in the auth-code flow, the node authenticates as a public client using its `did:web` base URL as `client_id` + `client_id_scheme=entity_id`:

- Authorization request: `auth/api/iam/openid4vci.go:131-140`
- Token request: `auth/client/iam/openid4vp.go:227` (`client_id` only, no secret)

`entity_id` is Nuts-specific. Third-party issuers expect a registered `client_id` + secret. AET (OpenIddict) **requires** client auth -- a public `client_id` is refused (`invalid_client`) and `client_secret_post` is accepted. No way to configure this today, so the auth-code flow can't complete against any non-`entity_id` AS.

## Solution

Per-server OAuth client config under `auth.experimental`. On a match, present the configured `client_id` + `client_secret` (`client_secret_post`) instead of the `did:web` + `entity_id` defaults.

```yaml
auth:
  experimental:
    clients:
      - serverurl: https://issuer.example.com/oauth   # OAuth AS identifier
        clientid: nuts-wallet-123
        clientsecret: s3cr3t        # optional
```

```
NUTS_AUTH_EXPERIMENTAL_CLIENTS_0_SERVERURL=https://issuer.example.com/oauth
NUTS_AUTH_EXPERIMENTAL_CLIENTS_0_CLIENTID=nuts-wallet-123
NUTS_AUTH_EXPERIMENTAL_CLIENTS_0_CLIENTSECRET=s3cr3t
```

Keys are single tokens (`serverurl`, not `server_url`): env vars map `_` to the `.` delimiter (`core/config.go:60`), so `..._SERVER_URL` would nest as `server.url`. No match -> current behavior unchanged (additive).

### Config

Add to the existing `ExperimentalConfig` (`auth/config.go`), which already documents "subject to change without notice and may be removed":

```go
type OAuthClientConfig struct {
    ServerURL    string `koanf:"serverurl"`
    ClientID     string `koanf:"clientid"`
    ClientSecret string `koanf:"clientsecret"`
}
// ExperimentalConfig: Clients []OAuthClientConfig `koanf:"clients"`
```

[...]

PR #4327, opened 2026-06-04
feat(auth): configurable OAuth client authentication for OpenID4VCI auth-code flow

reinkrul:
Implements #4316.

**EXPERIMENTAL** — this configuration may change or be removed without further notice (flagged in code doc comments and nested under `auth.experimental`).

## What

Per-server OAuth client configuration so the node can authenticate to external authorization servers in the OpenID4VCI authorization code flow with a registered `client_id` + `client_secret`, instead of the Nuts-specific `did:web` + `client_id_scheme=entity_id` defaults that third-party servers (e.g. AET/OpenIddict) reject.

```yaml
auth:
  experimental:
    clients:
      - serverurl: https://issuer.example.com/oauth   # OAuth AS issuer
        clientid: nuts-wallet-123
        clientsecret: s3cr3t        # optional
```

Env: `NUTS_AUTH_EXPERIMENTAL_CLIENTS_0_SERVERURL` / `_CLIENTID` / `_CLIENTSECRET` (single-token keys round-trip through the `_`→`.` env delimiter).

## Behavior

- Match by **OAuth AS issuer** (`authzServerMetadata.Issuer`), trailing-slash-normalized — not `credential_issuer` (AET splits them: `.../aet` vs `.../aet-oauth`).
- **Authorization request**: configured `client_id`, drop `client_id_scheme`.
- **Token request**: configured `client_id`; `client_secret` via `client_secret_post` (TODO: derive method from AS metadata + support `client_secret_basic`). Empty secret → public client (PKCE only).
- No matching entry → unchanged (`did:web` + `entity_id`). Fully additive.
- Validated at startup: `serverurl` required + valid URL (HTTPS in strict mode), `clientid` required, no duplicate `serverurl`.
- `clientsecret` redacted from logged/printed config (whole `auth.experimental.clients` subtree).

## Scope

Only the OpenID4VCI auth-code flow is wired; the config shape is protocol-neutral for later reuse (plain OAuth2, OpenID4VP). Not a pflag (no `[]struct` flag type), so it doesn't appear in `server_options.rst`; documented via experimental doc comments + #4316.

Already merged into `project-gf-pilot`; this PR tracks the same change against `master`.

## Tests

- `auth`: lookup match/normalize/miss, validation cases.
- `auth/api/iam`: authorization request (configured `client_id`, no scheme) and token request (configured `client_id` + secret) vs unchanged default path.
- `auth/client/iam`: `client_secret_post` on the wire; no secret for public client.
- `core`: config redaction of descendant/array keys.

Verified empirically against AET: public `client_id` → `invalid_client`; `client_secret_post` accepted.

🤖 Assisted by AI


reinkrul:
Keeping it in draft for now, as we don't know the exact form of client authentication against the AET IDP.

Issue #4328, opened 2026-06-05
OpenID4VCI: add authorization_request_params to requestCredential for issuer-specific authorization parameters


reinkrul:
## Context

Follow-up to the OpenID4VCI client work (#4316). While integrating a Nuts node with the AET ZORG-ID issuer, we found that AET's authorization server (OpenIddict) needs an extra query parameter on the authorization request to start smartcard login. From AET's docs:

> In order to use the ZORG-ID smartcard an extra redirect parameter needs to be added to the redirect in the authorize call: `auth_method=SmartCard`.

The Nuts `requestCredential` API could not add this parameter, so the redirect to AET never carried `auth_method=SmartCard` and the smartcard app (`zorgid://`) was never launched.

## Problem

The `requestOpenid4VCICredentialIssuance` request body has one field for passing extra parameters, `credential_request_params`, but it only affects the **Credential Request** (the call to the credential endpoint, `auth/openid4vci/client.go` -> `RequestCredential`). It does not affect the **authorization request**.

The authorization request query is a fixed list, built in `RequestOpenid4VCICredentialIssuance` (`auth/api/iam/openid4vci.go:131`): `response_type, state, client_id, client_id_scheme, authorization_details, redirect_uri, code_challenge, code_challenge_method`. There is no way for the caller to add anything to it.

| Field | Affects which request | Can add to the authorize redirect? |
|---|---|---|
| `credential_request_params` | Credential Request (POST to credential endpoint) | No |
| *(nothing today)* | Authorization request | -- |

So any issuer that needs an extra authorization parameter (AET's `auth_method`, and probably others) cannot be used through Nuts.

The ideal way to pass such information is the standard `authorization_details` (RFC 9396), which Nuts already sends. AET does not accept the smartcard selection there; it requires a separate top-level `auth_method` query parameter. This field is the workaround for issuers that, like AET, do not support the standard mechanism.

## Proposed change

Add a new optional field `authorization_request_params` to the `requestCredential` body. The node adds these key/value pairs to the authorization request, the same way `credential_request_params` works for the credential request. They are applied after the node's own parameters, so if the caller passes a key the node also sets, the caller's value is used. The caller is responsible for the result.

OpenAPI (`docs/_static/auth/v2.yaml`, `requestOpenid4VCICredentialIssuance` body):
```yaml
authorization_request_params:
  type: object
  additionalProperties:
    type: string
  description: |
[...]

PR #4333, opened 2026-06-05
feat(auth): add authorization_request_params to OpenID4VCI requestCredential API

reinkrul:
Implements #4328.

## What

Adds an optional `authorization_request_params` field to the `requestCredential` API body. The node merges these key/value pairs into the OpenID4VCI **authorization request** (the redirect to the Authorization Server's `authorization_endpoint`), mirroring the existing `credential_request_params` field for the Credential Request.

```json
{
  "issuer": "https://issuer.example.com/oauth",
  "wallet_did": "did:web:example.com",
  "authorization_details": [ { "type": "openid_credential", "credential_configuration_id": "..." } ],
  "redirect_uri": "https://example.com/oauth2/org1/callback",
  "authorization_request_params": { "auth_method": "SmartCard" }
}
```
-> redirect: `.../oauth/connect/authorize?...&auth_method=SmartCard`.

## Why

The AET ZORG-ID issuer requires a non-standard `auth_method=SmartCard` query parameter on the authorization request to start smartcard login. The authorization request query was a fixed list with no caller hook (`auth/api/iam/openid4vci.go`). The standard mechanism (`authorization_details`, RFC 9396) is preferred and already sent, but AET does not accept the smartcard selection there.

## Behavior

- May only **add** parameters. Attempting to override an OpenID4VCI parameter the node sets (`response_type`, `state`, `client_id`, `client_id_scheme`, `authorization_details`, `redirect_uri`, `code_challenge`, `code_challenge_method`) fails loud with a 400 -- no silent override.
- Values are strings (`map[string]string`) -- URL query parameters.
- Omitted/absent -> no change to today's behavior. Fully additive.

## Changes

- `docs/_static/auth/v2.yaml` -- new request-body field; regenerated `auth/api/iam/generated.go` and `e2e-tests/browser/client/iam/generated.go` via `make gen-api`.
- `auth/api/iam/openid4vci.go` -- merge the params in `RequestOpenid4VCICredentialIssuance`, rejecting overrides of node parameters.
- `auth/api/iam/openid4vci_test.go` -- assert the param lands in the redirect query, and that overriding a node parameter is rejected.

## Notes

Also running on `project-gf-pilot` (where the field was first added as `authorization_params`, then renamed here to align with `credential_request_params`). The override-protection change in this PR is not yet on `project-gf-pilot`.

🤖 Assisted by AI

### DIFF
--- a/auth/api/iam/openid4vci.go
+++ b/auth/api/iam/openid4vci.go
@@ -128,7 +128,7 @@ func (r Wrapper) RequestOpenid4VCICredentialIssuance(ctx context.Context, reques
 	if err != nil {
 		return nil, fmt.Errorf("failed to parse the authorization_endpoint: %w", err)
 	}
-	redirectUrl := nutsHttp.AddQueryParams(*authorizationEndpoint, map[string]string{
+	authzParams := map[string]string{
 		oauth.ResponseTypeParam:         oauth.CodeResponseType,
 		oauth.StateParam:                state,
 		oauth.ClientIDParam:             clientID.String(),
@@ -137,7 +137,15 @@ func (r Wrapper) RequestOpenid4VCICredentialIssuance(ctx context.Context, reques
 		oauth.RedirectURIParam:          redirectUri.String(),
 		oauth.CodeChallengeParam:        pkceParams.Challenge,
 		oauth.CodeChallengeMethodParam:  pkceParams.ChallengeMethod,
-	})
+	}
+	// Optional caller-supplied authorization request parameters, for issuers that need extras
+	// (e.g. auth_method=SmartCard). Applied after the node's own parameters, so caller values win.
+	if request.Body.AuthorizationRequestParams != nil {
+		for key, value := range *request.Body.AuthorizationRequestParams {
+			authzParams[key] = value
+		}
+	}
+	redirectUrl := nutsHttp.AddQueryParams(*authorizationEndpoint, authzParams)
 
 	return RequestOpenid4VCICredentialIssuance200JSONResponse{
 		RedirectURI: redirectUrl.String(),
--- a/auth/api/iam/openid4vci.go
+++ b/auth/api/iam/openid4vci.go
@@ -139,9 +139,13 @@ func (r Wrapper) RequestOpenid4VCICredentialIssuance(ctx context.Context, reques
 		oauth.CodeChallengeMethodParam:  pkceParams.ChallengeMethod,
 	}
 	// Optional caller-supplied authorization request parameters, for issuers that need extras
-	// (e.g. auth_method=SmartCard). Applied after the node's own parameters, so caller values win.
+	// (e.g. auth_method=SmartCard). These may only add parameters; they must not override the
+	// OpenID4VCI parameters set by the node above, which are essential to the flow.
 	if request.Body.AuthorizationRequestParams != nil {
 		for key, value := range *request.Body.AuthorizationRequestParams {
+			if _, isNodeParam := authzParams[key]; isNodeParam {
+				return nil, core.InvalidInputError("authorization_request_params may not override the '%s' parameter set by the node", key)
+			}
 			authzParams[key] = value
 		}
 	}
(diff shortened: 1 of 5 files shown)


docs/_static/auth/v2.yaml
```diff
@@ -187,6 +187,19 @@ paths:
                     {
                       "some-requested-credential-attribute": "900184590"
                     }
+                authorization_request_params:
+                  type: object
+                  additionalProperties:
+                    type: string
+                  description: |
+                    Optional key/value pairs added to the OpenID4VCI authorization request (the redirect to the
+                    Authorization Server's authorization_endpoint). If a key is also set by the node, the value given
```
stevenvegt: "Provided values have precedence over existing calculated or default values."

docs/_static/auth/v2.yaml
```diff
@@ -187,6 +187,19 @@ paths:
                     {
                       "some-requested-credential-attribute": "900184590"
                     }
+                authorization_request_params:
+                  type: object
+                  additionalProperties:
+                    type: string
+                  description: |
+                    Optional key/value pairs added to the OpenID4VCI authorization request (the redirect to the
```
stevenvegt, line 195: It can overwrite ANY query param in the authz request?

auth/api/iam/openid4vci.go
```diff
@@ -137,7 +137,15 @@ func (r Wrapper) RequestOpenid4VCICredentialIssuance(ctx context.Context, reques
 		oauth.RedirectURIParam:          redirectUri.String(),
 		oauth.CodeChallengeParam:        pkceParams.Challenge,
 		oauth.CodeChallengeMethodParam:  pkceParams.ChallengeMethod,
-	})
+	}
+	// Optional caller-supplied authorization request parameters, for issuers that need extras
+	// (e.g. auth_method=SmartCard). Applied after the node's own parameters, so caller values win.
+	if request.Body.AuthorizationRequestParams != nil {
+		for key, value := range *request.Body.AuthorizationRequestParams {
+			authzParams[key] = value
```
stevenvegt, line 149: I'm not sure if this is a good idea. This allows a caller to overwrite any of the node-generated or other in the request provided values. Including the pkcePramas, the redirect URI, clientID etc.
If this PR only serves to add values, then change the order: first the autzRequestParams and the apply the authzParams. If you want to allow certain fields to be overridden, make an allow list which can be overridded or a list of preserved values which never can be overwritten.

stevenvegt (changes requested):

auth/api/iam/openid4vci.go
reinkrul: > I'm not sure if this is a good idea. This allows a caller to overwrite any of the node-generated or other in the request provided values. Including the pkcePramas, the redirect URI, clientID etc. If this PR only serves to add values, then change the order: first the autzRequestParams and the apply the authzParams. If you want to allow certain fields to be overridden, make an allow list which can be overridded or a list of preserved values which never can be overwritten.

Why?

There's a strict trust between the caller of this API (EHR) and the Nuts node, so I think it's not a security concern. It would even allow use cases the Nuts node doesn't explicitly support (can't think of one), or is that something we explicitly don't want?

docs/_static/auth/v2.yaml
reinkrul: yes

auth/api/iam/openid4vci.go
stevenvegt: It requires the caller to sanitize its input. There is a risk that an attacker uses an operation to manipulate the contents of this object. It's a footgun. And if we already know that some of these values should never be changed (like the pkce values) we should not allow it. I would like to see it as additive and perhaps mark fields as over-writable, but not trust the client to never make a mistake.

stevenvegt: See the issue's considerations:
> * Because caller values are applied last, a caller could overwrite parameters the node sets (client_id, redirect_uri, etc.). This matches how credential_request_params already works. If we want to prevent this, we can deny a small set of keys (for example code_challenge).

When we say we trust the caller, we don't say we trust that every value is attacker-free. The reason many oauth params are often retrieved from a config is because is removes the risk of a trusted but vulnerable client.
Also, we currently only need additional params, no need to change existing. So I would argue to implement it like that.

stevenvegt (approved):

PR #4335, opened 2026-06-08
test(storage): fix flaky Test_logLeiaQueryStats

reinkrul:
## Summary

`Test_logLeiaQueryStats` intermittently fails (seen on #4331's CI run).

## Root cause

`logLeiaQueryStats` logs via `log.Logger()`, which wraps the **global** `logrus.StandardLogger()`. The test hooks that global logger and asserts on `hook.LastEntry()`. Other tests and background goroutines (e.g. BBolt store backups) write to the same global logger concurrently, so `LastEntry()` can return an unrelated entry — failing the level/message/field assertions.

## Fix

Match the specific entry by message substring via `hook.AllEntries()` instead of relying on `LastEntry()`. No production code change.

Verified stable with `go test ./storage/ -run Test_logLeiaQueryStats -count=20`.

Assisted by AI

### DIFF
--- a/storage/engine_test.go
+++ b/storage/engine_test.go
@@ -334,6 +334,17 @@ func Test_engine_sessionDatabase(t *testing.T) {
 }
 
 func Test_logLeiaQueryStats(t *testing.T) {
+	// logLeiaQueryStats logs to the global logrus logger, which other tests and background
+	// goroutines also write to concurrently. Match the specific entry by message rather than
+	// relying on LastEntry(), which is racy.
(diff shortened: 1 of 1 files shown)


Dirklectisch (approved):
The code looks good to me and running the suggested test works.

I couldn't verify on main that the suggested command broke even at much higher concurrency levels than 20.

`go test ./storage/ -run Test_logLeiaQueryStats -count=2000`

Probably only surfaces in more complicated scenarios where other threads are logging as well.

PR #4346, opened 2026-06-10
feat(auth): support both insert and append metadata discovery (#4310)

### LINKED ISSUE #4310: Metadata discovery: support both insert (RFC 8414) and append (OIDC Discovery) well-known forms
Related: #4248, #4233, #4156

## Problem Statement

Nuts derives metadata URLs using only the RFC 8414 **insert** convention: the well-known segment is placed at the authority root and the issuer/AS path is appended after it (`https://host/.well-known/<doc>/<path>`). This is implemented in `credentialIssuerWellKnown` (`auth/openid4vci/client.go:248`) for Credential Issuer Metadata and in `oauth.IssuerIdToWellKnown` (`auth/oauth/types.go:255`) for OAuth Authorization Server Metadata.

Many issuers and authorization servers publish metadata only under the OIDC Discovery **append** convention, where the well-known segment comes *after* the path (`https://host/<path>/.well-known/<doc>`). Against those servers Nuts gets a 404 (or 401/403/405) on metadata discovery and cannot complete issuance or token flows at all. Today the only workaround is per-deployment nginx rewriting that bridges insert -> append, which every operator integrating with such a partner has to set up and maintain by hand.

Both metadata documents are affected:
- **Credential Issuer Metadata** (`openid-credential-issuer`) -- `auth/openid4vci/client.go`
- **OAuth Authorization Server Metadata** (`oauth-authorization-server`) -- `auth/client/iam/client.go` via `auth/oauth`

## Solution

When the issuer/AS identifier has a non-empty path, try the candidate well-known locations in priority order and take the first `2xx` response:

1. **insert** (RFC 8414, current behavior, spec-preferred)
2. **append** (OIDC Discovery style)

This is spec-defensible: RFC 8414 §5 acknowledges the OIDC Discovery append style for compatibility, and OpenID4VCI permits retrieving AS metadata via either RFC 8414 or OIDC Discovery. The insert form stays first so spec-compliant servers are unaffected and no extra request is made against them on the happy path.

When the identifier has **no path**, insert and append collapse to the same URL -- derivation is unchanged and only one request is made. The fallback only widens behavior; it never narrows it. Each candidate URL is independently validated for SSRF (`core.ParsePublicURL` / `validateURL`) before any request, and the existing identifier-match check (`credential_issuer` / `issuer` must equal the requested identifier) is applied to whichever candidate returns `2xx`, so the fallback cannot be steered to an attacker-chosen document.

We considered making the convention an explicit per-issuer config flag but rejected it: operators integrating with these partners can't be expected to know which convention a given server uses, and the identifier-match check already makes "try both" safe.

## User Stories

<details>
<summary>User Stories</summary>

1. As a node operator integrating with an append-convention authorization server, I want Nuts to discover its metadata automatically, so that I don't have to run nginx rewrite rules per deployment.
2. As a wallet/holder node, I want credential issuance against an append-convention Credential Issuer to succeed, so that issuance isn't blocked by a 404 on metadata discovery.
3. As a developer, I want spec-compliant (insert) servers to keep working with a single request on the happy path, so that the fallback adds no latency or load against them.

</details>

### Candidate derivation and ordering

Add a new helper in `auth/oauth` that, given an identifier and a well-known document name, returns the ordered candidate URLs. `IssuerIdToWellKnown` is left untouched for its existing single-URL callers.

```go
// wellKnownCandidates returns the metadata URLs to try, in priority order:
//  1. insert (RFC 8414):  https://host/.well-known/<doc>/<path>
//  2. append (OIDC):       https://host/<path>/.well-known/<doc>
// When the identifier has no path, both collapse to the same URL and a single
// candidate is returned. Caller fetches in order and takes the first 2xx.
func wellKnownCandidates(identifier, wellKnown string) ([]string, error)
```

Insert form (candidate 1) is exactly today's derivation, preserving the existing `RawPath`/`%2F` double-escape handling already in `credentialIssuerWellKnown`.

### Fetch-with-fallback and failure modes

Each fetch wrapper iterates candidates and stops at the first `2xx` whose body passes the identifier-match check:

- Fall through to the next candidate on **any non-2xx** status (servers commonly use `401`/`403`/`405`, not `404`, for a path they don't serve) or an identifier-match failure on a `2xx` body.
- **Accumulate the failures across all attempted candidates.** When all are exhausted, **fail loud** with an error that names the identifier and reports the **non-404 failures only** -- a `404` just means "not at this location" and is noise, whereas a `403`/`405`/`5xx` is the diagnostic worth surfacing. If every candidate returned `404`, report a plain "metadata not found at any candidate" naming the identifier.
- Preserve the existing severity mapping: if any accumulated failure was `>= 500`, surface it as `502 Bad Gateway` (matching `OAuthAuthorizationServerMetadata` today at `auth/client/iam/client.go:78`). No silent empty-metadata fallback.
- Redirects (`3xx`) are left to the HTTP client's default follow behavior; only the final status surfaces to this loop.

## Testing Decisions

A good test asserts external behavior: given a server that serves metadata only at the append URL (and 404s the insert URL), discovery succeeds; given an insert-only server, discovery succeeds with a single request; given a server that 404s every candidate, discovery fails loud with an error naming the identifier; given a candidate that returns `403`/`500`, that status is preserved in the final error after the remaining candidates are exhausted. Test against an `httptest.Server` with per-path handlers rather than asserting derived URL strings.

## Impact Assessment

**Backwards compatibility**: fully additive. Insert-convention servers see identical behavior and a single request on the happy path.
**Versioning**: minor.
**Configuration/deployment**: no new config keys. Operators can remove per-deployment nginx metadata-rewrite rules once on this version.
**Security**: no new key material. New outbound requests against the same host are still SSRF-validated per candidate (`core.ParsePublicURL`), and the identifier-match check is enforced on the accepted document, so the fallback can't be redirected to an attacker-chosen metadata file. The widened candidate set means up to 2-3 outbound GETs against partner hosts on a miss -- a minor amplification, bounded and same-origin.
**Spec stability**: RFC 8414 (final) §5; OpenID4VCI 1.0 metadata retrieval.

## Out of Scope

- Per-issuer explicit convention configuration -- rejected above; revisit only if the identifier-match check proves insufficient.
- Caching of which convention a given host uses across requests -- possible later optimization, not needed for correctness.
- Changing how Nuts *serves* its own metadata (this PRD is client-side discovery only).
- Consolidating SSRF validation into the shared HTTP client (already tracked as the TODO at `auth/openid4vci/client.go:104`).

## Further Notes

The two documents live in different packages with different fetch wrappers (`openid4vci.client` uses a raw `http.Client`; `iam.HTTPClient` uses `doGet`), so the shared piece is candidate *derivation*, not the fetch loop. Each wrapper keeps its own loop to preserve its existing error mapping.

reinkrul:
Closes #4310

## Problem
Metadata discovery derived URLs using only the RFC 8414 **insert** convention (well-known segment at the authority root, issuer/AS path appended: `https://host/.well-known/<doc>/<path>`). Servers that publish metadata only under the OIDC Discovery **append** convention (`https://host/<path>/.well-known/<doc>`) returned 404/401/403/405, so issuance and token flows could not complete without per-deployment nginx rewrite rules.

Affected:
- Credential Issuer Metadata (`openid-credential-issuer`) — `auth/openid4vci`
- OAuth Authorization Server Metadata (`oauth-authorization-server`) — `auth/client/iam`

## Change
Discovery tries both placements in priority order and takes the first matching document:

1. **insert** (RFC 8414) — unchanged happy path, spec-preferred
2. **append** (OIDC Discovery)

When the identifier has no path the two collapse to one URL, so spec-compliant servers still make a single request — no added latency on the happy path.

Implemented by consolidating the previously-duplicated per-type fetch loops into one generic helper `oauth.FetchMetadata[T]`:
- Each metadata type declares its well-known path via `WellKnownPath()` and its issuer via `GetIssuer()`; the helper derives the candidate URLs (preserving the `%2F`/`RawPath` handling), fetches the first that returns 200 and JSON-decodes, and validates the issuer match.
- Identifier match is byte-exact, per RFC 8414 §3.3 / OpenID4VCI §12.2.4 (no trailing-slash or other normalization).
- Each candidate shares the identifier's host and is SSRF-validated via `core.ParsePublicURL`; the identifier-match check (`issuer` / `credential_issuer` must equal the requested identifier) is enforced on the accepted document, so the fallback cannot be steered to an attacker-chosen file.
- Every candidate is always tried, even if an earlier one fails with a 5xx. On exhaustion the returned error joins every candidate's failure (`errors.Join`), naming the identifier and the tried locations; a per-candidate `core.HttpError` stays discoverable via `errors.AsType` so callers can still detect an upstream 5xx.
- `iam.HTTPClient.OAuthAuthorizationServerMetadata` returns `FetchMetadata`'s error as-is, no longer rewriting an upstream 5xx to a `502 Bad Gateway` `core.HttpError` (which discarded the real upstream status). `FetchMetadata` tracks, while iterating candidates, whether every failure was a clean HTTP 4xx (not a 5xx, network/decode failure, or identifier mismatch); when so, the returned error also wraps the new `oauth.ErrAllCandidates4xx`. The iam client uses that to still classify that narrow case as `ErrInvalidClientCall` (400 Bad Request) — every candidate outright rejecting the request suggests the identifier itself is wrong. Any other failure (5xx, network error, identifier mismatch) falls through to the API's default `500 Internal Server Error` instead.
- `IdentifiersMatch` is now package-internal. The signed-JWT `OpenIDConfiguration` keeps its own (master) implementation.

## Compatibility
Fully additive. Insert-convention servers behave identically with a single request. No new config keys; operators can remove per-deployment nginx metadata-rewrite rules.

Note: the iam path now enforces the `issuer` identifier-match check on the accepted AS metadata document (previously absent), with a byte-exact comparison. Nuts-served metadata sets this correctly, but a partner whose metadata `issuer` field is not byte-identical to the requested identifier (e.g. a trailing-slash difference) would now be rejected rather than silently accepted.

## Tests
The insert/append fallback, identifier-match, and error-joining behavior is exhaustively covered once, at the `oauth.FetchMetadata` level; the two callers only test their own wiring on top of it (well-known constant, error-message prefix, `strictMode`), not that behavior again.
- `auth/oauth`: `FetchMetadata` — insert single-request, append fallback, identifier-mismatch-falls-through, byte-exact issuer match (trailing-slash rejected), all-404 names tried locations and wraps `ErrAllCandidates4xx`, a candidate's `core.HttpError` stays recoverable through `errors.Join` and a 5xx does not wrap `ErrAllCandidates4xx` (also proves both candidates are tried, not short-circuited on a 5xx), an identifier mismatch does not wrap `ErrAllCandidates4xx` either, invalid JSON, invalid identifier; plus `wellKnownCandidates` ordering.
- `auth/openid4vci`: fetch/parse from the well-known path, issuer-path derivation, percent-encoding preserved, `credential_issuer` mismatch rejected, all-candidates-404 (error carries the `openid4vci:` prefix), strict-mode/query-fragment URL validation, bad JSON.
- `auth/client/iam`: append-form fallback (proves end-to-end wiring), upstream errors returned as-is without rewriting the status code, all-candidates-4xx classifies as `ErrInvalidClientCall`.

Assisted by AI

### DIFF
--- a/auth/client/iam/client.go
+++ b/auth/client/iam/client.go
@@ -92,7 +92,7 @@ func (hb HTTPClient) OAuthAuthorizationServerMetadata(ctx context.Context, oauth
 		// Identifier-match check (RFC 8414 §3.3): the returned issuer MUST equal the
 		// requested identifier, so the fallback cannot be steered to a document the
 		// host serves under a different issuer. A mismatch falls through.
-		if metadata.Issuer != oauthIssuer {
+		if !oauth.IdentifiersMatch(metadata.Issuer, oauthIssuer) {
 			failures = append(failures, candidateFailure{
 				url: candidate,
 				err: fmt.Errorf("issuer %q does not match requested %q", metadata.Issuer, oauthIssuer),
--- a/auth/client/iam/client.go
+++ b/auth/client/iam/client.go
@@ -71,12 +71,11 @@ func (hb HTTPClient) OAuthAuthorizationServerMetadata(ctx context.Context, oauth
 	// Many authorization servers publish metadata only under the append convention.
 	metadata, err := oauth.FetchMetadata[oauth.AuthorizationServerMetadata](ctx, hb.httpClient, oauthIssuer, hb.strictMode)
 	if err != nil {
-		// An upstream server error (5xx) is surfaced as a core.HttpError; map it to 502
-		// Bad Gateway. Any other failure is a bad client call.
+		// An upstream server error (5xx) is returned as-is, carrying its original status.
+		// Any other failure is a bad client call.
 		httpErr, ok := errors.AsType[core.HttpError](err)
 		if ok && httpErr.StatusCode >= 500 {
-			httpErr.StatusCode = http.StatusBadGateway
-			return nil, httpErr
+			return nil, err
 		}
 		return nil, errors.Join(ErrInvalidClientCall, err)
 	}
--- a/auth/client/iam/client.go
+++ b/auth/client/iam/client.go
@@ -41,6 +41,9 @@ import (
 	"github.com/nuts-foundation/nuts-node/vcr/pe"
 )
 
+// ErrInvalidClientCall is returned when the node makes a http call as client based on wrong information passed by the client.
+var ErrInvalidClientCall = errors.New("invalid client call")
+
 // ErrBadGateway is returned when the node makes a http call as client and the upstream returns an unexpected result.
 var ErrBadGateway = errors.New("upstream returned unexpected result")
 
@@ -66,7 +69,13 @@ func (hb HTTPClient) OAuthAuthorizationServerMetadata(ctx context.Context, oauth
 	//  1. insert (RFC 8414):  https://host/.well-known/oauth-authorization-server/<path>
 	//  2. append (OIDC Disc): https://host/<path>/.well-known/oauth-authorization-server
 	// Many authorization servers publish metadata only under the append convention.
-	return oauth.FetchMetadata[oauth.AuthorizationServerMetadata](ctx, hb.httpClient, oauthIssuer, hb.strictMode)
+	metadata, err := oauth.FetchMetadata[oauth.AuthorizationServerMetadata](ctx, hb.httpClient, oauthIssuer, hb.strictMode)
+	if err != nil && errors.Is(err, oauth.ErrAllCandidates4xx) {
+		// Every candidate rejected the request outright (no 5xx, no network/decode failure, no
+		// identifier mismatch): the identifier itself is most likely wrong.
+		return nil, errors.Join(ErrInvalidClientCall, err)
+	}
+	return metadata, err
 }
 
 // ClientMetadata retrieves the client metadata from the client metadata endpoint given in the authorization request.
--- a/auth/client/iam/client.go
+++ b/auth/client/iam/client.go
@@ -41,9 +41,6 @@ import (
 	"github.com/nuts-foundation/nuts-node/vcr/pe"
 )
 
-// ErrInvalidClientCall is returned when the node makes a http call as client based on wrong information passed by the client.
-var ErrInvalidClientCall = errors.New("invalid client call")
-
 // ErrBadGateway is returned when the node makes a http call as client and the upstream returns an unexpected result.
 var ErrBadGateway = errors.New("upstream returned unexpected result")
 
@@ -69,17 +66,7 @@ func (hb HTTPClient) OAuthAuthorizationServerMetadata(ctx context.Context, oauth
 	//  1. insert (RFC 8414):  https://host/.well-known/oauth-authorization-server/<path>
 	//  2. append (OIDC Disc): https://host/<path>/.well-known/oauth-authorization-server
 	// Many authorization servers publish metadata only under the append convention.
-	metadata, err := oauth.FetchMetadata[oauth.AuthorizationServerMetadata](ctx, hb.httpClient, oauthIssuer, hb.strictMode)
-	if err != nil {
-		// An upstream server error (5xx) is returned as-is, carrying its original status.
-		// Any other failure is a bad client call.
-		httpErr, ok := errors.AsType[core.HttpError](err)
-		if ok && httpErr.StatusCode >= 500 {
-			return nil, err
-		}
-		return nil, errors.Join(ErrInvalidClientCall, err)
-	}
-	return metadata, nil
+	return oauth.FetchMetadata[oauth.AuthorizationServerMetadata](ctx, hb.httpClient, oauthIssuer, hb.strictMode)
 }
 
 // ClientMetadata retrieves the client metadata from the client metadata endpoint given in the authorization request.
--- a/auth/client/iam/client.go
+++ b/auth/client/iam/client.go
@@ -68,84 +68,19 @@ func (hb HTTPClient) OAuthAuthorizationServerMetadata(ctx context.Context, oauth
 	// returns a matching document:
 	//  1. insert (RFC 8414):  https://host/.well-known/oauth-authorization-server/<path>
 	//  2. append (OIDC Disc): https://host/<path>/.well-known/oauth-authorization-server
-	//  3. append openid-configuration: https://host/<path>/.well-known/openid-configuration
 	// Many authorization servers publish metadata only under the append convention.
-	candidates, err := oauth.WellKnownCandidates(oauthIssuer, oauth.AuthzServerWellKnown, hb.strictMode)
+	metadata, err := oauth.FetchMetadata[oauth.AuthorizationServerMetadata](ctx, hb.httpClient, oauthIssuer, hb.strictMode)
 	if err != nil {
-		return nil, err
-	}
-	// OpenID4VCI also permits retrieving AS metadata via OIDC Discovery's openid-configuration
-	// document; add its append form (the last candidate) as a final fallback.
-	oidcCandidates, err := oauth.WellKnownCandidates(oauthIssuer, oauth.OpenIdConfigurationWellKnown, hb.strictMode)
-	if err != nil {
-		return nil, err
-	}
-	candidates = append(candidates, oidcCandidates[len(oidcCandidates)-1])
-
-	var failures []candidateFailure
-	for _, candidate := range candidates {
-		var metadata oauth.AuthorizationServerMetadata
-		if getErr := hb.doGet(ctx, candidate, &metadata); getErr != nil {
-			failures = append(failures, candidateFailure{url: candidate, err: getErr, status: statusCodeOf(getErr)})
-			continue
+		// An upstream server error (5xx) is surfaced as a core.HttpError; map it to 502
+		// Bad Gateway. Any other failure is a bad client call.
+		httpErr, ok := errors.AsType[core.HttpError](err)
+		if ok && httpErr.StatusCode >= 500 {
+			httpErr.StatusCode = http.StatusBadGateway
+			return nil, httpErr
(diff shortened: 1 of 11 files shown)


auth/client/iam/client.go
```diff
@@ -68,84 +68,19 @@ func (hb HTTPClient) OAuthAuthorizationServerMetadata(ctx context.Context, oauth
 	// returns a matching document:
 	//  1. insert (RFC 8414):  https://host/.well-known/oauth-authorization-server/<path>
 	//  2. append (OIDC Disc): https://host/<path>/.well-known/oauth-authorization-server
```
JorisHeadease, line 70: nit: #4310 also mentions adding the `openid-configuration` form, but it was removed in the PR as it kept its own implementation. Might be good to update the ticket body for future references

auth/oauth/metadata.go
```diff
@@ -0,0 +1,147 @@
...
+	if err := json.NewDecoder(resp.Body).Decode(&metadata); err != nil {
+		return nil, resp.StatusCode, fmt.Errorf("decoding metadata: %w", err)
+	}
+	// The issuer in the document MUST match the requested identifier, so a host cannot
+	// steer discovery to metadata it serves under a different issuer. A mismatch falls
+	// through to the next candidate.
+	if !identifiersMatch(metadata.GetIssuer(), identifier) {
+		return nil, resp.StatusCode, fmt.Errorf("issuer %q does not match requested identifier %q", metadata.GetIssuer(), identifier)
+	}
+	return &metadata, resp.StatusCode, nil
```
JorisHeadease: Do we always want to throw an error here, or just in `strictMode`?

auth/oauth/metadata.go
```diff
@@ -0,0 +1,147 @@
...
+	for _, candidate := range candidates {
+		metadata, status, fetchErr := fetchMetadataCandidate[T](ctx, httpClient, candidate, identifier)
+		if fetchErr == nil {
+			return metadata, nil
+		}
+		// Surface an upstream server error as-is (it is a core.HttpError carrying the
+		// status) so callers serving an API can map it to 502 Bad Gateway.
+		if status >= 500 {
+			return nil, fetchErr
+		}
```
JorisHeadease: Shouldn't we try other candidates over directly returning the error? The function states 
> When every candidate fails it returns...

which is not true in this case.

auth/oauth/metadata.go
```diff
@@ -0,0 +1,147 @@
...
+
+// identifiersMatch reports whether two issuer / credential-issuer identifiers are
+// equal, tolerating a trailing-slash difference. RFC 8414 §3.3 and OpenID4VCI
+// §12.2.4 require the identifier in a metadata document to be byte-identical to
+// the requested identifier, but some servers (e.g. IdentityServer) normalize it
+// with a trailing slash. Treat those as a match so discovery does not reject
+// otherwise-valid metadata over a trailing slash.
+func identifiersMatch(a string, b string) bool {
+	return strings.TrimRight(a, "/") == strings.TrimRight(b, "/")
+}
```
JorisHeadease: Do we really want this? v1 explicitly states NOT to use this:

> credential_issuer: REQUIRED. The Credential Issuer's identifier, as defined in [Section 12.2.1](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#credential-issuer-identifier). The value MUST be identical to the Credential Issuer's identifier value into which the well-known URI string was inserted to create the URL used to retrieve the metadata. If these values are not identical (**when compared using a simple string comparison with no normalization**), the data contained in the response MUST NOT be used.

Feels like this problem should be solved elsewhere. And if we do decide to allow it, maybe make the normalization configurable (and off by default), or explicitly state that we deliberately deviate from the spec

JorisHeadease (changes requested):

auth/oauth/metadata.go
reinkrul: I agree, I'll talk to AET about it

auth/oauth/metadata.go
reinkrul: My thought was that > 500 normally indicates service being down, so it might still be the right endpoint, but the cost of this approach is potential false negatives and complexity here.

The downside of just trying all candidates, regardless of the error is that we potentially try one URL too much (because the first is down, the second one doesn't exist), but that's an edge case and just temporary (until service is restored).

I'll revert back to the simpler flow.

auth/oauth/metadata.go
reinkrul: Reverting this

auth/oauth/metadata.go
```diff
@@ -0,0 +1,147 @@
...
+		if fetchErr == nil {
+			return metadata, nil
+		}
+		// Surface an upstream server error as-is (it is a core.HttpError carrying the
+		// status) so callers serving an API can map it to 502 Bad Gateway.
+		if status >= 500 {
+			return nil, fetchErr
+		}
+		if status != http.StatusNotFound {
+			diagnostics = append(diagnostics, fmt.Sprintf("%s: %v", candidate, fetchErr))
```
reinkrul: maybe we should add the response codes

auth/oauth/metadata.go
reinkrul: the EHR is the client, so they're within the same trust boundary, so it's fine

reinkrul:
Also removed the 502-rewrite in \`OAuthAuthorizationServerMetadata\`: on an upstream 5xx it used to overwrite \`core.HttpError.StatusCode\` with \`http.StatusBadGateway\`, discarding the real upstream status (e.g. a 503 became a 502 in the returned error). It now returns the upstream error as-is.

JorisHeadease (approved):

PR #4360, opened 2026-06-18
test(docs): fail when generated docs are out of date

reinkrul:
## Why
Contributors who add/change a config option or CLI command must run `make cli-docs`, but it's easy to forget — the docs then drift from the code. This adds a unit test that regenerates the docs in-memory and fails if they differ from what's committed, so CI catches it.

## Changes
- **Refactor** `docs/generate_docs.go`: split the generators into render-to-memory functions (`generatedDocFiles` returns `path → content`); `generateDocs()` now just writes that map to disk. Output is byte-identical to before.
- **Test** `TestGeneratedDocsAreUpToDate`: compares `generatedDocFiles()` against the committed files; on mismatch fails with `"<file> is out of date; run \`make cli-docs\` and commit the result"`.
- **Regenerated** `server_options.rst`, which had drifted from the `policy.authzen.endpoint` flag description (needed for the new test to pass).

## Scope
- Covers the files produced by the Go generator: `cli-reference.rst`, `server_options.rst`, `server_options_didnuts.rst`.
- Does **not** cover `README.rst` — that's assembled by the separate `rst_include` step in `make cli-docs`, not by the Go generator.
- Map-typed flag ordering (the old source of spurious diffs) is already handled by `normalizeDefaultValue`; verified the generator output is stable across runs.

🤖 Assisted by AI


.github/workflows/go-test.yaml
```diff
@@ -37,7 +41,14 @@ jobs:
           filters: |
             code:
               - '**'
-              - '!docs/**'
+              - '!docs/_static/**'
```
stevenvegt: can be brough back to `!docs/*/` right, since all `.go` files are in the `/docs` root. This give a risk on a false positive when one of the non code files change directly under root, but there practically are none. Or add `!/docs/*.rst` to exclude the index file.

Reason i bring this up, is keeping the filter list up to date feels like a chore and leaks too much information about our codebase layout into this workflow config.

stevenvegt (approved):
LGTM, one question about the filter.

.github/workflows/go-test.yaml
reinkrul: I like it

Issue #4383, opened 2026-07-01
Migrate from lestrrat-go/jwx/v2 (EOL) to v3


reinkrul:
As of `v2.1.7`, `github.com/lestrrat-go/jwx/v2` is marked deprecated by the maintainers via Go module deprecation — the v2 series is no longer maintained. See the [v2.1.7 release notes](https://github.com/lestrrat-go/jwx/releases/tag/v2.1.7). No further bug or security fixes will land on v2.

jwx is a core JWT/JWS/JWK/crypto dependency, so staying on an unmaintained version is a security-relevant liability.

## Target: v3 (not v4, for now)

We evaluated jumping straight to v4 (latest major, `v4.1.0`) but chose **v3** (`v3.1.1`) as the immediate target.

**Why not v4 yet — `GOEXPERIMENT=jsonv2`:**
- v4 hard-depends on the standard library `encoding/json/v2` package (15 import sites in v4). Its [`MIGRATION.md`](https://github.com/lestrrat-go/jwx/blob/v4.1.0/MIGRATION.md) lists `GOEXPERIMENT=jsonv2` as a required build prerequisite.
- In Go 1.26 (our current toolchain: `go 1.26.3`, Docker `golang:1.26.4`), `json/v2` is still gated behind that experiment flag. Shipping v4 would mean setting `GOEXPERIMENT=jsonv2` in the Dockerfile builder, every CI workflow, and local dev — i.e. an **experimental Go feature baked into the production build** of a crypto/auth-sensitive node. Not acceptable right now.
- v4 is also two major versions away (v2 -> v3 -> v4); each hop has its own breaking changes.

**Why v3 is the right step:**
- Gets us off the EOL v2 now (the actual security goal).
- Single major-version hop, no experimental build flag, no companion-module split.
- v3 is actively maintained (`v3.1.1`, not deprecated).

**Revisit v4 later** once `json/v2` graduates out of `GOEXPERIMENT` (expected in a near-future Go release). At that point v3 -> v4 is largely mechanical via the [`jwxmigrate`](https://github.com/jwx-go/jwxmigrate) tool. Tracked as a follow-up.

## v2 -> v3 breaking changes to handle
- `jwa.RS256` (and all algorithm constants) become function calls `jwa.RS256()`.
- `xxx.Get(name)` signature changes to `Get(name, &dst) error`.
- Convenience accessors (e.g. `token.Subject()`) now return `(T, bool)` instead of `T`.
[...]

PR #4391, opened 2026-07-03
chore: migrate from lestrrat-go/jwx v2 (EOL) to v3

reinkrul:
Migrates all direct usage of `github.com/lestrrat-go/jwx` from the **EOL v2** series to **v3**. Closes part of #4383 (the `master` target).

## Dependency
Requires **go-did v0.22.0** (jwx v3) — pinned in `go.mod`. This was the last thing keeping jwx v2 in the tree (see nuts-foundation/go-did#155); the earlier merge blocker (branch pseudo-version) is now resolved.

## Why
jwx v2 is deprecated (EOL as of v2.1.7) — no further bug/security fixes. jwx is a core JWT/JWS/JWK/crypto dependency, so staying on it is a security liability. We target **v3, not v4**: v4 hard-depends on `encoding/json/v2` behind `GOEXPERIMENT=jsonv2`, which we won't bake into a production build until that graduates from the Go experiment (tracked as a follow-up in #4383).

## Key changes
- Imports rewritten v2 → v3; jwx v2 fully removed from `go.mod`/`go.sum`.
- `jwa` algorithm constants → function calls (`jwa.ES256` → `jwa.ES256()`); `jwa` values are now comparable structs.
- `Get()`/accessor call sites adapted to v3 signatures (single-value accessors now return `(T, bool)`; `Get(name, &dst) error`).
- `jwk.FromRaw` → `jwk.Import`; `(jwk.Key).Raw(&x)` → `jwk.Export(key, &x)`.
- **`didkey` resolver**: X25519 now built with `crypto/ecdh` (jwx dropped its `x25519` package in v3).

## Notable non-mechanical fixes (root-cause, behavior-preserving)
- **Raw signature verification**: v3 no longer registers algorithms for the low-level `jws.NewVerifier`; migrated the detached-signature verification in `crypto/jwx.go` and `vcr/signature/proof/jsonld.go` to `jwsbb.Verify(...)`.
- **`AsMap` / `PrivateClaims` removal**: replaced the interim per-claim `Get` loops with JSON-based extraction in `auth/client/iam`, `auth/api/iam/jar.go`, and `auth/services/oauth/authz_server.go`. v3's per-claim `Get` **errors on a null-valued claim**, which would have rejected otherwise-valid external tokens; JSON extraction preserves v2's `AsMap`/`PrivateClaims` tolerance.
- **Azure Key Vault**: `parseKey` now checks the key type/curve **before** `jwk.Export`, so an unsupported (e.g. RSA) key returns the intended `only ES256 keys are supported` rather than a leaky jwx-internal export error.

## Behavior changes surfaced by v3 (kept intentionally)
- **RSA ≥ 2048-bit enforced**: v3 rejects RSA keys under 2048 bits by default. Kept (security-positive); small test fixtures regenerated. `TestInsecure1024BitRS512JWT` still rejects the weak key.
- **Unsigned/self-attested credentials must use `alg:none`**: v3 rejects an empty compact signature on any other alg at parse time. Production already handles `alg:none`; affected test fixtures updated. Any external issuer emitting unsigned credentials must set `alg:none`.
- No `jku` / remote JWK fetching is introduced anywhere (v3 flipped the fetch whitelist default to allow-all — audited, we don't use it).

## Testing
- `go build ./...` clean; `go test ./...` green (107 packages).
- Test updates are error-string/wording alignment to v3 and fixture regeneration only — no assertions weakened.

## Follow-up
- Apply the same migration to `v6.2` and `v5.4` (#4383).
- Revisit jwx v4 once `encoding/json/v2` leaves `GOEXPERIMENT`.

Assisted by AI


auth/api/iam/api.go
```diff
@@ -402,7 +402,8 @@ func (r Wrapper) introspectAccessToken(input string) (*ExtendedTokenIntrospectio
 	// SHA256 hashing won't fail.
 	var cnf *Cnf
 	if token.DPoP != nil {
-		hash, _ := token.DPoP.Headers.JWK().Thumbprint(crypto.SHA256)
+		key, _ := token.DPoP.Headers.JWK()
+		hash, _ := key.Thumbprint(crypto.SHA256)
```
reinkrul: let's not ignore these errors

auth/api/iam/dpop.go
```diff
@@ -87,13 +87,14 @@ func (r Wrapper) ValidateDPoPProof(_ context.Context, request ValidateDPoPProofR
 		return ValidateDPoPProof200JSONResponse{Reason: &reason}, nil
 	}
 	// check if the jti is already used, if not add it to the store for the duration of the access token lifetime
+	jti, _ := dpopToken.Token.JwtID()
```
reinkrul, line 90: what if an error is returned?

auth/api/iam/jar.go
```diff
@@ -183,8 +184,12 @@ func (j jar) validate(ctx context.Context, rawToken string, clientId string) (oa
...
 		return nil, oauth.OAuth2Error{Code: oauth.InvalidRequestObject, Description: "request signature validation failed", InternalError: err}
 	}
-	claimsAsMap, err := token.AsMap(ctx)
-	if err != nil {
+	// Convert the token's claims to a map via its JSON representation. This mirrors jwx v2's
+	// token.AsMap: it preserves all claims including null-valued ones (a per-claim Get loop
+	// errors on null values in v3).
+	claimsAsMap := make(map[string]interface{})
+	claimsJSON, _ := json.Marshal(token)
+	if err := json.Unmarshal(claimsJSON, &claimsAsMap); err != nil {
```
reinkrul: isn't this a pattern we often have, and should we make a util function for it somewhere?

auth/api/iam/s2s_vptoken.go
```diff
@@ -226,8 +226,7 @@ func extractNonce(presentation vc.VerifiablePresentation) (string, error) {
 	var nonce string
 	switch presentation.Format() {
 	case vc.JWTPresentationProofFormat:
-		nonceRaw, _ := presentation.JWT().Get("nonce")
-		nonce, _ = nonceRaw.(string)
+		_ = presentation.JWT().Get("nonce", &nonce)
```
reinkrul, line 229: is it right to ignore the error?

auth/api/iam/validation.go
```diff
@@ -54,7 +54,7 @@ func (r Wrapper) validatePresentationAudience(presentation vc.VerifiablePresenta
 	var audience []string
 	switch presentation.Format() {
 	case vc.JWTPresentationProofFormat:
-		audience = presentation.JWT().Audience()
+		audience, _ = presentation.JWT().Audience()
```
reinkrul, line 57: is it right to ignore the error?

auth/client/iam/client.go
```diff
@@ -264,12 +264,17 @@ func (hb HTTPClient) OpenIDConfiguration(ctx context.Context, issuerURL string)
...
 		return nil, fmt.Errorf("unable to parse response: %w", err)
 	}
-	claims, err := token.AsMap(ctx)
-	if err != nil {
+	// Convert the token's claims to a map via its JSON representation. This mirrors jwx v2's
+	// token.AsMap: it preserves all claims including null-valued ones (a per-claim Get loop
+	// errors on null values in v3).
+	claims := make(map[string]interface{})
+	claimsJSON, _ := json.Marshal(token)
+	if err = json.Unmarshal(claimsJSON, &claims); err != nil {
```
reinkrul: duplicate?

auth/client/iam/client.go
```diff
@@ -264,12 +264,17 @@ func (hb HTTPClient) OpenIDConfiguration(ctx context.Context, issuerURL string)
...
 		return nil, fmt.Errorf("unable to parse response: %w", err)
 	}
 	// hack, broken iat
-	claims["iat"] = token.IssuedAt().Unix()
+	iat, _ := token.IssuedAt()
+	claims["iat"] = iat.Unix()
```
reinkrul, line 270: is it right to ignore this error?

crypto/jwx/jwx_es256k.go
```diff
@@ -20,7 +20,7 @@
 
 package jwx
 
-import "github.com/lestrrat-go/jwx/v2/jwa"
+import "github.com/lestrrat-go/jwx/v3/jwa"
 
 func init() {
 	AddSupportedAlgorithm(jwa.ES256K)
```
JorisHeadease:
```
➜  nuts-node git:(chore/jwx-v3-migration) go build -tags jwx_es256k ./crypto/...
# github.com/nuts-foundation/nuts-node/crypto/jwx
crypto/jwx/jwx_es256k.go:26:24: cannot use jwa.ES256K (value of type func() jwa.SignatureAlgorithm) as jwa.SignatureAlgorithm value in argument to AddSupportedAlgorithm
```

```suggestion
	AddSupportedAlgorithm(jwa.ES256K())
```

auth/api/iam/jar.go
```diff
@@ -183,7 +184,7 @@ func (j jar) validate(ctx context.Context, rawToken string, clientId string) (oa
 	if err != nil {
 		return nil, oauth.OAuth2Error{Code: oauth.InvalidRequestObject, Description: "request signature validation failed", InternalError: err}
 	}
-	claimsAsMap, err := token.AsMap(ctx)
+	claimsAsMap, err := jwx.AsMap(token)
```
JorisHeadease: The audience check in `handleAuthorizeRequestFromHolder` now always fails, breaking the JAR authorization flow. This is what's causing the e2e failure on this PR (invalid audience, expected: https://nodeA/oauth2/subjectA, was:  ).

`jar.validate` now uses the new `jwx.AsMap()` helper, which does a JSON round-trip. jwx v3 always marshals `aud` as a JSON array (`FlattenAudience` is not enabled anywhere in the codebase), so after `json.Unmarshal` the claim comes back as `[]interface{}{"..."}`. The old v2 `token.AsMap(ctx)` returned the audience as `[]string.oauthParameters.get()` in params.go only handles `string` and `[]string`, so `params.get(jwt.AudienceKey)` returns `""` and the check at `openid4vp.go:99` rejects every request.

The existing unit test doesn't catch this because it puts `aud` into the `oauthParameters` map directly as `[]string`, bypassing the JWT parse path.

crypto/jwx.go
```diff
@@ -112,17 +113,24 @@ func (client *Crypto) DecryptJWE(ctx context.Context, message string) (body []by
...
 	}
-	headers, err = msg.ProtectedHeaders().AsMap(ctx)
-	if err != nil {
-		return nil, nil, err
+	// Iterate the headers with Get (rather than jwx.AsMap's JSON round-trip) so rich header
+	// values such as the x5c certificate chain keep their concrete Go types for callers.
+	headers = make(map[string]interface{})
+	for _, k := range protectedHeaders.Keys() {
+		var v interface{}
+		if err = protectedHeaders.Get(k, &v); err != nil {
```
JorisHeadease: The v2 version `protectedHeaders.AsMap(ctx)` properly returned null-members as a `nil` in the map. 

This new per-field Get loop instead fails on them (`failed to assign value for field "...": source value is invalid (<nil>)`), so a JWS/JWE with a null-valued custom protected header is now rejected. The doc comment on [crypto/jwx/claims.go](https://github.com/nuts-foundation/nuts-node/pull/4391/crypto/jwx/claims.go) AsMap() describes exactly this pitfall:

> v3's per-field Get returns an error for a null-valued member, whereas the JSON round-trip preserves null members as nil - matching v2's AsMap behaviour and avoiding rejection of otherwise-valid input.

We probably want to handle null-valued members separately in this loop, keeping the concrete Go types for rich headers like x5c.

crypto/jwx.go
```diff
@@ -405,10 +411,15 @@ func ExtractProtectedHeaders(jwt string) (map[string]interface{}, error) {
...
-			var err error
-			headers, err = message.Signatures()[0].ProtectedHeaders().AsMap(context.Background())
-			if err != nil {
-				return nil, err
+			protectedHeaders := message.Signatures()[0].ProtectedHeaders()
+			// Iterate with Get (rather than jwx.AsMap's JSON round-trip) so rich header values
+			// such as the x5c certificate chain keep their concrete Go types for callers.
+			for _, k := range protectedHeaders.Keys() {
+				var v interface{}
+				if err := protectedHeaders.Get(k, &v); err != nil {
```
JorisHeadease: Same as https://github.com/nuts-foundation/nuts-node/pull/4391/changes#r3543872179

vcr/signature/proof/jsonld.go
```diff
@@ -142,7 +141,7 @@ func (p LDProof) Verify(document Document, suite signature.Suite, key crypto.Pub
 		return fmt.Errorf("could not base64 decode signature: %w", err)
 	}
 	challenge := fmt.Sprintf("%s.%s", splittedJws[0], tbv)
-	if err = jswVerifier.Verify([]byte(challenge), sig, key); err != nil {
+	if err = jwsbb.Verify(key, alg.String(), []byte(challenge), sig); err != nil {
```
JorisHeadease: Are you sure we want to use this over `jws.VerifierFor(alg)` here? The building block mentions the following:

```
// Package jwsbb provides the building blocks (hence the name "bb") for JWS operations.
// It should be thought of as a low-level API, almost akin to internal packages
// that should not be used directly by users of the jwx package. However, these exist
// to provide a more efficient way to perform JWS operations without the overhead of
// the higher-level jws package to power-users who know what they are doing.
//
// This package is currently considered EXPERIMENTAL, and the API may change
// without notice. It is not recommended to use this package unless you are
// fully aware of the implications of using it.
//
// All bb packages in jwx follow the same design principles:
// 1. Does minimal checking of input parameters (for performance); callers need to ensure that the parameters are valid.
// 2. All exported functions are strongly typed (i.e. they do not take `any` types unless they absolutely have to).
// 3. Does not rely on other public jwx packages (they are standalone, except for internal packages).
//
// This implementation uses github.com/lestrrat-go/dsig as the underlying signature provider.
```

crypto/jwx.go
```diff
@@ -298,7 +303,7 @@ func ParseJWS(token []byte, f PublicKeyFunc) (payload []byte, err error) {
 		for _, part := range parts {
 			payload = append(payload, part...)
 		}
-		err = verifier.Verify(payload, signature.Signature(), key)
+		err = jwsbb.Verify(key, alg.String(), payload, signature.Signature())
```
JorisHeadease: Same as https://github.com/nuts-foundation/nuts-node/pull/4391/changes#r3543944990

JorisHeadease (changes requested):

JorisHeadease:
There are also a lot of resolved comments about not ignoring the errors, but the code still ignores them

crypto/jwx/jwx_es256k_test.go
JorisHeadease:
```suggestion
		signature, _ := jwt.Sign(token, jwt.WithKey(jwa.ES256K(), ecKey))
```

reinkrul:
> There are also a lot of resolved comments about not ignoring the errors, but the code still ignores them

I checked them on case-by-case basis, and where they're ignored, they were ignored before. The semantic didn't change, as previously the single call basically did what is now done in 2 calls (meaning nil/error-behavior will be the same).

Edit: I added an explicit check for 2 of `JWK()` cases where the safety comes from another package.

vcr/signature/proof/jsonld.go
reinkrul: I missed the experimental flag... I'll see if we can use the usual jws.Verifier

crypto/jwx/jwx_es256k.go
reinkrul: The es256k tests are already broken on master, I fixed the build and made an issue to consider removing the es256k support (as nobody uses it); https://github.com/nuts-foundation/nuts-node/issues/4399

JorisHeadease (approved):

PR #4420, opened 2026-07-22
fix(http): block SSRF to non-public addresses in strict mode

stevenvegt:
## What

Two hardening changes to the shared strict-mode HTTP client:

1. In strict mode, refuse at connect time to reach a non-public address, as defined by the IANA special-purpose registries: loopback, private (RFC1918), unique local (ULA), link-local, CGNAT, NAT64, 6to4 and the remaining reserved space. The check runs on the resolved IP address, so a hostname that resolves into one of these ranges is caught as well.
2. In strict mode, refuse to follow a redirect that downgrades from HTTPS to HTTP.
3. In strict mode, always refuse cloud provider metadata endpoints (per the OWASP SSRF prevention cheat sheet), and let operators block additional ranges with `http.client.deniedcidrs`, for publicly routable ranges that are internal-only in their infrastructure. Denied ranges take precedence over the allowlist.

Deployments that legitimately reach a private address for an internal flow (such as an internal credential offering or an internal OAuth user flow) can permit specific ranges with `http.client.allowedinternalcidrs`.

## Why

In strict mode the shared HTTP client only checked that the initial request used HTTPS. Two weaknesses followed:

- The SSRF URL check (`core.ParsePublicURL`) was applied by each caller separately, and the `vcr/openid4vci` credential-offer handler did not apply it. That check also only looks at the URL string, so a hostname that resolves to an internal address passed it.
- The HTTPS check only covered the first request. The client follows redirects, so a remote host could return a redirect to a plaintext HTTP address. On that hop there is no TLS, so neither the HTTPS requirement nor the truststore applied, and the client would send a blind request to the redirect target, including an internal endpoint.

Federation targets are participants' public endpoints, so an outbound request to a non-public address is normally either a misconfiguration or an attempt to reach internal infrastructure, which is exactly what SSRF abuses. Blocking those ranges closes the gap; the allowlist covers the deployments that legitimately need an internal target.

## How

- Add a check on the shared transport (`http/client/ssrf.go`) that inspects the resolved IP address before it connects. In strict mode it refuses any address that is not public internet space, following the IANA special-purpose registries. IPv4 is checked against an explicit denylist (loopback, RFC1918, CGNAT, link-local including the cloud metadata address 169.254.169.254, benchmarking, documentation and reserved ranges). IPv6 is checked allowlist-style: only global unicast (`2000::/3`) is permitted, minus the special-purpose carve-outs inside it (6to4 and Teredo, which embed an IPv4 address, and the documentation ranges), so ULA, link-local, NAT64 and any future reservation outside global unicast are refused by construction. Because the check sees the resolved IP, it also catches a hostname that resolves into one of these ranges, and it runs on redirect targets as well.
- The prefix lists are cross-checked by a test against `code.dny.dev/ssrf` (a test-only dependency, generated upstream from the IANA registries), so a dependency bump that changes the upstream list fails the test and points at the registry that changed.
- Add `http.client.deniedcidrs`: CIDR ranges that are always refused in strict mode, taking precedence over `allowedinternalcidrs`. The built-in denied set covers cloud metadata endpoints on public address space (Azure's 168.63.129.16; the other providers' endpoints already fall in blocked non-public ranges) and cannot be re-admitted through the allowlist.
- Add `http.client.allowedinternalcidrs`: CIDR ranges that are exempt from the guard, for internal flows that legitimately target a private address. Empty by default, so nothing internal is reachable unless an operator opts in.
- Add a redirect policy (`CheckRedirect`) that refuses a redirect to a non-HTTPS target in strict mode, so every hop stays on HTTPS and stays subject to certificate validation. Setting `CheckRedirect` replaces the standard library default, so the ten-redirect limit is reimplemented.
- The checks live on the shared transport and client, so every client created through `client.New*` is covered.
- Non-strict mode is unchanged, so local and development setups keep working.

## Tests

- The address check is covered for all denied range families (IPv4 and IPv6), including CGNAT, NAT64, 6to4, IPv4-mapped IPv6 and zoned addresses, with public addresses allowed, allowlisted ranges allowed, and the non-strict passthrough. The allowlist and denylist parsers are covered for valid, invalid, and empty input. A drift test compares the prefix lists against `code.dny.dev/ssrf`.
- The denylist is covered for blocking public ranges, precedence over the allowlist, the default cloud metadata entries (including that the allowlist cannot re-admit them), and the non-strict passthrough.
- An integration test stands up an HTTPS server that redirects to a plaintext HTTP endpoint, drives it through the real client, and confirms the client does not follow the redirect and never contacts the plaintext endpoint. A non-strict subtest confirms redirects are still followed there.
- The strict-mode OpenID4VCI e2e tests federate over a Docker network on a private range that the guard now blocks. The nodes permit all RFC1918 ranges via `http.client.allowedinternalcidrs` (Docker auto-assigns the subnet, so pinning a specific range is not portable across runners), which covers the allowlist config end to end; narrow-allowlist precision is covered by unit tests.

## Notes

Verified the affected packages (`http`, `vcr`, `auth`, `discovery`, `vdr`, `policy`) still pass. Storage clients that may talk to private addresses (Vault, external key store, SQL and Redis) do not use this transport, so they are unaffected.

Reported by @raysabee.


stevenvegt:
e2e tests are failing since they use strict mode and are connecting to local docker container via an 172.x.x.x address.

Dirklectisch (approved):
The measures here looks ok to me but I would additionally suggest:

- What if the network configuration uses some other IP configuration. For example the company might own a range of IP's that are publicly routable but only used internally. Should you be able to configure additional blocked ranges in such a situation?
- [The OWASP guide to SSRF](https://cheatsheetseries.owasp.org/cheatsheets/Server_Side_Request_Forgery_Prevention_Cheat_Sheet.html#deny-list-last-resort) suggests blocking metadata endpoints of big cloud providers.

stevenvegt:
The OWASP guide shows IPs that are already blocked, so that point is covered.
Blocking additional public addresses is indeed a feature to make this fix complete. Not sure it will ever be used, but that is besides the point. I'll add it.

stevenvegt:
@Dirklectisch both suggestions are in as of d81cc2e5:

- New config `http.client.deniedcidrs`: CIDR ranges that are always refused in strict mode, for publicly routable ranges that are internal-only in a deployment's infrastructure. Denied ranges take precedence over `allowedinternalcidrs`.
- Cloud metadata endpoints are now always blocked per the OWASP cheat sheet. Only Azure's wire server (168.63.129.16) needed an explicit entry; the others (169.254.169.254, Alibaba's 100.100.100.200, AWS's fd00:ec2::254) already fall inside blocked non-public ranges (link-local, CGNAT, ULA). The built-in entries cannot be re-admitted through the allowlist.

PR #4421, opened 2026-07-23
fix(auth): never reflect fetched response bodies in API responses

stevenvegt:
## What

Stop reflecting fetched HTTP response content back to API callers. Handlers that make an outbound request now return a static error description and put the underlying error detail where it is logged, not serialized. A fetched response body that is useful for debugging is debug-logged (clipped) instead of being placed in a returned error.

## Why

Several callback handlers formatted `err.Error()`, or a raw response body, into the OAuth2 `error_description` (or the RFC7807 `detail`) returned to the caller. On their own these are just error messages, but an outbound fetch can be redirected to an internal endpoint, so the reflected content could disclose part of an internal response to the party that triggered the fetch. This removes that reflection as defence in depth, independent of the transport-level SSRF hardening.

## How

- The `OAuth2Error` type already has an `InternalError` field that is `json:"-"` and is only logged, never serialized, and the central error handler already logs the full error. So the fix is to pass the detail as the internal error and keep the description static.
- Static context that is not fetched data, such as the endpoint that failed, is kept, because it helps diagnostics and cannot leak the internal target (in a redirect the stored endpoint is the counterparty's own URL).
- A new `core.ClipHTTPBody` clips a response body for safe debug logging. Where a body is useful to debug (the token client's unparseable response, the parsed credential, the did:web document), it is now debug-logged clipped rather than returned.

## Sites changed

- `auth/api/iam/openid4vci.go` (`handleOpenID4VCICallback`): access token, nonce fetch and retry, credential fetch, credential parse (body debug-logged), and credential verify.
- `auth/api/iam/openid4vp.go` (`handleCallback`): token exchange and token store.
- `auth/client/iam/client.go` (`AccessToken`): the unparseable-response error no longer contains the body; it is debug-logged instead.
- `vdr/didweb/web.go`: the JSON parse error no longer contains the fetched document; it is debug-logged instead.

## Not changed

- The did:web resolve endpoint still returns the resolved document on success, since that is the purpose of the resolver, and it is an internal API.
- The `request-credential` internal endpoint still wraps the metadata-fetch error, which carries endpoint and status metadata rather than a response body, and unwrapping it would break `errors.Is` chains that callers and tests rely on.

## Tests

Existing handler tests were tightened to assert the reflected description no longer contains the fetched content while the underlying cause is still present in the error for logging. The token client test asserts the body is absent from the error and present in a debug log.

Assisted-by: AI


Dirklectisch (approved):
Approved, but I do think there might be one more instance of this problem in the discovery service.

Here we use `problemResponseToError` to format the error and I believe it is sometimes returned to the caller.

https://github.com/nuts-foundation/nuts-node/blob/master/discovery/api/server/client/http.go#L109

stevenvegt:
Good catch! And is not only "sometimes": `problemResponseToError` embedded the remote error response (parsed problem details, or the raw body when it wasn't JSON, up to 1 MB) and that reached our own API responses.

Fixed in 42887733: the client no longer formats the response body into the returned error (only endpoint URL and status code remain), and relies on `TestResponseCodeWithLog` to log the truncated body for diagnostics.

vdr/didweb/web.go
```diff
@@ -106,7 +107,10 @@ func (w Resolver) Resolve(id did.DID, _ *resolver.ResolveMetadata) (*did.Documen
 	var document did.Document
 	err = document.UnmarshalJSON(data)
 	if err != nil {
-		return nil, nil, fmt.Errorf("did:web JSON unmarshal error: %w", err)
+		// Debug-log the (truncated) body for diagnostics, but do not return it: the document
+		// is fetched from an externally-controlled URL and must not be reflected to the caller.
+		log.Logger().Debugf("did:web document at %s could not be parsed (error: %s, body: %q)", targetURL, err, core.TruncateHTTPBody(data))
+		return nil, nil, errors.New("did:web document could not be parsed as JSON")
```
reinkrul, line 113: I think this is bad; we should wrap the error and return it; since it all depends on the caller on how it's used/logged and to whom the error is returned. If it's an internal caller, it's now potentially losing information.

If it's an external caller (meaning API layer), they should determine whether the error is safe to return on not. The practice we normally have is: we don't return errors that bubble up to external clients, unless they're specifially designed to be returned (e.g. OAuthError).

reinkrul: Discussed, we'll keep it this way for now.

PR #4426, opened 2026-07-27
docs: generate server options as list-table for stable diffs

stevenvegt:
## What

Changes the server options documentation generator to emit a `list-table` directive instead of an RST simple table, and adds a `docs-docker` Makefile target.

## Why

The simple-table format pads every cell to the widest value in its column. Adding one option whose key, default, or description becomes the new column maximum reformats every row in the file. In #4420, adding 2 options produced a 142-line diff in `server_options.rst` because the new description was 9 characters longer than the previous longest, adding 9 trailing spaces to all 70 rows.

A list-table has no cross-row alignment: each row is its own 3-line block, so adding an option is always a 3-line diff, regardless of value lengths.

## How

- `docs/rst_table.go`: `printRstTable` emits `* - key` / `- default` / `- description` blocks. Empty cells become a bare `-` (no trailing whitespace anywhere in the output).
- `docs/generate_docs.go`: directive changed to `.. list-table::` with `:header-rows: 1`, keeping the existing `:widths: 20 30 50` and `:class: options-table`. Also removes a stray debug `println` that wrote to stderr on every generation run.
- `docs/generate_docs_test.go`: new test pinning the output format, including bold engine rows, `\:8080` escaping, and empty cells.
- `server_options.rst` and `server_options_didnuts.rst` regenerated (one-time full rewrite).
- New `make docs-docker` target: builds the docs with the existing (previously unreferenced) `docs/Dockerfile`, so no Python/Sphinx tooling is needed on the host.

## Rendering is unchanged

Verified by running the containerized Sphinx build against both the old and the new RST and diffing the output: the rendered tables in `configuration.html` are byte-identical (both directives produce the same docutils table node, with the same caption, column widths, and CSS class). The build shows the same 4 pre-existing warnings in both runs, none related to these files.


reinkrul (approved):
Nice, this annoyed me a lot

PR #4429, opened 2026-07-27
fix(deps): upgrade golang.org/x/text to v0.39.0 to fix CVE

stevenvegt:
## Summary

- Upgrade `golang.org/x/text` to v0.39.0

Fixes the following vulnerability:

| Advisory     | Package | Description       |
| ------------ | ------- | ----------------- |
| GO-2026-5970 | `golang.org/x/text` | norm.Iter could enter an infinite loop on input containing invalid UTF-8 bytes |

GO-2026-5158 (`go.opentelemetry.io/otel`) does not apply to V5.4: this branch has no OpenTelemetry dependency.

Verified with `go build ./... && govulncheck ./...`: 0 vulnerabilities affecting code, the advisory is no longer reported.

No release notes entry added: this is unreleased work on a release branch, and it would only clutter the notes before the next release.

Assisted by AI


reinkrul (approved):

reinkrul (changes requested):
I don't think we need to add security updates to the unreleased section, as they're unreleased (and will probably be replaced by a newer version of the dependency anyway, before we actually release). It'll only clutter the release notes

docs/pages/release_notes.rst
```diff
@@ -3,6 +3,13 @@
 Release notes
 #############
 
+****************
+Unreleased
+****************
+
+## Security
+* Upgrade golang.org/x/text to v0.39.0 to address `GO-2026-5970 <https://pkg.go.dev/vuln/GO-2026-5970>`_ (norm.Iter could enter an infinite loop on input containing invalid UTF-8 bytes).
+
```
reinkrul, line 12: this doesn't belong in v5.4 branch, as there should be no unreleased work on release branches (just release it?). Because, when willw e then release?

reinkrul: Removed the release notes entry — agreed, unreleased security bumps don't belong on a release branch. The dependency upgrade itself is kept.

reinkrul (approved):

Issue #1017, opened 2022-04-21
CircleCI test failure


reinkrul:
Sometimes occurs:
```
time="2022-04-21T14:10:08Z" level=info msg="gRPC server started on :45969" module=Network
panic: implement me

goroutine 284 [running]:
github.com/nuts-foundation/nuts-node/network/transport/grpc.stubServerStream.SendMsg(...)
	/home/circleci/project/network/transport/grpc/connection_manager_test.go:764
github.com/nuts-foundation/nuts-node/network/transport/grpc.(*conn).startSending.func1()
	/home/circleci/project/network/transport/grpc/connection.go:251 +0x1a5
created by github.com/nuts-foundation/nuts-node/network/transport/grpc.(*conn).startSending
	/home/circleci/project/network/transport/grpc/connection.go:244 +0x265
```

See https://app.circleci.com/pipelines/github/nuts-foundation/nuts-node/3478/workflows/e13d9d3c-8964-4e44-9470-4465bb1f722c/jobs/3754

PR #4433, opened 2026-07-27
test(network): re-add Test_conn_startSending with deterministic ordering

stevenvegt:
## Summary

Re-adds `Test_conn_startSending/disconnect_does_not_panic`, which was removed in #2618 because it was flaky. The flake was one over-specified assertion: the test cancelled the stream before calling `disconnect()`, racing the receive goroutine (which stores the stream error as close status) against `disconnect()` cancelling the connection context.

Disconnecting first makes the test deterministic and restores coverage for goroutine exit and the disconnect panic guard (#1017). Companion PRs: same re-add on V6.2, reorder-only fix on V5.4 (which still had the racy version and flaked on #4425).

Verified with `go test -race -count=100`: 100/100 pass.

Assisted-by: AI


stevenvegt:
Wrong head branch (created from the V6.2 fix branch); replaced by a PR from fix/flaky-conn-test-master.

PR #4441, opened 2026-07-30
fix(http): limit inbound request body size to 1MB

stevenvegt:
Adds an inbound request body size limit of 1MB on both the public and internal interfaces, returning 413 (Request Entity Too Large) when exceeded. Until now nothing capped inbound bodies: the only body limits in the repository are outbound (`http/client`), and the deployment documentation's claim that "the Nuts node will also limit the size of request bodies" was false. The documentation now states the actual limit.

The limit is deliberately not configurable. The heaviest legitimate request is an OAuth POST carrying a Verifiable Presentation; a deliberately heavy case (5 JWT-encoded credentials, each with an RSA-4096 `x5c` chain of 4 certificates) computes to roughly 133 kB, and the two-VP `jwt-bearer` flow to roughly 260 kB, so 1MB keeps several times headroom over anything real. It also matches the `client_max_body_size 1M` the deployment documentation already recommends configuring on a reverse proxy, so node and proxy guidance agree.

Implementation: echo's `middleware.BodyLimit` installed through `MultiEcho.Use` with the other global middleware in `Engine.Configure`, after the request logger so rejected requests still produce a request log entry.

Testing: new `TestEngine_RequestBodyLimit` asserts a 512 kB body is accepted (200), a 2 MB body is rejected with 413 on the public interface, and a 2 MB body is rejected with 413 on a route bound to the internal interface. Before the fix, both 2 MB requests returned 200.


reinkrul:
I think there's not much hurry here, since HTTP request bodies are normally already limited by reverse proxies (NGINX has a default of 10mb, for instance). If this is still something we want to fix, it also needs to be backported to v5.4 and v6.4

reinkrul (approved):

PR #4442, opened 2026-07-30
fix(http): serve /status/diagnostics only on the internal interface

stevenvegt:
`/status/diagnostics` is now refused with 403 on any interface other than the internal one. The endpoint discloses the exact software version and git commit (which answers "is this node patched" for version-scoped advisories), the gRPC peer list including remote addresses, node DIDs and authentication state, and DID document/credential store counts. The deployment documentation has always listed it under internal endpoints, but it was actually served publicly as well.

Why middleware instead of binding the route internally: interface binding (`MultiEcho`) resolves on the first path segment only and rejects subpath binds, so `/status` and `/status/diagnostics` cannot be bound to different interfaces through the bind table; the route is registered on every interface that serves `/status`. Teaching the bind table longest-prefix subpath matching is the cleaner end state but changes how every route resolves to an interface, which is a refactor for a minor release, not a security patch. When that refactor lands, this path should become a regular internal-only bind and the middleware deleted. This is documented at the `diagnosticsPath` constant.

How the interface is identified: Go's net/http server stores the connection's local address in the request context under `http.LocalAddrContextKey`. Interfaces always listen on distinct ports (binding the same port twice fails at startup), so comparing the local address port with the configured internal address port identifies the arrival interface. Host is deliberately not compared, since the configured bind host (`:8081`) need not equal the connection's concrete local IP. If public and internal are configured to the same address there is one server and the ports always match, honoring that explicit setup. A missing or unparseable local address refuses the request (fail closed), though net/http always sets it.

Why 403 with a message instead of 404: hiding the endpoint gains nothing, its existence is public knowledge given the open source, and any behavior change equally reveals the patch level. The explicit "only available on the internal interface" message tells an operator whose external monitor probes diagnostics exactly why it broke after a patch upgrade. `/status` itself deliberately stays public as the liveness signal for load balancers and uptime monitors.

Testing: new `TestEngine_DiagnosticsInternalOnly` starts the engine on both interfaces with routes registered as the status engine does, and asserts `/status` returns 200 on the public interface, `/status/diagnostics` returns 200 with its body on the internal interface, and `/status/diagnostics` returns 403 without the diagnostics body on the public interface. Before the fix the public request returned 200 with the full body.

The monitoring documentation now states the endpoint is internal-only and why.


reinkrul (changes requested):
This doesn't solve the issue as this information is also broadcast on the gRPC network (see comment on issue). So we need a plan for this; part of this PR or outside of it, but at least part of the bigger picture.

PR #4455, opened 2026-08-18
ci: skip test and e2e-test jobs for docs-only changes

reinkrul:
## What

`test` and `e2e-test` are required status checks. Both now check whether the PR/push diff touches anything outside `docs/`, `*.md` and `*.rst` (via `dorny/paths-filter`), and skip their actual work when it doesn't.

## Why

All the code checks, especially the unit and e2e tests, are slow and consume resources that aren't required - not very sustainable.

Docs-only changes can't affect Go code or e2e behavior, so there's nothing for either job to catch — but skipping them entirely at the workflow-trigger level (`paths-ignore`) would leave the required check stuck pending forever and block merging, since GitHub doesn't auto-satisfy a required check that never ran. Gating the *steps* instead of the *trigger* keeps the job itself running (so the required check resolves quickly with a pass) while skipping the actual test/build/e2e work.

This also sidesteps spending CI minutes, and occasionally hitting flaky tests (e.g. the embedded-NATS timing flake in `TestNetwork_Reprocess`), on changes that can't have broken anything either job tests.

## Scope

`.github/workflows/go-test.yaml` and `.github/workflows/e2e-tests.yaml` only. This PR itself touches `.github/**`, not `docs/**`, so it exercises the full test/e2e run as normal.

Backports to V6.2 and V5.4 to follow once this is reviewed.

### DIFF
--- a/.github/workflows/e2e-tests.yaml
+++ b/.github/workflows/e2e-tests.yaml
@@ -23,20 +23,41 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v7
+        with:
+          fetch-depth: 0
+
+      # Docs-only changes (docs/, markdown, rst) can't affect the e2e tests. Skipping them keeps
+      # this required check fast and avoids the cost of building/pushing images and running the
+      # full e2e suite for changes that can't have broken anything it tests.
+      - name: Check for non-doc changes
+        id: filter
+        uses: dorny/paths-filter@ceb8a2b8f2d89434be7ff52d3de7ec3738c5cc9d # v4.0.3
+        with:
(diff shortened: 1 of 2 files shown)


stevenvegt (approved):
Agreed with the changes.
I find all the if statements a bit ugly, but that might be the only way?
I saw a "credential persistence through GitHub Actions artifacts" warning that might be something to look into?

reinkrul:
> Agreed with the changes. I find all the if statements a bit ugly, but that might be the only way? I saw a "credential persistence through GitHub Actions artifacts" warning that might be something to look into?

Simplified it

PR #4467, opened 2026-09-02
fix(network): back off when a peer rejects the connection, lower maxbackoff

stevenvegt:
Rescoped: the idle-timeout half moved to #4562, as suggested in review. This PR is now only about the backoff on a rejected connection, and the logging that follows from it.

## Problem

A node kept logging this every few seconds for the same four peers, and never got connected to them:

```
level=warning msg="Peer connection error" error="rpc error: code = Unknown desc = already connected"
```

The peer had a stale stream open for our node and therefore rejected every new connection. Our node treated that rejection as a clean disconnect and reconnected after 1 to 5 seconds, forever. The backoff never kicked in. (#4562 addresses the other side of that incident, the stale stream itself.)

## Changes

**Back off on "already connected".** The rejection is now handled like an authentication failure: exponential backoff instead of a retry every few seconds. Fixing this exposed a race between the receive loop storing the peer's close status and the stream context cancelling the connection, which also affected the existing `Unauthenticated` path. The close status is now read only after the gRPC connection is closed and the receive loops have exited.

**`network.maxbackoff` default lowered from `24h` to `1h`.** The backoff is persisted across restarts and only reset when a peer's NutsComm address changes, so a peer that was unreachable for a few days could go unattempted for up to a day after it came back.

**Less logging, not more.** A lower maximum means more retries, so failed attempts now log at debug level, and a warning is logged once: on the attempt that reaches `network.maxbackoff`. An unreachable peer no longer repeats the same warning on every retry. For a peer that stays down this is 2 lines at warning level per outage instead of 24 a day. The current error and the next attempt stay visible in the contact diagnostics throughout.

## Review comments addressed

- **`ErrAlreadyConnected` matched by string.** The check now lives in one place, `isPeerRejection`, used by both the dialer and the receive loop. `Test_isPeerRejection` pins that the sending side keeps the message bare, so wrapping it anywhere fails the test instead of silently reverting the bug. A dedicated status code would be better still, but an older peer keeps sending the plain message, so the match has to stay for now.
- **`closedByPeer` contradicted the control flow.** Renamed to `closedWithError`, which is what it tests. "Peer closed connection" moved from info to debug: it is regular application flow.
- **`waitForReceivers` could hang indefinitely.** It now waits at most 5 seconds and warns if the receive loops have not exited, so a handler that does not return cannot block the disconnect.
- **`grpcConn.Close()` duplicates the deferred close in `connect()`.** The explicit close has to stay, because the receive loops must be unblocked before waiting for them and the defer runs too late. Commented as intentional and idempotent.
- **`receivers` duplicates `activeGoroutines`.** They track different sets: `activeGoroutines` counts the receive loops *and* the send loops, while `waitForReceivers` must wait only for the receive loops. Both are now commented to say so.
- **`atomic.Value` for `lastReceived`.** Applied in #4562, where that field lives.
- **Close status on a remote shutdown.** A graceful remote close arrives as `io.EOF`, so `closedWithError` is false and the status is not stored. No change needed.

## Tests

- a client rejected with "already connected" calls `Backoff()` and not `Reset()`
- `Test_isPeerRejection`: the message match, a wrapped error not matching, `Unauthenticated`, an unrelated status, and nil
- `Test_recordFailure`: debug while the backoff climbs, one warning on the attempt that reaches the maximum and none after, and the error stored on the contact
- `TestDefaultConfig` asserts the new `maxbackoff` default
- `Test_grpcConnectionManager_openOutboundStreams/ok` waits for the peer gauge decrement instead of asserting it immediately; the added `grpcConn.Close()` plus wait widened a latent race in that assertion

`go test -race ./network/...` passes, apart from `TestNetwork_checkHealth`, which fails locally for environment reasons on master too.


reinkrul (approved):

network/transport/grpc/connection.go
```diff
@@ -217,35 +235,49 @@ func (mc *conn) registerStream(protocol Protocol, stream Stream) bool {
...
+						log.Logger().
+							WithError(err).
+							WithField(core.LogFieldProtocolVersion, protocol.Version()).
+							WithFields(peer.ToFields()).
+							Warn("Peer connection error")
+					} else {
+						log.Logger().
+							WithField(core.LogFieldProtocolVersion, protocol.Version()).
+							WithFields(peer.ToFields()).
+							Info("Peer closed connection")
```
reinkrul: I don't think we should log this on info, as its regular application flow?

Also, the if-statement flow (`if closedByPeer`) doesn't line up with the control flow blocks, as "Peer closed connection" is logged when `closedByPeer == false`, which contradicts

network/transport/grpc/connection.go
```diff
@@ -217,35 +235,49 @@ func (mc *conn) registerStream(protocol Protocol, stream Stream) bool {
...
+					}
+				}
+				if closedByPeer {
+					// Record the peer's close status even if the connection was already cancelled (e.g. because the stream's context is done),
+					// so the caller can decide whether the peer rejected the connection.
+					mc.status.Store(errStatus)
```
reinkrul, line 265: is this also executed when there's no error, but just the remote server shutting down?

network/transport/grpc/connection.go
```diff
@@ -86,20 +87,32 @@ type Connection interface {
...
 	return result
 }
 
 type conn struct {
-	peer             atomic.Value
+	peer atomic.Value
+	// idleTimeout is the period without any received message after which the connection is closed. Zero disables the check.
+	idleTimeout time.Duration
+	// lastReceived holds the time (unix nanoseconds) a message was last received on any of the connection's streams.
+	lastReceived atomic.Int64
```
reinkrul: isn't it more semantic to store time.Time with `atomic.Value[time.Time]` as type?

network/transport/grpc/connection_manager.go
```diff
@@ -462,9 +462,17 @@ func (s *grpcConnectionManager) openOutboundStreams(connection Connection, grpcC
 	// Function must block until streams are closed or disconnect() is called.
 	connection.waitUntilDisconnected()
 
-	if st := connection.closeError(); st != nil && st.Code() == codes.Unauthenticated {
-		// return error so entire connection will be tried anew. Otherwise, backoff isn't honored
-		return st.Err()
+	// Close the gRPC connection so blocked receive loops return, then wait for them:
+	// only then is the close status (as sent by the peer) final.
+	_ = grpcConn.Close()
+	connection.waitForReceivers()
```
reinkrul, line 501: can this hang indefinetly, and should we guard against that?

reinkrul (changes requested):

reinkrul (commented):
A few additional findings from a separate pass, on top of the review already posted.

Assisted by AI

network/transport/grpc/connection_manager.go
```diff
@@ -462,9 +462,17 @@ func (s *grpcConnectionManager) openOutboundStreams(connection Connection, grpcC
...
+
+	if st := connection.closeError(); st != nil {
+		// Peer rejected the connection: return the error so the backoff is honored instead of reconnecting within seconds.
+		// ErrAlreadyConnected arrives as codes.Unknown (plain error returned by the peer's stream handler), so match on the message.
+		if st.Code() == codes.Unauthenticated || st.Message() == ErrAlreadyConnected.Error() {
```
reinkrul: `ErrAlreadyConnected` is detected here by string-matching `st.Message()` against `ErrAlreadyConnected.Error()`, not a status code — the comment above even notes it "arrives as codes.Unknown (plain error returned by the peer's stream handler)". Any future `fmt.Errorf("...: %w", ErrAlreadyConnected)` wrap, or an interceptor that prefixes the message, silently breaks this match and reverts to the pre-PR bug (reconnect every 1-5s instead of backing off), with nothing to catch it. Worth a dedicated gRPC status code/detail instead, or at least a test asserting the match survives a wrapped error.

network/transport/grpc/connection_manager.go
```diff
@@ -462,9 +462,17 @@ func (s *grpcConnectionManager) openOutboundStreams(connection Connection, grpcC
 	// Function must block until streams are closed or disconnect() is called.
 	connection.waitUntilDisconnected()
 
-	if st := connection.closeError(); st != nil && st.Code() == codes.Unauthenticated {
-		// return error so entire connection will be tried anew. Otherwise, backoff isn't honored
-		return st.Err()
+	// Close the gRPC connection so blocked receive loops return, then wait for them:
+	// only then is the close status (as sent by the peer) final.
+	_ = grpcConn.Close()
```
reinkrul, line 500: `grpcConn.Close()` here duplicates the existing `defer grpcClient.Close()` in `connect()` (its only caller, a few lines below where `openOutboundStreams` is invoked). Harmless today since `ClientConn.Close()` is idempotent and both call sites discard the error, but if either site starts checking the returned error it'll surface a confusing "already closing"-type error on every normal outbound disconnect.

network/transport/grpc/connection.go
```diff
@@ -86,20 +87,32 @@ type Connection interface {
...
+	// receivers tracks the receive loops, so callers can wait for the close status to be final.
+	receivers sync.WaitGroup
```
reinkrul, line 109: The new `receivers sync.WaitGroup` tracks the same goroutine lifecycle the existing `activeGoroutines` atomic counter already tracks (both incremented/decremented around the same receive-loop goroutines). Now there are two mechanisms to keep in sync by hand — a future added/removed goroutine is easy to wire into one and forget the other.

network/transport/grpc/connection.go
```diff
@@ -321,3 +348,24 @@ func (mc *conn) IsAuthenticated() bool {
...
+		close(done)
+	}()
+	select {
+	case <-done:
+	case <-time.After(receiverShutdownTimeout):
+		log.Logger().
+			WithFields(peer.ToFields()).
+			Warn("Timeout waiting for the receive loops to exit, the peer's close status may be incomplete")
+	}
+}
```
reinkrul, line 371: Driveby fix, but ok.

reinkrul (approved):

Issue #4469, opened 2026-09-03
OpenID4VCI: make server-to-server credential-offer delivery reliable (retry, DLQ, base-URL setup)


reinkrul:
Related: #2825, #2318

## Problem Statement

Private credentials are delivered from a credential issuer's node to a credential receiver's node two ways: primarily via an OpenID4VCI server-to-server credential offer (a synchronous HTTP push at issuance time, pre-authorized_code grant — `vcr/issuer/openid.go`), falling back to publishing the encrypted transaction on the Nuts gRPC network for the receiver to fetch (`vcr/issuer/network_publisher.go`, `network/transport/v2/protocol.go`) whenever the OpenID4VCI push fails or isn't supported by the receiver.

Both paths have reliability gaps that, together, can silently drop a credential with no trace above Trace-level logging:

1. **No retry for the OpenID4VCI push.** `Issue()` (`vcr/issuer/issuer.go:137-203`) makes exactly one synchronous HTTP attempt (`issueUsingOpenID4VCI`, line 180) and falls straight through to the gRPC/DAG fallback on any failure. There's no queue, no backoff, no persistence of the failed attempt — unlike the private-payload-fetch mechanism on the receiving side, which already has a `go-retry`-based notifier with a `payload_fetch_dlq` diagnostic (`network/dag/notifier.go`, `network/transport/v2/protocol.go:332-345`).

2. **The issuer's own base-URL identifier resolution can get permanently stuck empty.** `tlsIdentifierResolver` (`vcr/openid4vci/identifiers.go`) caches whatever it resolves — including an empty string — the first time it runs, and never invalidates it before process restart: the cache read (`identifiers.go:102-106`) only checks for a non-nil pointer, not a non-empty value, and the cache write (`identifiers.go:116-125`) stores on `err == nil` regardless of whether the identifier is non-empty. One bad resolution (missing `node-http-services-baseurl` service at the time, or a transient failure of the TLS-certificate self-discovery fallback) permanently breaks every subsequent OpenID4VCI offer from that issuer DID for the life of the process. Fixing (1) alone wouldn't help — every retry would hit the same poisoned cache.

   Observed on the **sender** side (issuer), for a `credential_offer` sent with `credential_issuer:""`:
   ```
[...]

PR #4470, opened 2026-09-04
fix(vcr): bound negative OpenID4VCI identifier caching, guard against empty offers

reinkrul:
Related: #4469

## Problem

`tlsIdentifierResolver` (`vcr/openid4vci/identifiers.go`) cached its resolution result as soon as it completed without erroring — including an empty string when neither the DID document nor TLS-certificate-derived candidates yielded a base URL. Once that happened, every later `Resolve()` call for that DID short-circuited on the cached empty value for the life of the process, even after the missing `node-http-services-baseurl` service was added, permanently breaking OpenID4VCI credential-offer delivery until restart.

Observed on the **sender** side (issuer), for a `credential_offer` sent with `credential_issuer:""`:
```
level=warning msg="Couldn't publish credential over OpenID4VCI, fallback to publish over Nuts network" credentialID="did:nuts:ISSUER_DID_REDACTED#REDACTED-CREDENTIAL-UUID" error="unable to offer the credential over OpenID4VCI to (wallet: https://receiver.example.com/n2n/identity/did:nuts:RECEIVER_DID_REDACTED): unable to offer credential (client-metadata-url=https://receiver.example.com/n2n/identity/did:nuts:RECEIVER_DID_REDACTED/openid4vci/credential_offer): offer credential error: unexpected http response code (...): 500" module=VCR
```
Observed on the **receiver** side, rejecting that same offer:
```
time="2026-09-03T08:16:47Z" level=error msg="HandleCredentialOffer failed" error="server_error - unable to create issuer client: empty Credential Issuer Identifier" module=VCR/OpenID4VCI operation=HandleCredentialOffer operationID=HandleCredentialOffer requestURI="/n2n/identity/did:nuts:RECEIVER_DID_REDACTED/openid4vci/credential_offer?credential_offer=..." user="<nil>"
```

## Fix

**Bound the negative cache instead of removing it.** An earlier version of this fix stopped caching empty results at all, but this resolver can be called on every OpenID4VCI request, so re-running resolution (including the DID document lookup) on every single call for a DID that never gets fixed isn't free — the original negative caching was likely deliberate. Instead, reuse the existing `lastAttempt`/`tlsAttemptInterval` throttle (already used to rate-limit the expensive TLS-certificate-derived resolution) to also bound how long an empty result is treated as cached, rather than adding a separate cache window. A successful (non-empty) identifier is still cached indefinitely, unchanged.

**Guard against sending a broken offer.** Nothing previously stopped an OpenID4VCI credential offer from being sent with an empty `credential_issuer` — which is what produced the receiver-side "empty Credential Issuer Identifier" rejection shown above. Introduced `openid4vci.ErrIdentifierNotConfigured`, returned when resolution succeeds but yields no identifier, and `issueUsingOpenID4VCI` now treats it like an unsupported wallet (fallback to the network) instead of constructing and sending the broken offer.

**Warn the operator, unlike the unsupported-wallet case.** An unsupported wallet is the other party's problem — nothing this operator can act on, so it stays silent. A missing `node-http-services-baseurl` service is *this* node's own misconfiguration, so it now logs a `Warn` pointing the operator at what to search the documentation for, instead of failing the same silent way.

**Return a 4xx, not a 500, for inbound requests hitting an unconfigured DID.** `GetOpenIDIssuer`/`GetOpenIDHolder` previously returned the bare `ErrIdentifierNotConfigured` for this case. `protocolErrorWriter` didn't recognize that as a protocol error, so an inbound OpenID4VCI request against such a DID got a `500 server_error`, where master returned a `200` with broken metadata URLs. Wrapped it in an `openid4vci.Error` (`invalid_request`, 404), like the not-owned case, and gave `openid4vci.Error` an `Unwrap()` so the `errors.Is` check in the issuer's fallback-to-network path still matches through the wrapping.

## Test plan

- [x] `TestTLSIdentifierResolver/empty_result_is_cached_briefly,_not_forever` (new) — asserts a cache hit on an immediate second call, then a re-check once the throttle window elapses.
- [x] `Test_vcr_GetOIDCIssuer/found_DID,_owned,_but_no_identifier_configured` (new) — asserts `ErrIdentifierNotConfigured` when resolution yields an empty identifier.
- [x] `TestWrapper_GetOpenID4VCIIssuerMetadata/no_identifier_configured` (new) — asserts the inbound API returns a 404 protocol error, not a 500, and that `errors.Is` still matches `ErrIdentifierNotConfigured` through the wrapping.
- [x] `Test_issuer_Issue/OpenID4VCI/ok_-_OpenID4VCI_issuer_identifier_not_(yet)_configured_-_fallback_to_network` (new) — asserts a quiet fallback to network publish, no offer sent, and the new Warn log.
- [x] Existing `vcr`, `vcr/issuer`, `vcr/openid4vci` suites still pass.

Assisted by AI


vcr/vcr.go
```diff
@@ -141,6 +141,9 @@ func (c *vcr) resolveOpenID4VCIIdentifier(ctx context.Context, id did.DID) (stri
 			StatusCode: http.StatusNotFound,
 		}
 	}
+	if identifier == "" {
+		return "", openid4vci.ErrIdentifierNotConfigured
```
stevenvegt: This error is not known to the API layer, so inbound OpenID4VCI requests for such a DID now get a 500 server_error (see protocolErrorWriter in
  vcr/api/openid4vci/v0/api.go), where master returned a 200 with broken metadata URLs. Better to wrap it in an openid4vci.Error with a 404 like the
  not-owned case above, so the caller sees a 4xx and doesn't retry. Keep ErrIdentifierNotConfigured as the Err so the errors.Is check in the issuer still
  works. Either way, worth a line in the PR description since it changes inbound behavior.

stevenvegt (changes requested):

stevenvegt (approved):

PR #4471, opened 2026-09-04
feat(vcr): retry + DLQ for the OpenID4VCI credential-offer push

reinkrul:
Related: #4469

## Problem

When pushing a credential offer to a wallet/issuer over OpenID4VCI fails, the node immediately falls back to the synchronous Nuts network (DAG) publish path. That fallback publishes the credential publicly-encrypted-for-participants over gRPC, which has its own delivery assumptions (e.g. PAL decryption via an exportable private key) that don't hold for every deployment — see #4469. A transient OpenID4VCI failure (wallet momentarily down, network blip) shouldn't immediately trigger that fallback.

## Solution

Add a persistent, retrying offer queue (`vcr/issuer/offer_queue.go`):

- On a genuine OpenID4VCI delivery failure, `Issue()` schedules the offer on the queue and returns immediately, instead of synchronously falling back to the network publisher.
- The queue retries with exponential backoff (`avast/retry-go`) for up to 24h (`offerRetryWindow`), persisted in a stoabs shelf so retries survive a node restart (`Run()` resumes pending jobs on startup).
- If the retry window is exhausted, or the attempt returns an unrecoverable error (e.g. the wallet no longer supports OpenID4VCI), the offer is dead-lettered (`GivenUp: true`) and only then does the node fall back to the network publisher (`giveUpOffer`).
- If no network publisher is configured, delivery failure after give-up is logged as an error (no silent drop).
- `GetFailedOffers()` exposes dead-lettered jobs for future diagnostics/requeue tooling (#4469 item 3).

`Issuer` gained `Start()`/`Shutdown()` to run/stop the queue's background goroutines; wired into `vcr.go`'s module lifecycle. Requires a persistent KV store (`openid4vci-offer-queue`), only allocated when OpenID4VCI is enabled.

Does not depend on #4470 — verified standalone (cherry-picked onto `master`, builds and tests pass in isolation).

## E2E test

`e2e-tests/openid4vci/offer-retry`: node B's OpenID4VCI endpoint is taken down before an authorization credential is issued from node A, confirming the offer is accepted (queued) rather than blocking; node B is then brought back up and the test polls until the credential is delivered via retry.

## Test plan

- [x] `go test ./vcr/...` (including `-race`, `-count=15` on the previously-flaky give-up subtest)
- [x] `go build ./...`
- [x] e2e test run locally (`e2e-tests/openid4vci/offer-retry/run-test.sh`, exit 0)

Assisted by AI


vcr/issuer/issuer.go
```diff
@@ -179,11 +204,21 @@ func (i issuer) Issue(ctx context.Context, template vc.VerifiableCredential, opt
 		if i.openidHandlerFn != nil && !options.Public {
 			success, err := i.issueUsingOpenID4VCI(ctx, *createdVC)
 			if err != nil {
```
stevenvegt, line 206: This start retrying even on a non related errors like an invalid subject DID?

Feels like we should add more detail to the error reasons and try to be smart on when to retry and not.

vcr/issuer/issuer.go
stevenvegt, line 245: should not retry

stevenvegt, line 262: really depends on the error if we should retry. 404 vs 5xx vs timeout etc. 
Some are unrecoverable and we should inform the user right away instead of letting them wait for 24 hours to come to the inevitable conclusion.

vcr/issuer/issuer.go
```diff
@@ -229,6 +264,40 @@ func (i issuer) issueUsingOpenID4VCI(ctx context.Context, credential vc.Verifiab
...
+	}
+	return nil
+}
+
+// giveUpOffer is called once an offer's retry window has been exhausted: it's the deferred equivalent of
+// the immediate fallback Issue() performs when no retry queue is configured.
+func (i issuer) giveUpOffer(ctx context.Context, credential vc.VerifiableCredential) {
+	log.Logger().
+		WithField(core.LogFieldCredentialID, credential.ID.String()).
+		Warn("Giving up on delivering credential over OpenID4VCI, falling back to publish over the Nuts network")
```
stevenvegt, line 286: weird to say you go use the nuts-publisher while you only check in the next step if that is available. Better guard this log by that check.

vcr/issuer/issuer.go
```diff
@@ -229,6 +264,40 @@ func (i issuer) issueUsingOpenID4VCI(ctx context.Context, credential vc.Verifiab
...
+	if i.networkPublisher == nil {
+		log.Logger().
+			WithField(core.LogFieldCredentialID, credential.ID.String()).
+			Error("No Nuts network publisher configured either; credential delivery has permanently failed")
```
stevenvegt, line 290: I wonder, is this a log level warn or error? This is expected behavior, not really an error right?

stevenvegt (changes requested):

vcr/issuer/issuer.go
stevenvegt, line 197: so what happens to a failed VC, it remains in the store right?

vcr/issuer/issuer.go
```diff
@@ -179,11 +204,21 @@ func (i issuer) Issue(ctx context.Context, template vc.VerifiableCredential, opt
...
+				// falling back immediately: publishing to the Nuts network is irreversible and
+				// network-wide replicated, so it shouldn't pay that cost for a failure that a retry
+				// might resolve. If no retry queue is configured (offerQueueStore was nil), fall back
+				// immediately instead, same as before.
 				log.Logger().
 					WithField(core.LogFieldCredentialID, createdVC.ID.String()).
 					WithError(err).
-					Warnf("Couldn't publish credential over OpenID4VCI, fallback to publish over Nuts network")
+					Warnf("Couldn't publish credential over OpenID4VCI, will retry in the background")
+				if i.offerQueue != nil {
```
stevenvegt, line 216: This does a retry per VC, but we can also do it per wallet. If we know a wallet is down, we can fail directly. Also, if you issue a lot to a wallet, this would bundle them instead of all trying in their own time.

reinkrul:
We're taking a different approach: detect undeliverable network issuance due to missing decryption key, make sure we return an error when credential can't be delivered.

Redesign this when moving away from gRPC network (see https://github.com/nuts-foundation/nuts-node/issues/4515)

PR #4475, opened 2026-09-07
docs(deployment): document node-http-services-baseurl as a required setup step

reinkrul:
Related: #4469

## Problem

`did:nuts` DIDs have historically relied on the gRPC/Nuts network to exchange credentials. We want to move away from that in favor of server-to-server issuance over HTTP instead, but nothing documents `node-http-services-baseurl` - the DID service required for a node to be reachable that way - as a setup step at all.

## Solution

Adds a "Registering `node-http-services-baseurl`" section to `docs/pages/deployment/configuration.rst`, right after the existing did:nuts/gRPC options table:

- Frames the mechanism as "server-to-server issuance" (currently OpenID4VCI, but that's an implementation detail that may change) rather than centering on OpenID4VCI by name, and the gRPC/Nuts network as the fallback we're moving away from.
- Lists the concrete `/n2n` endpoints involved, and notes the whole `/n2n` prefix must be proxied through, not just these specific paths.
- Explains that the service is normally registered automatically by GoldenHammer, and precisely how: it probes each hostname in the node's own TLS certificate (not `config.URL` - verified GoldenHammer predates that, from v5.4) with a HEAD request to the credential-issuer-metadata well-known path, registering the first one that responds. Points at the usual causes (DNS/firewall/reverse-proxy) when it's missing.
- Shows manual registration via the DIDMan API, for both a vendor DID (concrete URL) and a subject DID referencing a vendor's `NutsComm` service (matching reference, not a duplicate URL).

Also cross-references the new section from the `/n2n` endpoint description in `docs/pages/deployment/recommended-deployment.rst`'s "Legacy Endpoints" section, since server-to-server issuance runs over that same endpoint.

Verified with a clean sphinx build using `docs/Dockerfile` - 0 new warnings; the 4 pre-existing ones (`release_notes.rst`, `discovery.rst`) are unrelated.

## Test plan

- [x] `sphinx-build -M html . _build -W` via `docs/Dockerfile`, clean

Assisted by AI

### DIFF
--- a/docs/pages/deployment/configuration.rst
+++ b/docs/pages/deployment/configuration.rst
@@ -65,6 +65,38 @@ the options table must be regenerated using the Makefile:
 
     $ make docs
 
+.. _node-http-services-baseurl:
+
+Registering ``node-http-services-baseurl``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+``did:nuts`` DIDs receive private Verifiable Credentials over OpenID4VCI: a synchronous HTTP push, from the issuer's node directly to the holder's node, at issuance time.
+For a node to be reachable this way (as issuer or holder), its ``did:nuts`` DID document needs a ``node-http-services-baseurl`` service, pointing other nodes at the public base URL where its OpenID4VCI (and other ``/n2n``) HTTP endpoints can be reached.
+Without it, the node can't be discovered over OpenID4VCI: offers to it fail, and offers from it fall back to the deprecated gRPC/Nuts network, or fail outright if that's disabled too.
+
+This service is **not** registered automatically.
+GoldenHammer (a background module that periodically fixes a handful of common DID document issues) will add it if it detects it's missing, but GoldenHammer is scheduled for removal once OpenID4VCI becomes the only way to exchange credentials
+(see `issue #2318 <https://github.com/nuts-foundation/nuts-node/issues/2318>`_) and, being a periodic best-effort fix, can't guarantee the service is present the moment it's actually needed.
+Register it explicitly as part of node setup, the same way the required ``NutsComm`` service is registered, using the DIDMan API's endpoint:
+
+.. code-block:: shell
+
+    $ curl -X POST https://internal.example.com/internal/didman/v1/did/<did>/endpoint \
+        -H "Content-Type: application/json" \
+        -d '{"type": "node-http-services-baseurl", "endpoint": "https://your-node.example.com"}'
+
+For a vendor DID document (the one whose ``NutsComm`` service is a concrete URL, e.g. ``grpc://host:5555``), register the node's actual public base URL as the ``endpoint`` value, as above.
+
+Any care-organization/subject DID document whose ``NutsComm`` service is instead a *reference* to that vendor DID should get a matching reference for ``node-http-services-baseurl``, rather than a duplicate concrete URL, so all subject DIDs of the same vendor stay in sync with a single registration:
+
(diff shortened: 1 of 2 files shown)


docs/pages/deployment/configuration.rst
```diff
@@ -98,4 +98,43 @@ and the ``internalratelimiter`` is always on.
...
\ No newline at end of file
+This applies to every outbound HTTP call made via the shared HTTP client (OpenID4VCI, OAuth relying-party, IAM, Discovery, did:web resolution, etc.) and to every redirect target along the way.
+
+.. _node-http-services-baseurl:
+
+Registering ``node-http-services-baseurl``
+********************************************
+
+``did:nuts`` DIDs receive private Verifiable Credentials via server-to-server issuance: the issuer's node pushes the credential directly to the holder's node over HTTP, at issuance time.
+If a node can't be reached for server-to-server issuance, delivery falls back to the older gRPC/Nuts network instead, which publishes credentials (encrypted) for the holder to fetch rather than pushing them directly - a mechanism we're moving away from in favor of server-to-server issuance.
```
stevenvegt: >which publishes credentials (encrypted)

which publishes an encrypted reference to a credential

docs/pages/deployment/configuration.rst
```diff
@@ -98,4 +98,43 @@ and the ``internalratelimiter`` is always on.
...
+For a node to be reachable for server-to-server issuance (as issuer or holder), its ``did:nuts`` DID document needs a ``node-http-services-baseurl`` service, pointing other nodes at the public base URL where its HTTP services (including server-to-server issuance, and other ``/n2n`` endpoints) can be reached.
```
stevenvegt: `service` could be ambiguous, since the DID document term might not be know to all and could read like an actual service you need.

docs/pages/deployment/configuration.rst
```diff
@@ -98,4 +98,43 @@ and the ``internalratelimiter`` is always on.
...
+Given ``<base-url>`` and a DID, the endpoints it needs to be reachable at are:
+
+- ``<base-url>/n2n/identity/<did>/.well-known/openid-credential-issuer`` - credential issuer metadata
+- ``<base-url>/n2n/identity/<did>/.well-known/oauth-authorization-server`` - OAuth/provider metadata
+- ``<base-url>/n2n/identity/<did>/openid4vci/credential`` - credential endpoint (issuer side)
+- ``<base-url>/n2n/identity/<did>/openid4vci/credential_offer`` - credential offer endpoint (wallet/holder side)
+
+This isn't an exhaustive list to configure individually: any reverse proxy or firewall in front of the node must forward the entire ``/n2n`` path prefix to it, not just these specific paths.
+
+This service is normally registered automatically (by a background module, GoldenHammer), as long as the node can reach itself: for each hostname in its own TLS certificate, it does a ``HEAD`` request to ``https://<hostname>/n2n/identity/<did>/.well-known/openid-credential-issuer`` and registers the first hostname that responds with ``200 OK`` and ``Content-Type: application/json``.
```
stevenvegt: "This service is normally registered automatically" -> The DID Document service entry is added automatically when possible to ease configuration, but this process can fail when node is unable to reach itself:...

stevenvegt (approved):
Small changes you can make without requesting a new review.

docs/pages/deployment/configuration.rst
reinkrul: Actually, both not true; the transaction's receiver metadata is encrypted, the rest is not. The TX payload is ACL'd. I'll make that clear in the docs.

Issue #4485, opened 2026-09-09
Add /v5 and /v6 major version suffix to the module path so advisories and scanners can match releases


stevenvegt:
## Problem

The module path is `github.com/nuts-foundation/nuts-node` while releases are tagged v5.x and v6.x. Go module rules say a module with a go.mod file must carry the major version in its path from v2 on, so the Go module proxy (proxy.golang.org, the source of truth for which versions of a module exist) rejects every v5 and v6 tag as a version of this module. The highest version the ecosystem knows is v1.1.0.

This breaks security advisories, not just builds. A scanner match is "installed version falls inside the advisory's vulnerable range", and that range is stored as boundaries: an introduced version and a fixed version. GitHub and the Go vulnerability database only store boundaries that the proxy confirms are real versions of the module. Our advisory GHSA-9hmg-827w-9rhj (CVE-2026-41164) was filed as `<= v6.2.2, fixed in v6.2.3` and `<= v5.4.30, fixed in v5.4.31`. Since none of those are versions of the module, it is stored as:

- GitHub global advisory (used by Trivy, Grype, Docker Scout): `<= 1.1.0`, no fix. The range was clamped to the newest version the proxy knows.
- Go vulnerability database GO-2026-5291 (used by govulncheck, OSV): all versions from 0, no fix.

Consequences, verified with Trivy and Syft against the 6.2.10 image and a locally built binary:

- Trivy and Grype never match a v6 image at any version, because 6.x is above the stored `<= 1.1.0` range. A customer running 6.1.0 gets a clean scan. This is a false negative on a real CVE.
- govulncheck and OSV-based tools flag every version, including fixed ones, and no release can make that go away.
- Downstream repos (nuts-knooppunt, nuts-node-private, data-viewer, and others) can only depend on pseudo-versions such as `v1.0.1-0.20260713064648-0b7d3a55aed2`, so Dependabot and govulncheck cannot reason about their nuts-node dependency either.

The pending pipeline changes (ldflags version key, `-buildvcs=false`, tag-derived version) make the binary report the correct version to scanners. They cannot fix the advisory side: the databases must be able to store our fix versions as boundaries, and with the current module path they cannot. Build-time import rewriting, attached SBOMs and image labels have the same limitation, because they change what the binary says, not what the databases can store.

## Proposal

[...]

PR #4486, opened 2026-09-09
build: make the main module version readable by image scanners

stevenvegt:
Docker image scanners could not read the version of the nuts-node main module from the image. Trivy showed no version at all, because it only reads `-X` ldflag keys ending in `.version` and ours was `core.GitVersion`. The GitHub release binaries carry `v0.0.0-<time>-<rev>`, stamped by Go's VCS support, because the module path without `/v6` makes the v6.x tags invalid module versions.

Changes:

- Rename `core.GitVersion` to `core.Version`, so the ldflag key ends in `.version`. That name was taken by the function `core.Version()`, and a Go package cannot have a variable and a function with the same name, so the function is renamed to `core.VersionOrBranch()`. Its behaviour is unchanged: it returns the tagged version if set, otherwise the branch. Tests updated.
- Build with `-buildvcs=false` in the Dockerfile, makefile and binaries workflow, so no pseudo-version can be stamped. The ldflag is the intended source of the version.
- Derive the version in the images workflow from `GITHUB_REF_TYPE` and `GITHUB_REF_NAME` instead of `git name-rev`, so tag builds get a clean `vX.Y.Z` and everything else gets `undefined`.

Verified on a binary built with the new flags: Trivy and Syft both report `pkg:golang/github.com/nuts-foundation/nuts-node@v6.2.10`. Before the change Trivy reported no version for the published 6.2.10 image.

This does not make advisories match yet. Both GitHub and the Go vulnerability database store our advisory ranges as "up to 1.1.0" or "all versions", because the module path has no major version suffix. See #4485 for that. 

Related: #4451 attaches the SBOM to the image; this PR fixes what that SBOM says about the nuts-node module. The two are independent and touch neighbouring lines of `build-images.yaml`, so whichever merges second needs a trivial rebase.


stevenvegt:
Converted to draft. #4487 (module path rename on V6.2) supersedes this PR: once the module path carries the /v6 suffix, go build stamps the version from the tag itself and scanners read it from the build info, so the ldflags key and -buildvcs=false here are no longer needed. The -buildvcs=false flag would even block that stamping. If #4487 is approved and ported to master, close this PR. If #4487 is rejected, this PR is the fallback and can be marked ready again.

PR #4490, opened 2026-09-09
build: add /v6 major version suffix to the module path

### LINKED ISSUE #4485: Add /v5 and /v6 major version suffix to the module path so advisories and scanners can match releases

stevenvegt:
Implements #4485 for the master branch: the module path becomes `github.com/nuts-foundation/nuts-node/v6`, which makes the v6.x tags valid module versions. That is what lets security advisories name our fix versions and lets scanners match images against them. See the issue for the reasoning.

This PR is the rename only:

- `gomajor path -version v6` rewrote go.mod and every import.
- The existing ldflags `-X` keys in the build files and the `go_package` options in the two `.proto` files carry the new path. gomajor does not touch those.

Nothing else changes. How the version reaches the binary (git stamping in the Docker build, `.dockerignore`, checkout depth, makefile and workflow adjustments) is a separate concern and follows in its own PR, per review feedback on #4490. The branch history shows those commits and their removal; the squash merge keeps only the rename.

Verified locally: the repo builds, `go mod tidy` changes nothing beyond the module line, and no Go file imports the un-suffixed path.

Follow-ups: downstream importers must switch to the `/v6` import path to receive newer versions; after the first tag on this branch, the repository advisory can name the `/v6` package with that tag as the patched version.


.github/workflows/build-images.yaml
```diff
@@ -29,6 +29,10 @@ jobs:
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
         with:
           persist-credentials: false
+          # Full history so go build can derive the version from the nearest
```
reinkrul: won't this be a lot slower and wasteful CPU/network traffic? Can't we just fetch what we need?

.dockerignore
```diff
@@ -1,15 +1,19 @@
...
 data
-e2e-tests
+**/data
+docs/_build
+*.pem
+*.key
+*.db
+nuts
+nuts-node
+nuts.yaml
```
reinkrul: It was so because we optimized for build speed and Docker layer cache, do you realize you're changing that trade-off?

Why is "get tag from Git instead of env params "even part of this PR? Feels like scope creep

reinkrul (changes requested):

reinkrul (approved):

PR #4508, opened 2026-09-09
build: harden docker image, fix latest tag, scan published images

stevenvegt:
Follow-up to a Trivy scan of the published `v5.4.38` and `6.2.11` images (49 and 54 fixable Alpine findings plus Go 1.26.5 stdlib CVEs). Companion PRs on the release branches: #4506 (V5.4) and #4507 (V6.2). Four independent commits:

**Dockerfile**: builder to `golang:1.26.8-alpine`, runtime stage runs `apk -U upgrade --no-cache` before installing tzdata and curl. Alpine point-release images lag the repos (3.24.1 ships older openssl than the 3.24 repo), so the tag bump alone never clears a scan.

**`latest` tag**: `actions-ecosystem/action-get-latest-tag` sorts tags by creation date, not version, so whichever release line is tagged last wins `latest`. v5.4.38 was tagged 50 minutes after v6.2.11 and `latest` on Docker Hub still points at the 5.4 image. Replaced with `git tag --sort=-v:refname`; checkout now fetches tags. The same change needs to land on V5.4 and V6.2 since the tag build runs the workflow from the tagged branch.

**Dependabot**: added `target-branch` entries for V6.2 and V5.4 (docker and gomod). Dependabot only reads this file from master and only targets master without them; the release branches never received bump PRs. Go module updates on release branches are limited to patch versions.

**Scheduled image scan**: new `image-scan-cron-schedule.yaml` runs Trivy against the published `master` image and the highest release tag of v5.4 and v6.2, fixable HIGH/CRITICAL only, and posts to the same Slack webhook as the govulncheck cron. Nothing is built; it scans Docker Hub directly. Hub tag naming differs per line (`v5.4.38` vs `6.2.11`), the resolve step tries both forms.

Verification: workflow YAML parses; the tag resolution and highest-tag logic were run locally against the current tag list (v5.4 -> `v5.4.38`, v6.2 -> `6.2.11`, highest -> `v6.2.11`). The scheduled scan will fail until the release branches are rebuilt with the companion PRs and the Go module bumps; that is the intended signal.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -59,3 +59,75 @@ updates:
       interval: "weekly"
     # Allow up to 0 open pull requests for pip dependencies
     open-pull-requests-limit: 0
+
+  # Release branches: Dependabot only reads this file from the default branch
+  # and only targets that branch unless told otherwise. Keep the base image
+  # and Go modules on the supported release lines patched as well.
+  # Only patch-level Go module updates are proposed there.
+  - package-ecosystem: "docker"
+    directory: "/"
+    target-branch: "V6.2"
+    schedule:
+      interval: "weekly"
+      day: "sunday"
+      time: "23:00"
+      timezone: "Europe/Amsterdam"
+    groups:
+      docker:
+        patterns:
+          - "*"
+        update-types:
+          - "minor"
+          - "patch"
+
+  - package-ecosystem: "gomod"
+    directory: "/"
+    target-branch: "V6.2"
+    schedule:
+      interval: "weekly"
+      day: "sunday"
+      time: "23:00"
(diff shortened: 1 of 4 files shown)


reinkrul (commented):
Do we need both govulncheck and Trivy?

stevenvegt:
> Do we need both govulncheck and Trivy?

Trivy scans our docker image, including installed binaries and the base image. This PR adds that check on the currently published images of master, v5.4 and v6.2. govulncheck checks our codebase for new vulnerabilities. Trivy won't find problems in unpublished code, govulncheck won't find a stale openssl in the latest patch release.

reinkrul (approved):

PR #4510, opened 2026-09-09
fix(deps): upgrade go to 1.26.8 and vulnerable modules

stevenvegt:
## Summary

- Go directive to 1.26.8 (the Dockerfile builder bump is in #4507; the go-test and govulncheck workflows read the version from go.mod, so this is what makes the scheduled govulncheck green again)
- Upgrade golang.org/x/crypto v0.53.0 to v0.55.0, golang.org/x/net v0.55.0 to v0.58.0 (required by x/crypto and grpc), google.golang.org/grpc v1.82.1 to v1.83.2, github.com/labstack/echo/v4 v4.15.0 to v4.15.3 and github.com/go-chi/chi/v5 v5.2.2 to v5.2.4
- Release notes under Unreleased

Fixes the following vulnerabilities:

| Advisory | Package | Description |
| --- | --- | --- |
| GO-2026-6218 | `net/url` | Go stdlib, fixed in 1.26.6 |
| GO-2026-6091 | `html/template` | Go stdlib, fixed in 1.26.6 |
| GO-2026-6090 | `crypto/tls` | Go stdlib, fixed in 1.26.6 |
| GO-2026-6089, GO-2026-5026 | `net/http` | Go stdlib, fixed in 1.26.6 |
| GO-2026-6088 | `encoding/xml` | Go stdlib, fixed in 1.26.6 |
| GO-2026-5972 | `encoding/asn1` | Go stdlib, fixed in 1.26.6 |
| CVE-2026-56854 | `golang.org/x/crypto` | fixed in v0.55.0 |
| CVE-2026-46600 | `golang.org/x/net` | fixed in v0.56.0 |
| CVE-2026-84303, CVE-2026-84304, CVE-2026-84445 | `google.golang.org/grpc` | fixed in v1.83.2 |
| CVE-2026-55677 | `github.com/labstack/echo/v4` | fixed in v4.15.3 |
| CVE-2025-69725 | `github.com/go-chi/chi/v5` | fixed in v5.2.4 |

The test job on V6.2 stays red until #4494 merges: it carries the fix for the stale `nuts-node/cmd` import in the docs test that has broken go-test on the branch since #4487.

Verification: `go build ./...` ok; `govulncheck ./...` reports 0 reachable vulnerabilities (was 7 stdlib). The module CVEs were never reachable according to govulncheck; they are fixed to clear image scanner reports. `go test ./...` passes except TestNetwork_checkHealth, which fails on this machine regardless of branch (local network environment).

[...]

PR #4511, opened 2026-09-09
fix(docs): import cmd package through the /v6 module path

stevenvegt:
The docs up-to-date test backported in #4493 still imports `github.com/nuts-foundation/nuts-node/cmd` after #4487 moved the module path to `/v6`. The go-test workflow on V6.2 has failed on every push since that merge (`no required module provides package github.com/nuts-foundation/nuts-node/cmd`), and `go mod tidy` on the branch pulls in `nuts-node v1.1.0` as a dependency.

Verification: `go vet ./docs/...` compiles the test file again. Split out of #4510 so it can merge first and unblock #4507 and #4510.

Assisted-by: AI

### DIFF
--- a/docs/generate_docs_uptodate_test.go
+++ b/docs/generate_docs_uptodate_test.go
@@ -23,7 +23,7 @@ import (
 	"path/filepath"
 	"testing"
 
-	"github.com/nuts-foundation/nuts-node/cmd"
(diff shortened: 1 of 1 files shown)


stevenvegt:
Superseded: #4494 already contains this import fix as part of the docs backport.

Issue #4515, opened 2026-09-10
DRAFT: OpenID4VCI delivery: classify errors, let the caller choose delivery mode, make delivery status visible


stevenvegt:
**Status: first draft.** Written up from the review of #4471 to have something concrete to discuss. Nothing in here is decided. Expect it to change after discussion.

Related: #4469, #4470, #4471

## Context

#4469 decided to add a retry queue with a dead-letter queue (DLQ) for the OpenID4VCI credential offer, gate the gRPC fallback behind a 24 hour retry window, and backport that as a patch to v5.4 and v6.2. #4471 implements the queue. Reviewing #4471 raised questions that the retry queue alone does not answer. This issue writes down those questions and a proposal, so we can decide before #4471 merges rather than fix it afterwards.

The short version: the queue retries every failure the same way, the caller cannot see that a credential was queued, and the default behaviour changes for every existing integration. The care workflow case makes the second point urgent. When a call center grants night care access and the credential silently enters a retry queue, nobody at the desk knows that access is not going to work.

## Problems with #4471 as it stands

- Every error is retried for 24 hours: a credential without a valid subject DID, a local store failure, a wallet that returns HTTP 400, and a wallet that is restarting. Some can never succeed. Others need a person to fix configuration, and nobody is told.
- The HTTP status code is lost. `httpDo` in `vcr/openid4vci/issuer_client.go` turns a non-2xx response into a plain string error. No caller can distinguish 404 from 503.
- Queueing is invisible to the caller. `Issue()` returns the credential with a nil error whether it was delivered or queued. The API returns 200 either way.
- The default changes for every existing integration. Today a misconfigured counterpart still gets the credential immediately over the gRPC fallback. With #4471 the same call silently takes up to 24 hours. Because the fallback hides misconfiguration, those counterparts exist and do not get fixed.
- A store failure after a successful offer is treated as a delivery failure. The next attempt sends the offer again. The receiver's store deduplicates identical credentials, so this is harmless today, but it makes a local problem look like a delivery problem.
- One goroutine per job, each with its own backoff. A setup run issuing hundreds of credentials to a wallet that is down produces hundreds of loops against the same endpoint.
- Give-up means gRPC publish, which is the behaviour #4469 wants to move away from.
- The job holds the full signed credential in a KV shelf. Dead-lettered jobs are kept forever.

## Two kinds of issuance

Issuance happens in two situations with different expectations.

[...]

Issue #1948, opened 2023-03-06
Support fine grain control of encryption key algorithms


woutslakhorst:
https://www.sogis.eu/documents/cc/crypto/SOGIS-Agreed-Cryptographic-Mechanisms-1.2.pdf specifies that only RSA keys may be used for asymmetric encryption.

The node currently has code support for RSA and EC, but uses EC by default. There's no option to create RSA keys via the API.

This option should be added so parties can generate RSA encryption keys when required to conform to SOGIS (and others)

woutslakhorst:
still relevant

woutslakhorst:
API already supports this. Encryption is not used at the moment.

PR #4522, opened 2026-09-10
fix(vdr): reflect key store backend capability in DID document key usage

### LINKED ISSUE #4518: vdr: reflect key usage in DID document verification methods on key pair creation (+ backports)
Tracks the first checklist item from nuts-foundation/nuts-node#4515 (comment):

> When creating a key pair, let the usage (Verification Methods) in the DID document reflect the allowed usage (patch for master, v6.2 and v5.4). EC keys in Azure Key Vault can't be used for decryption (KeyAgreement key)

Context: an EC key generated in Azure Key Vault can't be used for decryption, so it should not be published as a KeyAgreement verification method. Today key pair creation doesn't restrict a DID document's verification methods to what the underlying key store can actually do with that key, so a KeyAgreement entry can end up pointing at a key that can't decrypt — breaking gRPC network private-transaction delivery for the sender resolving the recipient's KeyAgreement key.

## Tasks

- [x] Patch for master: #4522
- [x] Backport to v6.2: #4550
- [ ] ~~Backport to v5.4~~ — not needed, v5.4 doesn't support the Azure Key Vault backend

Assisted by AI

reinkrul:
Closes #4518.

## Problem

did:nuts's and did:web's key/VerificationMethod creation always tagged a newly generated key with whatever `DIDKeyFlags` were requested, without checking whether the key store backend can actually back them. did:web already hard-rejects `KeyAgreementUsage` for an unrelated reason (RSA not implemented there). did:nuts is where it bit: a DID document could advertise KeyAgreement for a key that can't decrypt — e.g. an Azure Key Vault key, which can only sign — silently breaking gRPC private-transaction delivery for anyone sending to it (see nuts-foundation/nuts-node#4515).

**Scope note:** only affects *newly generated* keys/documents going forward. Doesn't retroactively fix did:nuts documents already published with a KeyAgreement entry that can't decrypt — already anchored in the DAG. We know of (almost certainly) one affected party in production, handled directly with them; no network-wide migration planned.

## Fix

- `crypto.Crypto.New()` takes the required `orm.DIDKeyFlags` and checks it against what the configured backend can back — Azure Key Vault: signing only; every other backend: signing + decryption — *before* creating any key. If unsupported, no key is created and `crypto.ErrKeyUsageNotSupported` is returned (mapped to 400 in both `vdr/api/v1` and `v2`).
- did:nuts's and did:web's `NewVerificationMethod`/`NewDocument` pass the caller's requested flags straight through as a hard requirement instead of silently granting less than what was asked.
- did:nuts's `NewDocument` no longer forces `DefaultKeyFlags()` regardless of what was requested (that's what caused the original bug). It now honors `keyFlags` like did:web already did: the default subject-creation request (`AssertionKeyUsage` only, unless `EncryptionKeyCreationOption` is given) reaches did:nuts too, so a default did:nuts document creates fine on Azure Key Vault (just without KeyAgreement) — this is safe now that OpenID4VCI is an alternative to gRPC/DAG delivery. An explicit request for KeyAgreement against a backend that can't back it still fails loudly.

No new persisted state, no migration: nothing needs to know a key's usage after the moment it's created.

## Test plan

- `TestCrypto_New/required_usage_not_supported_by_backend` — refuses before calling the backend.
- `TestManager_NewDocument/key_store_backend_can't_back_KeyAgreement` — default document creation succeeds without KeyAgreement; explicit `DefaultKeyFlags()`/`EncryptionKeyUsage` requests fail with `ErrKeyUsageNotSupported`.
- Existing `crypto`, `vdr`, `vdr/didnuts`, `vdr/didweb`, `vdr/didsubject`, `network`, `vcr` suites still pass.

Assisted by AI


stevenvegt (approved):
Code for prevention of newly created documents looks good.

Should we also add detection of existing incompatible keys, e.g. during startup?

reinkrul:
> Code for prevention of newly created documents looks good.
> 
> Should we also add detection of existing incompatible keys, e.g. during startup?

As it's only 1 vendor we know of, I think we should just contact them to get it fixed. Because it might become more complex later; RSA keys supporting decryption, EC keys not. So it could become more complex than "this key store supports encryption", making the startup check complicated and expensive.

reinkrul:
## Consideration: V1 `CreateDID` needs to explicitly request KeyAgreement again — but that exposes a pre-existing bug

While investigating the `e2e-test` failure on this PR, found a regression in the deprecated VDR V1 API, and a pre-existing bug it makes newly reachable.

### The regression

`didnuts.Manager.NewDocument` used to ignore its `keyFlags` argument and always request `DefaultKeyFlags()` (Assertion + KeyAgreement), regardless of what the caller asked for. This PR fixes that (correctly, for Azure Key Vault support) — but `vdr/api/v1/api.go:CreateDID` never explicitly requested `EncryptionKeyCreationOption`; it relied on did:nuts's old bug to always get a KeyAgreement key. Once that bug is fixed, V1-created did:nuts documents silently lose their KeyAgreement key, breaking gRPC/DAG private-transaction delivery (PAL header encryption). `docs/_static/vdr/v1.yaml` documents `keyAgreement = true` as the fixed default regardless of request body, so this is a real behavior change. Fix: add `.With(didsubject.EncryptionKeyCreationOption{})` to V1's `CreateDID`, restoring the documented default.

This is what's failing `e2e-tests/storage/backup-restore` and `e2e-tests/nuts-network/private-transactions` (both use the V1 `CreateDID` endpoint via `util.sh:setupNode`).

### Two different constraints, only one should block

- **Azure Key Vault genuinely can't do it.** An EC key from Azure Key Vault can't do ECDH/decryption — a hard, physical backend limitation. Requesting KeyAgreement for did:nuts on that backend should fail loudly (`ErrKeyUsageNotSupported`, 400). This PR already implements that correctly.
- **did:web never supports it**, regardless of backend — an RSA/SOGIS constraint (#1948), not a capability gap in any particular key store.

`SqlManager.Create` currently conflates these: it has a guard that aborts creation of the **entire subject** — did:nuts included — whenever KeyAgreement is requested and did:web is among the configured methods:

```go
if keyFlags.Is(orm.KeyAgreementUsage) && method == "web" {
    return nil, ErrKeyAgreementNotSupported
}
```

That's too broad. A party must be able to create a subject with *any* combination of did:nuts/did:web enabled, request KeyAgreement, and have did:nuts get a working key (backend permitting) while did:web simply doesn't get one — silently, not as an error. Making V1 explicitly request KeyAgreement without addressing this would newly break every V1 caller with did:web enabled (plausibly common among v5-era integrations), which today "works" only because did:nuts's bug bypassed this guard entirely.

Recommendation: don't drop `ErrKeyAgreementNotSupported`'s underlying policy (did:web still never gets a KeyAgreement key — no change to the #1948/SOGIS position), just stop erroring the whole subject over it. For a minimal, temporary fix: keep the existing hardcoded `method == "web"` check in `SqlManager.Create` (consistent with what's already there today), but change what it does — instead of aborting the whole call with `ErrKeyAgreementNotSupported`, strip the KeyAgreement bit from that method's flags before calling `NewDocument`, so did:web's document is created without it while did:nuts still gets its key. No compliance trade-off, and Azure Key Vault stays correctly enforced exactly where this PR already put it, in `Crypto.New`. This stays a hardcoded, temporary check until we relax SOGIS conformance for did:web (#1948) — at which point did:web could get a real KeyAgreement key like any other method, and the special-casing goes away entirely rather than needing a cleaner capability-declaring design.

### Impact

Once this is fixed, V1 callers with did:web enabled shouldn't need to change anything. The remaining question is only for Azure Key Vault + did:nuts (the #4518 scenario): if that party needs a working KeyAgreement key, no API version can provide one on that backend. **Correction**: V1 has no opt-out — its request body is ignored entirely, `keyAgreement` is a fixed default — so V1 + Azure Key Vault always fails. The only paths are V2 with `keys.encryptionKey: false`, or moving the key(s) backing did:nuts off Azure Key Vault.

Edit: see the follow-up comment below for the full verified behavior matrix across V1/V2, with and without Azure Key Vault.

Assisted by AI

reinkrul:
## Verified behavior matrix (did:nuts KeyAgreement / did:web)

Documenting the scenarios discussed on this PR, double-checked against the code as it now stands (through the V2-default fix).

| Call | did:nuts | did:web | Result |
|---|---|---|---|
| `POST /internal/vdr/v1/did` | gets KeyAgreement | no KeyAgreement | 200 |
| same, key store backend can't back it (e.g. Azure Key Vault) | — | — | 400 `ErrKeyUsageNotSupported`, no documents created |
| `POST /internal/vdr/v2/subject`, no body | gets KeyAgreement | no KeyAgreement | 200 |
| same, w/ Azure Key Vault | — | — | 400 `ErrKeyUsageNotSupported` |
| `POST /internal/vdr/v2/subject`, `keys.encryptionKey: true` | gets KeyAgreement | no KeyAgreement | 200 |
| same, w/ Azure Key Vault | — | — | 400 `ErrKeyUsageNotSupported` |
| `POST /internal/vdr/v2/subject`, `keys.encryptionKey: false` | no KeyAgreement | no KeyAgreement | 200 |
| same, w/ Azure Key Vault | no KeyAgreement | no KeyAgreement | 200 — this is the working opt-out for Azure Key Vault parties on V2 |

Notes:
- V1 has no opt-out (request body is ignored entirely), so V1 + Azure Key Vault always fails; the only paths are switching to V2 with `keys.encryptionKey: false`, or moving the key(s) backing did:nuts off Azure Key Vault.
- did:web never gets a KeyAgreement key regardless of any of the above — that's the pre-existing #1948/SOGIS constraint, not something either API can override, and it never produces an error on its own.
- V2's default (no body) now mirrors V1's: both request KeyAgreement unless explicitly told not to, matching the behavior every did:nuts document used to get before key usage started reflecting backend capability.

Assisted by AI

reinkrul:
Backported to V6.2: #4550.

PR #4535, opened 2026-09-11
ci: skip test, e2e-test, govulncheck, docker build, and CodeQL (go) on docs-only changes (backport #4455 and #4459 to V5.4)

reinkrul:
Backport of #4455 and #4459 to V5.4. Neither had been backported here before, so this brings both over in one PR.

- #4455: `test` and `e2e-test` (required status checks) skip their actual work on docs-only changes (docs/, `*.md`, `*.rst`), gated via a `dorny/paths-filter` `changes` job + job-level `if:`. A job skipped this way still reports as a passing check, unlike a trigger-level path filter, which would leave a required check stuck pending.
- #4459: extends the same pattern to `govulncheck.yaml`, `build-images.yaml` (docker), and `codeql-analysis.yml` (Analyze (go)). None of these three are required status checks, but they still burn CI time on changes that can't affect Go code or the built image.

Assisted by AI

### DIFF
--- a/.github/workflows/build-images.yaml
+++ b/.github/workflows/build-images.yaml
@@ -12,7 +12,33 @@ on:
       - master
 
 jobs:
+  # Docs-only changes (docs/, markdown, rst) can't affect the built image, so building/pushing one
+  # is redundant. Detected once here so the job below can skip entirely on them.
+  changes:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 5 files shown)


stevenvegt (approved):
But still some open warnings about permissions in the workflows.

PR #4552, opened 2026-09-17
test(network): wait for peer gauge decrement instead of asserting immediately

stevenvegt:
## Problem

`Test_grpcConnectionManager_openOutboundStreams/ok` asserts the peer gauge is back to zero immediately after the disconnect observer has fired:

```go
c.disconnect()
disconnectedWG.Wait()

// Assert peer gauge is decremented
_ = client.peersCounter.Write(metric)
assert.Equal(t, float64(0), *metric.Gauge.Value)
```

Those two events are not ordered. The observer is notified from the per-protocol goroutine that watches `clientStream.Context().Done()` (`connection_manager.go:448`), while the gauge is decremented by the `defer s.peersCounter.Dec()` in `openOutboundStreams`, which only runs once that function returns. The assertion wins the race on master today, so the test is green, but the margin is whatever the main goroutine has to do between `waitUntilDisconnected()` returning and the deferred call.

That margin is not guaranteed to stay zero. #4467 adds a `grpcConn.Close()` plus a wait for the receive loops in exactly that spot, and the test then fails in roughly 1 out of 10 to 1 out of 30 runs, in CI and locally.

## Change

Wait for the decrement with `test.WaitFor`, the same way the test already waits for the increment a few lines above. Test-only, no production code touched.

## Verification

This is a latent race, so there is no RED on master: the assertion passes 100 out of 100 runs both before and after this change. The failure is only reproducible on the #4467 branch, where I measured it before proposing the fix:

- master, `-count=100`: 0 failures
- #4467 branch, `-count=30`: 1 to 3 failures
- #4467 branch with just the two added lines stubbed out, `-count=100`: 0 failures
- #4467 branch with this test change, `-count=100`: 0 failures

`go test -race ./network/transport/grpc/... -count=1` passes on this branch.

Assisted-by: AI


stevenvegt:
Folded into #4467 as 87bd74fb. The race only fires on that branch (the `grpcConn.Close()` plus wait for the receive loops is what widens the window), so as a standalone PR against master this change has no failing test to justify it. #4467 cannot go green without it, which is where it belongs.

PR #4585, opened 2026-09-21
test(network): fix race in TestNotifier_Notify/OK_-_updates_DB (backport of #4274)

reinkrul:
- Backport of #4274 to V6.2. Clean cherry-pick, test code only.
- The test waited for the callback counter and then read the event from the database. The notifier saves the retry count and timestamp after the callback returns, so the read could come too early: `Retries` was still 0 and `Latest` was nil, which made the test panic.
- The test now waits until the saved event has `Latest` set.
- The same failure showed up in CI on V5.4 (#4582).
- Checked locally: 30 runs with the race detector pass.

Assisted by AI


stevenvegt:
The test error "Message:"Slow query (took 210.833361ms): INSERT INTO `key_reference" might be interesting to look into. Seems unrelated to this change.

stevenvegt (approved):

